PR #45, opened 2026-04-14
feat(web): add headless mode

powerzist:
Now you can run the web interface with automatically opening the browser by default. If you want to disable this feature, you can use the `--headless` flag when starting the web interface.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/web.py
+++ b/packages/memtomem/src/memtomem/cli/web.py
@@ -56,15 +56,16 @@ def web(host: str, port: int, headless: bool) -> None:
         while not server.started:
             await asyncio.sleep(0.1)
         import webbrowser
+
         webbrowser.open(f"http://{host}:{port}")
-    
+
     async def start_server() -> None:
         web_config = uvicorn.Config(create_app(lifespan=_lifespan), host=host, port=port)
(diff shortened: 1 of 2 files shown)


memtomem (commented):
Thanks for this — auto-opening the browser is a nice QoL improvement!

Two things I'd like to discuss before merging:

**1. Default behavior — opt-in vs opt-out**

Currently `mm web` doesn't open a browser, so making it the new default is a breaking change for scripts, SSH sessions, etc. Most tools use an opt-in flag for this (`jupyter notebook --no-browser`, `vite --open`). Could we flip it to `--open` instead of `--headless`?

**2. Implementation complexity**

Replacing `uvicorn.run()` with `asyncio.run()` + `asyncio.gather()` is a lot of machinery just to open a browser. A simpler approach:

```python
if open_browser:
    import threading, webbrowser
    threading.Timer(1.0, webbrowser.open, args=[f"http://{host}:{port}"]).start()
uvicorn.run(...)
```

This also avoids the infinite polling loop (`while not server.started`) which has no timeout.

Minor: please remove trailing whitespace on the blank lines, and add a test that `webbrowser.open` is/isn't called depending on the flag.

powerzist:

PR #84, opened 2026-04-14
fix: log dropped shutdown sentinel in watcher when queue is full

### LINKED ISSUE #83: indexing: log dropped shutdown sentinel when watcher queue is full
## Summary

`packages/memtomem/src/memtomem/indexing/watcher.py:96-97` silently swallows `asyncio.QueueFull` when enqueuing the shutdown sentinel, similar to the pattern fixed in #78.

## Context

```python
async def stop(self) -> None:
    if self._task is not None:
        # Signal graceful shutdown — flush pending before exit
        try:
            self._queue.put_nowait(_STOP_SENTINEL)
        except asyncio.QueueFull:
            pass   # <-- silent
```

## Problem

If the queue is full when `stop()` tries to enqueue the sentinel, the shutdown signal is lost. The consumer task never sees the sentinel, so graceful shutdown degrades into timeout-based cancellation (the `asyncio.wait_for(..., timeout=5.0)` immediately below). The operator only observes a 5-second hang followed by cancellation — with no log line explaining why the graceful path failed.

## Suggested fix

Add a `logger.warning` — consistent with the same file's other `asyncio.QueueFull` handler at line 42-43, which already uses `logger.warning`:

```python
except asyncio.QueueFull:
    logger.warning("File watcher queue full; could not signal graceful shutdown")
```

`logger.warning` (not `debug`) is correct here — a dropped sentinel is a real degradation that the operator needs to see at the default log level when diagnosing a hang.

## Getting started

1. Read [CONTRIBUTING.md](../CONTRIBUTING.md) for setup.
2. Edit `packages/memtomem/src/memtomem/indexing/watcher.py:96-97`.
3. Run `uv run ruff check packages/memtomem/src && uv run ruff format --check packages/memtomem/src`.
4. Run `uv run pytest packages/memtomem/tests/ -m "not ollama" -k watcher`.

memtomem:
## Summary

- Replace silent `pass` with `logger.warning` when the watcher queue rejects the shutdown sentinel in `AsyncFileWatcher.stop()`.
- Message mirrors the sibling `asyncio.QueueFull` handler at `watcher.py:42-43` (also `logger.warning`).
- Operators diagnosing a mysterious 5-second graceful-shutdown hang now see the cause at the default log level.

## Test plan

- [x] `uv run ruff check packages/memtomem/src && uv run ruff format --check packages/memtomem/src` passes
- [x] `uv run pytest packages/memtomem/tests/ -m "not ollama" -k watcher` (2 passed)
- [ ] CI green

Closes #83

### DIFF
--- a/packages/memtomem/src/memtomem/indexing/watcher.py
+++ b/packages/memtomem/src/memtomem/indexing/watcher.py
@@ -94,7 +94,7 @@ class FileWatcher:
             try:
                 self._queue.put_nowait(_STOP_SENTINEL)
             except asyncio.QueueFull:
-                pass
+                logger.warning("File watcher queue full; could not signal graceful shutdown")
             try:
                 await asyncio.wait_for(self._task, timeout=5.0)
             except (TimeoutError, asyncio.CancelledError):


memtomem:
Closing — yielding to @K-Tanish who requested to take #83 before this PR was opened. Context in the issue.

PR #85, opened 2026-04-14
fixed watcher queue full logging

### LINKED ISSUE #83: indexing: log dropped shutdown sentinel when watcher queue is full

K-Tanish:
## Description
Fix file watcher shutdown logging so a dropped shutdown sentinel is no longer silent.

## Changes made
- updated ` watcher.py `
- When ` FileWatcher.py ` fails to enqueue ` _STOP_SENTINEL ` because  the queue is full, it now logs a warning as "File watcher queue full; could not signal graceful shutdown"

Closes #83

### DIFF
--- a/packages/memtomem/src/memtomem/indexing/watcher.py
+++ b/packages/memtomem/src/memtomem/indexing/watcher.py
@@ -94,7 +94,7 @@ class FileWatcher:
             try:
                 self._queue.put_nowait(_STOP_SENTINEL)
             except asyncio.QueueFull:
-                pass
+                logger.warning("File watcher queue full; could not signal graceful shutdown")
             try:
                 await asyncio.wait_for(self._task, timeout=5.0)
             except (TimeoutError, asyncio.CancelledError):


K-Tanish:
I have read the CLA document and I hereby sign the CLA

memtomem:
recheck

memtomem (approved):
LGTM. Matches sibling handler at watcher.py:43. Thanks @K-Tanish!

Issue #67, opened 2026-04-14
Add logging to silent except-pass blocks in sqlite_backend.py and status_config.py


memtomem:
## Summary

Four `except Exception: pass` blocks silently swallow errors without any logging. This hides potential problems — corrupted embeddings, failed WAL checkpoints, resource leaks, and incomplete status reports all go unnoticed.

### `packages/memtomem/src/memtomem/storage/sqlite_backend.py`

1. **Line 148** — read pool connection close:
```python
try:
    rconn.close()
except Exception:
    pass
```

2. **Line 155** — WAL checkpoint before close:
```python
try:
    self._db.execute("PRAGMA wal_checkpoint(TRUNCATE)")
except Exception:
    pass
```

3. **Line 617** — embedding deserialization (most impactful — silently drops corrupted embeddings):
```python
try:
    result[row[0]] = deserialize_f32(row[1])
except Exception:
    pass
```

### `packages/memtomem/src/memtomem/server/tools/status_config.py`

4. **Line 90** — orphan detection in `mem_status`:
```python
try:
    source_files = await app.storage.get_all_source_files()
    orphaned = sum(1 for sf in source_files if not sf.exists())
    ...
except Exception:
    pass
```

## Proposed fix

Replace `pass` with `logger.debug(...)` or `logger.warning(...)` depending on severity:

- **Connection close** (line 148): `logger.debug` — best-effort cleanup, low risk
- **WAL checkpoint** (line 155): `logger.debug` — expected to fail occasionally (e.g., read-only mode)
- **Embedding deserialization** (line 617): `logger.warning` — data is silently lost
- **Orphan detection** (line 90): `logger.debug` — informational, doesn't affect core functionality

Both files already import `logging` and have a module-level `logger`.

## Getting started

1. Read [CONTRIBUTING.md](../CONTRIBUTING.md) for setup instructions
2. Edit the two files listed above
3. Run `uv run ruff check packages/memtomem/src && uv run ruff format --check packages/memtomem/src`
4. Run `uv run pytest packages/memtomem/tests/ -x`

PR #91, opened 2026-04-15
fix: log swallowed query_history error in get_knowledge_gaps (follow-up #67)

memtomem:
## Summary

- Add `logger.debug(..., exc_info=True)` to the broad `except Exception` in `AnalyticsMixin.get_knowledge_gaps`.
- Keep `return []` so callers (`server/tools/reflection.py`) continue to see the empty-list sentinel on DB failure.

Missed instance of the #67 pattern. #67 fixed `sqlite_backend.py` and `status_config.py`; this one lives in `storage/mixins/analytics.py` and survived the original sweep.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_analytics.py -q` (16 passed)
- [x] `uv run ruff check` + `ruff format --check` on the edited file

### DIFF
--- a/packages/memtomem/src/memtomem/storage/mixins/analytics.py
+++ b/packages/memtomem/src/memtomem/storage/mixins/analytics.py
@@ -132,6 +132,7 @@ class AnalyticsMixin:
             ).fetchall()
             return [{"query": r[0], "count": r[1]} for r in rows]
         except Exception:
+            logger.debug("query_history query failed (table may not exist yet)", exc_info=True)
             return []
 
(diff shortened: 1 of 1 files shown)


memtomem:
Superseded by #117 which landed the same fix on main (0a843b1 range) — same file, same function, same approach (add `logger.debug` with `exc_info=True` to the swallowing `except Exception: return []`). Closing as duplicate. Thanks for the work though — the fact that two of us independently arrived at the same remediation confirms the #67 follow-up was the right call.

PR #131, opened 2026-04-15
fix(test): make pytest work on Windows

powerzist:
- fix the temporary config dir for tests
- fix the path formatter to use forward slashes only (even on Windows)
- fix OS-dependent path separator handling
- fix OS-dependent encoding ambiguity by enforcing UTF-8
- apply ruff formatting to affected files Tested on Windows 10.


packages/memtomem/src/memtomem/cli/init_cmd.py
```diff
@@ -330,9 +330,9 @@ def _step_mcp(state: dict) -> None:
 # ── Write config & summary ────────────────────────────────────────────
 
 
-def _write_config_and_summary(state: dict) -> None:
+def _write_config_and_summary(state: dict, base_dir: Path = Path("~/").expanduser()) -> None:
     """Write config files and show summary (runs after all steps)."""
-    config_dir = Path("~/.memtomem").expanduser()
+    config_dir = base_dir / ".memtomem"
```
memtomem: Nit: `Path("~/").expanduser()` here evaluates at module import and is captured in the function object, so later env tweaks (e.g. `monkeypatch.setenv("HOME", ...)`) won't affect it. A `None` sentinel with `Path.home()` resolved per call plays nicer with standard test-isolation patterns, and is a bit more idiomatic for optional DI. Feel free to take or leave — non-blocking.

```suggestion
def _write_config_and_summary(state: dict, base_dir: Path | None = None) -> None:
    """Write config files and show summary (runs after all steps)."""
    if base_dir is None:
        base_dir = Path.home()
    config_dir = base_dir / ".memtomem"
```

memtomem (approved):
Thanks for this — much appreciated as a first contribution. All CI green, and the fixes hold up well on review:

- `formatters._display_path` routing through `Path(path).as_posix()` actually removes a latent Windows bug: `_short_path` delegates to `PurePosixPath(_display_path(path)).name`, and `PurePosixPath` doesn't recognize `\` as a separator, so on Windows the whole path was being returned as the "filename." Nice catch.
- `ast.parse(..., encoding="utf-8")` closes a platform-default encoding trap in the test helpers.
- darwin `/private/tmp/` handling is preserved, so no regression there.
- The `base_dir` injection on `_write_config_and_summary` is a real test-isolation win beyond just Windows.

Approving. Two non-blocking notes:

1. **`base_dir` default evaluates at import time.** `Path("~/").expanduser()` is captured once when the module loads, so it won't pick up a later `monkeypatch.setenv("HOME", ...)`. I've left an inline suggestion for a `None` sentinel + per-call `Path.home()` — apply it in this PR if you like, or we can follow up separately. Doesn't block merging.

2. **Notebook reformat is outside the Windows-fix scope.** Our CI ruff pass only covers `packages/memtomem/{src,tests}` (see `ci.yml`), so `examples/notebooks/` reformatting isn't strictly required here. No harm this time — for future PRs please keep unrelated formatting in its own PR so the review stays tight.

Thanks again — looking forward to more.

Issue #63, opened 2026-04-14
Add debug logging to silent exception catches in search/expansion.py


memtomem:
## Context

In `packages/memtomem/src/memtomem/search/expansion.py`, two functions silently catch exceptions and return the original query with no logging:

**Line 25-26** (`expand_query_tags`):
\`\`\`python
except Exception:
    return query
\`\`\`

**Line 55-56** (`expand_query_headings`):
\`\`\`python
except Exception:
    return query
\`\`\`

Meanwhile, the successful expansion paths in the same file already use `logger.debug(...)` (lines 40, 71). The inconsistency makes it hard to diagnose when expansion silently fails.

## Task

Add `logger.debug("...", exc_info=True)` before each `return query` in the except blocks, matching the existing logging pattern in the file.

## Acceptance criteria

- Both catches log at DEBUG level with `exc_info=True`
- No behavior change (still returns `query` on failure)
- `uv run ruff check` and `uv run ruff format --check` pass

PR #141, opened 2026-04-15
fix(indexing): add the warning log in is_duplicate exception (#139)

### LINKED ISSUE #139: indexing: log warning when is_duplicate silently returns False on exception
## Context

`IndexEngine.is_duplicate()` in `packages/memtomem/src/memtomem/indexing/engine.py` catches **all** exceptions and returns `False` with no log line. Any failure during the embedder call, the storage `dense_search`, or the namespace parse is silently treated as "not a duplicate" — so the caller proceeds to ingest, and a duplicate chunk may be stored without anyone noticing.

**Handled today**: successful duplicate detection returns `True`/`False` based on score threshold.
**Unhandled**: embedder failure, storage outage, network error — all collapse to `return False`.

**Current code** (`indexing/engine.py:147-155`):
\`\`\`python
try:
    embedding = await self._embedder.embed_query(text)
    ns_filter = NamespaceFilter.parse(namespace) if namespace else None
    results = await self._storage.dense_search(
        embedding, top_k=1, namespace_filter=ns_filter
    )
    return bool(results and results[0].score >= threshold)
except Exception:
    return False
\`\`\`

This repeats the silent-except pattern that #63, #67, and #83 each cleaned up in their respective files. `is_duplicate` is the next case.

## Proposed fix

Add a warning-level log with `exc_info=True` before returning `False`. Behavior stays the same (still returns `False` on failure); only observability changes.

\`\`\`python
except Exception:
    logger.warning("is_duplicate failed; treating as non-duplicate", exc_info=True)
    return False
\`\`\`

## Acceptance criteria

- `except Exception` block logs at `warning` level with `exc_info=True`
- Return value unchanged (still `False` on failure)
- `uv run ruff check packages/memtomem/src && uv run ruff format --check packages/memtomem/src` pass
- `uv run pytest -m \"not ollama\"` stays green

## Files touched

- `packages/memtomem/src/memtomem/indexing/engine.py` (1 file, ~2 lines)

## Notes

Re-raising on non-transient errors is an interesting longer-term direction but is **out of scope** for this issue — future discussion.

Um-E-Salma:
## Summary
This PR adds a warning log inside the `is_duplicate` exception handler to improve observability when failures occur.

## Changes
- Added `logger.warning(..., exc_info=True)` inside exception block
- No change in behavior (still returns False on failure)

## Related Issue
Closes #139

## Notes
This improves debugging and follows existing logging patterns in the codebase.

### DIFF
--- a/packages/memtomem/src/memtomem/indexing/engine.py
+++ b/packages/memtomem/src/memtomem/indexing/engine.py
@@ -152,6 +152,7 @@ class IndexEngine:
             )
             return bool(results and results[0].score >= threshold)
         except Exception:
+            logger.warning("is_duplicate failed; treating as non-duplicate", exc_info=True)
             return False
 
     def _resolve_namespace(self, file_path: Path, explicit_ns: str | None) -> str | None:


Um-E-Salma:

memtomem (approved):
Welcome and thanks for picking up your first issue here, @Um-E-Salma!

The change matches the proposed fix in #139 exactly:
- `logger.warning(..., exc_info=True)` added inside the `except Exception` block
- Return value preserved — observability-only change

Verified:
- `logger` is already defined at `packages/memtomem/src/memtomem/indexing/engine.py:23`
- Continues the same silent-except cleanup pattern as #63 / #67 / #83

The `CLA Assistant` check is showing failure even though you've signed it (your sign-off comment is recorded above, and the bot's follow-up confirms "All contributors have signed the CLA"). That's a known bot regression on our side, not yours — merging via admin override.

PR #130, opened 2026-04-15
fix: tighten llm_provider type from object to LLMProvider

armorbreak001:
## Summary

Fixes **#118** — Narrow `LLMProvider` type annotations from `Any` to `LLMProvider | None` across 4 declarations, eliminating 2 mypy errors.

## Changes

| File | Line(s) | Change |
|------|---------|--------|
| `search/pipeline.py` | 67 | `Any` → `LLMProvider \| None` |
| `tools/policy_engine.py` | 270, 572, 618 | `Any` → `LLMProvider \| None` |

Uses existing `TYPE_CHECKING` import pattern (already in place) to avoid circular dependency with `memtomem.llm.base.LLMProvider` Protocol.

## Mypy Error Comparison

**As requested by @memtomem — before/after mypy error count:**

\`\`\`
# On main:
Found 40 errors in 16 files (checked 186 source files)

# On this branch (fix/tighten-llm-provider-type):
Found 38 errors in 16 files (checked 186 source files)
\`\`\`

**Delta: -2 errors** (40 → 38). The 2 eliminated errors are the `arg-type` errors at the narrowed declaration sites. No unexpected arg-type cascade appeared at call sites — the remaining 38 errors are pre-existing and unchanged.

## Verification

- [x] `auto_tag.py:244/276` already on `LLMProvider | None` (not changed, as noted in #118)
- [x] `LLMProvider` Protocol confirmed at `packages/memtomem/src/memtomem/llm/base.py:8`
- [x] `TYPE_CHECKING` import pattern correctly avoids circular dependency
- [x] All 4 declarations narrowed as described
- [x] No new errors introduced

## CLA
✅ Signed — see comment above.

PR #145, opened 2026-04-16
typing: tighten 4 incomplete list/dict element-type annotations (#138)

### LINKED ISSUE #138: typing: tighten 4 incomplete list/dict element-type annotations
## Context

Four local variables are declared with bare `list` / `dict` (no element
type parameters), so mypy falls back to `list[Any]` / `dict[Any, Any]`.
No mypy error is produced today, but future refactors lose a type-
precision signal, and these stand out next to sibling variables that
are already fully annotated (e.g. `list[Chunk]`, `dict[str, int]`).

- `packages/memtomem/src/memtomem/indexing/engine.py:98` —
  `all_new_chunk_ids: list = []` (neighboring vars on lines 86-87 use
  `list[dict[str, object]]` / `list[str]`)
- `packages/memtomem/src/memtomem/storage/sqlite_namespace.py:113` —
  `params: list = []` (SQL placeholder params)
- `packages/memtomem/src/memtomem/cli/search.py:82` —
  `groups: dict[str, list] = {}` (inner `list` missing element type)
- `packages/memtomem/src/memtomem/web/routes/context_gateway.py:38` —
  `result: dict = {}` (both key and value types missing)

## Not a runtime change

Purely annotation tightening. Runtime semantics unchanged.

## How to pick the right element type

For each line, scan the subsequent `.append(...)` / `[key] = ...`
assignments to infer the concrete type. If still ambiguous:

- SQL params → often `list[object]` (sqlite3 accepts mixed).
- `all_new_chunk_ids` → check the consumer signature (likely `UUID` or
  `str`).
- `groups` / `result` → follow the dict build-up and use the most
  specific type that still covers every assignment.

## Acceptance

- Each of the 4 annotations gets a concrete element type.
- `uv run mypy packages/memtomem/src` error count stays at or below
  the current baseline (39).
- `uv run pytest -m "not ollama"` passes.
- No new `# type: ignore` introduced.

armorbreak001:
Fixes #138 — Four bare list/dict annotations now have concrete element types.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/search.py
+++ b/packages/memtomem/src/memtomem/cli/search.py
@@ -1,6 +1,10 @@
 """CLI: memtomem search <query>."""
 
 from __future__ import annotations
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from memtomem.models import SearchResult
 
 import asyncio
 import json
@@ -79,7 +83,7 @@ async def _search(
         if not results:
             return
         # Group by namespace, show tags, adjust detail by relevance
-        groups: dict[str, list] = {}
+        groups: dict[str, list[SearchResult]] = {}
         for r in results:
             ns = r.chunk.metadata.namespace or "default"
             groups.setdefault(ns, []).append(r)
--- a/packages/memtomem/src/memtomem/indexing/engine.py
+++ b/packages/memtomem/src/memtomem/indexing/engine.py
@@ -6,6 +6,7 @@ import asyncio
 import logging
 import time
 from pathlib import Path
+from uuid import UUID
 from typing import TYPE_CHECKING
 
 from memtomem.chunking.markdown import MarkdownChunker
@@ -95,7 +96,7 @@ class IndexEngine:
 
         # Aggregate new_chunk_ids across all files — preserves per-file order
         # so callers that sort/filter by source get a consistent ordering.
-        all_new_chunk_ids: list = []
+        all_new_chunk_ids: list[UUID] = []
         for r in file_results:
             ids = r.get("new_chunk_ids", ())
             if ids:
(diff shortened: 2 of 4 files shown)


armorbreak001:
I have read the CLA Document and I hereby sign the CLA on this PR

memtomem:
Hi @armorbreak001 — thanks for the two new PRs! Same heads-up as before: we recently migrated CLA signature storage to a dedicated branch (context in #143). Your sign-off from #130 didn't persist because the old flow was broken at the time. Could you re-post the sign statement once more on this PR? It'll cover both #145 and #146 — we can `recheck` the other one afterwards.

`I have read the CLA Document and I hereby sign the CLA`

memtomem (approved):
Thanks again, @armorbreak001! All four annotation refinements look correct:

1. `cli/search.py`: `dict[str, list]` → `dict[str, list[SearchResult]]` — `TYPE_CHECKING` import avoids circular dep ✓
2. `indexing/engine.py`: `list` → `list[UUID]` — matches the return type from storage insert operations ✓
3. `sqlite_namespace.py`: `list` → `list[object]` — heterogeneous query params, widest concrete type ✓
4. `context_gateway.py`: `dict` → `dict[str, dict[str, int | bool | str]]` — matches `_count_statuses` shape ✓

No runtime changes, just tighter static annotations. `from __future__ import annotations` is present in all modified files, so the `UUID` import in `engine.py` could technically live in the `TYPE_CHECKING` block, but it's harmless at runtime (stdlib) — not blocking.

CLA note: please re-sign the CLA per the comment on #145 — the earlier signature from #130 didn't persist due to a now-fixed infrastructure issue (#143).

memtomem:
recheck

PR #146, opened 2026-04-16
cli(session): defer module-level _STATE_DIR expansion to runtime (#137)

armorbreak001:
## Summary

Fixes **#137** — Replace module-level `_STATE_DIR = Path("~/.memtomem").expanduser()` with a lazy `_state_dir() -> Path` helper that resolves `Path.home()` at each call.

## Problem

The old code captured `~` expansion once at import time:
```python
_STATE_DIR = Path("~/.memtomem").expanduser()  # frozen at import!
```

Any subsequent `monkeypatch.setenv("HOME", ...)` in test code cannot relocate it, limiting test isolation for the session CLI.

## Solution

```python
def _state_dir() -> Path:
    """Return the memtomem state directory, resolving HOME at call time."""
    return Path.home() / ".memtomem"
```

- `_STATE_FILE` still computed at module level (harmless — path construction only)
- `_STATE_DIR.mkdir(...)` call site updated to `_state_dir().mkdir(...)`
- Zero behavior change for production usage

## Test Plan

- [x] Import succeeds: `from memtomem.cli.session_cmd import _state_dir; _state_dir()` returns valid Path
- [x] `uv run mypy packages/memtomem/src/memtomem/cli/session_cmd.py` — 0 errors
- [x] Existing tests pass: `uv run pytest -m "not ollama"` — 1272/1272 OK
- [x] No new `# type: ignore` lines

## AI Usage Disclosure
**None — human-authored.**

## Files Changed (1 file, +6/-3)
- `packages/memtomem/src/memtomem/cli/session_cmd.py`

PR #157, opened 2026-04-16
Log warning instead of silently swallowing exceptions in activity hook

### LINKED ISSUE #149: cli(session): add warning log to silent except in activity hook
## Problem

`cli/session_cmd.py:216` swallows all exceptions silently:

```python
try:
    asyncio.run(_log_event(session_id, event_type, content, metadata))
except Exception:
    pass  # hooks must never block
```

If activity logging fails (e.g. SQLite corruption, missing state file),
the hook silently stops recording session events with no diagnostic
output. At the default log level, this is completely invisible.

Per project convention (see #83, #139), dropped operations should log at
**warning** level — `debug` is invisible at default and provides no
diagnostic value.

## Suggested fix

Replace `pass` with `logger.warning(...)`. The hook must still never
block (no re-raise).

```python
import logging

logger = logging.getLogger(__name__)

# ... inside log_activity():
except Exception:
    logger.warning("Activity hook failed", exc_info=True)
```

Note: `session_cmd.py` does not currently import `logging` — the import
needs to be added.

## Scope

- 1 file (`session_cmd.py`)
- Add `logging` import + `logger` + replace `pass` with `logger.warning`
- No behavior change (hook still catches and continues)

impsislegobatman:
## Summary
- Replace bare `except: pass` in `session_cmd.py:log_event()` with `logger.warning(...)` so activity hook failures are logged instead of silently swallowed.
- Add `import logging` and a module-level `logger` to the file.

Fixes #149

## Test plan
- [ ] Verify `uv run ruff check packages/memtomem/src` passes (lint)
- [ ] Verify `uv run pytest -m "not ollama"` passes (tests)
- [ ] Trigger an activity hook failure and confirm the warning appears in logs

### DIFF
--- a/packages/memtomem/src/memtomem/cli/session_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/session_cmd.py
@@ -4,10 +4,13 @@ from __future__ import annotations
 
 import asyncio
 import json
+import logging
 import uuid
 from pathlib import Path
 
 import click
+
+logger = logging.getLogger(__name__)
 
 # Session state file — stores active session UUID.
 _STATE_DIR = Path("~/.memtomem").expanduser()
@@ -214,7 +217,7 @@ def log_event(event_type: str, content: str, meta: str | None) -> None:
     try:
         asyncio.run(_log_event(session_id, event_type, content, metadata))
     except Exception:
-        pass  # hooks must never block
+        logger.warning("activity hook failed for session %s", session_id, exc_info=True)
 
 
 async def _log_event(session_id: str, event_type: str, content: str, metadata: dict | None) -> None:


memtomem:
This is the cleanest of your five PRs — spot-on match for #149 (`logger.warning` + `exc_info=True` + warning level matches the convention from #83/#139), and the `session %s` detail is a nice step beyond the issue's suggested wording.

**One small nit** — please keep the original intent comment:

```python
except Exception:
    logger.warning("activity hook failed for session %s", session_id, exc_info=True)
    # hooks must never block
```

The `logger.warning(...)` alone doesn't signal why we're catching-and-continuing. A future reader might see the warning and think "shouldn't we re-raise?" The comment preempts that.

Non-blocking — if you'd rather not bother, I can add it on merge. CLA + CI approval note same as the other PRs.

memtomem (commented):
Code looks correct — replacing a silent `pass` with `logger.warning(...)` is the right call for the activity hook path.

Note: this PR will have merge conflicts with #146 and #162 which also modify `session_cmd.py`. Please sign the CLA first, then we can coordinate the merge order.

To sign, post exactly:

> I have read the CLA Document and I hereby sign the CLA

memtomem:
Hi @impsislegobatman — closing this as superseded, with thanks and an apology for the CLA friction that turned out to be moot.

The same fix landed on `main` in [b3311c0](https://github.com/memtomem/memtomem/commit/b3311c0) (`fix(cli): replace silent except with warning log in activity hook`). Current code at `packages/memtomem/src/memtomem/cli/session_cmd.py:290`:

```python
except Exception:
    logger.warning("Activity hook failed", exc_info=True)
```

The message string and surrounding flow differ slightly from your patch (owner's version also wires `as_json` output on failure), but the core ask — replace silent `pass` with a warning — is in. Keeping this PR open behind a CLA check for a fix that's already shipped isn't useful.

Thanks for catching the silent-except and filing the fix. Good defensive-noise instinct on what was a real observability gap.

PR #158, opened 2026-04-16
refactor: extract importance decay multiplier into named constant

impsislegobatman:
Fixes #148

Extracted the hardcoded `0.5` decay multiplier into `IMPORTANCE_DECAY_MULTIPLIER` constant in `consolidation_engine.py`, consistent with the existing `DECAY_FLOOR` pattern. Updated both occurrences (engine and server tool).

PR #159, opened 2026-04-16
refactor: extract confidence thresholds into named constants

### LINKED ISSUE #150: entity_extraction: extract confidence thresholds into named constants
## Problem

`tools/entity_extraction.py` uses 6 distinct confidence values across
~11 call sites, all as bare numeric literals:

| Value | Used for | Lines |
|-------|----------|-------|
| `0.95` | ISO date matches | 213 |
| `0.9`  | Action items, known tech | 233, 247 |
| `0.85` | Decision patterns | 224 |
| `0.8`  | Person name context, natural dates | 204, 215 |
| `0.7`  | @mentions, concepts, LLM fallback default | 206, 264, 314, 317 |
| `0.5`  | PascalCase heuristic | 253 |

These thresholds control entity quality filtering but are scattered
through individual `_extract_*` functions with no documentation of what
the values mean or how they relate to each other.

## Suggested fix

Define named constants at the top of `entity_extraction.py`, near the
existing `_VALID_ENTITY_TYPES` block:

```python
_CONFIDENCE_EXACT = 0.95    # high-precision pattern (ISO dates)
_CONFIDENCE_HIGH = 0.9      # reliable pattern (action items, known tech)
_CONFIDENCE_DECISION = 0.85 # decision keyword + context
_CONFIDENCE_CONTEXT = 0.8   # context-dependent match (person names, natural dates)
_CONFIDENCE_DEFAULT = 0.7   # moderate confidence (mentions, concepts, LLM fallback)
_CONFIDENCE_HEURISTIC = 0.5 # low-confidence heuristic (PascalCase guess)
```

Names are suggestions — any clear naming scheme is fine.

## Out of scope (follow-up)

- Minimum length filters (`3`, `4`, `5`) — separate concern, separate issue
- Text truncation limits (`[:200]`, `max_tokens=256`) — separate concern

## Scope

- 1 file (`entity_extraction.py`)
- ~6 constants defined, ~11 call sites updated
- No behavior change

impsislegobatman:
Fixes #150

Extracted all hardcoded confidence values in `entity_extraction.py` into named constants (e.g. `CONFIDENCE_PERSON_CONTEXT = 0.8`, `CONFIDENCE_TECH_INFERRED = 0.5`). Makes tuning thresholds easier and the code more self-documenting.

### DIFF
--- a/packages/memtomem/src/memtomem/tools/entity_extraction.py
+++ b/packages/memtomem/src/memtomem/tools/entity_extraction.py
@@ -17,6 +17,18 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+# ── Confidence thresholds ─────────────────────────────────────────────
+CONFIDENCE_PERSON_CONTEXT = 0.8
+CONFIDENCE_PERSON_MENTION = 0.7
+CONFIDENCE_DATE_ISO = 0.95
+CONFIDENCE_DATE_NATURAL = 0.8
+CONFIDENCE_DECISION = 0.85
+CONFIDENCE_ACTION_ITEM = 0.9
+CONFIDENCE_TECH_KNOWN = 0.9
+CONFIDENCE_TECH_INFERRED = 0.5
+CONFIDENCE_CONCEPT = 0.7
+CONFIDENCE_LLM_FALLBACK = 0.7
+
 _MONTHS = (
     "January",
     "February",
@@ -201,18 +213,18 @@ def _extract_persons(text: str) -> list[ExtractedEntity]:
     for m in _PERSON_CONTEXT_RE.finditer(text):
         name = m.group(1).strip()
         if len(name) > 3:
-            results.append(ExtractedEntity("person", name, 0.8, m.start(1)))
+            results.append(ExtractedEntity("person", name, CONFIDENCE_PERSON_CONTEXT, m.start(1)))
     for m in _MENTION_RE.finditer(text):
-        results.append(ExtractedEntity("person", f"@{m.group(1)}", 0.7, m.start()))
+        results.append(ExtractedEntity("person", f"@{m.group(1)}", CONFIDENCE_PERSON_MENTION, m.start()))
     return results
 
 
 def _extract_dates(text: str) -> list[ExtractedEntity]:
     results: list[ExtractedEntity] = []
     for m in _ISO_DATE_RE.finditer(text):
-        results.append(ExtractedEntity("date", m.group(1), 0.95, m.start()))
+        results.append(ExtractedEntity("date", m.group(1), CONFIDENCE_DATE_ISO, m.start()))
     for m in _NATURAL_DATE_RE.finditer(text):
-        results.append(ExtractedEntity("date", m.group(1).strip(), 0.8, m.start()))
+        results.append(ExtractedEntity("date", m.group(1).strip(), CONFIDENCE_DATE_NATURAL, m.start()))
     return results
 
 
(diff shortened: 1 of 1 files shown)


memtomem:
The call-site-based split is a reasonable choice — the PR body calls out "tuning thresholds easier" as the goal, and per-call-site constants map more directly to that than value tiers would. The 0.7 coincidence between `PERSON_MENTION`, `CONCEPT`, and `LLM_FALLBACK` is probably accidental overlap between unrelated extractors rather than a shared semantic tier, so keeping them separate leaves the optionality open to tune them independently later.

One convention fix before merge:

**Add the `_` prefix to match the rest of the module**

Every other module-level name in `entity_extraction.py` is underscore-prefixed as module-private: `_MONTHS`, `_MONTHS_PATTERN`, `_KNOWN_TECH`, `_PERSON_CONTEXT_RE`, `_MENTION_RE`, `_ISO_DATE_RE`, `_NATURAL_DATE_RE`, `_DECISION_RE`, `_ACTION_RE`, `_PASCAL_CASE_RE`, `_QUOTED_TERM_RE`, `_VALID_ENTITY_TYPES`, `_ENTITY_SYSTEM_PROMPT`. The 10 new `CONFIDENCE_*` constants would be the only names exported as public from the module, which breaks that pattern. Please rename to `_CONFIDENCE_PERSON_CONTEXT`, `_CONFIDENCE_DATE_ISO`, etc.

**Optional** — a one-line header comment that explains the grouping axis would future-proof against a reviewer later suggesting "collapse the three 0.7s into one":

```python
# ── Confidence thresholds (per call site; values may coincide) ──
_CONFIDENCE_PERSON_CONTEXT = 0.8
...
```

Not required for merge, just a nice guard. Happy to re-review once the prefix is in.

(Same CLA + CI-approval note from #160 applies here — one sign comment covers all your open PRs.)

memtomem (changes requested):
Thanks for the refactoring effort!

After reviewing, these confidence values are per-entity-type independent thresholds — e.g. `CONFIDENCE_DATE_ISO = 0.95` and `CONFIDENCE_TECH_INFERRED = 0.5` will never need to change together. This is coincidental value duplication, not semantic coupling, so extracting them into top-level constants doesn't add clarity — the values are already self-documenting at their call sites (e.g. `0.95` next to a date ISO regex is immediately clear).

I'd suggest closing this PR. If a threshold needs tuning later, making it configurable at that point is more useful than a named constant.

memtomem:
Hi @impsislegobatman — closing this one, with thanks but a different outcome from your other refactor PRs.

The review feedback on 2026-04-16 (`CHANGES_REQUESTED`) flagged the refactor axis: the `0.8 / 0.7 / 0.95 / 0.85 / 0.9 / 0.5 / ...` values across `_extract_persons`, `_extract_dates`, `_extract_decisions`, `_extract_action_items`, `_extract_tech_terms`, and `_extract_concepts` are **per-extractor independent thresholds**, not a shared concept — `CONFIDENCE_DATE_ISO = 0.95` and `CONFIDENCE_TECH_INFERRED = 0.5` will never need to move together. Hoisting them into module-level constants creates a false coupling seam that implies future maintainers should consider them as a group.

This is a judgment-call distinction from the `DECAY_FACTOR` extraction in your #158 (which we did accept independently): there the same `0.5` literal appeared twice with identical semantics (importance decay during consolidation) — genuine semantic coupling. Here, the shared shape is coincidental.

No action needed on your side. The original review is the final one on this PR — closing for housekeeping since there's been no response in 8 days. If a related refactor of just the genuinely-coupled values in this file comes up later, we'd welcome a follow-up.

Thanks for the careful eye across all four of your PRs.

PR #174, opened 2026-04-16
fix: use shared constants for input length validation

### LINKED ISSUE #151: server/tools: extract shared query/content length limits into named constants
## Problem

Input length validation limits are hardcoded in individual tool files:

| Limit | Used for | Location |
|-------|----------|----------|
| `10_000` | Question text | `server/tools/ask.py:52` |
| `10_000` | Search query  | `server/tools/search.py:61` |
| `100_000` | Memory content | `server/tools/memory_crud.py:72` |

The `10_000` query limit is duplicated across two files with no shared
constant. If one is updated without the other, users would see
inconsistent behavior between `mem_ask` and `mem_search`.

## Out of scope

`top_k` upper bounds differ by design: `mem_ask` caps at 20 (LLM
context constraint), `mem_search` at 100 (exploration use case), and
the web API at 500. These are intentional per-tool limits, not
duplication.

## Suggested fix

Define shared constants in a common location (e.g.
`server/tools/_constants.py` or `server/validation.py`) and import in
each tool file:

```python
MAX_QUERY_LENGTH = 10_000
MAX_CONTENT_LENGTH = 100_000
```

## Scope

- 1 new file or addition to existing module
- 3 file edits (import + use constant)
- No behavior change

whaiman:
Closes #151

Deduplicates hardcoded input length limits by introducing shared constants in server/validation.py and reusing them across tools.

No behavior change.


whaiman:

memtomem (changes requested):
Thanks for the PR — the constant extraction is clean and the query-length dedup is the right call.

One issue: **lint is failing** because `validation.py` is missing a trailing newline:

```diff
-MAX_CONTENT_LENGTH = 100_000
\ No newline at end of file
+MAX_CONTENT_LENGTH = 100_000
```

Once fixed, this should be good to go.

memtomem (approved):
Newline fixed, CI all green. Thanks for the quick turnaround!

PR #221, opened 2026-04-18
fix(embedding): re-register BAAI/bge-m3 via fastembed add_custom_model

memtomem:
## Summary

- fastembed 0.8.0 removed `BAAI/bge-m3` from `TextEmbedding`'s built-in catalog — every existing `mm init` install with the multilingual default currently breaks on first query (fastembed raises "unsupported model"). Verified 0.5.1 is equally affected, so a version pin is not a viable fix.
- Re-register bge-m3 via fastembed's own escape hatch (`TextEmbedding.add_custom_model`) against the official HF ONNX export: `onnx/model.onnx` + `onnx/model.onnx_data`, CLS pooling, 1024-dim normalized output. Helper is idempotent (guarded by `list_supported_models()` to avoid the hard `ValueError` fastembed raises on duplicate registration).
- Users keep the same config (`embedding.model = "bge-m3"`), the same vector dim, the same CLS pooling semantics.

## Background — why add_custom_model, not version pin

The `fastembed>=0.4` range in `pyproject.toml` has no upper bound, so any fresh install picks 0.8.x. Probing the catalog directly:

| fastembed | `BAAI/bge-m3` in `TextEmbedding` | `LateInteractionTextEmbedding` |
|---|---|---|
| 0.5.1 | ❌ | colbert-ir / answerai / jinaai only |
| 0.8.0 | ❌ | colbert-ir / answerai / jinaai only |

Pinning to 0.5.1 does not recover bge-m3 — the regression predates it. `add_custom_model` is the only path that keeps the model accessible via `TextEmbedding` without changing the user-facing model name.

## Changes

- `packages/memtomem/src/memtomem/embedding/onnx.py` — new `_register_custom_models_if_needed()` helper called from `_get_model()`; idempotent guard via `list_supported_models()` membership.
- `packages/memtomem/tests/test_onnx_custom_model.py` — new, 3 tests covering registration, idempotency, and metadata shape. Does **not** download or load the 2.3 GB ONNX export (that stays on the Ollama-gated e2e path).
- `packages/memtomem/tests/test_golden_path.py` — docstring corrected (the old "bge-m3 not supported by TextEmbedding" note is now obsolete; kept MiniLM-L12 for the CI cache-size reason only).
- `CHANGELOG.md` — one entry for the bge-m3 re-register, plus a separate note on the upstream MiniLM-L12 CLS → MEAN pooling change in fastembed 0.8.0 (cannot override without a model-name collision; users who indexed earlier should re-index).

## Smoke test (local)

```
cos(KO paraphrase):     0.8818  (high, as expected)
cos(KO ↔ EN unrelated): 0.2989
cos(KO ↔ EN unrelated): 0.3434
dim: 1024
first load:             31.3 s (2.3 GB download)
cache hit re-load:       0.55 s
```

## Test plan

- [x] `uv run ruff check packages/memtomem/src packages/memtomem/tests` — passes.
- [x] `uv run ruff format --check packages/memtomem/src packages/memtomem/tests` — passes.
- [x] `uv run pytest packages/memtomem/tests -m "not ollama"` — **1550 passed**, 46 deselected, 2 warnings. The 2 warnings are the upstream fastembed MiniLM pooling notice that CHANGELOG documents.
- [x] Manual `add_custom_model` smoke test (KO paraphrase + KO↔EN similarity, 1024-dim shape).

## Follow-ups (out of scope)

- Record `fastembed_version` in the embedding metadata so `mem_status` can warn on silent pooling drift (currently the MiniLM CLS → MEAN upstream change is only surfaced via CHANGELOG).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   responses.
 
 ### Fixed
+- **ONNX `bge-m3`**: fastembed 0.8.0 dropped `BAAI/bge-m3` from its built-in
+  `TextEmbedding` catalog — re-registered via `add_custom_model` against the
+  official HF ONNX export (1024-dim, CLS pooling, normalized). Existing
+  `mm init` users keep working with no config change.
 - **Async I/O**: 6 blocking file read/write calls in MCP tool handlers
   (`mem_add`, `mem_edit`, `mem_delete`, `mem_context_*`) wrapped with
   `asyncio.to_thread` to prevent event loop starvation.
@@ -41,6 +45,11 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   tightened (#145).
 - Dead config sections removed (`conflict`, `entity_extraction`,
   `timezone`) — `extra="ignore"` ensures old config files still load.
+- **MiniLM pooling upstream change**: fastembed 0.8.0 switched
+  `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2` from CLS to
+  mean pooling. Users who indexed with fastembed <0.5.1 and this model
+  should re-index for consistent dense-search quality; new installs are
+  unaffected.
 - Refactored truncation magic numbers in consolidation engine;
   watcher queue maxsize extracted to constant.
 - CI: ruff lint/format scope extended to `tests/`; notebooks CI job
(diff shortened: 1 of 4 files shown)


tsdata (approved):
Verified: scope matches planned design (add_custom_model with CLS pooling, 1024d, onnx/model.onnx + onnx/model.onnx_data, idempotent guard via list_supported_models). Helper placement in _get_model() is correct. Tests gate appropriately (registration-only, 2.3GB download stays on the Ollama-gated e2e path). Smoke test numbers (cos KO paraphrase 0.88, KO↔EN 0.30) confirm CLS pooling + normalization semantics. CHANGELOG accurately separates the bge-m3 fix from the upstream MiniLM pooling notice. LGTM.

PR #249, opened 2026-04-18
config: env wins over config.json + config.d/ fragment layer (fixes #248)

### LINKED ISSUE #248: config: ~/.memtomem/config.json silently overrides MEMTOMEM_* env vars, contradicting all .mcp.json docs
## Summary

`~/.memtomem/config.json` (written by `mm init`) silently overrides any `MEMTOMEM_*` env var at server startup. Every integration doc we ship tells users to configure per-client settings like `MEMTOMEM_INDEXING__MEMORY_DIRS` / `MEMTOMEM_STORAGE__SQLITE_PATH` in an `.mcp.json` / `claude_desktop_config.json` / `mcp_config.json` `env` block — but for any field persisted to `config.json` by the init wizard, those env entries are ignored. Users get the wizard-saved value regardless.

This is the **opposite** of the convention most `pydantic-settings`-based tools follow (env wins over file) and is undocumented.

## Repro

1. `mm init` → wizard writes `~/.memtomem/config.json` with `indexing.memory_dirs = ["/tmp/memories"]` (or whatever the user picked).
2. Register the server with a per-project env override, e.g.:
   ```jsonc
   // .mcp.json
   {
     "mcpServers": {
       "memtomem": {
         "command": "uvx",
         "args": ["--from", "memtomem", "memtomem-server"],
         "env": {
           "MEMTOMEM_INDEXING__MEMORY_DIRS": "[\"/Users/me/Drive/memtomem-memories\"]"
         }
       }
     }
   }
   ```
3. Restart the MCP client, call `mem_status` → the server reports `/tmp/memories`, not the Drive path. The env override was silently dropped.

## Root cause

Startup sequence — `packages/memtomem/src/memtomem/server/component_factory.py:41-47`:

```python
config = config or Mem2MemConfig()      # (1) pydantic-settings binds MEMTOMEM_* env
load_config_overrides(config)           # (2) ~/.memtomem/config.json setattr()s on top
ensure_auto_discovered_dirs(config)     # (3) append-only auto-discovery
```

`load_config_overrides` (`packages/memtomem/src/memtomem/config.py:542-588`) iterates every section/key in `config.json` and unconditionally `setattr(section_obj, key, value)`, so any field present in the file clobbers the env-var-bound value. There is no "env wins" path.

## Impact — docs vs reality

Env-var overrides in MCP `env` blocks are the **documented** way to configure per-client paths. Pages that currently promise this works:

- `docs/guides/mcp-clients.md` — Cursor / Windsurf / Claude Desktop / Gemini CLI / Antigravity sections all show `MEMTOMEM_INDEXING__MEMORY_DIRS` examples
- `docs/guides/integrations/claude-desktop.md:41-62` — both macOS and Windows blocks
- `docs/guides/integrations/claude-code.md` (after #247) — `.mcp.json` example
- `docs/guides/getting-started.md:180,195`
- `docs/guides/cloud-sync.md:8` — explicitly tells users to point `MEMTOMEM_INDEXING__MEMORY_DIRS` at a cloud-synced folder
- `docs/guides/use-cases.md:54,74`

For any user who ran `mm init` first (the recommended path), **none of these examples work as documented** for the fields the wizard persists (`memory_dirs`, `sqlite_path`, `embedding.*`, etc.).

Two real user scenarios this breaks:

- **Per-project memory_dirs**: users with multiple MCP clients pointed at the same `memtomem-server` with different `.mcp.json` env blocks — all collapse to the wizard's single path.
- **Cloud-sync onboarding** (`cloud-sync.md`): the doc tells users to set `MEMTOMEM_INDEXING__MEMORY_DIRS` to their Drive folder; post-`mm init` this is a no-op.

Separately but related: the bare-string-vs-JSON-array gotcha for `list[Path]` env vars (`mcp-clients.md:274-278`) becomes a secondary footgun — users who notice their env has no effect will try to "fix" it by unquoting the array, and then get a pydantic-settings crash instead of the same silent override.

## Options for resolution

Not prescribing a fix — listing the alternatives for discussion:

1. **Invert precedence: env wins over `config.json`.** Most intuitive, matches pydantic-settings default semantics. Breaking change for anyone currently relying on `config.json` as the authoritative source after a `mm config set` followed by a stale env var — but in practice that combo seems rare (env vars in `.mcp.json` are long-lived; users don't typically leave stale ones around).
2. **Only persist wizard fields to `config.json` if no matching env var is set** — shifts the precedence decision to write-time instead of load-time. Less invasive at runtime but makes `mm init` non-deterministic (depends on env at wizard time).
3. **Merge semantics for list fields** (`memory_dirs` = env ∪ config.json). Solves the per-project case but adds a new mental model users have to learn. Doesn't help scalar fields like `sqlite_path`.
4. **Keep current semantics, document them loudly.** Update every `MEMTOMEM_*` env-var example in docs with a "⚠ only takes effect if this key is absent from `~/.memtomem/config.json`" warning, and add a `mm config unset <key>` CLI for users who want to fall back to env. Minimal code change, maximal docs churn.
5. **Config isolation by scope** — support `MEMTOMEM_CONFIG_PATH` env var so per-project `.mcp.json` can point at a project-local `config.json`. Enables real per-project configs but grows surface area.

Any of 1–3 also needs a migration / release-notes entry since `mm init`-driven setups would change behavior.

## Acceptance criteria (once a direction is picked)

- [ ] Behavior matches what docs claim (or docs are updated to match behavior, explicitly and per-page)
- [ ] `tests/test_config.py` (or a new test) pins the chosen precedence so it can't silently flip again
- [ ] `mcp-clients.md` / `cloud-sync.md` / `integrations/*.md` / `getting-started.md` examples are verified end-to-end against the chosen behavior before PR merge (docs-as-tests)

memtomem:
Closes #248.

## Summary

Resolves the precedence bug where `~/.memtomem/config.json` silently
overrode `MEMTOMEM_*` env vars set in `.mcp.json` / MCP client `env`
blocks — i.e. every example our docs ship. Also adds the `config.d/`
drop-in layer discussed in #248 so integration installers can drop a
fragment without mutating shared state.

Five commits, each independently reviewable:

1. **Phase 1 — `fix(config): env vars win over ~/.memtomem/config.json`** — `load_config_overrides` now checks whether the matching `MEMTOMEM_<SECTION>__<FIELD>` env var is set and skips the override if so. Regression test pins PR #247's exact `.mcp.json` env block (1 scalar `SQLITE_PATH` + 1 list `MEMORY_DIRS`) — both must survive a full `mm init` + `config.json`.
2. **Phase 2a — `feat(config): add MergeStrategy annotations to list fields`** — co-locates the merge strategy with the field via `Annotated[list[X], APPEND | REPLACE]`. 4 APPEND (memory_dirs, exclude_patterns, system_namespace_prefixes, webhook.events), 2 REPLACE (rrf_weights, importance.weights) — REPLACE on positional tuning knobs where appending would misalign element slots. No runtime change on its own.
3. **Phase 2b — `feat(config): add ~/.memtomem/config.d/ fragment loader`** — scans `config.d/*.json` in lex order, respects per-field strategy, skips fields where env is set. `config.json` remains the REPLACE-on-set user-override layer (`mm init` UX unchanged).
4. **`test(config): enforce merge strategy declaration on all list fields`** — walks every `Mem2MemConfig` section and fails if a `list[*]` lacks a `MergeStrategy` in its metadata. Guards against silent "default-to-REPLACE" drift when new fields get added.
5. **`docs(config): document precedence and list merge strategies`** — adds a "Precedence and merge behaviour" section to `docs/guides/configuration.md` covering the full resolution order and the APPEND/REPLACE table. Other integration docs already advertised the env-wins semantic, so they now work as written.

## Resolution order (lowest → highest priority)

```
defaults → config.d/*.json (lex order, per-field strategy)
         → config.json (REPLACE for every field)
         → MEMTOMEM_* env vars
```

## Testing

- 14 new tests in `test_config_overrides.py` covering Phase 1 precedence, Phase 2b merge semantics (APPEND dedup, REPLACE last-wins, scalar last-wins, env-wins-over-fragments, unknown section warn, malformed JSON warn, non-JSON files ignored), and the enforcement test.
- Full suite: 1633 passed, 46 deselected (ollama marker), 0 failures.
- `ruff check` and `ruff format --check` both clean.
- `mypy` clean on the touched files.

## Release notes

> **Breaking:** `MEMTOMEM_*` environment variables now take precedence over `~/.memtomem/config.json`. Previously `config.json` silently clobbered env overrides set in `.mcp.json` blocks. Any setup that was actively relying on file-wins-over-env behaviour will see env take effect — delete the env var or remove the field from `config.json` to restore the old outcome.

> **New:** `~/.memtomem/config.d/*.json` drop-in fragments are now loaded at startup. Integration installers can drop a per-client fragment (e.g. `claude-desktop.json`) instead of mutating the shared `config.json`; removing the file reverses the change.

## Test plan

- [x] Unit tests: precedence + merge semantics
- [x] Regression test pinning #247's advertised `.mcp.json` env block
- [x] Enforcement test on list-field metadata
- [ ] Manual smoke: restart memtomem-server with a `config.d/test.json` + overlapping `config.json` + env, confirm resolution matches the documented order end-to-end

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -12,6 +12,43 @@ export MEMTOMEM_EMBEDDING__API_KEY=sk-...
 
 For interactive setup, run `mm init` instead of editing env vars by hand.
 
+## Precedence and merge behaviour
+
+memtomem resolves each field from up to four sources at startup, in order
+of increasing priority:
+
+1. **Built-in defaults** — the values in `config.py`.
+2. **`~/.memtomem/config.d/*.json`** — drop-in fragments, applied in
+   lexicographic filename order. Intended for integration installers
+   (`mm init <client>` drops one fragment; removing the file reverses
+   the change). For `list[*]` fields, each fragment respects a per-field
+   merge strategy (see below).
+3. **`~/.memtomem/config.json`** — the user-managed override layer that
+   `mm init` writes to. Every key here replaces whatever earlier layers
+   produced for that field (REPLACE semantics across the board).
+4. **`MEMTOMEM_*` environment variables** — highest priority. If an
+   env var is set, the corresponding entries in `config.d/` and
+   `config.json` are skipped.
+
+### List field merge strategies
+
+`list[*]` fields declare either `APPEND` or `REPLACE` in the type
+annotation, and that strategy governs how `config.d/` fragments layer on
+top of the default:
+
+| Field | Strategy | Notes |
+|-------|----------|-------|
+| `indexing.memory_dirs` | APPEND | Each fragment contributes more roots, dedup by path string |
+| `indexing.exclude_patterns` | APPEND | Multiple denylists merge cleanly |
+| `search.system_namespace_prefixes` | APPEND | Integrations can add further hidden namespaces on top of `archive:` |
+| `webhook.events` | APPEND | Fragments can subscribe to additional event types |
+| `search.rrf_weights` | REPLACE | Positional tuning knob — appending would misalign `[BM25, Dense]` slots |
+| `importance.weights` | REPLACE | Same positional constraint |
+
+`config.json` always replaces, regardless of strategy — it's the
+explicit-user-override layer. Use a fragment in `config.d/` if you want
+APPEND semantics.
+
 ## Storage
 
 | Variable | Default | Description |
@@ -140,10 +177,11 @@ AI tool directories to `memory_dirs` when they exist on the machine:
 | `~/.gemini` | Gemini CLI | global `GEMINI.md` |
 | `~/.codex/memories` | Codex CLI | global memories |
 
-Auto-discovered directories are appended after `config.json` overrides, so
-they are always available even if you override `memory_dirs` manually. This
-means `mem_index` and the file watcher accept paths under these directories
-without requiring explicit `MEMORY_DIRS` configuration.
+Auto-discovered directories are appended after `config.d/` and
+`config.json` overrides, so they are always available even if you
+override `memory_dirs` manually. This means `mem_index` and the file
+watcher accept paths under these directories without requiring explicit
+`MEMORY_DIRS` configuration.
 
 > **Tip:** Use `mm ingest claude-memory`, `mm ingest gemini-memory`, or
 > `mm ingest codex-memory` for richer ingestion with per-tool tagging and
--- a/packages/memtomem/src/memtomem/cli/config_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/config_cmd.py
@@ -23,9 +23,10 @@ def config() -> None:
 @click.option("--format", "fmt", type=click.Choice(["table", "json"]), default="table")
 def config_show(fmt: str) -> None:
     """Show current configuration (API keys masked)."""
-    from memtomem.config import Mem2MemConfig, load_config_overrides
+    from memtomem.config import Mem2MemConfig, load_config_d, load_config_overrides
 
     cfg = Mem2MemConfig()
+    load_config_d(cfg)
     load_config_overrides(cfg)
     data = cfg.model_dump()
 
--- a/packages/memtomem/src/memtomem/cli/embedding_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/embedding_cmd.py
@@ -27,10 +27,11 @@ def embedding_reset(mode: str) -> None:
 
 
 async def _run(mode: str) -> None:
-    from memtomem.config import Mem2MemConfig, load_config_overrides
+    from memtomem.config import Mem2MemConfig, load_config_d, load_config_overrides
     from memtomem.storage.sqlite_backend import SqliteBackend
 
     cfg = Mem2MemConfig()
+    load_config_d(cfg)
     load_config_overrides(cfg)
 
     storage = SqliteBackend(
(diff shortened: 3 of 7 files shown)


memtomem:
## Manual smoke — PASS

3 scenarios with `HOME=/tmp/mm-smoke-…`, `config.json` + `config.d/claude-desktop.json`, exercised via `mm config show`:

**S1** — no env, both layers:
- `sqlite_path = /from/cfgjson.db` ✓ (config.json only touches)
- `rrf_weights = [0.9, 0.1]` ✓ (config.json REPLACE beats fragment `[0.5, 0.5]`)
- `memory_dirs = ['/from/cfgjson']` ✓ (config.json REPLACE wipes fragment's `/from/fragment`)
- `exclude_patterns = ['*.tmp']` ✓ (fragment APPEND; config.json silent)

**S2** — `MEMTOMEM_STORAGE__SQLITE_PATH=/from/env.db` set:
- `sqlite_path = /from/env.db` ✓ (env wins)
- Other fields unchanged from S1 ✓

**S3** — config.json deleted, fragment only:
- `rrf_weights = [0.5, 0.5]` ✓ (fragment REPLACE over default `[1.0, 1.0]`)
- `memory_dirs = [~/.memtomem/memories, /from/fragment]` ✓ (fragment APPEND to default)
- `exclude_patterns = ['*.tmp']` ✓ (fragment APPEND to empty default)

Followup commit `fix(cli): load config.d fragments in read paths` — `mm config show` / `reset` / `embedding-reset` were only calling `load_config_overrides`, so fragments were invisible to the CLI even though the server saw them. `config set` left alone intentionally (see commit body — avoids baking fragment state into `config.json` via `save_config_overrides`' read-merge-write).

Test-plan checkbox for manual smoke now checked.

PR #267, opened 2026-04-19
fix(web): hot-reload config.json + extend write lock to all mutators

memtomem:
## Summary

Fix the save-overwrite regression where `mm config set` edits on disk are
invisible to the running Web UI and get silently clobbered on the next UI
save. Also closes a pre-existing locking gap where only PATCH was serialized.

- Re-read `~/.memtomem/config.json` + `config.d/*.json` on every `GET /api/config`
  and at the top of every config-writing endpoint (PATCH, POST /config/save,
  POST /memory-dirs/add, POST /memory-dirs/remove). Detection is a cheap
  `os.stat` signature — no watchdog Observer needed.
- Rename `_config_patch_lock` → `_config_lock` and extend it from PATCH to
  all four write handlers (closes a pre-existing gap).
- Fail-closed on broken disk: keep last-known-good config, surface
  `config_reload_error` on `GET /api/config`, refuse writes with HTTP 409.
  Never overwrite a broken disk state — that would destroy the recovery trail.
- Extract tokenizer+cache fanout into \`apply_runtime_config_changes\` shared
  by the PATCH handler and the reload path, so a disk-triggered tokenizer
  change fires the same FTS rebuild + cache invalidation as an in-process
  PATCH.
- FE: \`settings-config.js\` shows a banner when mtime changes or when disk is
  invalid, and re-fetches on \`visibilitychange\` so CLI edits made while the
  tab was hidden become visible on focus. No WebSocket/polling — single-user
  local tool.

### Scope disclosure

- **Covered**: all 4 web config write paths (PATCH, /config/save, /memory-dirs/add,
  /memory-dirs/remove) + GET read-through. CLI writes go through
  \`save_config_overrides\`'s own read-merge-write path (PR #258), so they
  were already safe from this specific race.
- **Not covered (V1 explicit non-goals, documented in CHANGELOG/docs)**:
  auto-reindex on external \`memory_dirs\` changes, FS watcher push-to-browser,
  multi-tab sync (single-user tool).

### Why strict pre-parse in hot_reload

\`load_config_overrides\` historically logs-and-returns on JSON parse errors
(startup wanted to boot with defaults rather than crash on a bad file). That
silent fallback breaks fail-closed: a truncated \`config.json\` would be read
as "empty overrides", the UI would swap to a default config, and the next
save would write the defaults back over whatever the user was trying to fix.
So \`_build_fresh_config\` pre-parses \`config.json\` strictly before delegating.

## Test plan

- [x] \`uv run ruff check packages/memtomem/src\`
- [x] \`uv run ruff format --check packages/memtomem/src\`
- [x] \`uv run mypy packages/memtomem/src/memtomem/web/hot_reload.py packages/memtomem/src/memtomem/web/routes/system.py\`
- [x] \`uv run pytest packages/memtomem/tests -m \"not ollama\"\` → **1735 passed, 46 deselected**
- [x] 20 new tests in \`test_web_hot_reload.py\` cover:
  1. read-through on stale GET
  2. PATCH re-reads before merge
  3. all 4 write handlers honor reload
  4. \`config.d/\` fragment change detection
  5. invalid JSON → 200 with \`config_reload_error\` + stale config preserved
  6. post-fix reload clears the error
  7. tokenizer change via reload triggers fanout (FTS rebuild + cache invalidate)
  8. concurrent PATCH + disk edit are serialised by the lock
  - plus \`current_signature\` / \`reload_if_stale\` / \`apply_runtime_config_changes\` unit tests + a lock-rename regression guard.
- [ ] Manual smoke: start web UI, \`mm config set mmr.enabled true\` in a
      terminal, refocus browser tab, confirm banner + updated value.
- [ ] Manual smoke: write \`{\"search\":\` (truncated) to \`config.json\`, confirm
      Config tab shows red banner and Save returns 409.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,16 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   `config.d/` fragment.
 
 ### Fixed
+- **Web UI config hot-reload**: `~/.memtomem/config.json` and `config.d/*.json`
+  are re-read on every `GET /api/config` and at the top of every config-
+  writing endpoint (`PATCH /api/config`, `POST /api/config/save`,
+  `POST /api/memory-dirs/add|remove`). Previously, external edits
+  (`mm config set`, manual editor) were invisible to the running server
+  and got silently clobbered on the next UI save. The writer lock was
+  extended from PATCH to all four write handlers (closing a pre-existing
+  gap). If `config.json` becomes invalid on disk, the UI keeps the last
+  known-good config, surfaces `config_reload_error` in the response, and
+  refuses writes with HTTP 409 until the file is fixed.
 - **ONNX `bge-m3`**: fastembed 0.8.0 dropped `BAAI/bge-m3` from its built-in
   `TextEmbedding` catalog — re-registered via `add_custom_model` against the
   official HF ONNX export (1024-dim, CLS pooling, normalized). Existing
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -49,6 +49,28 @@ top of the default:
 explicit-user-override layer. Use a fragment in `config.d/` if you want
 APPEND semantics.
 
+### External edits while the Web UI is running
+
+The Web UI server re-reads `config.json` and `config.d/*.json` on every
+`GET /api/config` and at the top of every config-writing endpoint
+(`PATCH /api/config`, `POST /api/config/save`, `POST /api/memory-dirs/add`,
+`POST /api/memory-dirs/remove`). This means:
+
+- `mm config set ...` or a manual editor save while the server is running
+  becomes visible on the next UI interaction (or when the tab regains
+  focus), without a restart.
+- A subsequent UI save merges against the *current* disk state rather
+  than overwriting the external change with a stale in-memory copy.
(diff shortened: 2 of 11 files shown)


memtomem:
Follow-up to review comments:

In-PR (commit 4e4dbbe):
- [1] Test 8 now actually proves lock serialisation via (enter, exit) span instrumentation on `save_config_overrides`.
- [2] `commit_writer_signature(app)` public helper extracted; four `_set_last_signature` cross-module calls replaced.
- [5] Added `test_non_json_structural_error_also_surfaces` covering the non-JSONDecodeError failure mode.
- [6] Docstring clarifications: FTS rebuild out-of-band vs `_config_lock` (different SQLite objects, no file-level race), and the error-clear condition (mtime change OR vanish).

Deferred to follow-up (#268):
- [Medium] V2 race: GET's lock-free `reload_if_stale` can overwrite a concurrent writer's signature bump. 3-line compare-and-swap fix + 50-line test. V1 tradeoff documented in the module docstring per the bridge memo D3.

Skipped:
- [4] `_stat_mtime_ns` OSError branch coverage — defensive path, defer per `feedback_defensive_noise.md`.

All CI checks green after 4e4dbbe: lint / test / typecheck / golden-path / notebooks / CLA ✅. 1736 passed (+1 new), 0 regressions.

PR #269, opened 2026-04-19
fix(web): eliminate GET-path signature overwrite race in hot_reload

### LINKED ISSUE #268: web: eliminate GET-path signature-overwrite race in hot_reload (V2 of #267)
## Context

PR #267 shipped Web UI config hot-reload with D3 "readers are lock-free,
writers hold `_config_lock`" per the bridge memo. PR review flagged a
medium-severity race in the lock-free GET path that was not in the
original 5 design decisions:

```
1. GET: sig_current = T1, enters _build_fresh_config (file I/O, ~ms)
2. Writer: acquires _config_lock, reloads, merges, saves at T2,
   calls commit_writer_signature() → signature = T2
3. GET: returns from _build_fresh_config with cfg_T1_or_T2, executes
   app.state.config = ... and _set_last_signature(T1) → reverts
   the writer's signature bump
4. Next GET: sees sig mismatch → spurious re-reload
```

**Worst case**: transient stale in-memory view for one request, disk
remains authoritative. No data loss, no config corruption — just an
extra file-read on the next GET.

For a single-user local tool the V1 tradeoff is acceptable (and
explicitly documented in `hot_reload.py`'s module docstring). This
issue tracks the V2 fix.

## Proposed fix (3 lines in `reload_if_stale`)

Add a compare-and-swap check right before the mutation:

```python
# Re-check signature before committing the swap; another writer may
# have advanced state.config between our _build_fresh_config() entry
# and return.
if _get_last_signature(app) != last:
    # Someone else reloaded while we were building; their view is
    # at least as fresh as ours. Discard our rebuild.
    return False

app.state.config = new_cfg
_set_last_signature(app, sig)
```

## Test design

Reuse the monkeypatch pattern from `test_concurrent_patches_are_
serialised_by_lock` (#267 Test 8), but instrument `_build_fresh_config`
instead of `save_config_overrides`:

1. Slow a GET's `_build_fresh_config` via `time.sleep(0.05)`
2. Fire a PATCH concurrently that completes within the sleep window
3. Assert `app.state.config_signature` after both = writer's signature
   (not the reader's). Without the fix, the reader's assignment would
   have overwritten the writer's bump.

Scope: +50 lines test, +3 lines prod, no API change.

## Not a blocker for V1

PR #267 merged with the race acknowledged. This issue is enhancement-
labeled and can ship any time; nothing depends on it.

## References

- PR #267 — V1 implementation
- `packages/memtomem/src/memtomem/web/hot_reload.py:182-207` — current
  `reload_if_stale` docstring (D3 tradeoff documented there)
- Bridge memo `project_web_hot_reload_bridge.md` — D3 "readers tolerate
  stale view for one request"

memtomem:
## Summary

V2 of PR #267. Closes #268.

The hot-reload read path in `reload_if_stale` is intentionally lock-free
per bridge memo D3 ("readers tolerate stale view for one request").
PR #267 review identified a medium-severity race that goes beyond that
tradeoff: the reader's trailing state mutation could *overwrite* a
concurrent writer's signature bump, not merely serve stale data.

### The race (from #268)

1. GET: reads `sig = current_signature()` → T1, enters `_build_fresh_config` (file I/O, ~ms).
2. Writer: acquires `_config_lock`, reloads + merges + saves at T2, calls `commit_writer_signature()` → signature = T2.
3. GET: returns from `_build_fresh_config` with a cfg built from T1 disk, blindly assigns `app.state.config = that_cfg` and `_set_last_signature(T1)` — reverts the writer's bump to T2.
4. Next GET: sig mismatch → spurious re-reload from identical disk.

Worst case was one wasted file read on the next GET. No corruption,
but a violation of D3's "tolerate one-request staleness" contract
(the reader wasn't tolerating — it was actively regressing shared state).

### The fix (3 lines)

In `reload_if_stale`, after `_build_fresh_config` returns and before
the state mutation:

```python
# Compare-and-swap: while we were in _build_fresh_config (file I/O, can
# take milliseconds), a writer inside _config_lock may have already
# committed a fresh reload + signature bump via commit_writer_signature.
# That view is at least as fresh as ours; discard our rebuild so we
# don't revert the writer's signature and force a spurious next-GET
# reload. Race eliminated per issue #268.
if _get_last_signature(app) != last:
    return False
```

## Test plan

- [x] `uv run ruff check packages/memtomem`
- [x] `uv run ruff format --check packages/memtomem`
- [x] `uv run mypy packages/memtomem/src/memtomem/web/hot_reload.py`
- [x] `uv run pytest packages/memtomem/tests -m "not ollama"` → **1737 passed (+1 new), 46 deselected**
- [x] New `test_reload_if_stale_cas_yields_to_concurrent_writer_bump`:
  - Unit-level (no HTTP, no asyncio) so the race window is deterministic.
  - Monkeypatches `_build_fresh_config` to mutate `app.state.config_signature` via `commit_writer_signature` *between* call and return, simulating "writer committed while reader was in file I/O."
  - Asserts writer's bump survives + reader doesn't swap `app.state.config`.
- [x] **Negative-tested**: removing the CAS line flips the test from pass to fail, so it's a real regression guard not a no-op.

## Scope disclosure

- **Covered**: the GET path's CAS. PATCH/save/memory-dirs handlers run inside `_config_lock` so they don't need CAS — the lock serialises writer-vs-writer.
- **Not covered (explicit non-goals, same as V1)**: FS watcher push, multi-tab FE sync, auto-reindex on memory_dirs change.

### DIFF
--- a/packages/memtomem/src/memtomem/web/hot_reload.py
+++ b/packages/memtomem/src/memtomem/web/hot_reload.py
@@ -240,6 +240,15 @@ def reload_if_stale(
         _set_last_signature(app, sig)
         return False
 
+    # Compare-and-swap: while we were in _build_fresh_config (file I/O, can
+    # take milliseconds), a writer inside _config_lock may have already
+    # committed a fresh reload + signature bump via commit_writer_signature.
+    # That view is at least as fresh as ours; discard our rebuild so we
+    # don't revert the writer's signature and force a spurious next-GET
+    # reload. Race eliminated per issue #268.
+    if _get_last_signature(app) != last:
+        return False
+
     old_cfg = getattr(app.state, "config", None)
     app.state.config = new_cfg
     _set_last_signature(app, sig)
--- a/packages/memtomem/tests/test_web_hot_reload.py
+++ b/packages/memtomem/tests/test_web_hot_reload.py
@@ -15,6 +15,8 @@ Covers (numbering matches ``project_web_hot_reload_bridge.md`` test plan):
 6. after fix → GET clears the error
 7. tokenizer change via reload triggers fanout (FTS rebuild + cache invalidate)
 8. concurrent PATCH + disk edit: lock serialisation
+9. V2 (issue #268): GET's lock-free reload yields to a concurrent writer's
+   signature bump via compare-and-swap
 """
 
 from __future__ import annotations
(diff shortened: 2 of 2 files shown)


memtomem (commented):
## Overview
Closes the GET-path signature-overwrite race discovered after #267 with a 3-line CAS. Matches issue #268's proposed fix exactly and preserves the D3 invariant ("readers are lock-free, writers hold `_config_lock`").

## Strengths
- **Minimum scope**: +3 prod / +73 test, 1:1 with the issue proposal.
- **Correct CAS placement** (`hot_reload.py:249`) — right after `_build_fresh_config` succeeds and right before the commit; exactly at the tail end of the race window.
- **Deterministic unit test**: no HTTP / no asyncio, uses `monkeypatch` to interleave the writer from inside `_build_fresh_config`. Scheduler-independent, fixed race window (as the Test 9 docstring spells out).
- **Strong WHY comment** at `hot_reload.py:243-248`: 5 lines cover the scenario, the yield rationale, and the issue reference. A future reviewer won't need `git blame` to understand "why the `return False` here".
- **CI fully green**: lint / typecheck / test / golden-path / notebooks all SUCCESS.

## Observations

### 1. PR body is stored as the literal `$(gh pr view 269 --json body -q .body)` (recommend fixing immediately)
GitHub renders the PR description as the raw shell-template string — the shell substitution was saved unescaped, so the intended content (issue #268 context / 3-line diff / test design) never made it in.
```bash
gh pr edit 269 --body-file <actual-body.md>
```
Worth fixing before merge — this also risks ending up in the squash commit body.

### 2. Mirror race in the `except` branch (follow-up candidate, out of scope for this PR)
The success path is now CAS-protected, but the same race shape still lives in the error path:
```python
# hot_reload.py:240
_set_last_signature(app, sig)  # can overwrite a writer's bump with sig(T1)
return False
```
Scenario (disk transitioning broken → valid):
1. Reader enters, snapshots sig_T1 (disk broken)
2. Writer repairs disk + `commit_writer_signature` → sig_T2
3. Reader's `_build_fresh_config` still fails (its read raced the broken state)
4. Error branch sets `_set_last_signature(app, sig_T1)` → writer's T2 bump is lost

Worst case matches the original V1 issue (one spurious reload, no data loss). Fine to keep out of this PR to respect the stated scope — a follow-up issue + a single extra CAS line would close it symmetrically.

### 3. Narrow window between CAS check and assignment (acceptable, noting only)
```python
if _get_last_signature(app) != last:   # (A)
    return False
# ← theoretically a writer could land here
app.state.config = new_cfg             # (B)
_set_last_signature(app, sig)          # (C)  sig = at-entry snapshot
```
Under the GIL, (A)→(C) spans only a handful of bytecodes — a nanosecond-scale window. Closing it would require the reader to briefly acquire `_config_lock`, which directly contradicts the "lock-free readers" design. The current tradeoff is the right one.

### 4. Merge state: `BEHIND main` (rebase needed, no conflicts)
`mergeable: MERGEABLE`, so a fast-forward rebase is sufficient.

### 5. Minor nits
- The module-level docstring at `test_web_hot_reload.py:8-18` still lists only tests 1–8; the new Test 9 isn't reflected. Easy to slip into a future PR.
- `from fastapi import FastAPI` inside the new test function is fine in isolation, but if more tests start needing it, lifting it to the top-level import block keeps the file consistent.

## Verdict
A focused fix that closes the reported race while preserving the design invariant. Mergeable once the PR body is restored (and optionally the error-path follow-up issue is filed). LGTM.

memtomem:
리뷰 감사합니다. 처리 내역:

**In-PR (commit 8db3ac6, force-pushed)**:
- [1] PR body 복구 — `gh pr edit --body-file` 로 파일 전달 방식으로 고침 (shell substitution escape 실패가 원인).
- [5a] `test_web_hot_reload.py` 모듈 docstring 에 Test 9 항목 추가 (numbering 1-9).
- [4] `origin/main` 위로 rebase 완료. #272 (reset button) + #271 (exclude guard) 결과 1742 passed / 0 regression.

**Follow-up issue**:
- [2] Error-branch mirror race → **#273** 로 분리. 같은 3-line CAS 패턴을 error path 에 mirror 하는 작업. `feedback_one_change_per_pr.md` 준수를 위해 분리, #269 scope 는 success-path 만.

**No change (의견 수용)**:
- [3] CAS-check-vs-assignment ns-window — 잔여 위험 인정, lock-free reader design 을 바꾸지 않는 한 닫을 수 없는 범위. 현 선택 유지.
- [5b] Test 9 내부 `from fastapi import FastAPI` — 현재 이 테스트 1곳만 사용하므로 함수 내부 import 유지. 후속 PR 에서 다른 test 가 같은 import 를 쓰면 그때 top-level 로 승격.

CI 재실행 대기 중.

PR #294, opened 2026-04-19
fix: update pyproject.toml

powerzist:
### Problem

Running Ruff formatting in bulk with commands such as `ruff format .` can also affect files that do not need to be reformatted.

Although formatting can be applied selectively, using a single `ruff format .` command is much faster and more convenient.

### Change

This PR adds an exclusion rule to `pyproject.toml` so that unnecessary files are automatically excluded from Ruff formatting. At the moment, this applies to `*.ipynb`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,8 @@ memtomem = { workspace = true }
 [tool.ruff]
 target-version = "py312"
 line-length = 100
+extend-exclude = ["*.ipynb"]
+force-exclude = true
 
 [tool.ruff.lint.per-file-ignores]
 "packages/memtomem/src/memtomem/server/__init__.py" = ["E402", "F401"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ memtomem = { workspace = true }
 target-version = "py312"
 line-length = 100
 extend-exclude = ["*.ipynb"]
-force-exclude = true
 
 [tool.ruff.lint.per-file-ignores]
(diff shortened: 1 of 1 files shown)


memtomem (changes requested):
Thanks @powerzist — the motivation is real (verified locally: ruff 0.15 picks up `examples/notebooks/01_hello_memory.ipynb` in `ruff format .`), and CI isn't affected since CI passes explicit paths.

One small ask: could we drop `force-exclude = true`? `extend-exclude` alone already achieves the stated goal of skipping `*.ipynb` during `ruff format .` bulk discovery. `force-exclude = true` adds a stronger behavior — it also blocks explicit invocations like `ruff format examples/notebooks/01_hello_memory.ipynb`, which means anyone who *wants* to format a specific notebook can no longer do so without overriding config. That's a broader policy than the PR body argues for; the bulk-format motivation alone doesn't quite reach there.

Suggested final state:

```toml
[tool.ruff]
target-version = "py312"
line-length = 100
extend-exclude = ["*.ipynb"]
```

If there's a specific reason explicit invocations should also be blocked (e.g. you want pre-commit hooks to skip ipynb even when given the path directly), happy to keep `force-exclude = true` — just add a one-line note to the PR body so the rationale is on record. Either way is fine.

powerzist:
Of course — no problem at all. I removed the `force-exclude = true` option as requested.

There was no special reason for adding `force-exclude = true`; I had only included it as an extra safeguard in case Ruff formatting was applied to those files by mistake.

memtomem (approved):
LGTM — thanks for the quick turnaround!

Issue #331, opened 2026-04-20
feat(cli): add `--format [table|json]` to session list/events and activity log


memtomem:
## Problem

`mm search`, `mm recall`, and `mm config show` all support `--format [table|json|...]` for scripting. Three sibling commands do not:

```
$ mm session list --help
Options:
  -a, --agent-id TEXT
  --since TEXT
  -l, --limit INTEGER
  -h, --help
```

No `--format` / `--json`. Same for `mm session events` and `mm activity log`. Callers must parse text output.

Affected commands:

- `mm session list`
- `mm session events`
- `mm activity log`

## Proposed fix

Add `--format [table|json]` to each of the three commands, following the existing `search` / `recall` precedent. Default remains `table` — no behavior change for existing callers.

Files to touch: `packages/memtomem/src/memtomem/cli/session_cmd.py` for all three commands.

## Out of scope

- Unifying `--format` vs `--json` convention across the CLI (#332).
- Adding JSON output to `watchdog` (already has `--json` flag) or other subcommands not listed above.

## Test plan

Per command, add a `--format json` test that parses stdout with `json.loads` and asserts required keys exist (matching the table output's column set where applicable).

memtomem:
Scope narrowed after implementation review: `mm activity log` is a write-only silent command (intentional for hook callers — must not fail or emit noise), so adding `--json` to it first requires deciding what "success" output looks like. That's a design question, not a parity addition.

Splitting it out as #335 so this issue stays focused on the pure read-side parity for `session list` / `session events` — both of which already have structured table output and land cleanly under the #332 rule.

PR #337, opened 2026-04-21
feat(cli): add `--json` ack to `mm activity log`

### LINKED ISSUE #335: mm activity log — define success output shape (silent ack vs `{ok, event_id}`)
## Problem

`mm activity log` currently writes an event to the active session and produces **no output** — success is silent, "no active session" silently returns, and write failures are logged as warnings but still produce no stdout. This is intentional (`activity log` is called from hooks, which must not fail even when no session is active), but it leaves two open questions:

1. Is the silent-success behavior actually the desired contract? From a hook author's perspective, no output is fine; from a scripter's perspective, they can't confirm the event landed.
2. If we ever want JSON output for `activity log` (for symmetry with `session list`/`events` after #331), we first need to decide what "success" looks like.

Split from #331 — that issue covers read-side parity (`session list`, `session events`), which is a pure `--json` flag addition with no shape question. `activity log` is write-side and needs the shape decided first.

## Options for success output

**A. Keep silent by default, opt-in verbose flag**
```
$ mm activity log --type tool_call -c "ran tests"
# (no output, exit 0)

$ mm activity log --type tool_call -c "ran tests" --verbose
event logged: <event_id> session=<sid>
```
No breaking change for hook callers. `--verbose` gives humans a confirmation line.

**B. Default silent, add `--json` returning ack shape**
```
$ mm activity log --type tool_call -c "ran tests" --json
{"ok": true, "event_id": "...", "session_id": "...", "event_type": "tool_call"}

$ mm activity log --type tool_call -c "x" --json     # no active session
{"ok": false, "reason": "no_active_session"}
```
Consistent with the `--json` rule from #332 (binary human/machine scenario). "No active session" becomes a parseable error shape, matching `session events --json`'s treatment of the same class of error.

**C. Do nothing — the silent contract is correct and wanted**
Close this issue. Hook authors don't need output; scripters who want confirmation can `mm session events --json | jq` after the fact.

## Non-goals

- Changing the silent-on-no-session behavior (hooks depend on it).
- Changing the warning-only behavior for write failures (hooks must not fail hard).

## Recommendation

Lean toward **B** once `session events --json` lands via #331 — both are write-adjacent read-backs and share a "no active session" failure mode, so the shapes should match. But the value here is low compared to #331's read-side parity, so this is worth parking until there's a concrete caller who wants it.

memtomem:
## Summary

Follow-up to #331. `activity log` was split out of that issue (which shipped in #336 for the read-side commands) because it's write-side and needed a shape decision before implementation. This closes #335 by adopting **option B** from the issue body — silent by default, `--json` emits an ack with an explicit `ok` discriminator.

## Shape

| scenario           | output                                                         |
|--------------------|----------------------------------------------------------------|
| success            | `{"ok": true, "session_id": ..., "event_type": ...}`           |
| no active session  | `{"ok": false, "reason": "no_active_session"}`                 |
| write failure      | `{"ok": false, "reason": "write_failed"}`                      |

Exit code is always 0. The silent-by-default hook contract is preserved when `--json` is not set, so existing hook callers keep working unchanged.

## Shape choice vs `session events --json`

`session events --json` uses `{"error": "no_session"}` (no `ok` field). This PR deliberately does **not** match that shape:

- `session events` is a **read** command. Its success payload (`events: [...], count: N`) is a natural disambiguator from the error shape, so `"error" in data` is a clean check.
- `activity log` is a **write** command. Its success ack has no natural disambiguator beyond the keys themselves, so an explicit `ok` discriminator lets consumers branch on `data["ok"]` instead of presence-of-a-key.

Documented in the test class docstring and CHANGELOG so the asymmetry is a deliberate design choice, not drift.

## Scope

The issue body's example ack included `event_id`, but `storage.add_session_event` currently returns `None` — surfacing the row id would widen the storage protocol across base + mixin + 4 call sites. Out of scope for a CLI flag addition. Callers who need the id can follow up with `mm session events --json`; a focused follow-up issue is the right shape if a concrete caller later needs it in the ack.

Closes #335.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_session_cli.py` — 11 passed (5 new TestActivityLogJson cases)
- [x] Broader regression — 102 passed across `test_session_cli`, `test_sessions`, `test_tool_registration`, `test_cli`
- [x] `uv run ruff check packages/memtomem/{src,tests}` + `ruff format --check` — clean
- [x] `uv run mypy packages/memtomem/src/memtomem/cli/session_cmd.py` — 0 errors
- [ ] Manual: `uv run mm activity log -c "test" --json` (no session) → `{"ok": false, "reason": "no_active_session"}`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,14 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   `{error: "no_session"}` on stdout (exit 0) so scripts can parse the
   failure instead of getting a Click exit-1 + stderr line. Text path is
   unchanged. (#331)
+- **`mm activity log --json`** — scriptable ack output for hook-driven
+  event writes. Success: `{ok: true, session_id, event_type}`. No active
+  session: `{ok: false, reason: "no_active_session"}`. Write failure:
+  `{ok: false, reason: "write_failed"}`. Exit code is always 0 (the
+  silent-by-default hook contract is preserved without the flag). The
+  ok-flag shape intentionally differs from `session events --json`'s
+  `{error: ...}` — write acks have no natural disambiguator, so an
+  explicit `ok` discriminator is clearer for consumers. (#335)
 
 ### Changed
 
--- a/packages/memtomem/src/memtomem/cli/session_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/session_cmd.py
@@ -257,17 +257,33 @@ def activity() -> None:
 )
 @click.option("--content", "-c", required=True, help="Event description")
 @click.option("--meta", default=None, help="JSON metadata")
-def log_event(event_type: str, content: str, meta: str | None) -> None:
-    """Log an activity event to the current session."""
+@click.option("--json", "as_json", is_flag=True, help="Output a JSON ack for scripting.")
+def log_event(event_type: str, content: str, meta: str | None, *, as_json: bool = False) -> None:
+    """Log an activity event to the current session.
+
+    Silent by default so hook callers never fail. ``--json`` emits an ack
+    shape on stdout: ``{"ok": true, ...}`` on success, ``{"ok": false,
+    "reason": ...}`` when there is no active session or the write failed.
+    Exit code is always 0.
+    """
     session_id = _read_current_session()
     if not session_id:
-        # No active session — silently skip (hooks should not fail)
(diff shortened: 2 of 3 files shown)


tsdata (approved):
Ship it — 3-surface shape-choice defense (tests / CHANGELOG / PR body) holds up, scope discipline on event_id is right, control flow verified.

PR #340, opened 2026-04-21
docs(contrib): document JSON error shape convention (read vs write)

### LINKED ISSUE #339: docs(contrib): document JSON error shape convention (read `{error}` vs write `{ok, reason}`)
## Problem

`CONTRIBUTING.md` (after #334) documents when to use `--json` vs `--format [table|json|...]`, but it doesn't say anything about the **JSON error shape** convention. As a result, memtomem CLI now has two `--json` error patterns in the wild:

- **Read commands** (`session events --json`, `watchdog status --json`) — `{"error": "<reason>"}`. The success payload (`events: [...], count: N`) is a natural disambiguator, so a bare `error` key is enough.
- **Write commands** (`activity log --json` from #337) — `{"ok": false, "reason": "<reason>"}`. Write acks have no natural disambiguator beyond the keys themselves, so an explicit `ok` discriminator lets consumers branch on `data["ok"]` instead of presence-of-a-key.

The asymmetry is defensible (documented in #337 PR body, test class docstring, and CHANGELOG), but without a rule in CONTRIBUTING.md future contributors will re-litigate the shape every time a new `--json` flag lands.

## Proposed fix

Add a short subsection to the `--json` / `--format` rule in
`CONTRIBUTING.md` with the read/write distinction. Rough shape:

> ### JSON shape for error paths
>
> - **Read commands** (`list`, `get`, `show`, `events`): emit
>   `{"error": "<reason>"}` on stdout with exit 0. The success
>   payload's own keys (e.g. `events: [...]`) disambiguate error
>   from success, so an explicit `ok` field is redundant.
>
> - **Write commands** (`log`, `add`, `set`): emit
>   `{"ok": false, "reason": "<reason>"}` on stdout with exit 0,
>   and `{"ok": true, ...}` on success. Write acks have no natural
>   disambiguator, so an explicit `ok` flag is clearer for
>   consumers than `"error" in data`.
>
> Both shapes keep exit code 0 under `--json` — the JSON body
> carries the outcome, not the exit code, so `cmd --json | jq`
> pipelines don't break on failure.

## Why this belongs in CONTRIBUTING.md

Shape consistency is a contributor-facing rule, not a user-facing
one. Consumers read their own shape off `--help` / per-command
docs; contributors need to know the rule when they add the next
`--json` flag (and there will be more — see #335's recommendation
framing).

## Out of scope

- Unifying the two existing shapes retroactively. Both are already
  released; changing either is a breaking change for downstream
  scripts.
- Docstring-level documentation of the shape per-command. That's
  handled per-PR as commands get `--json`.

## Context

From #337 review (2026-04-21):
> Convention sprawl risk: memtomem CLI now has two `--json` error
> patterns. When you update CONTRIBUTING.md for #332's `--json` /
> `--format` rule, add a sentence on this — otherwise future
> contributors will re-litigate the shape per command.

memtomem:
## Summary

Adds a `JSON error shape` subsection to `CONTRIBUTING.md`'s existing CLI output convention, anchoring the shape choice made across #336 (read-side) and #337 (write-side) so future contributors don't re-litigate the shape per new `--json` command. Flagged explicitly in the #337 review as "convention sprawl risk."

## The rule

| Command kind | Success shape | Error shape |
|--------------|---------------|-------------|
| **Read** (`list`, `get`, `show`, `events`, `status`) | `{"<items>": [...], "count": N}` — keys self-disambiguate | `{"error": "<reason>"}` |
| **Write** (`log`, `add`, `set`, `run`) | `{"ok": true, ...}` — explicit flag | `{"ok": false, "reason": "<reason>"}` |

Both shapes exit 0 under `--json` so `cmd --json | jq` pipelines don't break on handled failures.

## Why asymmetric

Read success payloads already disambiguate from `{error: ...}` by presence-of-key (`"events" in data`, `"sessions" in data`). An `ok` flag would be redundant noise.

Write acks have no such natural disambiguator — consumers would have to discriminate by the specific keys present (`"event_id" in data` vs `"error" in data`), which gets awkward as ack shapes grow. An explicit `ok` flag is clearer for writes.

## No new tests

Existing `test_session_cli.py::TestSessionEventsJson` and `TestActivityLogJson` already pin the exact shapes the subsection describes — they're fail-if-false tests of the rule. Added a cross-reference comment to the `TestActivityLogJson` docstring so the docs/test link is discoverable from either side. Adding a separate convention-anchor test would be pure redundancy.

## Scope

- **Pure docs + one test-file comment.** No code changes.
- **No retroactive unification** of the two existing shapes. Both are released; changing either breaks downstream scripts. Rules apply forward.
- Intentionally does NOT document success shapes exhaustively (that lives in per-command `--help` + PR-time discussion). This is the *error*-shape rule, which is the one that caused the sprawl risk.

Closes #339.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_session_cli.py -q` — 11 passed (unchanged)
- [x] `uv run ruff check` on touched files — clean
- [x] Manual read-through of the new subsection in rendered GitHub preview (table + paragraphs)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,15 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   (genuine non-JSON modes like `plain` / `context` / `smart`), with a
   forward-compatibility guidance to prefer `--format` when a command might
   grow additional modes later. (#332)
+- **Documented JSON error shape convention** — `CONTRIBUTING.md` now
+  specifies the `--json` error shape by command kind: read commands
+  (`list`, `get`, `show`, `events`, `status`) emit `{"error": "<reason>"}`
+  since their success payloads self-disambiguate; write commands (`log`,
+  `add`, `set`, `run`) emit `{"ok": false, "reason": "<reason>"}` with an
+  explicit discriminator since write acks have no natural key-based
+  disambiguator. Both shapes exit 0 so `--json` pipelines don't break
+  on handled failures. Anchors the shape choice made in #336 / #337 so
+  future `--json` flags don't re-litigate per-command. (#339)
 
 ## [0.1.14] — 2026-04-21
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,6 +101,33 @@ plausible a `context` / `digest` / `diff` mode gets added later — choose
 `--format` from the start. Migrating from `--json` flag to `--format` is a
 breaking change for scripts; going the other way isn't necessary.
 
+### JSON error shape
+
+Pick the `--json` error shape by whether the command's success payload
+already disambiguates from an error:
+
+| Command kind | Success shape                       | Error shape                              |
+|--------------|-------------------------------------|------------------------------------------|
+| **Read** (`list`, `get`, `show`, `events`, `status`) | `{"<items>": [...], "count": N}` — the keys self-disambiguate | `{"error": "<reason>"}` |
+| **Write** (`log`, `add`, `set`, `run`) | `{"ok": true, ...}` — explicit flag | `{"ok": false, "reason": "<reason>"}` |
+
+Both shapes keep **exit code 0** under `--json`; the JSON body carries
+the outcome, not the exit code, so `cmd --json | jq` pipelines don't
+break on a handled failure. Unhandled exceptions (programmer errors,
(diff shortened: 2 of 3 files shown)


tsdata (approved):
Docs-only, scope clean. Rule grounded in the real asymmetry, cross-ref comment prevents docs/test drift via the seam that could introduce it. Ship.

PR #341, opened 2026-04-21
fix(cli): `mm activity log --meta <bad> --json` emits invalid_meta ack

### LINKED ISSUE #338: feat(cli): `mm activity log --meta <bad-json> --json` should emit `{ok: false, reason: "invalid_meta"}` instead of throwing
## Problem

`mm activity log --meta '{oops' --json` produces a traceback + exit 1 even though `--json` is set. `json.loads(meta)` in `session_cmd.py:log_event` sits outside the try/except that wraps the storage write, so malformed `--meta` throws through Click's default exception handler.

This is a minor consistency gap surfaced in the #337 review — `--json` callers reasonably expect all error paths to emit the ack shape rather than a Click traceback.

```
$ mm activity log --content x --meta '{oops' --json
Traceback (most recent call last):
  ...
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: ...
# exit 1
```

Expected under `--json`:

```json
{"ok": false, "reason": "invalid_meta"}
# exit 0
```

## Proposed fix

Move `json.loads(meta)` inside the existing try/except in
`packages/memtomem/src/memtomem/cli/session_cmd.py::log_event` (the one
that already wraps `asyncio.run(_log_event(...))`), with a distinct
reason for `json.JSONDecodeError` so the ack shape distinguishes "bad
input" from "storage error":

- `json.JSONDecodeError` → `{"ok": false, "reason": "invalid_meta"}`
- Other `Exception` → `{"ok": false, "reason": "write_failed"}` (unchanged)

Under the text path (no `--json`), the behavior question is trickier
— current code lets `JSONDecodeError` bubble up to Click's default
handler. Options:

1. Keep text path unchanged (traceback stays, matches pre-existing
   behavior). `--json` callers get the ack, text callers keep what
   they had.
2. Catch under text path too and log a warning (matches the
   silent-by-default hook contract). Downside: a hook author who
   genuinely mistyped their meta never sees why.

Lean toward option 1 — `--json` is the scripting surface where
parseable errors are load-bearing; text path traceback is actually
useful for a human diagnosing a hook config.

## Test plan

- Add a case to `TestActivityLogJson` with malformed `--meta`
  asserting `{"ok": false, "reason": "invalid_meta"}` and `exit 0`.
- Pin the text path traceback behavior with a separate test so the
  design choice is explicit.

## Non-goals

- Validating the **contents** of parsed meta. Only JSON parse is in
  scope.

## Context

From #337 review (2026-04-21):
> `--meta '{oops'` with `--json` still throws a traceback. Not a
> regression, but a `--json` caller would reasonably expect
> `{ok: false, reason: "invalid_meta"}` for consistency.

memtomem:
## Summary

Follow-up to #337. Before this fix, `mm activity log --meta '{oops' --json` threw a `json.JSONDecodeError` traceback and exit 1 through Click because `json.loads(meta)` sat outside the try/except that wrapped the storage write. `--json` callers reasonably expect all handled failure modes to emit the ack shape.

## Behavior change

| Invocation                                    | Before                                        | After                                                       |
|-----------------------------------------------|-----------------------------------------------|-------------------------------------------------------------|
| `--meta '{oops' --json`                       | traceback + exit 1                            | `{"ok": false, "reason": "invalid_meta"}` + exit 0          |
| `--meta '{oops'` (text)                       | traceback + exit 1                            | **unchanged** — still traceback + exit 1                    |
| `--meta '{"valid": true}' --json`             | `{"ok": true, ...}` + exit 0                  | **unchanged**                                               |

Text-path behavior is deliberately preserved per #338 option 1 — the silent-by-default hook contract applies to *write* failures (DB locked, schema drift, etc.), not programmer input errors. A hook author mistyping meta wants to see the traceback.

## Implementation

Split the parse into its own try block catching `json.JSONDecodeError` narrowly:

```python
try:
    metadata = json.loads(meta) if meta else None
except json.JSONDecodeError:
    if as_json:
        click.echo(json.dumps({"ok": False, "reason": "invalid_meta"}))
        return
    raise
try:
    asyncio.run(_log_event(session_id, event_type, content, metadata))
except Exception:
    ...  # existing write_failed handler
```

Two blocks instead of one nested handler so that if storage ever raises a `JSONDecodeError` internally (today it doesn't — sqlite3 + `json.dumps` don't), the parse-error ack doesn't mask it as `invalid_meta`. Per failure-mode classification discipline.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_session_cli.py -q` — 13 passed (2 new: `test_invalid_meta_emits_invalid_meta_ack`, `test_invalid_meta_text_path_bubbles`)
- [x] `uv run ruff check packages/memtomem/{src,tests}` + `ruff format --check` — clean
- [x] `uv run mypy packages/memtomem/src/memtomem/cli/session_cmd.py` — 0 errors

Closes #338.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   ok-flag shape intentionally differs from `session events --json`'s
   `{error: ...}` — write acks have no natural disambiguator, so an
   explicit `ok` discriminator is clearer for consumers. (#335)
+- **`mm activity log --json` invalid-meta ack** — malformed `--meta`
+  JSON now emits `{ok: false, reason: "invalid_meta"}` under `--json`
+  (exit 0) so scripts can distinguish bad input from a write failure.
+  Without `--json` the `json.JSONDecodeError` still bubbles to Click
+  (traceback + exit 1) so a hook author mistyping meta sees why. (#338)
 
 ### Changed
 
--- a/packages/memtomem/src/memtomem/cli/session_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/session_cmd.py
@@ -274,7 +274,17 @@ def log_event(event_type: str, content: str, meta: str | None, *, as_json: bool 
         if as_json:
             click.echo(json.dumps({"ok": False, "reason": "no_active_session"}))
         return
-    metadata = json.loads(meta) if meta else None
+    try:
+        metadata = json.loads(meta) if meta else None
+    except json.JSONDecodeError:
+        # Malformed --meta: under --json emit the error ack (exit 0) so
+        # scripts can distinguish "bad input" from "write failed". Under
+        # text path, let Click surface the traceback — a hook author
+        # mistyping meta wants to see why.
+        if as_json:
+            click.echo(json.dumps({"ok": False, "reason": "invalid_meta"}))
+            return
+        raise
     try:
         asyncio.run(_log_event(session_id, event_type, content, metadata))
     except Exception:
--- a/packages/memtomem/tests/test_session_cli.py
+++ b/packages/memtomem/tests/test_session_cli.py
@@ -196,3 +196,31 @@ class TestActivityLogJson:
         result = runner.invoke(cli, ["activity", "log", "-c", "quiet"])
         assert result.exit_code == 0
         assert result.output == ""
+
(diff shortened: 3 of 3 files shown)


tsdata (approved):
Narrow exception + two-block classification + preserved text-path traceback. Clean #337-review follow-through. Ship.

PR #376, opened 2026-04-22
feat(init): prompt + progress bar for large memory_dirs too

memtomem:
## Summary

Follow-up to #375. Previously the large-case branch (file count > 10 OR
total bytes > 64 KB) silently skipped the seed prompt and just printed a
`mm web Reindex / mm index` hint. Users with a pre-existing memory dir
(Obsidian vault, synced notes folder, migrated `~/memories`) had no
in-wizard option to kick off the initial index — they had to exit and
re-run `mm index <dir>` separately.

This PR unifies small-case and large-case paths into a **single prompt**
(TTY-only, default No), with a **visible click progress bar** + Ctrl-C
resume hint so minutes-long runs on CPU embedders don't look hung. The
large-case prompt adds a provider-specific advisory so the user knows
roughly what to expect before committing.

## UX flow (large case)

```
  Found 423 file(s) (12 MB) in /Users/x/memories.
  bge-m3 / CPU embedder → may take several minutes.
  Ctrl-C cancels; resume later with `mm index <dir>` (hash-dedup safe).
  Start seeding now? [y/N]: y

  Seeding  [####................]  120/423  notes/day12.md
```

On Ctrl-C:
```
  Cancelled. Resume with: mm index /Users/x/memories
```

## Provider-specific advisories

| `state["provider"]` | Advisory |
|---|---|
| `onnx` (bge-m3 etc) | "may take several minutes" |
| `ollama` | "time varies by model; watch the progress bar" |
| `openai` | "mostly network-bound, usually under a minute" |
| `none` | "BM25-only provider → tokenize-only, usually fast" |
| unknown / absent | (no advisory — falls through to generic Ctrl-C hint) |

## Why default No (both cases)

Blindly pressing Enter through the wizard still shouldn't trigger a
multi-minute embedder job. The visible progress bar (when the user
opts in) is the PR #295 mitigation for the "is it hung?" failure mode
that killed the earlier startup-scan attempt. If the user actively
types `y`, they've implicitly consented to the wait and now they have:

1. Progress visibility (can see it's not stuck)
2. Ctrl-C escape hatch (clean abort, resumable)
3. Provider advisory (know roughly what to expect)

## Changes

- **Replaced** `_seed_initial_index` (one-shot `index_path`) with
  `_seed_with_progress` using `IndexEngine.index_path_stream`. Per-file
  progress dicts feed a `click.progressbar` lazily initialized on the
  first `progress` event (so we know `files_total`). Bar exit is
  handled via a small helper so KeyboardInterrupt / exception paths
  don't leak an open bar into the wizard's subsequent output.
- **`_format_size` + `_provider_seed_hint` helpers**: extracted from
  the large-case branch so the logic is tested in isolation.
- **`_maybe_seed_initial_index` now takes `state`**: needed for the
  provider advisory. Non-TTY skips both branches silently now (large
  case previously emitted a cyan hint even without a TTY — redundant
  with the Next-steps `mm index` line).
- **Cancellation**: `KeyboardInterrupt` during the async stream
  escapes `asyncio.run` as-is and is caught at the sync boundary to
  print a yellow resume hint. `mm index` is content-hash idempotent
  so resuming is safe.

## Tests

12 total in `TestInitialSeedThreshold` (was 8):

- `test_provider_seed_hint_variants` — pins all 4 known providers +
  unknown → `None` fallback.
- `test_maybe_seed_large_by_count_prompts_with_advisory` — was the
  "large skips silently" regression; now asserts prompt fires + advisory
  appears.
- `test_maybe_seed_large_by_bytes_also_prompts` — same for bytes-axis
  trigger.
- `test_maybe_seed_large_accept_invokes_seed` — large + Yes → seeder
  called.
- `test_maybe_seed_large_non_tty_silent_skip` — consistent skip semantic
  with small-case non-TTY.
- `test_seed_with_progress_keyboard_interrupt_is_graceful` — Ctrl-C →
  yellow "Cancelled. Resume with: mm index <dir>", no traceback. Primary
  "escape hatch must land cleanly" pin.
- Plus renames of existing `_seed_initial_index` references to
  `_seed_with_progress`.

## Test plan

- [x] `uv run pytest -m "not ollama"` → 2100 passed (+4 net)
- [x] `uv run ruff check` + `format --check` → clean (288 files)
- [x] `uv run mypy packages/memtomem/src/memtomem/cli/init_cmd.py` → clean
- [x] Non-TTY large smoke (`mm init -y` + 15 `.md` files + piped stdin)
  → clean Next-steps output, no prompt leak, no "Cancelled" noise.
- [ ] CI green.
- [ ] (Reviewer) TTY smoke: run `mm init --preset english` with 15+
  `.md` files in `~/memories` and press `y` to verify the progress
  bar renders, then Ctrl-C to verify graceful cancel.

## Out of scope

- Background subprocess seeding (Option G from planning) — foreground
  streaming with Ctrl-C escape is simpler, avoids log-file /
  orphan-process / crash-notification infra. Revisit if user feedback
  says "I want wizard to exit while seeding continues".
- Threshold promotion to `config.indexing.seed_max_*` fields —
  module constants are tuned for the 90% case and can grow into
  config surface later if needed.
- `CHANGELOG.md` — user-visible but small; rolls into the next
  release's `[Unreleased]` block.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -1291,10 +1291,18 @@ def _collect_seed_scale(memory_dir: Path) -> tuple[int, int]:
 
 
 def _format_size(total_bytes: int) -> str:
-    """Human-readable size — KB below 1024 KB, MB above (integer-only,
-    avoids fractional noise in the wizard's one-line advisory)."""
+    """Human-readable size — ``<1 KB`` for sub-KB totals (15 tiny memo
+    files shouldn't display as "0 KB"), ``N KB`` up to 1024 KB, ``N MB``
+    above. Integer-only beyond the sub-KB floor to avoid fractional noise
+    in the wizard's one-line advisory."""
+    if total_bytes == 0:
+        return "0 bytes"
     kb = total_bytes // 1024
-    return f"{kb // 1024} MB" if kb >= 1024 else f"{kb} KB"
+    if kb == 0:
+        return "<1 KB"
+    if kb >= 1024:
+        return f"{kb // 1024} MB"
+    return f"{kb} KB"
 
 
 def _provider_seed_hint(provider: str) -> str | None:
@@ -1388,10 +1396,34 @@ def _seed_with_progress(memory_dir: Path) -> bool:
         return False
 
     click.echo()
+    total_files = result["total_files"]
+    indexed = result["indexed_chunks"]
+    skipped = result["skipped_chunks"]
+
+    # Defensive: if the stream processed files but landed zero chunks
+    # (neither new nor skipped-as-unchanged), something went wrong
+    # silently — per-file errors are logged but the `complete` event
+    # aggregates to zero counters. Known trigger: provider=none
+    # (NoopEmbedder dim=0) which leaves the ``chunks_vec`` virtual
+    # table uncreated, so ``upsert_chunks`` rolls back every insert
+    # (``feedback_chunks_vec_dim0_legacy.md``). Return False so the
+    # Next-steps step 1 stays unmarked and the user knows to investigate
+    # rather than seeing a false green success.
+    if total_files > 0 and indexed == 0 and skipped == 0:
+        click.secho(
+            f"  Seeded {total_files} file(s) but 0 chunks were indexed — "
+            "check logs for upsert errors.",
+            fg="yellow",
+        )
+        click.secho(
+            "  If you switched embedders recently, try `mm embedding-reset --mode apply-current`.",
+            fg="yellow",
+        )
+        return False
+
     click.secho(
-        f"  Seeded initial index: {result['total_files']} file(s), "
-        f"{result['indexed_chunks']} new chunk(s), "
-        f"{result['skipped_chunks']} unchanged.",
+        f"  Seeded initial index: {total_files} file(s), "
+        f"{indexed} new chunk(s), {skipped} unchanged.",
         fg="green",
     )
     return True
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -1290,60 +1290,137 @@ def _collect_seed_scale(memory_dir: Path) -> tuple[int, int]:
     return count, total
 
 
-def _seed_initial_index(memory_dir: Path) -> bool:
-    """Run one-shot :meth:`IndexEngine.index_path` on ``memory_dir``.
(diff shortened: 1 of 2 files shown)


memtomem:
## TUI smoke review (pty-based)

Ran pty smoke for scenarios S1–S4 + S8 on this branch. Two issues
surfaced and fixed in commit 5017fa5:

### Issue A: `(0 KB)` display for sub-KB totals

15 tiny memo files totalled ~300 bytes → advisory showed
`"Found 15 file(s) (0 KB)"` — looks like a display bug. `_format_size`
now returns `"<1 KB"` for sub-KB sums.

### Issue B: Zero-chunks seed printed green "success" 🚨

With `provider=none` (NoopEmbedder, dim=0), the `chunks_vec` virtual
table is never created so every `upsert_chunks` rolls back with
`"no such table: chunks_vec"`. Per-file errors are logged but the
stream's `complete` event aggregates to zero counters, and the wizard
previously printed:

```
  Seeded initial index: 3 file(s), 0 new chunk(s), 0 unchanged.
```

...in **green** — user thinks it worked. In reality, 0 chunks landed.

This is the pre-existing NoopEmbedder dim=0 bug documented in
`feedback_chunks_vec_dim0_legacy.md` (surfaced today by this PR's live
run). Direct `mm index <dir>` hits the same failure, so **not caused
by this PR** — but this PR's seed UX was masking it.

**Fix in this PR**: defensive gate — when `total_files > 0 AND
indexed_chunks == 0 AND skipped_chunks == 0`, print a **yellow**
warning + `mm embedding-reset` suggestion, and return `False` so the
Next-steps step 1 stays un-annotated. Catches any future silent
zero-result path, not just this specific bug.

**Followup**: filing a separate issue for the underlying
`provider=none` + `chunks_vec` schema gap.

### Scenario matrix (all verified)

| # | Scenario | Result |
|---|---|---|
| S1 | TTY + small + Y | prompt → progress bar → yellow warning (chunks_vec bug) |
| S2 | TTY + small + N | prompt → skip, Next-steps unchanged |
| S3 | TTY + large + Y | advisory → prompt → progress bar → yellow warning |
| S4 | TTY + large + N | advisory (`<1 KB` now correct) → prompt → skip |
| S6 | non-TTY + small | silent ✅ |
| S7 | non-TTY + large | silent ✅ |
| S8 | TTY + empty | silent (no prompt) ✅ |
| S5 | Ctrl-C during seed | covered by unit test (no pty signal test here) |
| S9 | cli_components raises | covered by unit test |

Ready to merge once CI on the new commit is green.

PR #377, opened 2026-04-22
fix(storage): gate chunks_vec writes on dim=0 BM25-only installs

memtomem:
## Summary

Fresh `mm init --provider none` uses NoopEmbedder (dim=0). The schema gate at `sqlite_schema.py:202-206` correctly skips the `chunks_vec` virtual table when dim=0, but the writer never learned about this — `upsert_chunks`, `delete_chunks`, `delete_by_source`, and `delete_by_namespace` all issue unconditional SQL against `chunks_vec`. Every `mm index` call rolled back with `no such table: chunks_vec`, indexed 0 chunks, and left BM25 search empty because the FTS write shared the same rolled-back transaction. PR #376 added a wizard-only yellow warning; bare `mm index` / MCP `mem_index` stayed broken.

This PR gates every chunks_vec touch on `SqliteBackend._has_vec_table`, a cached boolean primed from a one-time `sqlite_master` probe at `initialize()` and refreshed by `reset_embedding_meta` / `reset_all` whenever the underlying DROP/CREATE runs. `NamespaceOps` receives the flag as a **required** live callable (no default) so recovery transitions dim=0 → real provider flip cleanly without reconstruction.

Repro (pre-fix):
```bash
rm -rf /tmp/mm-bug-home && mkdir -p /tmp/mm-bug-home/memories
echo "hello world" > /tmp/mm-bug-home/memories/test.md
HOME=/tmp/mm-bug-home uv run mm init -y --provider none \
  --memory-dir /tmp/mm-bug-home/memories --mcp skip
HOME=/tmp/mm-bug-home uv run mm index /tmp/mm-bug-home/memories
# upsert_chunks failed, transaction rolled back: no such table: chunks_vec
```

Post-fix:
```
Indexed 1 file(s): 1 new, 0 unchanged, 0 deleted (1ms)
```

## Why this approach (Option B, not A or C)

- **A — schema creates a `float[1]` stub vtable at dim=0.** The legacy DB sniff at `sqlite_schema.py:127-143` reads `float[N]` from the existing chunks_vec SQL and caches it as `stored_dimension=1`. On any later provider switch, the fail-fast gate at line 184 fires `EmbeddingDimensionMismatchError` with a misleading remediation hint. Schema would lie about the real state. Rejected.
- **B — gate the writer (this PR).** Schema stays honest. One cached boolean + 7 one-line guards. Existing dim=0 DBs need no migration.
- **C — add a `bm25_only` config flag.** `enable_dense=False` already exists in `search/pipeline.py`. A new flag would duplicate surface and require reading from 3 layers. Rejected.

## Exhaustiveness verification

`rg -n "chunks_vec" packages/memtomem/src packages/memtomem/tests` — the 8 references outside `storage/` are all comments / docstrings / error-message strings (no SQL):

- `errors.py:18,20` — `EmbeddingDimensionMismatchError` docstring
- `server/context.py:41`, `server/lifespan.py:109,139`, `server/component_factory.py:40` — mismatch-handling comments
- `cli/init_cmd.py:958, 1407, 1409` — wizard zero-chunks defense comments
- `server/tools/memory_crud.py:77` — crash-risk comment

Storage layer is the complete write surface. `NamespaceOps` has a single caller at `sqlite_backend.py:148` — internal API, making the required-parameter constructor safe.

## PR-level notes (design decisions preserved from plan)

1. **`chunks_vec_info` explicit DROP** in reset paths is existing defensive behavior. sqlite-vec's `vec0` typically cascades shadow tables (`_chunks`, `_rowids`, `_info`, …) on virtual-table DROP. Left unchanged per "one change per PR"; if truly redundant, a separate cleanup PR can evaluate.
2. **Single-writer invariant** — `SqliteBackend` has no explicit writer lock; serialization relies on a single `self._db` connection. `_has_vec_table` mutation sits on top of this pre-existing assumption and does not alter it.
3. **Atomicity** — `reset_embedding_meta` / `reset_all` DROP+CREATE are not wrapped in an explicit `BEGIN`/`COMMIT` (existing semantics). If CREATE fails after DROP commits, the flag stays at its old value and every chunks_vec write short-circuits safely; next startup re-probes `sqlite_master` and self-heals.

## Test plan

- [x] `test_storage_noop.py` — 3 unit tests:
  - `test_upsert_chunks_noop_embedder_no_vec_table` — fresh dim=0 schema has no chunks_vec, upsert + BM25 search + `get_chunk` + `get_embeddings_for_chunks(...)=={}` + `dense_search(...)=[]`
  - `test_delete_paths_noop_mode` — `delete_chunks` + `delete_by_source` + `delete_by_namespace` all succeed without `OperationalError`
  - `test_reset_embedding_meta_creates_vec_table_and_preserves_fts` — dim=0 → real provider reset: chunks_vec appears, FTS entries survive, `NamespaceOps` callable picks up the live flag flip
- [x] `test_cli_index_noop_e2e.py` — inline (`CliRunner`) + subprocess (`mm` binary via `sys.executable` bindir) variants of `init → index → search`
- [x] `test_init_cmd.py::test_seed_with_progress_happy_path_suppresses_zero_chunks_warning` — happy-path pin that the yellow "0 chunks were indexed" warning does NOT fire when `indexed > 0`
- [x] Manual repro: pre-fix crashes with "no such table", post-fix "1 new"; `mm embedding-reset` recovery confirmed (flag flips True after dim=0→1024)
- [x] Full suite: `uv run pytest -m "not ollama"` → 2108 passed, 0 regressions
- [x] `uv run ruff check` + `uv run ruff format --check` pass
- [x] `uv run mypy` on changed files: no issues

### DIFF
--- a/packages/memtomem/src/memtomem/storage/sqlite_backend.py
+++ b/packages/memtomem/src/memtomem/storage/sqlite_backend.py
@@ -490,10 +490,6 @@ class SqliteBackend(
                         f"DELETE FROM chunks_vec WHERE rowid IN ({placeholders(len(new_rowids))})",
                         new_rowids,
                     )
-                else:
-                    logger.debug(
-                        "bm25-only mode (dimension=0); skipping chunks_vec write in upsert_chunks"
-                    )
 
                 db.executemany(
                     "INSERT INTO chunks_fts(rowid, content, source_file) VALUES (?,?,?)",
--- a/packages/memtomem/src/memtomem/storage/sqlite_backend.py
+++ b/packages/memtomem/src/memtomem/storage/sqlite_backend.py
@@ -97,6 +97,11 @@ class SqliteBackend(
         self._meta: MetaManager | None = None
         self._ns: NamespaceOps | None = None
         self._in_transaction: bool = False
+        # Invariant: _has_vec_table is True iff sqlite_master contains 'chunks_vec',
+        # which holds iff self._dimension > 0. Maintained by initialize(),
+        # reset_embedding_meta(), and reset_all() — all three must update this
+        # flag in lockstep with the underlying DROP/CREATE.
+        self._has_vec_table: bool = False
 
     async def initialize(self) -> None:
         db_path = Path(self._config.sqlite_path).expanduser()
@@ -140,7 +145,7 @@ class SqliteBackend(
 
         try:
             self._meta = MetaManager(self._get_db)
-            self._ns = NamespaceOps(self._get_db)
+            self._ns = NamespaceOps(self._get_db, lambda: self._has_vec_table)
 
             self._dimension, self._dim_mismatch, self._model_mismatch = create_tables(
                 self._db,
@@ -149,6 +154,12 @@ class SqliteBackend(
                 self._embedding_provider,
                 self._embedding_model,
                 strict_dim_check=self._strict_dim_check,
+            )
+            self._has_vec_table = (
+                self._db.execute(
+                    "SELECT 1 FROM sqlite_master WHERE type='table' AND name='chunks_vec'"
+                ).fetchone()
+                is not None
             )
         except Exception:
             await self.close()
@@ -284,10 +295,14 @@ class SqliteBackend(
             self._embedding_provider = provider
         if model:
             self._embedding_model = model
-        db.execute(f"""
-            CREATE VIRTUAL TABLE IF NOT EXISTS chunks_vec
-            USING vec0(embedding float[{self._dimension}])
-        """)
+        if self._dimension > 0:
+            db.execute(f"""
+                CREATE VIRTUAL TABLE IF NOT EXISTS chunks_vec
+                USING vec0(embedding float[{self._dimension}])
+            """)
+            self._has_vec_table = True
+        else:
+            self._has_vec_table = False
         db.commit()
         self.clear_embedding_mismatch()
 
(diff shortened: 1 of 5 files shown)


memtomem:
Thanks for the thorough review. Responses to each point:

## 🟡 #1 `reset_embedding_meta` DROP asymmetry

This is a misread. `sqlite_backend.py:290-291` runs `DROP TABLE IF EXISTS chunks_vec` unconditionally **before** the `if/else` block (existing code, preserved as-is):

```python
db.execute("DROP TABLE IF EXISTS chunks_vec")        # line 290 — always runs
db.execute("DROP TABLE IF EXISTS chunks_vec_info")   # line 291 — always runs
...
if self._dimension > 0:
    db.execute("CREATE VIRTUAL TABLE ...")
    self._has_vec_table = True
else:
    self._has_vec_table = False
```

All four reset scenarios stay consistent:
- real → dim=0: DROP (removes), skip CREATE, flag=False. sqlite_master empty ✓
- real → real: DROP + CREATE, flag=True ✓
- dim=0 → real: DROP (no-op), CREATE, flag=True ✓
- dim=0 → dim=0: DROP (no-op), skip CREATE, flag=False ✓

No divergence. No code change needed.

## 🟡 #2 Debug log asymmetry → removed

Agreed. Removed the lone `logger.debug` call (commit e36a010). `_has_vec_table` sits on the hot path, and per-call logging of the normal-operation branch is noise (consistent with `feedback_defensive_noise.md`). All skip sites are now silent for consistency.

## 🟡 #3 Subprocess test skip → fail

Agreed. `pytest.skip` → `pytest.fail` with a helpful message (commit e36a010). Any valid test environment (`uv run pytest` or `uv pip install -e`) must provide the `mm` entry point; if it's missing that's a test-env misconfiguration, and a loud fail beats a silent skip that would turn this regression guard into a CI false-green.

## 🟢 #4 Test comment cleanup

Agreed. Cleaned up the misleading comment and added the empty-namespace no-op assertion (`delete_by_namespace("nonexistent-ns") == 0`) so both populated and empty paths are exercised (commit e36a010).

## `chunks_vec_info` DROP redundancy follow-up

I'll track this as a separate issue after merge (keeping one change per PR).

Please re-review 🙏

Issue #387, opened 2026-04-22
`memtomem-server` leaves stale `.server.pid` on exit — risks `mm uninstall` liveness false-positive via PID recycling


memtomem:
## Problem

`memtomem-server` writes `~/.memtomem/.server.pid` on startup but **does not unlink it on exit**. After any non-clean termination (SIGKILL, parent-process death, system restart, crash) — and possibly after clean SIGTERM too — the file lingers on disk pointing at a PID that is no longer the server.

This is the mirror image of #384 (liveness check only sees MCP server pid):

- **#384** → real writer present but no pid file → liveness check false-negative → uninstall proceeds, risks WAL corruption
- **This** → no writer present but stale pid file → combined with PID recycling, liveness check can false-positive on an unrelated process that happened to inherit the recycled PID → `mm uninstall` refuses when it shouldn't

## Reproduction

```bash
# 1. Start a server (any path — a client-spawned one works too)
uvx --from memtomem memtomem-server </dev/null &
SERVER_PID=$!

# 2. Wait for pid file, then kill server abruptly
until [ -f ~/.memtomem/.server.pid ]; do sleep 0.2; done
cat ~/.memtomem/.server.pid                  # → 12345 (matches $SERVER_PID)
kill -9 $SERVER_PID                          # SIGKILL, no cleanup opportunity

# 3. Observe stale pid file
ls -la ~/.memtomem/.server.pid               # → still present
cat ~/.memtomem/.server.pid                  # → 12345, even though PID 12345 is dead
```

Observed in the wild on 2026-04-23 during first-time-user smoke testing. After `pkill -f memtomem-server`, `~/.memtomem/.server.pid` remained on disk with the dead PID written to it.

## Why it matters

`mm uninstall` probes liveness via `os.kill(pid, 0)` against the pid file (per #379 / `project_mm_uninstall_cmd.md`'s HIGH 1 review item). Today, `os.kill(<dead_pid>, 0)` raises `ProcessLookupError` and the check correctly concludes "not alive" — so a freshly-stale pid file doesn't immediately break uninstall.

The failure mode arrives with **PID recycling**: once the kernel hands the stale PID to an unrelated process (a browser helper, a shell, anything), `os.kill(<recycled_pid>, 0)` succeeds, the liveness check says "alive", and `mm uninstall` refuses to proceed until `--force`. The user gets a confusing refusal for a process that has nothing to do with memtomem.

## Suggested directions

[...]

PR #397, opened 2026-04-23
fix(uninstall): detect active DB writers via BEGIN IMMEDIATE probe

### LINKED ISSUE #384: `mm uninstall` liveness check only sees MCP server pid — silently ignores `mm web` and other DB writers
## Problem

`mm uninstall` (shipped in v0.1.23, #379) refuses to proceed when it detects a running server — good design, protects against deleting SQLite out from under a live WAL handle. The liveness probe is `os.kill(pid, 0)` against `~/.memtomem/.server.pid`.

**Gap**: `.server.pid` is written by the MCP `memtomem-server` entrypoint **only**. Every other process that can hold a live handle on `~/.memtomem/memtomem.db` is invisible to the check:

- `mm web` (FastAPI uvicorn) — no pid file
- `mm watchdog` long-running checks — no pid file
- Ad-hoc `sqlite3` connections, user scripts, test harnesses — no pid file

Result: liveness check false-negatives, `mm uninstall` proceeds without `--force`, DB file is unlinked while another process is still writing to it. On POSIX this doesn't surface immediately (the other process keeps writing to the unlinked inode), so the uninstall reports success and the user has no signal that anything went wrong — until a re-install inherits a state that doesn't match the live writer's view.

## Reproduction (observed 2026-04-22)

```bash
# Terminal A — start mm web against the default DB
mm web &  # PID 74828, opens ~/.memtomem/memtomem.db

# Terminal B — run uninstall without --force
mm uninstall -y
# Output:
#   memtomem state inventory:
#     Database     ~/.memtomem/memtomem.db    200.0 KB
#     ...
#   Removed: config.json, database, state dir.
#
# No warning. No refusal. No `--force` required.
# DB file is gone, but mm web (PID 74828) still holds the inode.

ls -la ~/.memtomem            # -> No such file or directory
ps -p 74828                   # -> still running, still writing
```

The `.server.pid` file was absent throughout — `mm web` doesn't write one.

## Suggested directions

Pick one or combine:

1. **Extend pid-file scheme to all long-lived entrypoints.** Every process that holds a DB connection for more than one request writes `~/.memtomem/.<name>.pid` at startup, removes it on clean shutdown. `mm uninstall` scans `~/.memtomem/.*.pid`. Simple, but requires discipline in every entrypoint forever.

2. **Fall back to `lsof` / `fuser` when `.server.pid` is absent or stale.** Platform-specific but reliable for any process that actually has the file open. `lsof -t ~/.memtomem/memtomem.db` returns PIDs.

3. **Try a `BEGIN IMMEDIATE` write lock on the DB with a short `busy_timeout`.** If another process holds a write lock, you get `SQLITE_BUSY` and can refuse cleanly. Platform-independent, reasoning-from-semantics rather than scanning process tables.

My preference: (3) as the primary check (captures *any* writer regardless of how it was launched), keep `.server.pid` as a fast-path / hint for the error message ("looks like `memtomem-server` PID X is holding the lock").

## Severity

**High** — silent data-corruption path triggered by the documented uninstall command. The current `--force` flag exists precisely to let the user override refusals, so a working liveness check is load-bearing.

## Context

Discovered during a first-time-user smoke test of v0.1.23, 2026-04-22. See also:
- #381 (docs: verify command side effects)
- #382 (CLI: missing `mm status`)
- #383 (MCP serverInfo.version mismatch)

memtomem:
## Summary

Adds a `BEGIN IMMEDIATE` probe against the SQLite DB in `mm uninstall` so that any active writer — `mm web`, `mm watchdog`, an ad-hoc script, anything — blocks the uninstall, not just the MCP `memtomem-server` whose pid we track in `.server.pid`. Closes #384.

2 files, +183 / −10.

## Why

The liveness check shipped in #379 / v0.1.23 was `os.kill(pid, 0)` against `~/.memtomem/.server.pid`. That pid file is written by the MCP entrypoint only. Every other process that can hold a live SQLite handle is invisible to that scheme, which is exactly the silent-data-corruption path #384 reproduced: `mm web` was running, `mm uninstall -y` succeeded without `--force`, the DB inode was unlinked while another process was still writing to it.

## Design

**Primary check — `BEGIN IMMEDIATE` with short `busy_timeout`** (direction 3 from #384, per reporter's stated preference):

- Catches any connection that holds RESERVED / PENDING / EXCLUSIVE at probe time.
- Process-agnostic: doesn't care how the writer was launched.
- Zero side effects: `ROLLBACK` immediately on success; no data modified.

**Scoping tradeoff, documented in the function docstring**:

- A process that holds only a SHARED lock (pure reader at probe time) will *not* block `BEGIN IMMEDIATE` in WAL mode. That reader can still end up with the inode after uninstall unlinks the directory entry, leading to UI/DB desync once the user re-installs.
- The severe case — active writer → WAL corruption — is what this PR guards. Full reader detection needs either `lsof` fallback (POSIX-only) or extending the pid-file scheme to every long-lived entrypoint; both are separate follow-ups and worth doing, but out of scope here.

**Header gate** — only probe when the file's first 16 bytes are `SQLite format 3\x00`:

- Discovered while fixing: opening a non-SQLite file with `mode=rw` can trigger sibling-file side effects (an existing test with fake `b"sqlite-fake"` content started losing its fake `-wal` file after my probe ran). Gating on the magic bytes keeps the probe side-effect-free and makes the existing fake-DB tests continue to exercise the probe-skipped path.

**`.server.pid` behavior preserved**: still checked first; if `.server.pid` is valid and alive, the existing "Server still running (pid X)" message wins. The new DB-lock message only fires when there's no live pid file but the DB is actually locked.

**Error handling is permissive**: if the probe can't run (DB missing, corrupt, permission denied, `sqlite3.Error` of any kind), return `locked=False` with `probe_error` set. Uninstall is itself a recovery path and must not be blocked by unrelated integrity issues — same principle as the existing `_load_config_safely` that swallows config parse errors.

## Test plan

- [x] `test_refuses_when_writer_holds_lock` — open a real SQLite DB, `BEGIN IMMEDIATE` in the test process, verify uninstall exits 2 with "holds a write lock" + `lsof` hint, verify DB still exists.
- [x] `test_force_overrides_db_lock` — same setup, `--force`, verify uninstall proceeds and the DB file is unlinked (even though the test's connection still holds the inode on POSIX — that's the documented escape hatch).
- [x] `test_proceeds_when_db_exists_but_no_writer` — valid SQLite DB with no open connection, uninstall proceeds.
- [x] All 22 existing uninstall tests continue to pass.
- [x] `ruff check` + `ruff format --check` clean.
- [x] `mypy` advisory clean on the changed file.
- [x] Broader `pytest -k "cli or uninstall"`: 222 passed.

## Out of scope — tracked as follow-ups

- **Reader-only detection** (the desync case). Options in the issue discussion: `lsof` fallback or pid-file expansion. Pick later.
- **#387** — `memtomem-server` leaves stale `.server.pid` on exit. Separate issue, adjacent but not blocking this fix.
- **Behavior when `--force` proceeds with an active writer**: we print the usual summary but don't escalate beyond what `--force` already signals ("you said so"). If we want a post-hoc "the other process is now orphaned at inode N" message, that's a separate UX call.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/uninstall_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/uninstall_cmd.py
@@ -12,9 +12,12 @@ Design notes:
   (uv-tool / uvx / venv-relative / system / unknown) need different
   uninstall commands; we print the right one but never execute, since the
   package manager owns its own permissions and lifecycle.
-* If the MCP server is still running we refuse — deleting an open SQLite
-  file under WAL mode risks corruption. ``--force`` overrides for the
-  rare case the user knows the pid is wrong.
+* If the MCP server is still running (``.server.pid``) OR any other
+  process holds a SQLite write lock on the DB (``mm web``, ``mm
+  watchdog``, ad-hoc connections) we refuse — deleting an open SQLite
+  file under WAL mode risks corruption. ``--force`` overrides. The
+  write-lock probe uses ``BEGIN IMMEDIATE`` to catch writers the pid
+  file doesn't know about (see ``_check_db_lock``).
 * If config.json is corrupted we fall back to defaults rather than
   aborting — uninstall is itself a recovery path, so it cannot depend on
   a valid config.
@@ -24,6 +27,7 @@ from __future__ import annotations
 
 import fcntl
 import shutil
+import sqlite3
 import sys
 from dataclasses import dataclass
 from pathlib import Path
@@ -82,6 +86,20 @@ class _ServerState:
     alive: bool
     pid: int | None
     pid_file: Path | None
+
+
+@dataclass(frozen=True)
+class _DbLockState:
+    """Result of probing the SQLite DB for an active writer.
+
+    ``locked`` is True only when another connection holds a RESERVED /
+    PENDING / EXCLUSIVE lock at probe time — i.e. an active writer.
+    Pure readers (SHARED locks only) are not detected; that's an
+    accepted tradeoff (see ``_check_db_lock``).
+    """
+
+    locked: bool
+    probe_error: str | None
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
## LGTM after rebase — design is right, verified locally

Pulled the branch and rebased onto current `main` (`2c153b8` → `d325324`). Conflicts are mechanical: add `sqlite3` alongside the existing `Iterator` import in `test_uninstall_cmd.py`, and keep main's no-arg `_check_server_liveness()` + add `db_lock = _check_db_lock(db_path)` at the `uninstall()` call site. No semantic conflict.

Post-rebase locally:
- `pytest packages/memtomem/tests/test_uninstall_cmd.py` — 27 passed (the 5 flock probe tests from #387 coexist with the new 3 DB-lock tests)
- `pytest` across uninstall + `test_runtime_paths.py` + `test_server_sigterm.py` — 50 passed
- `ruff check` / `ruff format --check` / `mypy uninstall_cmd.py` — clean

## Design

Complementary to existing checks, not duplicative:

- `.server.pid` flock probe (#387) catches the MCP `memtomem-server`.
- The new `BEGIN IMMEDIATE` probe catches every other writer (`mm web`, watchdog, ad-hoc scripts) — which #387's own commit message explicitly called out as "needs a separate, wider change". This PR is that change.

Worth preserving:

- **Header-magic gate** (`b"SQLite format 3\x00"`) prevents the WAL/shm side-effect trap the PR body describes. Small but load-bearing — without it, probing a corrupt/non-SQLite file with `mode=rw` can trigger sibling-file deletions.
- **Permissive error handling**: a non-busy `OperationalError` or sibling error returns `locked=False` so uninstall as a recovery path still runs on a corrupt DB. Mirrors `_load_config_safely`.
- **Reader-only case is explicitly out-of-scope** with pointers to the two follow-up options from #384. Good honesty about the tradeoff.

## Non-blocking nits

1. **Prefer `sqlite_errorcode` over substring match.** `sqlite3.OperationalError.sqlite_errorcode` (Py3.11+, available in our `py312` target) returns the numeric code directly — `SQLITE_BUSY = 5`, `SQLITE_LOCKED = 6`. More robust than `"locked" in msg or "busy" in msg` to any future wording/localization shift in the underlying SQLite C lib. Optional.

2. **PR body is slightly stale re: #387 / #413.** The body describes the baseline as "the liveness check shipped in #379 / v0.1.23 was `os.kill(pid, 0)`", but `main` has since moved to `fcntl.flock` (#387) and `$XDG_RUNTIME_DIR` (#413). A one-line "builds on the flock probe (#387) and runtime_paths relocation (#413)" in the body would make the stack visible to future archaeologists.

3. **Test-cleanup catch widening.** `test_force_overrides_db_lock`'s `finally` only catches `sqlite3.ProgrammingError`, but after the directory entry is unlinked a platform-specific `sqlite3.OperationalError` is also plausible from an open WAL/shm handle. Widening to `except sqlite3.Error:` would be defensive. Micro.

None are blockers. The diff against `84a6424` after rebase is zero functional change — only the trivial conflict resolution. Happy to push the rebased commit if helpful, or will leave that to you.

pandas-studio:
Rebased onto current `main` (d325324) to clear the conflict noted in the review — mechanical resolution, zero functional change vs. 84a6424. CI re-running.

Issue #399, opened 2026-04-23
Lazy DB init: defer ~/.memtomem/memtomem.db creation until first tool call


memtomem:
Continuation of the docs-only fix in #381 — addressing the broader behavior the comment thread on that PR identified.

## Problem

Any MCP client connecting to memtomem — not just the `memtomem-server` verify command — creates `~/.memtomem/memtomem.db` on handshake. #381 addressed only the docs verify command; the broader vector is every client (Claude Code, Cursor, Windsurf, Gemini CLI, any MCP session) that connects to memtomem after registration instantiates the DB before the user runs `mm init`.

Reproduction (observed 2026-04-23, memtomem==0.1.23):

```bash
rm -rf ~/.memtomem
claude mcp add memtomem -s user -- uvx --from memtomem memtomem-server
ls ~/.memtomem                  # still absent
claude mcp list                 # → ~/.memtomem suddenly populated
```

Observed: the directory appeared at the moment of `claude mcp list`. The exact spawner is not yet pinpointed — `claude mcp list` itself does a health-check spawn, and a concurrent Claude Code session may also rescan and reconnect when registration lands in `~/.claude.json`. Both paths execute the same `memtomem-server` startup.

## Root cause

Startup eagerly creates state at two layers:

1. **`server/__init__.py`** — line 149 `pid_dir.mkdir(...)` creates `~/.memtomem/`; lines 150–173 open + `flock` `.server.pid` (advisory lock). Runs on every `memtomem-server` spawn, including short-lived health-check spawns.
2. **`server/lifespan.py:105`** — `app_lifespan` calls `create_components()` which runs `await storage.initialize()` (`server/component_factory.py:66`), creating `~/.memtomem/memtomem.db` with all schemas, **before** the MCP `initialize` handshake yields.

MCP `initialize` + `tools/list` + `resources/list` do not need the DB — tool metadata is bound at import time in `server/__init__.py:42-105` (decorators) and filtered at lines 127–139, both lazy-independent. Only tool handlers and resource handlers actually touch the DB (`mem_add` / `mem_index` / `mem_search` / `mem_recall` / `mem_list` / `mem_read` and the `memtomem://*` resources in `server/resources.py:12-78`, all `app.storage.*` calls).

[...]

PR #413, opened 2026-04-23
server: relocate .server.pid to $XDG_RUNTIME_DIR

### LINKED ISSUE #412: server: relocate `.server.pid` to $XDG_RUNTIME_DIR so `~/.memtomem/` stays lazy
## Problem

`memtomem-server` writes `~/.memtomem/.server.pid` at startup to hold an
advisory `fcntl.flock` for the server-lifetime and to give the uninstall
command a liveness probe target. That single write is the last thing
keeping `~/.memtomem/` from being fully lazy — #399 Phase 3 (ea `00e948b`)
moved the SQLite DB + scheduler startup to the first tool call, but
`Path("~/.memtomem").mkdir(...)` in `server/__init__.py:main` still fires
on every MCP handshake (`claude mcp list`, Cursor/Windsurf probe, etc.),
even for a client that connects and never calls a tool.

Secondary concern: `.server.pid` is transient **runtime** state (auto-release
on process death), but it lives beside **persistent** data
(`memtomem.db`, `config.json`, `memories/`). Mixing the two means a
`rm -rf ~/.memtomem` workflow also deletes runtime-state knowledge about
live writers, and tooling has to special-case `.server.pid` when walking
the state dir.

This was called out as out-of-scope in the #399 Phase 3 plan and
referenced in #384 / #387 as "relocate to `$XDG_RUNTIME_DIR`".

## Proposal

Relocate the server pid / lock file from `~/.memtomem/.server.pid` to:

- **Linux w/ systemd**: `$XDG_RUNTIME_DIR/memtomem/server.pid`
  (e.g. `/run/user/1000/memtomem/server.pid`). Kernel auto-cleans at
  logout; no stale files after reboot.
- **macOS, Linux w/o `$XDG_RUNTIME_DIR`, BSD**: `{tempfile.gettempdir()}/memtomem-{uid}/server.pid`.
  On macOS this resolves to the per-user `/var/folders/.../T/memtomem-{uid}/`
  (already user-owned); on Linux fallback it's `/tmp/memtomem-{uid}/`
  created with mode 0700.
- **Windows**: server is POSIX-only (`fcntl`), so not relevant for the
  write side. Uninstall probe is a no-op there.

With this change, `~/.memtomem/` is created only by code paths that
actually need persistent storage — index, add, first tool call. A user
who only ever runs `claude mcp list` against a clean machine gets a
truly empty home, closing the last gap from #399.

## Design

1. New helper `memtomem._runtime_paths.runtime_dir() -> Path`:
   - Check `$XDG_RUNTIME_DIR`; if set and is a dir, use `{XDG}/memtomem`.
   - Otherwise `{tempfile.gettempdir()}/memtomem-{uid}`.
   - `mkdir(mode=0o700, exist_ok=True)` on first access.
2. Server (`server/__init__.py:main`) writes `runtime_dir() / "server.pid"`
   (drops the leading dot — dedicated runtime dir, no need to hide).
3. Uninstall (`cli/uninstall_cmd.py:_check_server_liveness`) probes the
   new location; during the transition period also probes the legacy
   `~/.memtomem/.server.pid` so a mixed-version upgrade (old server
   running, new uninstall) still refuses correctly.
4. Uninstall cleanup removes the legacy file if present but does NOT
   rmdir `$XDG_RUNTIME_DIR/memtomem` — the kernel handles it.

## Non-goals

- **Does not fix #384** (liveness probe only sees MCP server, ignores
  `mm web` / `mm watchdog`). That's a separate expansion — register
  per-process lock files and scan the directory. Out of scope here; the
  new runtime dir is the right *place* for those future lock files,
  but adding them is its own PR.
- **No change to `.current_session`**. Session state is persistent
  (mm remembers the user's current session across reboots); it stays
  in the state dir.

## Backward compat

- Legacy servers (pre-relocation) still write `~/.memtomem/.server.pid`.
  New uninstall probe checks both locations for one release cycle.
- Next-major drops the legacy probe.

## Related

- #384 — liveness probe only sees MCP server (this PR prepares the
  ground; expansion is separate)
- #387 — CLOSED, stale pid file after SIGTERM (addressed by flock +
  SIGTERM handler in `81bbdd5`)
- #399 Phase 3 — lazy lifespan (this closes the last non-lazy write)

memtomem:
## Summary

- Relocate `memtomem-server`'s pid / flock file from
  `~/.memtomem/.server.pid` to `$XDG_RUNTIME_DIR/memtomem/server.pid`
  (with per-user `$TMPDIR/memtomem-$UID/` fallback on macOS / BSD /
  non-systemd Linux).
- Combined with #399 Phase 3's lazy DB creation, an idle MCP handshake
  against a fresh machine now leaves `~/.memtomem/` untouched entirely
  — the last write that forced the dir into existence is gone.
- `mm uninstall` probes both the new and legacy locations during the
  transition window so mixed-version upgrades (pre-#412 server still
  running + new CLI) still refuse correctly.

## Details

New helper `memtomem._runtime_paths`:

- `runtime_dir()` — pure path resolver (no mkdir). Checks
  `$XDG_RUNTIME_DIR`; falls back to `tempfile.gettempdir()/memtomem-$UID`.
- `ensure_runtime_dir()` — explicit opt-in that creates the dir with
  `mode=0o700`. Never re-`chmod`s an existing dir (a `root`-owned
  leftover surfaces via the caller's `open()` rather than silently
  adjusted).
- `server_pid_path()` / `legacy_server_pid_path()` — typed handles for
  the two probe targets.

`server/__init__.py:main` now calls `ensure_runtime_dir() / "server.pid"`
instead of `Path("~/.memtomem").expanduser().mkdir(...) / ".server.pid"`.

`cli/uninstall_cmd.py`:

- `_check_server_liveness` walks `(server_pid_path(), legacy_server_pid_path())`
  and returns `alive=True` for the first flock holder.
- The new runtime pid file is included in the "Other" inventory group
  so it is cleaned alongside `.current_session` + legacy `.server.pid`.
- If the cleanup empties the runtime subdir, it is `rmdir`'d (we own
  the subdir; the XDG parent is kernel-managed and never touched).

## Non-goals

- Does not extend the liveness probe to other DB writers (`mm web`,
  `mm watchdog`) — that is #384 and needs per-process lock
  registration of its own. The new runtime dir is the right *place*
  for those future lock files; adding them is a separate PR.
- `.current_session` stays in `~/.memtomem/`. That is persistent
  session state across reboots, not runtime ephemera.

## Test plan

- [x] `uv run pytest packages/memtomem/tests -m "not ollama"` — 2191
      passed, 0 regressions.
- [x] New `test_runtime_paths.py` covers XDG set / unset / stale,
      0o700 creation, idempotency, uid suffix on fallback.
- [x] `test_uninstall_cmd.py` now covers refusal at both the new and
      legacy flock locations, PID recycling at both locations, and
      runtime-subdir cleanup.
- [x] `test_server_sigterm.py` integration test spawns a real
      `memtomem-server` with isolated `$HOME` + `$XDG_RUNTIME_DIR` and
      verifies SIGTERM cleanup at the new location.
- [x] `uv run ruff check` + `uv run ruff format --check` — clean.
- [x] `uv run mypy` advisory on the three changed source files — clean.
- [x] Manual smoke: `memtomem-server` writes and cleans up
      `server.pid` at the new location on macOS; `~/.memtomem/` is not
      created.

Closes #412.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### Changed
 
+- **`memtomem-server` pid / flock file moved to `$XDG_RUNTIME_DIR/memtomem/server.pid`.**
+  Previously the server wrote `~/.memtomem/.server.pid` at startup, which
+  forced `~/.memtomem/` into existence on every MCP handshake (even for
+  clients that connect but never call a tool). Runtime state now lives
+  on `$XDG_RUNTIME_DIR/memtomem/` when the platform provides it
+  (Linux w/ systemd), or `$TMPDIR/memtomem-$UID/` otherwise (macOS, BSD,
+  Linux without systemd). Combined with #399 Phase 3's lazy DB creation,
+  an idle MCP handshake against a fresh machine now leaves `~/.memtomem/`
+  untouched entirely. `mm uninstall` probes both the new and legacy
+  locations during the transition window, so a mixed-version upgrade
+  (pre-#412 server still running + new uninstall CLI) still refuses
+  correctly. (#412)
+
 - **MCP handshake no longer creates `~/.memtomem/memtomem.db`.** Previously,
   every MCP client that connected to memtomem (Claude Code's `claude mcp list`,
   Cursor, Windsurf, Gemini CLI) instantiated the SQLite database on handshake —
@@ -31,10 +44,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
     flagging for any maintenance schedule that assumed
     background-without-tool-calls semantics.
 
-  Note: `~/.memtomem/` itself is still created at process start by the
-  `memtomem-server` `.server.pid` advisory-lock acquisition; relocating
-  that to `$XDG_RUNTIME_DIR` is tracked separately in #384/#387.
-  (#399, #411; builds on #400 plumbing and #410 handler migration.)
+  Paired with #412 (runtime pid file relocation), an idle MCP handshake
+  now leaves `~/.memtomem/` untouched altogether — the advisory-lock
+  write that forced its creation moved to `$XDG_RUNTIME_DIR`. (#399,
+  #411; builds on #400 plumbing and #410 handler migration.)
 
 ### Fixed
 
--- a/docs/guides/uninstall.md
+++ b/docs/guides/uninstall.md
@@ -84,7 +84,13 @@ This removes:
 | `memories/` | User-created memories from `mem_add` |
 | `uploads/` | Files uploaded via the Web UI |
 | `.current_session` | Active session marker |
-| `.server.pid` | MCP server advisory lock |
+| `.server.pid` | Legacy MCP server advisory lock (pre-#412 installs only) |
+
+The running server's pid/flock file lives **outside** `~/.memtomem/`
+under `$XDG_RUNTIME_DIR/memtomem/server.pid` (Linux w/ systemd) or
+`$TMPDIR/memtomem-$UID/server.pid` (macOS, BSD). It is cleaned by
+`mm uninstall` automatically; a manual cleanup is rarely needed but
+equivalent to `rm -rf "${XDG_RUNTIME_DIR:-/tmp}/memtomem"*`.
 
 ## 4. Clean up project-scoped files (optional)
 
--- a/packages/memtomem/src/memtomem/_runtime_paths.py
+++ b/packages/memtomem/src/memtomem/_runtime_paths.py
@@ -1,0 +1,84 @@
+"""Runtime-state path resolution.
+
+Runtime files (pid files, locks, sockets) belong on ``$XDG_RUNTIME_DIR``
+when the platform provides one — the kernel auto-cleans them at logout,
+no stale artifacts survive a reboot, and they never mingle with the
+user's persistent data under ``~/.memtomem/``.
+
+Resolution order:
+
+1. ``$XDG_RUNTIME_DIR/memtomem`` — Linux + systemd (and any other OS that
+   exports the var). Per-user, ``tmpfs``-backed, kernel-managed lifecycle.
+2. ``{tempfile.gettempdir()}/memtomem-{uid}`` — macOS (where
+   ``gettempdir()`` resolves to a per-user ``/var/folders/.../T/`` already)
+   and Linux without systemd. ``uid`` suffix disambiguates a shared
+   ``/tmp``; ``mode=0o700`` at creation keeps it private to the user.
+
+The directory is created on first access. We never try to ``chmod`` an
+existing directory — if a user ran memtomem as root once and left behind
+a ``root``-owned runtime dir, fixing it silently would be worse than
+letting the open fail loudly.
+"""
+
+from __future__ import annotations
+
+import os
+import tempfile
+from pathlib import Path
+
+
+def runtime_dir() -> Path:
+    """Return the memtomem runtime directory path *without* creating it.
+
+    See module docstring for resolution order. Use :func:`ensure_runtime_dir`
+    when the directory needs to exist (e.g. opening a pid file for write);
+    the plain form is safe to call during read-only introspection such as
+    the uninstall inventory walk, which must not leave behind an empty dir.
+    """
+    xdg = os.environ.get("XDG_RUNTIME_DIR", "")
+    if xdg:
+        base = Path(xdg)
+        if base.is_dir():
+            return base / "memtomem"
+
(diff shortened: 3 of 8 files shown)


memtomem:
Thanks for the thorough pass — pushed `cdb1423` addressing every item
above B1 through NIT.

### B1 (BLOCKER) — legacy flock in `main()`

New `_try_hold_legacy_flock` acquires a **lifetime** flock on
`~/.memtomem/.server.pid`:

- Old server holds it → stderr `error: another memtomem-server holds a
  lock at …` + `sys.exit(1)` (matches reviewer's recommended refuse
  path — warn+continue is the wrong policy when the two writers are
  different versions hitting the same DB).
- Legacy path missing → skip entirely (fresh install, no upgrade
  history; touching `~/.memtomem/` would re-pollute the dir #412
  specifically keeps out of handshake).
- Acquired → hold for process lifetime via `atexit`, so a future old
  server starting after us also hits this lock via its own
  concurrent-detection path. Bidirectional mutex during the transition
  window.

Subprocess test `test_server_refuses_when_legacy_lock_held` pins it:
test holds `fcntl.flock` on the legacy file, spawns the real server
binary, asserts non-zero exit with the legacy pid path in stderr.

### M2 — XDG sanity tightened

`runtime_dir()` now gates on `_is_safe_dir(Path(xdg))` which `lstat`s
(never follows the symlink), checks `st.st_uid == os.geteuid()`, and
requires `mode & 0o077 == 0`. Any failure silently falls through to
the tempdir form so a misconfigured `XDG_RUNTIME_DIR=/tmp` or
symlinked export never places the pid file in a world-readable
location. Three dedicated tests: symlink, loose mode, wrong owner.

### M3 — docstring and enforcement now agree

`test_does_not_chmod_existing_dir` (which pinned the loose behavior)
is flipped to `test_refuses_existing_loose_mode` →
`pytest.raises(PermissionError, match="unsafe permissions")`. Full
matrix on `ensure_runtime_dir` refusal paths:

- symlink → `PermissionError("... is a symlink; refusing to follow")`
- wrong owner → `PermissionError("... owned by uid N (expected M)")`
- group/world bits → `PermissionError("... unsafe permissions 0o755")`
- non-directory → `PermissionError("... is not a directory")`

All error messages end with a `rm -f <path>` / `rm -rf <path>` hint so
the user can unstick themselves without reading the source.

### M1 — symlink path covered by the same matrix

The `_is_safe_dir` gate already rejects a symlinked XDG base (M2's
test suite proves it); `ensure_runtime_dir` rejects a symlinked
memtomem subdir (dedicated test). The tempdir fallback case is
structurally equivalent once the gate falls through to it: the same
validator runs. Umask belt-and-suspenders: explicit `os.chmod(target,
0o700)` after `mkdir` with its own test (`umask 0o177` scenario).

### Minor

- **m1 umask**: covered by new `test_explicit_chmod_survives_wild_umask`.
- **m2 non-empty rmdir pin**:
  `test_runtime_subdir_preserved_when_unrelated_files_present` —
  pre-seeds `someone-elses.pid` in the runtime subdir, runs `mm
  uninstall`, asserts sibling survives and runtime subdir stays.
  Invert regression of the `not any(iterdir())` guard would break
  this.
- **m3 fallback subprocess**: new `test_server_uses_tempdir_fallback_
  when_xdg_unset` spawns the server with `XDG_RUNTIME_DIR` unset +
  isolated `TMPDIR`, asserts the pid file lands on the expected
  per-uid path at mode 0o700. The `~/.memtomem/` must-not-exist
  assertion is now directly in
  `test_sigterm_unlinks_pid_file_end_to_end` too.
- **m4**: `test_falls_back_when_xdg_points_at_nonexistent_dir` renamed
  to `test_falls_back_when_xdg_path_is_missing_or_stale` with both
  semantic cases (never-materialized + reaped) noted in the docstring.

### Nit

- **n2**: `from memtomem._runtime_paths import … runtime_dir …` hoisted
  to module level in `uninstall_cmd.py`.
- **n3**: `_check_server_liveness` no longer takes `_state_dir`.
  Docstring updated to explain why the absence of a parameter is the
  right shape: both probes use canonical absolute paths, and the
  future #384 expansion will keep that posture (glob the runtime dir,
  not the state dir).

### n1 left for a separate pass

Declining the `sys.platform == "win32"` raise in `_runtime_paths.py` —
the module is imported by `uninstall_cmd.py` which also needs to load
cleanly on Windows (for the detect-and-report paths that don't touch
flock). The POSIX-only gate already lives on `fcntl` imports inside
both callers; adding it to `_runtime_paths` would surface a different
error than the one tests exercise. Happy to reconsider if this comes
up again.

### CI

Local: 2201 passed, 0 regressions. `ruff check` + `ruff format --check`
+ `mypy` on the three changed source files all clean. Pushing and
watching CI.

PR #418, opened 2026-04-23
fix(init): "b" at Memory Directory navigates back to preset picker (#371)

### LINKED ISSUE #371: mm init: "b" does not work at step 3 (Memory Directory)
## Description
In `mm init`, the "b" input does not work at step 3 (Memory Directory).

## Steps to Reproduce
1. Run `mm init` in your terminal or cmd
2. At step 3 (Memory Directory), enter "b"
3. It does not return to the previous step

## Expected Behavior
Entering "b" should return to the previous step. Entering "q" should exit the process.

## Actual Behavior
Entering "b" does not return to the previous step. However, entering "q" exits as expected.

## Environment
- OS: Windows 10

pandas-studio:
## Summary

Closes #371. Reported by @powerzist.

In the default interactive `mm init` path, typing `b` at step 3 ("Memory Directory") did nothing — the step re-prompted instead of returning to the preset picker. Root cause was platform-independent: the picker and its follow-up steps lived in two separate `run_steps` calls, so `_step_memory_dir` was **index 0** of the second call and `run_steps` took its "already at first step" branch.

## Why this reproduces only in the default path

Three entry shapes exist today:

1. `mm init --advanced` → `run_steps(advanced_steps)` — `_step_memory_dir` is index 2, "b" works.
2. `mm init --preset <name>` → explicit preset, memory_dir is the first interactive step. No picker to return to, so "b" is legitimately a no-op.
3. `mm init` (default) → `run_steps([_step_preset_picker])` **then** `run_steps([_step_memory_dir, _step_provider_dirs_auto, _step_mcp])` — this is the broken shape the reporter hit. Memory_dir is index 0 of the second list.

The issue title says "step 3" because `_step_memory_dir`'s `step_header(3, …)` hard-codes the label — that display is correct for the `--advanced` shape but misleading in the preset shapes. The fix below doesn't touch step numbering; narrower and scope-contained.

## Fix

- Combine the two `run_steps` calls in the default interactive path into one:
  `[_step_preset_picker, _step_memory_dir, _step_provider_dirs_auto, _step_mcp]`.
- `_step_preset_picker` now applies the chosen preset inline via `_apply_preset`. That's idempotent, so re-picking a different preset after "b" overwrites the previously applied state cleanly.
- Advanced choice is no longer encoded as a sentinel in `state["_preset_choice"]` — it raises a new `_AdvancedSelected` exception that the caller catches to dispatch the 10-step wizard. This lets us break out of the combined `run_steps` without smuggling control flow through state.
- CLI flag overrides (`_override_from_flags`) now run after the combined `run_steps` on the preset branch; advanced prompts per-field so there's no preset baseline to override there.
- Added a module-level `_isatty()` seam so the picker's non-TTY gate is patchable from tests without fighting CliRunner's substituted stdin (per `feedback_clirunner_isatty_seam.md`). Only the picker gate routes through this; the two legacy `sys.stdin.isatty()` call sites elsewhere in `init_cmd.py` are untouched and keep their own test paths.

`mm init --preset <name>` (path 2) is explicitly **not** changed — there's no picker to go back to from memory_dir there, and the current "already at first step" behavior is correct for that flow.

## Review follow-ups (9900208)

- **`_apply_preset` now drops any stale `rerank_model`** before reapplying, so re-picking from a rerank-enabled preset (english/korean) to a rerank-disabled one (minimal) after "b" doesn't leave the previous model string in state. Benign today (all downstream readers gate on `rerank_enabled` first) but the docstring promises a full preset-surface overwrite.
- **CHANGELOG: user-visible precedence change documented**. Combining the two `run_steps` also moved `_override_from_flags` from *before* the prompts to *after* them on the default-interactive branch, which means `mm init --memory-dir /x` (no `--preset`, no `-y`) finally respects `/x` — previously the prompt silently clobbered the flag. Same rule that already held for `--preset` / `-y` paths now also holds for the default path.

## Test plan

- [x] `TestPresetWizardBackNav::test_back_at_memory_dir_returns_to_preset_picker` — end-to-end: pick English, hit "b" at memory-dir, re-pick Korean. Final config writes `bge-m3` + `kiwipiepy`, proving the picker re-ran and the new preset fully overwrote the old one.
- [x] `TestPresetWizardBackNav::test_advanced_from_picker_routes_to_full_wizard` — picking `[4] Advanced` still dispatches the 10-step wizard via `_AdvancedSelected`. Guards against losing the advanced dispatch when refactoring the flow.
- [x] `TestPresetWizardBackNav::test_preset_picker_applies_preset_inline_on_non_advanced` — unit test pinning that `_apply_preset` runs inside the picker now.
- [x] `TestPresetWizardBackNav::test_preset_picker_advanced_choice_raises` — unit test pinning the `_AdvancedSelected` exception path.
- [x] `pytest packages/memtomem/tests/test_init_cmd.py` → 206 passed (202 original + 4 new; no existing test regressed).
- [x] `pytest packages/memtomem/tests -m "not ollama"` → 2215 passed.
- [x] `ruff check` + `ruff format --check` + `mypy init_cmd.py` → clean.

## Out of scope (noted, not fixed here)

- **Step-number display drift**: `step_header(3, "Memory Directory")` hard-codes "3" regardless of the flow, so preset paths show "3" even when it's the first interactive step the user sees. That's a UX polish issue orthogonal to "b" navigation — worth a separate cleanup.
- **Back-through-silent-steps**: `_step_provider_dirs_auto` prints a banner but doesn't prompt, so "b" at `_step_mcp` re-runs the banner and re-prompts MCP instead of jumping past the silent step to memory_dir. Also out of scope.
- **Path 2 `--preset X`**: still advertises "(b: back)" on the memory-dir step_header even though "b" is a no-op there (no picker to return to). Cosmetic; deferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,21 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### Fixed
 
+- **`mm init`: "b" (back) now works at step 3 in the default interactive
+  path.** The preset picker and its follow-up steps (memory dir, provider
+  dirs auto-detect, MCP config) used to run in two separate `run_steps`
+  calls, which made `_step_memory_dir` index 0 of the second call — so
+  hitting "b" at the "Memory Directory" prompt showed `(already at first
+  step)` and re-prompted the same step instead of returning to the preset
+  picker. The two calls are now combined into one: `_step_preset_picker`
+  applies the chosen preset inline and raises a new `_AdvancedSelected`
+  exception when the user picks Advanced, which the caller catches to
+  dispatch the full 10-step wizard. "b" from memory-dir decrements back
+  into the picker, and re-picking a different preset overwrites the
+  previously applied state cleanly (via the idempotent `_apply_preset`).
+  The explicit `--preset <name>` flag path is unchanged — there's no
+  picker to return to there. (#371)
+
 - **MCP `serverInfo.version` now reports the memtomem package version.**
   Previously `FastMCP.__init__` left the underlying `Server.version`
   at `None`, so the lowlevel server fell back to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,20 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   The explicit `--preset <name>` flag path is unchanged — there's no
   picker to return to there. (#371)
 
+- **`mm init` default-interactive CLI flag precedence is now "flag wins
+  over prompt", matching the documented behavior of
+  `_override_from_flags`.** The default path used to run the override
+  pass before the memory-dir / MCP prompts, so prompts silently
+  clobbered explicit flags: `mm init --memory-dir /x` (no `--preset`,
+  no `-y`) would apply `/x`, then immediately ask for a directory and
+  overwrite with whatever the user typed (usually the default
+  `~/memories`). The override pass now runs after the combined
+  `run_steps` on the preset branch, so flag values win — same rule that
+  already held for `--preset <name>` (path 2) and `-y` (non-interactive).
+  User-visible effect: `mm init --memory-dir /x` now respects `/x`
+  without requiring `--preset`. Advanced (`--advanced`) is unchanged —
+  advanced prompts per-field and has no baseline to override. (#371)
+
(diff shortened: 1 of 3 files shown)


tsdata (approved):
CI green, scope narrow, review follow-ups (9900208) address stale rerank_model + CHANGELOG precedence note. Approving to unblock merge.

PR #419, opened 2026-04-23
chore(init): preset-picker review follow-ups — dead write, docstring, coverage

pandas-studio:
## Summary

Addresses the remaining non-blocking nits from the #418 review round — all scoped to the preset picker and its tests.

## Changes

- **m1 — Drop `state["_preset_choice"] = "advanced"` dead write.** Control flow into the advanced branch goes through the `_AdvancedSelected` exception alone; nothing downstream reads the sentinel. Test updated to pin the "state untouched on advanced" contract explicitly (user-supplied sentinel key survives, both `_preset_choice` and `_preset_applied` stay `None`).

- **n3 — Shorten `_step_preset_picker` docstring + first-step comment.** The combined-`run_steps` change in #418 made the multi-paragraph rationale redundant. Comment is now one line: `# First step — "b" is a no-op here, so only advertise "q: quit".`

- **m3 — Parametrize `test_preset_picker_applies_preset_inline_on_non_advanced`** across all three non-advanced presets (`minimal` / `english` / `korean`) with their expected `model` and `tokenizer`. Previously only `korean` was covered, so a preset-spec drift on `minimal` or `english` (including rerank wiring, tokenizer flip) would have slipped through silently.

- **m4 — New `test_repeated_back_to_same_preset_is_idempotent`** — same-preset repick after "b". Regression guard: a future `_apply_preset` that toggled rather than assigned flat keys would fail here (the `pick-back-pick-different` test from #418 doesn't catch the toggle case).

- **m4 — New `test_repeated_back_cycle_through_multiple_presets`** — multi-cycle: english → "b" → korean → "b" → minimal. Final config must be entirely minimal; any field leaked from the earlier picks (provider, rerank, tokenizer) fails the assertions. Also pins that the `rerank_model` stale-field fix from 9900208 holds across more than one back cycle.

## Test plan

- [x] `pytest packages/memtomem/tests/test_init_cmd.py::TestPresetWizardBackNav` → 8 passed (was 4 pre-PR: +2 m4 tests, +2 net from the parametrize — one test became three parametrize cases, minus one replaced).
- [x] `pytest packages/memtomem/tests/test_init_cmd.py` → 210 passed (+4 from this PR; no existing test regressed).
- [x] `pytest packages/memtomem/tests -m "not ollama"` → 2222 passed.
- [x] `ruff check` / `ruff format --check` / `mypy init_cmd.py` → clean.

## Note on n2

The three out-of-scope UX items from the #418 review (step-number display drift, back-through-silent-step, `--preset X` stale "b: back" advertisement) are being filed as separate follow-up issues rather than bundled here — each touches a different part of the wizard surface and deserves its own scoping discussion before a fix lands.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -2184,22 +2184,15 @@ class _AdvancedSelected(Exception):
 def _step_preset_picker(state: dict) -> None:
     """First step of the default interactive path — pick a preset or Advanced.
 
-    Writes ``state["_preset_choice"]`` = one of ``"minimal" | "english" |
-    "korean" | "advanced"``. The Advanced entry raises
-    :class:`_AdvancedSelected` so the caller can dispatch the full
-    10-step wizard; the other three apply the preset inline via
-    :func:`_apply_preset` so the rest of ``run_steps`` can continue with
-    ``_step_memory_dir``, ``_step_provider_dirs_auto``, and ``_step_mcp``
-    against fully populated state.
-
-    Re-entry via ``"b"`` from a later step is safe: :func:`_apply_preset`
-    overwrites the full preset surface, so re-picking a different preset
-    takes effect cleanly.
+    Advanced raises :class:`_AdvancedSelected` so the caller can dispatch
+    the full 10-step wizard. The three preset entries apply the preset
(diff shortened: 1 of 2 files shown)


tsdata (approved):
CI green, scope contained (m1/m3/m4/n3 from #418 review trail). 2222 total tests pass locally. Approving to close out the review round.

Issue #422, opened 2026-04-23
init --preset <name>: memory-dir step advertises `(b: back)` but `b` has nothing to return to


pandas-studio:
## Symptom

`mm init --preset <name>` (the explicit-CLI-flag preset path) runs `_step_memory_dir` as step 0 of a fresh `run_steps` call — there's no picker to return to, so "b" at that prompt is legitimately a no-op. But `step_header` still advertises `(b: back, q: quit)`, which is misleading: the hint promises a return path that doesn't exist in this flow.

## Repro

```
$ mm init --preset korean
  ... (no picker shown — preset came from the CLI flag)
3. Memory Directory
  (b: back, q: quit)         # ← "b" is claimed to work here but it's a no-op
  Directory [~/memories]: b
  (already at first step)
  Directory [~/memories]:
```

## Scope

UX bug (behavior correct, hint misleading) — there's nothing to back into, but the hint still advertises `(b: back)`. Split cleanly from #371 / #418 (which fixed the **default-interactive** path where "b" **should** work and didn't).

## Suggested directions

1. **Variant header for first-and-only-step-is-memory-dir flows.** `_step_memory_dir` could accept a `show_back: bool = True` kwarg that `step_header` honors; `--preset X` dispatch sets it to `False`. Three lines of change.
2. **`step_header_first(title)` helper.** Dedicated helper that only advertises `(q: quit)`, used at the entry of a `run_steps` list that starts with a real-prompt step. Consistent with how `_step_preset_picker` already handles its own entry.
3. **Fix at `run_steps` level**: `run_steps` tells each step "you're at position 0 of N" via `state["_wizard_position"]`, and `step_header` reads that. Overlaps with the step-number-display issue (separately tracked) — maybe worth doing together.

## Acceptance criteria

- [ ] `mm init --preset <name>` no longer advertises `(b: back)` on the memory-dir step.
- [ ] The default interactive path (where "b" DOES work) keeps the existing advertisement.
- [ ] A test greps the output of a `CliRunner` invocation for both flows and asserts the correct hint in each.

## Related

- #371 / #418 — default path, "b" now works. This issue is the cosmetic twin for path 2.

timthoithi100:
Hi, I'd like to work on this. I'm thinking of going with Variant 1 (the show_back: bool kwarg approach) as it seems the most targeted. Let me know if that direction is approved.

memtomem:
[...]

Issue #420, opened 2026-04-23
init: `step_header(N, ...)` hardcodes step number — preset flows show misleading positions


pandas-studio:
## Symptom

`_step_memory_dir` (and every other `_step_*` in `init_cmd.py`) hard-codes its step number via `step_header(N, title)`. That number is correct for the `--advanced` flow (where memory-dir is step 3 of 10) but misleading in the two preset flows where memory-dir is the **first** interactive step the user actually sees:

| Flow | `step_header` shows | Actually is |
|---|---|---|
| `mm init --advanced` | `3. Memory Directory` | step 3 of 10 ✓ |
| `mm init --preset korean` | `3. Memory Directory` | step **1** of 3 |
| `mm init` (default, preset chosen) | `3. Memory Directory` | step **2** of 4 (picker + 3 follow-ups) |

First-time users reading `3.` without seeing steps 1 and 2 get a nagging "did I miss something?" signal, and the surrounding log ordering looks bizarre. (This may have contributed to the confusion in #371 — `3.` implies there was a previous step to return to.)

## Root cause

`packages/memtomem/src/memtomem/cli/init_cmd.py` — each `_step_*` function opens with `step_header(<hardcoded_int>, "<title>")`. The integer is a property of the `--advanced` 10-step sequence, not of the function's position in whatever `run_steps` list actually invokes it. `packages/memtomem/src/memtomem/cli/wizard.py:step_header` takes the number as a parameter but has no way to know the step's real position inside the current `run_steps` queue.

## Suggested directions

Pick one:

1. **Pass position through `run_steps`.** `run_steps` knows `i` and `len(steps)`; it could stash `state["_wizard_position"] = (i + 1, len(steps))` before each step call. Step functions read that instead of hard-coding. One-time wiring, no signature change to step functions.
2. **Make `step_header` take a `title` only and derive the number from a `state["_wizard_position"]` seed.** Same idea, different seam — the number lives in the helper instead of the step.
3. **Drop the number entirely** from the preset flows (show only the title), or use word markers like `"First. …"` / `"Next. …"`. Less structured but zero risk of drift.

Direction 1 or 2 is the right choice if we want the full sequence numbered consistently. Direction 3 is the lowest-touch escape hatch.

## Acceptance criteria

- [ ] In all three flows (advanced, `--preset X`, default-interactive preset), step headers reflect the actual position within the flow the user is taking.
- [ ] No step function has to know or care which flow it runs in.
[...]

PR #423, opened 2026-04-23
fix(init): back-nav through silent step reaches previous prompt (#421)

### LINKED ISSUE #421: init wizard: `b` at `_step_mcp` silently re-runs `_step_provider_dirs_auto` instead of reaching a real prompt
## Symptom

In the `mm init` default interactive path (and `--preset X`), typing `b` at the MCP step (`_step_mcp`) should conceptually take the user back to the memory-directory step. Instead it lands on `_step_provider_dirs_auto`, which is a **silent** step (prints a "No provider memory folders detected." banner, or the "Auto-added N folder(s)" banner, but does not prompt), so it runs its `click.echo` calls and falls straight through to the next iteration — dropping the user back at `_step_mcp` with just a re-printed banner in between.

Net effect: "b" at `_step_mcp` appears to do almost nothing. The user has to hit "b" twice in a row to reach a prompt that can actually accept input, which no wizard documents.

Same shape as #371 but one step further along the flow.

## Repro (assuming #418 lands)

```
$ mm init
  Choose setup style: [2]
  Directory [~/memories]: /tmp/m
  'tmp/m' doesn't exist. Create it? [Y/n] y
  No provider memory folders detected (checked: ...)
  Run 'mm init --advanced' later to add custom paths.

1. MCP Configuration     # (or whatever step_header shows for mcp)
  (b: back, q: quit)
  Select [2]: b          # user hits b expecting to go back to memory-dir

  No provider memory folders detected (checked: ...)   # banner re-prints
  Run 'mm init --advanced' later to add custom paths.

1. MCP Configuration    # same prompt reappears, no actual back-nav
  (b: back, q: quit)
  Select [2]:
```

## Root cause

`packages/memtomem/src/memtomem/cli/wizard.py:run_steps`:

```python
while i < len(steps):
    try:
        steps[i](state)
        i += 1
    except StepBack:
        if i > 0:
            i -= 1
            ...
```

`run_steps` decrements `i` and re-runs `steps[i]`, but it doesn't distinguish "user hit b, went back one" from "normal forward step". If `steps[i]` is silent (no prompt, just echo), it runs its side effect and `i += 1` advances right back to where we came from — user never sees a prompt they can act on.

## Suggested directions

1. **Mark silent steps explicitly.** Give a step function a marker (attribute, decorator, or separate list) saying "on back-nav pass-through, not present". `run_steps` on `StepBack` keeps decrementing until it hits a non-silent step.
2. **Teach silent steps to skip their side effects when visited during back-nav.** Pass a direction hint from `run_steps` (`_from_back=True`) and steps can short-circuit. More plumbing.
3. **Fold silent steps into adjacent interactive ones.** `_step_provider_dirs_auto`'s side effect could run inside `_step_memory_dir`'s tail, or inside a wrapper around `_step_mcp`'s head. Removes the silent step entirely.

Direction 1 is the narrowest fix and keeps step composability. Direction 3 would be cleaner but may complicate the advanced-flow case where provider_dirs is also a step.

## Acceptance criteria

- [ ] `b` at `_step_mcp` lands the user back at the previous **prompt** (memory-dir), not on a silent step.
- [ ] Silent step side effects (banners) don't re-fire on a back-pass-through.
- [ ] New test that sequences `CliRunner().invoke(init, input="<preset>\n<dir>\ny\nb\n...")` and asserts the memory-dir prompt re-appears in the output.

## Related

- #371 / #418 — same UX family, one step earlier.

pandas-studio:
## Summary

Closes #421.

`_step_provider_dirs_auto` sits between `_step_memory_dir` and `_step_mcp`
but only emits a detection banner — it has no user prompt. Pre-fix, "b" at
`_step_mcp` decremented onto this silent step, which re-emitted its banner
and advanced straight back to `_step_mcp`, so "b" appeared to do nothing.
Same "boundary can't be crossed" shape as #371 / #418 but one step further
along the flow.

## Fix

- **`wizard.py`**: new `silent_step` decorator sets a `_silent_in_back_nav`
  attribute on the step function. `run_steps`'s `StepBack` handler now walks
  back past any consecutive silent steps until it reaches an interactive one
  (or echoes `(already at first step)` if none precedes the current
  position). Forward direction is unchanged — silent steps still run their
  banner on the way through.
- **`init_cmd.py`**: decorate `_step_provider_dirs_auto` with `@silent_step`
  and document the skip contract in the docstring. It is the only silent
  step today; the decorator is exposed generically from `wizard.py` so
  future silent steps opt in the same way.

## Acceptance criteria (from #421)

- [x] `b` at `_step_mcp` lands the user back at the previous **prompt**
  (memory-dir), not on a silent step.
- [x] Silent step side effects (banners) don't re-fire on a back-pass-through.
- [x] New test that sequences `CliRunner().invoke(init, input="...")` and
  asserts the memory-dir prompt re-appears in the output.

## Test plan

- [x] `tests/test_wizard.py` (new) — 9 unit tests covering the skip mechanism
  directly: single silent skip, multi-silent skip, forward-only flow,
  back-at-index-0, silent-at-position-zero, state persistence across back-nav.
- [x] `TestBackNavThroughSilentStep` in `test_init_cmd.py` — CliRunner regression
  that pins the real wiring. `Memory Directory` header must appear twice in
  captured output (initial visit + post-back re-entry) and the banner text
  must be absent from the slice between the first mcp prompt and the second
  memory-dir header.
- [x] `pytest packages/memtomem/tests -m "not ollama"` → 2233 passed, 46
  deselected.
- [x] `ruff check` + `ruff format --check` + `mypy init_cmd.py wizard.py` → clean.

## Out of scope (tracked separately)

- **Step-number display drift** (#420) — `step_header` still hardcodes its
  number regardless of the flow. Orthogonal UX polish; next PR.
- **`--preset <name>` false-advertises `(b: back)`** (#422) — memory-dir is
  the only step there, so "b" is legitimately a no-op but the hint lies.
  Cosmetic twin, good-first-issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,20 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### Fixed
 
+- **`mm init`: "b" (back) at the MCP step now reaches the memory-directory
+  prompt instead of stalling on a silent banner.** The wizard's
+  `_step_provider_dirs_auto` sits between `_step_memory_dir` and `_step_mcp`
+  but only prints a detection banner — no prompt. Pre-fix, "b" at the MCP
+  step decremented the step index onto that silent step, which re-emitted
+  its banner and advanced straight back to `_step_mcp` — so "b" appeared to
+  do nothing until the user hit it twice. `wizard.run_steps` now skips past
+  `@silent_step`-marked functions when handling `StepBack`, so "b" lands on
+  the previous *interactive* step (memory_dir). Banner does not re-fire on
+  the back-pass (it only runs forward, once after each memory_dir
+  confirmation). Unit coverage for the skip mechanism lives in
+  `test_wizard.py`; end-to-end CliRunner regression is in
+  `test_init_cmd.py::TestBackNavThroughSilentStep`. (#421)
+
 - **`mm init`: "b" (back) now works at step 3 in the default interactive
   path.** The preset picker and its follow-up steps (memory dir, provider
   dirs auto-detect, MCP config) used to run in two separate `run_steps`
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -15,7 +15,7 @@ from typing import Literal
 import click
 
 from memtomem.cli.init_presets import PRESETS, _VALID_PRESETS, get_preset
-from memtomem.cli.wizard import nav_confirm, nav_prompt, run_steps, step_header
+from memtomem.cli.wizard import nav_confirm, nav_prompt, run_steps, silent_step, step_header
 
 InstallType = Literal["source", "project", "tool", "uvx"]
 CwdInstallType = Literal["source", "project", "pypi"]
@@ -2249,6 +2249,7 @@ def _apply_preset(state: dict, preset_name: str) -> None:
     state["_preset_applied"] = preset_name
 
 
+@silent_step
 def _step_provider_dirs_auto(state: dict) -> None:
     """Auto-apply detected provider memory folders for preset 2/3.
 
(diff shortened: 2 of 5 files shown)


tsdata (approved):
LGTM — automated review pass.

tsdata (approved):
LGTM — automated review pass.

PR #424, opened 2026-04-23
fix(init): step_header renders the actual flow position, not a hardcoded N (#420)

### LINKED ISSUE #420: init: `step_header(N, ...)` hardcodes step number — preset flows show misleading positions

pandas-studio:
## Summary

Closes #420.

Every `_step_*` function used to call `step_header(N, title)` with an integer
matching the step's index in the 10-step `--advanced` sequence. Preset flows
re-use the same step functions in shorter sequences, so `_step_memory_dir`
advertised `3. Memory Directory` even when it was the first prompt the user
saw (`--preset korean`) or the second (default interactive after a preset
pick). First-time users reading "3" without seeing steps 1 and 2 got a
nagging "did I miss something?" signal.

## Fix

- **`wizard.py`**: `run_steps` now seeds
  `state["_wizard_position"] = (i + 1, len(steps))` before each step
  invocation. `step_header(state, title)` reads the current index from that
  seed rather than taking the integer as an argument. If the key is absent
  (standalone / test use), `step_header` falls back to an unnumbered title.
  The signature change is internal — no external callers use `step_header`.
- **`init_cmd.py`**: 10 call sites migrated from `step_header(N, title)` to
  `step_header(state, title)`.

## Numbers per flow post-fix

| Flow | `step_header` shows | Actually is |
|---|---|---|
| `mm init --advanced` | 1..10 | ✓ (was correct by coincidence) |
| `mm init --preset korean` | 1..3 | ✓ (was 3, 4, 10 — the visible bug) |
| `mm init` (default, preset chosen) | 1..4 | picker=1 (own header), memory_dir=2, silent=3 (no header), mcp=4 |

## Scope notes

- **`_step_preset_picker`** keeps its own custom header (`"Choose setup
  style:"` without a number and only `(q: quit)` since there's no prior
  step). It doesn't use `step_header`, so the picker's look is unchanged.
- **`_step_provider_dirs_auto`** is silent (no header) — its position
  counts (step 3 in default flow) but the user never sees "3." announced.
  Users see numbers 2 and 4 with a banner between them. This matches the
  issue's suggested Direction 1 ("stash `(i + 1, len(steps))`") — changing
  the semantics to "only count interactive steps" would be a separate
  design decision.
- **`(b: back)` hint behavior** is untouched in this PR — that's #422,
  marked good-first-issue for external contributors.

## Test plan

- [x] `TestStepHeaderPosition::test_run_steps_seeds_position_before_each_step`
  — unit: `(i+1, len)` tuple visible to each step *during* its execution.
- [x] `test_position_reseeded_after_back_nav` — unit: re-entry after
  `StepBack` shows the same position number the second time through.
- [x] `test_step_header_renders_number_from_state` — unit: seeded state
  → `"2. Title"` in output.
- [x] `test_step_header_renders_without_number_when_state_unseeded` —
  unit: graceful fallback to unnumbered title.
- [x] `test_advanced_flow_shows_numbers_1_through_10` — CliRunner:
  end-to-end `--advanced` shows 1..10 at each prompt.
- [x] `test_default_preset_flow_renumbers_from_picker` — CliRunner:
  default interactive shows `2. Memory Directory` and
  `4. Connect to AI Editor`, never the pre-fix `3.` / `10.`.
- [x] `test_explicit_preset_flow_renumbers_from_memory_dir` — CliRunner:
  `--preset korean` shows `1. Memory Directory` and `3. Connect to AI Editor`.
- [x] `pytest packages/memtomem/tests -m "not ollama"` → 2229 passed,
  46 deselected.
- [x] `ruff check` + `ruff format --check` + `mypy wizard.py init_cmd.py`
  → clean.

## Related

- #371 / #418 — default-interactive "b" navigation fix. This PR closes
  the step-number half of the follow-up that was explicitly deferred
  from #418's scope.
- #421 — silent-step back-nav fix (same family, different symptom,
  independent PR).
- #422 — `--preset <name>` false-advertises `(b: back)`. Orthogonal to
  the number-display problem; good-first-issue candidate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,21 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### Fixed
 
+- **`mm init`: step headers show the correct position in every flow —
+  not a hardcoded number correct only for `--advanced`.** Previously every
+  `_step_*` function called `step_header(N, title)` with a hardcoded `N`
+  that matched the step's index in the 10-step advanced sequence. Preset
+  flows re-use the same step functions in shorter sequences, so
+  `_step_memory_dir` advertised `3. Memory Directory` even when it was
+  the first prompt the user saw (`--preset korean`) or the second (default
+  interactive after a preset pick). `wizard.run_steps` now seeds
+  `state["_wizard_position"] = (current_index, total)` before each
+  invocation, and `step_header(state, title)` reads that instead of
+  taking the integer as an argument. Numbers: advanced 1–10 (unchanged),
+  `--preset X` 1–3, default-interactive 1–4 (picker counts as step 1, the
+  silent `_step_provider_dirs_auto` counts as step 3 but prints no header
+  of its own). (#420)
+
 - **`mm init`: "b" (back) at the MCP step now reaches the memory-directory
   prompt instead of stalling on a silent banner.** The wizard's
   `_step_provider_dirs_auto` sits between `_step_memory_dir` and `_step_mcp`
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -278,7 +278,7 @@ def _y_refuse_hint(flag: str, extra: str) -> str:
 
 
 def _step_embedding(state: dict) -> None:
-    step_header(1, "Embedding Provider")
+    step_header(state, "Embedding Provider")
     click.echo("  Choose how to generate embeddings:")
     click.echo("    [1] Quick start — keyword search only, no setup needed (recommended)")
     click.echo("    [2] Local ONNX — dense search, no server needed (pip install memtomem[onnx])")
@@ -440,7 +440,7 @@ def _step_embedding(state: dict) -> None:
 
 
 def _step_reranker(state: dict) -> None:
-    step_header(2, "Reranker (optional)")
+    step_header(state, "Reranker (optional)")
     click.echo("  Cross-encoder reranking sharpens search relevance after BM25+dense fusion.")
     click.echo("  Runs locally via fastembed ONNX — no API key, no server.")
     enabled = nav_confirm("  Enable reranker?", default=False)
@@ -468,7 +468,7 @@ def _step_reranker(state: dict) -> None:
 
 
 def _step_memory_dir(state: dict) -> None:
-    step_header(3, "Memory Directory")
+    step_header(state, "Memory Directory")
     click.echo("  Where are the files you want to index?")
     memory_dir = nav_prompt("  Directory", default="~/memories")
     memory_path = Path(memory_dir).expanduser()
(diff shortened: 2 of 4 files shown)


tsdata (approved):
LGTM — automated review pass.

tsdata (approved):
LGTM — automated review pass.

PR #429, opened 2026-04-23
docs(guides): post-0.1.25 polish — mm status parity + tool-mode footnote

pandas-studio:
## Summary

Post-0.1.25 audit of the public 17-doc surface — three cross-doc
consistency fixes, plus a source-scan regression guard.

- `integrations/{claude-desktop,cursor}.md` now surface `mm status` as
  the terminal mirror of `mem_status` (previously only `claude-code.md`
  did).
- `integrations/cursor.md` First Indexing example switched from a
  one-line summary to the multiline `Indexing complete:` block used by
  the other two integrations.
- `mcp-clients.md` Config tool group now marks `mem_config`,
  `mem_embedding_reset`, `mem_reset` with `\*` and carries the same
  `MEMTOMEM_TOOL_MODE=full` footnote as `reference.md`.

## Regression guard

`packages/memtomem/tests/test_docs_guards.py` — 10 source-scan tests
(3 × `mm status`, 3 × `Indexing complete:` block, 4 × tool-mode footnote
parity). Written TDD: 3/10 failed before the doc edits, 10/10 pass now.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_docs_guards.py -v` — 10/10 pass
- [x] `uv run pytest packages/memtomem/tests/test_i18n.py -q` — 12/12 pass (no collateral damage)
- [x] `uv run ruff check` + `ruff format --check` on the new test — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #430, opened 2026-04-23
test(docs-guards): derive tool-mode footnote from reference.md

pandas-studio:
## Summary

Follow-up to code-review item #1 on #429: `test_docs_guards.py`'s
`_TOOL_MODE_FOOTNOTE` was a hardcoded string duplicated in the test —
any wording tweak to `reference.md` would fail the parity check even
when parity itself was preserved.

Now `reference.md` is the canonical source:

- New `canonical_footnote` fixture extracts the footnote line (the first
  line starting with `\* Requires `MEMTOMEM_TOOL_MODE=full``) from
  `reference.md` at test time.
- If `reference.md` loses the footnote, the fixture calls `pytest.fail`
  with a scoped message — so parity-test failures stay limited to
  "target file drifted" instead of conflating it with "canonical gone".
- `test_mcp_clients_matches_reference_footnote` asserts the canonical
  line appears verbatim in `mcp-clients.md`. That's the whole parity
  check now.
- The old `test_reference_carries_footnote` is now redundant (the
  fixture returning a line is itself the smoke check) and was removed.

Net: 10 tests → 9 tests + 1 fixture.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_docs_guards.py -v` — 9/9 pass

### DIFF
--- a/packages/memtomem/tests/test_docs_guards.py
+++ b/packages/memtomem/tests/test_docs_guards.py
@@ -24,12 +24,8 @@ _REPO_ROOT = Path(__file__).resolve().parents[3]
 _GUIDES = _REPO_ROOT / "docs" / "guides"
 _INTEGRATIONS = _GUIDES / "integrations"
 
-_TOOL_MODE_FOOTNOTE = (
-    r"\* Requires `MEMTOMEM_TOOL_MODE=full`. "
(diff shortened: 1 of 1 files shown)


memtomem (approved):
Approved.

Issue #437, opened 2026-04-23
server: failed-handshake leaves legacy `.server.pid` flock locked; reconnects loop


memtomem:
## Summary

When a `memtomem-server` child process started by an MCP client (Claude Code) fails its stdio handshake but the process **stays alive**, it continues to hold the legacy `~/.memtomem/.server.pid` flock. Every subsequent reconnect attempt from the client spawns a fresh child that aborts immediately with:

```
error: another memtomem-server holds a lock at /Users/.../.memtomem/.server.pid (likely a pre-0.1.25 install). Stop it before starting a new server; `mm uninstall` will also refuse until it is gone.
```

Manual `ps`+`kill` is required to recover. The error message blames "pre-0.1.25" but the lock holder here is a current-build server — the legacy path just happens to still be the one that gets held.

## Repro (observed live; not yet minimized)

1. Claude Code opens a project with `.mcp.json` pointing at `memtomem-server`.
2. First server child spawns, acquires legacy flock at `~/.memtomem/.server.pid`.
3. Handshake fails for some reason (exact cause TBD — the child stayed up as an orphan), Claude Code's UI shows `✘ failed`.
4. Reconnect → new child hits the flock held by (1) → aborts with the message above.
5. Loop until the user manually `kill`s the orphan.

Reliable trigger path not yet isolated. Filed on one repro because the **recovery story is poor** regardless of how the handshake fails: a dead-on-arrival server handshake shouldn't produce a lock that requires manual cleanup.

## Where (starting points)

- `packages/memtomem/src/memtomem/server/__init__.py:241-255` — `_try_hold_legacy_flock(legacy_server_pid_path())` is acquired early, before the MCP stdio handshake. If the server exits via an unhandled path (or stays alive despite handshake failure) the lock can outlive a useful session.
- `_runtime_paths.py:148-161` — `server_pid_path()` (new `$XDG_RUNTIME_DIR/memtomem/server.pid`) vs `legacy_server_pid_path()` (`~/.memtomem/.server.pid`). Both are held in parallel for migration (#412).

## Suggested investigation (hypothesis)

1. Confirm whether orphan processes here are the original handshake failure or a reconnect storm that the lock itself perpetuates.
[...]

PR #442, opened 2026-04-24
feat(server): parent-death watchdog self-SIGTERMs on reparent (closes #440)

### LINKED ISSUE #440: server: legacy flock UX follow-up — liveness probe, clearer stderr, orphan lifecycle (follow-up to #437)
Follow-up to #437 (PR #439 closed that). PR #439 fixed the **stale-file** axis — when the server exits cleanly, `~/.memtomem/.server.pid` is now unlinked on both `atexit` and SIGTERM, so a later probe won't race on a leftover file.

But a second axis of the same user symptom remains: **live orphan holder**. Observed today immediately after the fix landed:

```
Memtomem MCP Server
Status:           ✘ failed
Command:          memtomem-server
Config location:  /Users/pdstudio/.claude.json
```

Diagnosis at that moment:

```
$ lsof ~/.memtomem/.server.pid
COMMAND     PID     USER   FD   TYPE ...
python3.1 98059 pdstudio    3u   REG ...
$ ps 98059
98059  memtomem-server    (started ~5 min earlier)
```

So the server process is **alive and holding the fd** — this is the real "lock is held by another process" case, not a stale-file race. Claude Code's handshake failed but the child stayed up (Claude Code doesn't SIGTERM the orphan on handshake failure — or it does and the child swallows it), and every subsequent `Reconnect` spawns a new child that legitimately contends with the orphan.

User-visible symptoms repeat indefinitely:
- `Status: ✘ failed` until the user manually `pkill memtomem-server`.
- `/mcp` → "Failed to reconnect to memtomem." on every retry.
- stderr of the new child says "pre-0.1.25 install" — which is **still misleading**: the holder is a current-build orphan, not an old install.

### 1. Liveness probe before aborting

When `_try_hold_legacy_flock`'s `fcntl.flock(LOCK_EX|LOCK_NB)` fails, read the PID from the file (`_lock_fp.read()` before close, or open-for-read the path) and check `os.kill(pid, 0)`:
- Process gone → file is stale (this path became rarer after #439, but can still happen if someone upgrades from 0.1.25 mid-run). Unlink + retry acquiring the flock.
- Process alive + is actually `memtomem-server` → abort with a *live-holder* message (see #2).
- Any error reading/parsing → conservative abort.

Guardrail: the PID we read could be wrapped to a different process after the original died. Second check: on macOS/Linux, `/proc/<pid>/comm` or `ps -p <pid> -o comm=` to confirm it's actually a memtomem server. Accepting some false-positive abort is OK — false-negative would let two writers run.

### 2. Clearer stderr message

Today's message blames "a pre-0.1.25 install". Two better shapes:

- **Stale file path**: `"error: stale pid file at ~/.memtomem/.server.pid (no live holder). Retrying."` — and self-heal via (1).
- **Real contention path**: `"error: another memtomem-server is already running (pid {N}). If this is an orphan from a failed MCP handshake, run `kill {N}` and reconnect."`

The "pre-0.1.25 install" wording should only fire when the caller can actually verify the holder is pre-0.1.25 — which isn't practical post-#412, so probably drop it entirely.

### 3. Child-lifecycle hygiene

When `memtomem-server` detects that its controlling MCP client is gone (stdin EOF, or stdio writes erroring), it should exit proactively. `mcp.run()` should do this already via the JSON-RPC `exit` notification, but some clients (Claude Code observed here) seem to just stop reading without sending `exit` when their own handshake judgment fires. Options:

- Add a stdin-watchdog: a thread/task that reads from stdin and exits the process on EOF.
- Use `SO_KEEPALIVE`-equivalent for stdio — poll stdin reads with a short timeout, if no traffic for N minutes exit.

Option 3 is the broadest fix but also most invasive; (1) + (2) together would resolve the user-visible loop without changing lifecycle semantics.

## Suggested scoping

- **Must**: (1) + (2) as one PR — liveness probe + message rewrite. Small and self-contained in `_try_hold_legacy_flock`.
- **Should**: (3) as a separate follow-up (or filed against the upstream `modelcontextprotocol/python-sdk` if the issue is the SDK not handling client-departure).

## Reference

- PR #439 (stale-file teardown): landed.
- Original issue: #437.
- Memory `feedback_advisory_flock_unlink_pair.md` captures the general principle.

memtomem:
## Summary

Closes #440. Makes `memtomem-server` self-terminate when its MCP stdio client parent process disappears, instead of lingering as an orphan holding `~/.memtomem/.server.pid`.

Background from the #440 investigation:

- **Server side is clean**: three shutdown tests (stdin EOF / init+EOF / SIGTERM) all show the server exits properly and unlinks its pid file post-#439.
- **Client side doesn't always close stdio**: live `lsof` shows the stdio unix sockets to `claude` stay open across the shutdown path that produces orphans. When the `claude` process disappears without a clean socket close, the server has no client-provided signal to act on.
- **POSIX gives us one anyway**: when a child's parent exits, the child is reparented (`os.getppid()` changes to 1 on Linux, launchd's PID on macOS). Polling that is a portable, client-agnostic orphan signal.

## Approach

10s-interval `asyncio.create_task` watchdog launched in `app_lifespan`. On reparent detection it sends `SIGTERM` to its own PID rather than `os._exit`. This is deliberate: `_install_sigterm_handler` (#439) unlinks every pid file we own. Going around it would re-create the stale-file class of bugs #437/#439 closed.

## Design choices

- **Gating**: `MEMTOMEM_PARENT_WATCHDOG` env (default `on`). Disable with `off` / `0` / `false` for supervised / daemonised deployments where reparenting is expected.
- **Interval**: `MEMTOMEM_PARENT_WATCHDOG_INTERVAL` (default `10`). Invalid values fall back to default, server never crashes on a bad env.
- **Exit path**: `os.kill(os.getpid(), SIGTERM)` → existing sigterm handler → pid files unlinked → `os._exit(0)`. No new teardown code.
- **Task lifecycle**: created/cancelled in the same `app_lifespan` context, cancellation wrapped so main teardown still runs.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_server_parent_watchdog.py -v` — **17 passed**
  - 6× env-gating parametrized
  - 3× interval (default / override / invalid fallback)
  - Coroutine no-ops while parent alive
  - Coroutine self-SIGTERMs when ppid changes (monkey-patched iterator)
  - Cancellation returns cleanly
  - **End-to-end**: grandparent spawns parent+server, `SIGKILL`s parent, asserts server exits within 5s and pid file is unlinked — exercises the full watchdog → self-SIGTERM → #439 handler → pid file unlink chain
- [x] `uv run pytest packages/memtomem/tests/ -m "not ollama"` — **2268 passed** (previously 2251; +17)
- [x] `ruff check` / `ruff format --check` / `mypy` on `server/lifespan.py` — clean

## Out of scope (stays on #440 wishlist, not in this PR)

- Stderr message cleanup for the legacy-flock error ("pre-0.1.25 install" phrasing). With this watchdog live, that error should become rare enough that rewording is lower priority.
- Liveness probe on `_try_hold_legacy_flock` failure (dead-pid detection and unlink). Same rationale — orphan rate should drop sharply.

Either follow-up can land in a separate PR if still worth doing after real-world usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### Added
 
+- **Parent-death watchdog — `memtomem-server` self-SIGTERMs when its
+  MCP client parent process disappears.** Investigation in #440 showed
+  that Claude Code (and likely other MCP stdio clients) sometimes
+  terminate without closing our stdio unix sockets or sending a signal,
+  leaving `memtomem-server` alive as an orphan holding
+  `~/.memtomem/.server.pid`. The next client start then loses the
+  flock race and reports "Failed to connect". The server now polls
+  `os.getppid()` every 10s (configurable via
+  `MEMTOMEM_PARENT_WATCHDOG_INTERVAL`) and self-SIGTERMs when the
+  parent has been reparented — a POSIX-portable, client-agnostic
+  orphan signal. Self-SIGTERM (not `os._exit`) so the #439 sigterm
+  handler still fires and the pid files are unlinked cleanly. Can be
+  disabled with `MEMTOMEM_PARENT_WATCHDOG=off` for supervised /
+  daemonised deployments where reparenting is expected. (#440)
+
 ### Changed
 
 ### Fixed
--- a/packages/memtomem/src/memtomem/server/lifespan.py
+++ b/packages/memtomem/src/memtomem/server/lifespan.py
@@ -2,9 +2,11 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import logging.config
 import os
+import signal
 import sys
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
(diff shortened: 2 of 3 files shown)


memtomem:
Closing after empirical validation that v0.1.26 already handles both orphan-generating paths without this watchdog:

## Experiments (all run on v0.1.26 from PyPI)

**1. Normal `/exit` shutdown (primary use case)**
- User `/exit`s a Claude Code session holding a `memtomem-server` child
- Observed: server process disappears + both pid files unlinked within 1s ✅

**2. Parent SIGKILL with `MEMTOMEM_PARENT_WATCHDOG=off` (worst-case simulation)**
- Parent python process gets `SIGKILL` — no chance to close stdio or send signal
- Observed: server exits + pid files unlinked within 1s ✅

Why both work in 0.1.26: kernel closes parent's fd on process death → unix socket peer closes → server's stdio read returns EOF → `mcp.run()` exits → `lifespan` finally → atexit unlinks. The chain from #439 is sufficient.

## So when would this watchdog have helped?

Only in a hypothetical scenario where:
- Parent `claude` process is alive
- But the stdio socket to the child is dead/leaked

I couldn't produce this state with any plausible user action. The empirical root cause of today's reported orphan turned out to be **pre-0.1.26 processes still running in memory after an in-place upgrade** (disk bytes new, process memory old — `uv tool install` doesn't touch running Python processes). That's an upgrade-hygiene problem, not a runtime-watchdog problem, and it's filed as [#443][1] for an `mm upgrade` subcommand.

## Decision

- **Close PR** — no demonstrated bug to fix.
- **Branch `fix/440-parent-watchdog` retained on remote** in case we observe a concrete live-orphan-without-parent-death scenario later. The implementation (parent-polling + self-SIGTERM reusing the #439 handler) is sound; it just lacks a real-world trigger right now.
- Issue #440 can be closed in parallel — the three sub-improvements it proposed (liveness probe, stderr message, child lifecycle) are all either landed (#439 made stale files self-healing) or empirically unnecessary (this experiment).

[1]: will link in follow-up

PR #453, opened 2026-04-24
Fixes #422

### LINKED ISSUE #422: init --preset <name>: memory-dir step advertises `(b: back)` but `b` has nothing to return to

timthoithi100:
What I've changed: 
step_header in wizard.py now reads _wizard_position (seeded by run_steps per #424) to decide whether to advertise (b: back). When the step is at position 1, only (q: quit) is shown. Any step beyond position 1 keeps the existing (b: back, q: quit) hint.
This is a single-file, 4-line change that covers the general case rather than patching _step_memory_dir specifically.

Tests added:
test_preset_cli_flag_omits_back_hint — asserts (b: back) is absent on the memory-dir step when invoked via mm init --preset <name>
test_interactive_path_keeps_back_hint_at_memory_dir — regression guard asserting (b: back, q: quit) is still shown on memory-dir in the default interactive path (#418)

### DIFF
--- a/packages/memtomem/src/memtomem/cli/wizard.py
+++ b/packages/memtomem/src/memtomem/cli/wizard.py
@@ -136,6 +136,8 @@ def step_header(state: dict, title: str) -> None:
     if position is not None:
         current, _total = position
         click.secho(f"{current}. {title}", fg="yellow", bold=True)
+        nav_hint = "(b: back, q: quit)" if current > 1 else "(q: quit)"
     else:
         click.secho(title, fg="yellow", bold=True)
-    click.echo(click.style("  (b: back, q: quit)", dim=True))
+        nav_hint = "(b: back, q: quit)"
+    click.echo(click.style(f"  {nav_hint}", dim=True))
--- a/packages/memtomem/tests/test_init_cmd.py
+++ b/packages/memtomem/tests/test_init_cmd.py
@@ -3864,6 +3864,65 @@ class TestPresetWizardBackNav:
         # No dead write, no preset application — just the exception.
         assert "raised preset_choice=None applied=None sentinel=untouched" in result.output
 
+class TestPresetBackHintNavHint:
+    """(#422) ``mm init --preset <name>`` must not advertise ``(b: back)`` on
+    the memory-dir step because there is no picker to return to in that flow.
+    The default interactive path must still advertise ``(b: back)`` on the
+    memory-dir step (regression guard for #418)."""
+
+    def test_preset_cli_flag_omits_back_hint(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """``mm init --preset korean`` must show ``(q: quit)`` and not
+        ``(b: back, q: quit)`` on the Memory Directory step."""
+        from click.testing import CliRunner
+
+        from memtomem.cli.init_cmd import init
+
+        monkeypatch.setenv("HOME", str(tmp_path))
+        monkeypatch.setattr("memtomem.cli.init_cmd._isatty", lambda: True)
+        monkeypatch.setattr(
+            "memtomem.config._detect_provider_dirs",
+            lambda: {"claude-memory": [], "claude-plans": [], "codex": []},
+        )
+
+        memory_dir = tmp_path / "memories"
+        result = CliRunner().invoke(
+            init,
+            ["--preset", "korean"],
+            input=f"{memory_dir}\ny\n3\n",
+        )
+        assert result.exit_code == 0, result.output
+        # The memory-dir step is first in this flow — back hint must be absent there.
+        memory_dir_section = result.output.split("Memory Directory")[1].split("\n")[1]
+        assert "(q: quit)" in memory_dir_section
+        assert "(b: back, q: quit)" not in memory_dir_section
+
+    def test_interactive_path_keeps_back_hint_at_memory_dir(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """Default interactive path must still show ``(b: back, q: quit)`` on
+        the Memory Directory step — regression guard for #418."""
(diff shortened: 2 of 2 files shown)


timthoithi100:

memtomem (commented):
Thanks for the quick turnaround on this, @timthoithi100 — and nice choice
going with the `_wizard_position` root-level fix over the kwarg-threading
Variant 1. It generalises cleanly to any future first-step-is-interactive
call site rather than patching `_step_memory_dir` specifically, which is
exactly what the issue wanted.

The direction and test structure look good. Just one blocking nit on
CI plumbing and two optional suggestions.

### 🔴 `ruff format` (blocking the `lint` CI job)

`packages/memtomem/tests/test_init_cmd.py` is missing the PEP 8 two-blank-line
separator before the new `TestPresetBackHintNavHint` class and before the
following `TestReinstallEmbeddingReconciliation` class. `ruff check` is
clean but `ruff format --check` fails — both need to pass independently
([feedback_docs_as_tests note in the contributing quickstart](https://github.com/memtomem/memtomem/blob/main/CONTRIBUTING.md)).

Quick fix:

```bash
uv run ruff format packages/memtomem/tests/test_init_cmd.py
git add packages/memtomem/tests/test_init_cmd.py
git commit --amend --no-edit    # or a fresh commit, either works
git push --force-with-lease
```

I'll approve the Actions run so CI can go green once the push lands.

### 🟡 Optional — tighten the regression-guard assertion

`test_interactive_path_keeps_back_hint_at_memory_dir` currently asserts
`(b: back, q: quit) in result.output`, which passes if the hint appears
on **any** step past position 1 (provider-dirs, mcp, …), not specifically
on the memory-dir step. If a future refactor accidentally suppressed the
hint on memory-dir but left it on subsequent steps, this test would still
pass.

Mirroring the first test's section-scoped slice would make this a true
regression guard for #418:

```python
memory_dir_section = result.output.split("Memory Directory")[1].split("\n")[1]
assert "(b: back, q: quit)" in memory_dir_section
```

Non-blocking — the single-seam fix makes a divergent regression unlikely,
but it'd name what the test actually guards.

### 🟡 Optional — extend the `step_header` docstring

The new conditional isn't documented. Two lines would help future readers:

```python
"""Print a step header with navigation hint.

The number is read from ``state["_wizard_position"]`` as seeded by
:func:`run_steps`. If the key is absent (standalone / test use), the
header is rendered without a number. When ``current == 1``, the back
hint is omitted since there is no prior step to return to. (#420, #422)
"""
```

### Summary

Format fix → approve CI → merge is the path. Ping me when the force-push
lands and I'll take it from there.

memtomem:
Hi @timthoithi100 — checking in on this one.

The review above flagged the `ruff format` lint failure (the only blocker — the two yellow items are optional and don't need to be addressed for merge). Today is the Monday I'd promised to circle back by on the parallel PR (#425), so I want to be transparent about timing rather than let this drift silently.

The fix is the one-liner from the previous comment:

```bash
uv run ruff format packages/memtomem/tests/test_init_cmd.py
git add packages/memtomem/tests/test_init_cmd.py
git commit --amend --no-edit
git push --force-with-lease
```

I'll keep this PR open until **end of week (Fri 2026-05-01)**. If the lint push lands by then I'll approve the CI run and we can merge from here. If not, I'll revisit options between this PR and #425 (which covers the same `_wizard_position` fix and is currently CI-green) — happy to keep this open longer if you're mid-rebase, just drop a note.

If anything is blocking you on the format push, just say so and I'm glad to help.

memtomem:
Thanks for this, @timthoithi100, and for the two-case regression tests covering preset flow vs. interactive mode.

This has been overtaken in the meantime: #422 was fixed and closed by #425 ("fix(wizard): hide 'b: back' on first step where back is a no-op", merged 2026-05-04). Current main already trims the hint to `(q: quit)` on the first step in `wizard.py:step_header` — the same behavior this PR adds, so the change is now redundant against main.

Closing as superseded. This is not a reflection on the patch, which correctly diagnosed #422. Thanks again for the contribution.

PR #461, opened 2026-04-24
feat(cli): mm agent register/list/share + hidden debug-resolve

memtomem:
## Summary

Adds operator-facing CLI mirrors of the multi-agent MCP tools so
running multi-agent workflows no longer requires an MCP client. The
hidden `debug-resolve` subcommand dumps the namespace filter
`mem_agent_search` would resolve as JSON, given simulated
`AppContext` state — so multi-agent integration scripts can verify
namespace resolution end-to-end without standing up an MCP server.

**Stacked on top of #459 (PR-2: session→agent_id inheritance)** because
`debug-resolve` reuses the `_resolve_agent_namespace` helper added
there. Merge order: #459 first, then this PR.

## Commands

- `mm agent register <id> [--description ...] [--color ...]` — mirrors
  `mem_agent_register`. Sanitizes the agent_id, creates the
  `agent-runtime:<id>` namespace, ensures the cross-agent `shared`
  namespace exists.
- `mm agent list [--json]` — table by default (agents section + shared
  section); `--json` for scripts. Non-agent namespaces (`default`,
  `claude-memory:*`, etc.) are intentionally not surfaced — the
  command is scoped to the multi-agent surface.
- `mm agent share <chunk-id> [--target ...]` — mirrors `mem_agent_share`.
  Reuses `_build_shared_tags` from #458 (with a fallback for branches
  that pre-date it) so the audit-tag dedup contract stays in lock-step.
- `mm agent debug-resolve [--agent-id ...] [--current-agent-id ...]
  [--current-namespace ...] [--include-shared/--no-include-shared]`
  — hidden, JSON-only. Re-uses `_resolve_agent_namespace` so the CLI
  reports exactly what the MCP tool would compute given the same
  inputs. Designed for use in PR-6's integration scripts and the
  manual e2e in the multi-agent guide.

## Test plan

- [x] 14 cases in `test_agent_cmd.py`:
  - `TestAgentRegister` — creates NS + ensures shared, skips shared
    when it already exists, rejects empty agent_id.
  - `TestAgentList` — table grouping, JSON shape, empty state.
  - `TestAgentDebugResolve` — explicit-id + shared, `current_agent_id`
    fallback, legacy `current_namespace` fallback, all-None → null
    filter.
- [x] CliRunner mocks `cli_components` per
  `feedback_clirunner_isatty_seam` guidance.
- [x] `uv run pytest -m "not ollama"` — 2349 passed, 46 deselected.
- [x] `uv run ruff check` + `ruff format --check` — clean.

The `share` command itself is exercised end-to-end once PR-6 lands
(needs real `index_engine` wiring); the unit-style coverage above
guards the command-construction surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,18 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   `BaseStore` (`aput` / `aget` / `alist_namespaces`); that surface
   remains a follow-up.
 
+- **`mm agent` CLI: `register`, `list`, `share`, plus hidden `debug-resolve`.**
+  Mirrors the `mem_agent_*` MCP tools so operators don't have to spin up
+  an MCP client for one-off agent setup. `mm agent list [--json]` groups
+  registered agents (`agent-runtime:` namespaces) with the cross-agent
+  `shared` namespace; `mm agent share <chunk-id> [--target ...]`
+  performs the same content copy + `shared-from=<src>` audit tag as the
+  MCP tool. The hidden `mm agent debug-resolve` dumps the namespace
+  filter `mem_agent_search` would resolve given simulated
+  `current_agent_id` / `current_namespace` / `--include-shared` inputs,
+  as JSON — for use in multi-agent integration scripts so they can
+  assert namespace resolution without standing up an MCP client.
+
 - **`mem_import` gains `on_conflict` and `preserve_ids` (bundle schema v2).**
   Bundles now carry per-chunk `chunk_id` + `content_hash` so importers can
   dedup by content across instances. `on_conflict` accepts `"skip"`
--- a/packages/memtomem/src/memtomem/cli/agent_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/agent_cmd.py
@@ -3,16 +3,20 @@
 from __future__ import annotations
 
 import asyncio
+import json as _json
 from typing import TYPE_CHECKING
 
 import click
 
-from memtomem.constants import AGENT_NAMESPACE_PREFIX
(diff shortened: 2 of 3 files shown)


memtomem:
**Merge-order coordination note** (raised in #457 review): once #457
lands on main, this PR should rebase and replace the local
`_CURRENT_PREFIX = "agent-runtime:"` and `_SHARED_NAMESPACE = "shared"`
constants in `cli/agent_cmd.py` with imports from `memtomem.constants`
(which #457 introduces as the canonical source for `AGENT_NAMESPACE_PREFIX`
and `SHARED_NAMESPACE`). The hardcoded literals in this PR were
necessary because the branch was opened before `constants.py` existed,
but leaving them after #457 lands would re-create the parallel-literal
seam that #457's `feedback_drift_close_must_derive` audit just closed.

I'll handle the rebase + literal replacement before requesting merge here.

memtomem:
Done — rebased onto the post-#457 main (via the rebased #459 base) and
replaced the local literals with imports from `memtomem.constants`:

* `_SHARED_NAMESPACE = "shared"` → dropped; `SHARED_NAMESPACE` imported
  from `memtomem.constants` and used at all 9 call sites
  (`get_namespace_meta`, click `--target` default, debug-resolve filter
  composition, etc.).
* `_CURRENT_PREFIX` kept as a local alias paired with `_LEGACY_PREFIX`
  so the migration mapping reads as `(old, new)`, but the value derives
  from `AGENT_NAMESPACE_PREFIX` (no parallel literal).

`uv run pytest -m "not ollama"` — 2355 passed, 46 deselected. CI
re-running on the force-pushed branch.

PR #475, opened 2026-04-25
fix(sessions): mem_session_start derives namespace from agent_id

tsdata:
## Summary

`mem_session_start(agent_id="planner")` previously stored the session row with `namespace="default"` because `session.py:82` only consulted explicit `namespace=`, then `app.current_namespace`, then `"default"`. The supplied `agent_id` was set on `AppContext.current_agent_id` for downstream `mem_agent_search` resolution but never flowed into the session row itself. The LangGraph adapter `MemtomemStore.start_agent_session` already auto-derived `agent-runtime:<id>`; the MCP tool now matches.

New priority order:

1. explicit `namespace=` (escape hatch)
2. `agent-runtime:<agent_id>` when `agent_id != "default"`
3. `app.current_namespace` (pre-multi-agent fallback)
4. `"default"`

Only the session record's namespace field is derived — `app.current_namespace` is **not** touched, so `mem_add` / `mem_search` without explicit `namespace=` keep the legacy fallback. The "namespace and agent_id are separate axes on AppContext" invariant is preserved.

## Backward compat

- Callers passing explicit `namespace=` are unaffected (priority 1).
- Callers not passing `agent_id` (defaults to `"default"`) are unaffected (priority 3 fallback).
- Only `agent_id=<non-default>` without `namespace=` sees new behavior, and the new behavior matches what the LangGraph adapter already does.

## Why now

Caught while walking the v0.1.28 multi-agent test scenarios from `memtomem-docs` — the tester ran `mem_session_start(agent_id="planner", title="Multi-agent guide review")` and saw:

```
Session started: 89b2c742-...
- Title: Multi-agent guide review
- Agent: planner
- Namespace: default
```

Reasonable expectation: if `agent_id` is recorded for downstream tools, it should also drive the session's namespace. The MCP/adapter divergence becomes confusing the moment a tester switches between LangGraph code and the raw MCP tool.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_sessions.py -v` — 14 / 14 pass (10 existing + 4 new in `TestSessionNamespaceDerivation`)
- [x] `uv run pytest -m "not ollama"` — 2445 / 2445 pass, 46 deselected, no regressions
- [ ] Manual smoke: `mm session start --agent-id planner` → `mm session list` shows `agent-runtime:planner` namespace

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,29 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Changed
+
+- **`mem_session_start(agent_id="<id>")` now derives the session
+  namespace from `agent_id` when the caller doesn't pass an
+  explicit `namespace=`.** Previously the session record's
+  namespace fell back to `app.current_namespace` or `"default"`,
+  ignoring the supplied `agent_id` even though the LangGraph
+  adapter `MemtomemStore.start_agent_session` already auto-derived
+  `agent-runtime:<id>`. The MCP tool now matches that semantic.
+  Priority order: explicit `namespace=` > `agent-runtime:<id>`
+  (when `agent_id != "default"`) > `app.current_namespace` >
+  `"default"`. Only the **session record's** namespace field
+  changes; `app.current_namespace` is untouched, so subsequent
+  `mem_add` / `mem_search` without explicit `namespace=` keep the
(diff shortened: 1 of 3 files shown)


memtomem (approved):
LGTM — small focused fix; priority ladder cleanly documented in docstring + CHANGELOG; 4 tests pin each tier; back-compat preserved. Manual smoke deferred to post-merge.

PR #478, opened 2026-04-25
fix(server): correct mem_agent_share example in instructions= field

tsdata:
## Summary

The recipe shipped in #477 showed `mem_agent_share(memory_id=...)`
but the real signature is `mem_agent_share(chunk_id=..., target=...)`
(see `packages/memtomem/src/memtomem/server/tools/multi_agent.py:162`).

**Impact, reproduced via 4-spawn A/B**:

- 2 worktrees (`v0.1.29` baseline vs current `main` w/ instructions=)
- 2 prompts (structured "set up multi-agent isolation" vs ambiguous
  "two AI helpers, keep notes private until handed off")
- 2 models (Sonnet 4.5, Haiku 4.5)
- Same `claude -p` harness, `--mcp-config` pointing at each worktree's
  `python -m memtomem.server`

| Setup | Symptom |
|---|---|
| Sonnet × structured | inferred past the typo, used `chunk_id`+`target` directly |
| Sonnet × ambiguous | same — emergent recovery |
| Haiku × ambiguous | **followed the typo verbatim** → arg-name reject → `mem_do(action="help")` → retried with correct args (extra round trip on the critical share path) |

So the bug is silent on stronger models but adds a recovery hop on
weaker ones. Sonnet's robustness is also why the existing pin test
+ E2E test passed: they only check token *presence*, not signature
parity.

## What changes

- **`server/instructions.py:33`** — `memory_id=...` → `chunk_id=..., target="shared"`,
  comment updated to "copy chunk to shared scope" (matches the
  function docstring's clarification that this is a content copy,
  not a reference link).
- **`tests/test_server_instructions.py`** — new `test_instructions_kwargs_match_signature`
  parametrized over the five multi-agent / session functions
  documented in INSTRUCTIONS (`mem_agent_register`, `mem_agent_search`,
  `mem_agent_share`, `mem_session_start`, `mem_session_end`). It uses
  `inspect.signature(...)` to extract real parameter names and
  cross-checks every `name(arg=...)` form found in INSTRUCTIONS via
  regex. The MCP `ctx` parameter is excluded.
- **CHANGELOG** — `[Unreleased] → Fixed` entry.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_server_instructions.py -v` — 17/17 pass (was 12; +5 parametrized signature-parity cases)
- [x] `uv run ruff check` + `ruff format --check` on changed files — clean
- [x] Sanity: temporarily reverted `chunk_id` → `memory_id` in instructions.py; new test fails with `INSTRUCTIONS shows mem_agent_share(memory_id=...) but the real signature has parameters ['chunk_id', 'target']`
- [ ] Full `pytest -m "not ollama"` in CI

## Why now

The bug is unreleased (PR #477 hasn't shipped to PyPI yet — current
`mm --version` is still 0.1.29). Fixing before the v0.1.30 cut
keeps the changelog narrative clean: the `instructions=` feature
ships correct rather than as "added in 0.1.30, fixed in 0.1.30.1".

## Out of scope

- Adding `agent_id="planner"` example for `mem_agent_register` etc. —
  out of scope; the new signature test would catch any future drift
  there.
- Deepening INSTRUCTIONS into a full reference (`USAGE-multi-agent.md`)
  — the A/B data showed weak ROI for general "deepen" beyond fixing
  this factual error; deferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Fixed
+
+- **`mem_agent_share` example in server `instructions=` had wrong
+  parameter name.** The recipe shipped in #477 showed
+  `mem_agent_share(memory_id=...)` but the real signature is
+  `mem_agent_share(chunk_id=..., target=...)`. Weak LLMs (reproduced
+  on Claude Haiku 4.5) followed the example verbatim, hit a
+  parameter-name rejection, then `mem_do(action="help")`-recovered to
+  the correct args — burning an extra round trip per multi-agent
+  share. Stronger LLMs (Sonnet 4.5+) inferred past the typo. Fixed
+  the example and tightened `tests/test_server_instructions.py` with
+  a signature-parity check: any `tool(arg=...)` form in INSTRUCTIONS
+  is now cross-checked against `inspect.signature(...)`, so future
+  parameter renames force the example to follow.
+
 ### Added
 
 - **MCP `initialize` response now carries server-level `instructions`.**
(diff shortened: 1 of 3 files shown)


memtomem (approved):
LGTM. Reviewed: 1-line fix matches the actual signature, parity pin test prevents recurrence, scope tight. CI green.

PR #483, opened 2026-04-26
feat(context)!: codex_agents project-scope default

pandas-studio:
## Summary

`mm context sync` previously fanned out Codex sub-agents to **user-scope** `~/.codex/agents/<name>.toml` regardless of the project the command ran from. This PR switches the default to **project-scope** `<project>/.codex/agents/<name>.toml`, making `codex_agents` symmetric with `claude_agents` and `gemini_agents`.

This is a **BREAKING CHANGE** but pre-1.0 and well-bounded — see the migration note in `CHANGELOG.md`.

## Why

User-scope-by-default had three concrete problems:

1. **Cross-project collision.** Two projects with the same canonical agent name silently overwrote each other in `~/.codex/agents/`.
2. **Worktree leakage.** Running `mm context sync` from a `git worktree` clone wrote real files into the host's home directory, defeating the isolation worktrees are supposed to provide.
3. **Asymmetry.** `claude_agents` and `gemini_agents` already wrote to project-scope. `codex_agents` was the only generator that ignored its `project_root` argument — fragile and confusing.

Upstream support is verified: the [OpenAI Codex CLI subagents docs](https://developers.openai.com/codex/subagents) explicitly say "standalone TOML files under `~/.codex/agents/` for personal agents or `.codex/agents/` for project-scoped agents." Project scope is the right default for a per-repo canonical source; user-scope can come back as an explicit `--scope=user` flag in a future PR.

## What changed

- `CodexAgentsGenerator.target_file()` now uses `project_root / .codex/agents / <name>.toml` (was `Path.home() / .codex/agents / ...`).
- `_runtime_agent_names("codex_agents", ...)` for `diff_agents()` reads the project-scope path.
- `AGENT_DIRS` adds `"codex_agents": [".codex/agents"]`; new `AGENT_FILE_SUFFIX` so `detect_agent_dirs` can match `.toml` alongside `.md`. `mm context detect` and the Web UI Context Gateway now list project-scope Codex agents.
- Module/function docstrings updated; `_names.py` example path corrected.
- 13 test assertions moved from `codex_home / ".codex/agents/..."` → `tmp_path / ".codex/agents/..."`. Two of those gain a defensive `assert not (codex_home / ".codex/agents").exists()` so a regression that called `Path.home()` would be caught. New `test_detects_codex_project_scope` covers the new detect path; the existing `test_codex_user_scope_not_in_project_detect` is kept (still important — user-scope must NOT be auto-detected).

## Migration

Existing `~/.codex/agents/<name>.toml` files are **left untouched** — memtomem just stops writing to that path. If you relied on user-scope output, copy what you want to keep:

```bash
cp ~/.codex/agents/*.toml <project>/.codex/agents/
```

A `--scope=user` flag is on the roadmap for users who genuinely want to keep writing to `~/.codex/agents/` (tracked separately as PR-8 of the context-gateway audit plan).

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_context_agents.py` — 61 passed (was 60; +1 = `test_detects_codex_project_scope`)
- [x] `uv run pytest -m "not ollama"` — 2477 passed (main: 2476; +1 from above)
- [x] `uv run ruff check …` + `ruff format --check …` — clean
- [x] End-to-end repro with `HOME=/tmp/throwaway uv run python -c "generate_all_agents(...)"` — writes only to `<project>/.codex/agents/sample.toml`; `HOME` directory is never created or touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,34 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Changed (BREAKING)
+
+- **`mm context sync` now writes Codex sub-agents to project-scope
+  `.codex/agents/<name>.toml` instead of user-scope
+  `~/.codex/agents/<name>.toml`.** Previously every `mm context sync`
+  wrote into the user's home directory regardless of which project the
+  command ran from, so two projects with same-named canonical agents
+  silently overwrote each other in `~/.codex/agents/`, and any
+  `mm context sync` from a worktree leaked onto the real host home —
+  defeating worktree isolation. The OpenAI Codex CLI [supports both
+  scopes](https://developers.openai.com/codex/subagents) ("standalone
+  TOML files under `~/.codex/agents/` for personal agents or
+  `.codex/agents/` for project-scoped agents"), and project scope is
+  the right default for a per-repo canonical source. This makes
+  `codex_agents` symmetric with `claude_agents` (`.claude/agents/`)
+  and `gemini_agents` (`.gemini/agents/`).
+
+  **Migration:** if you relied on the old user-scope output, copy any
+  agents you want to keep with
+  `cp ~/.codex/agents/*.toml <project>/.codex/agents/`, or restore
+  user-scope behavior on a future `mm context sync` once `--scope=user`
+  ships (tracked separately). Existing `~/.codex/agents/` files are
+  left in place; memtomem just stops writing there.
+
+  `detect_agent_dirs` now also picks up `.codex/agents/<name>.toml`
+  alongside Claude/Gemini, so `mm context detect` and the Web UI
+  Context Gateway list project-scope Codex agents in their inventory.
+
 ### Fixed
 
 - **`mm context` round-trip no longer silently drops `## <Agent>-Specific`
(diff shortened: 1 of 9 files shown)


pandas-studio:
Thanks for the thorough review — addressed all five items in babda0f:

**Must-fix**
- ✅ Item 1 — doc drift across 3 files
  - `docs/adr/0001-context-gateway-sync-policies.md` — split agents (project-scope) from prompts (user-scope); both still one-way / never imported.
  - `docs/guides/integrations/claude-code.md:350` and `docs/guides/getting-started.md:412` — example paths updated to `.codex/agents/`.
- ✅ Item 2 — `test_web_routes_context.py:552` comment rewritten to describe the actual purpose of the canonical filter (runtime-only entries without a `canonical_path`), no longer mentions user-scope Codex.

**Nice-to-haves**
- ✅ Item 3 — added `TestRuntimeAgentNames` with 5 cases that pin `_runtime_agent_names()` directly: codex project-scope read, ignores user-scope leak, claude/gemini parity, unknown runtime, missing dir.
- ✅ Item 4 — CHANGELOG migration block now tells users to delete the old `~/.codex/agents/*.toml` after copying (Codex CLI still loads them, memtomem no longer manages them, so they drift).
- ✅ Item 5 — `AGENT_FILE_SUFFIX` locked to `AGENT_DIRS` keys via a module-level `assert` (same style as `config.py` / `init_cmd.py`), and the lookup switched from `.get(agent, ".md")` to `[agent]` so a forgotten suffix fails loudly at import.

Local check: `uv run ruff check && ruff format --check` clean, 2482 tests pass.

PR #484, opened 2026-04-26
fix(context): confirm before settings sync writes outside the project

pandas-studio:
## Summary

- `mm context {generate,sync} --include=settings` now confirms before
  writing to a settings file that lives outside the project root
  (today only `~/.claude/settings.json`); declining leaves the host
  untouched.
- `--yes` / `-y` skips the prompt for scripted callers.
- Path-based gate (`target.is_relative_to(root)`) so future
  project-scope settings generators won't prompt at all.

## Why

Running `mm context sync --include=settings` from a worktree silently
edited the real `~/.claude/settings.json`. `setup-worktree.sh`
redirects storage and memory dirs but not HOME, so the host
pollution was easy to miss. This is one of the three P0 items called
out in the 2026-04-26 context-gateway audit (alongside the
round-trip data loss already fixed in #482 and the codex_agents
project-scope fix in #483).

## Test plan

- [x] New regression class `TestClaudeSettingsHostWritePrompt` in
      `test_context_settings.py` (7 cases): sync/generate × {decline,
      confirm, `--yes`} + 2 negatives (missing canonical / runtime
      unavailable) that assert no prompt fires.
- [x] `uv run pytest packages/memtomem/tests -m "not ollama"` — 2488
      passed locally.
- [x] `uv run ruff check` + `ruff format --check` clean.
- [x] `uv run mypy packages/memtomem/src/memtomem/cli/context_cmd.py`
      — no issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Added
+
+- **`mm context {generate,sync} --include=settings` now confirms before
+  writing settings files outside the project root** (today only
+  `~/.claude/settings.json`); pass `--yes` / `-y` to skip the prompt.
+  The same gate is enforced inside `generate_all_settings`, so callers
+  that bypass the CLI — `mem_context_generate` / `mem_context_sync`
+  (MCP) and `POST /settings-sync` (Web) — also refuse host writes
+  unless they pass `allow_host_writes=true`. Closes one of the
+  2026-04-26 audit P0 items: a stray sync from a worktree no longer
+  silently edits the real home directory regardless of which front-end
+  drove it. Symlink-aware (`Path.resolve()` both sides) so a
+  symlinked `.claude` cannot smuggle a host write past the gate.
+
 ### Changed (BREAKING)
 
 - **`mm context sync` now writes Codex sub-agents to project-scope
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -32,9 +32,9 @@ from memtomem.context.generator import (
 )
 from memtomem.context.parser import CONTEXT_FILENAME, parse_context, sections_to_markdown
 from memtomem.context.settings import (
-    SETTINGS_GENERATORS,
     diff_settings,
     generate_all_settings,
+    host_write_targets,
 )
 from memtomem.context.skills import (
     diff_skills,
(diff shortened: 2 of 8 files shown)


memtomem:
Thanks for the thorough review — addressed all 4 in-scope items in 3ed42f7. The big one (item 1) reshapes the API; rebased on the new main first.

**Item 1 — the gate now lives in `generate_all_settings`** (the I/O boundary). New keyword-only `allow_host_writes: bool = False` param: targets whose resolved path is outside the resolved project root come back with `status="needs_confirmation"` instead of being written. Front-ends each pass `True` only after acknowledgement:
- **CLI** — same interactive confirm / `--yes` / `-y` as before, threaded into `generate_all_settings(allow_host_writes=...)`.
- **MCP** — `mem_context_sync` and `mem_context_generate` accept `allow_host_writes: bool = False`. Default surfaces a `needs confirmation <name>: <path>` line so the agent can re-call after asking the user.
- **Web** — `POST /settings-sync` now takes a body `{"allow_host_writes": false}`. Default refuses; UI is expected to list targets and re-post with `true`.

Also exported a `host_write_targets(project_root)` helper so any caller (CLI prompt, MCP tool author, Web modal) gets the same list of pending host paths without re-implementing the rule.

**Item 2** — `--yes` help text now generic: `Skip confirmation prompts before writing settings files outside this project.`

**Item 3** — `CHANGELOG.md` `[Unreleased]` `Added` entry covers the prompt, `--yes`, the lib gate, the MCP/Web coverage, and the symlink fix.

**Item 4** — `Path.resolve(strict=False)` on both target and project root, so a symlinked `<project>/.claude → ~/.claude` no longer smuggles a host write past the gate. There's a regression test for that exact case.

(Item 5 — diff preview before the prompt — left as a follow-up; out of scope here.)

**Tests added** (9 lib + 4 MCP + 2 Web = 15 new cases):
- `TestGenerateAllSettingsHostWriteGate` (lib): default refuses; `allow_host_writes=True` proceeds; `diff_settings` unaffected; no-runtime / no-canonical → `skipped` (not `needs_confirmation`); `host_write_targets` lists/empties; symlink resolves to host write.
- `test_server_tools_context_settings_gate.py` (MCP): refuse / accept × `mem_context_sync` and `mem_context_generate`.
- `TestSettingsSync` (Web): POST default refusal; POST with `{"allow_host_writes": true}` proceeds.

**API note:** `generate_all_settings` now defaults to refusing host writes. Any library caller that imported it without the new kwarg goes from "always writes" to "needs_confirmation" — that's the desired behavior change. Captured in CHANGELOG.

Local: `ruff check` + `ruff format --check` + `mypy` clean on the four changed source files; `2509 passed` on the full suite.

PR #485, opened 2026-04-26
fix(cli): derive agent-runtime namespace in `mm session start --agent-id`

memtomem:
## Summary
PR #475 fixed `mem_session_start(agent_id=...)` for the MCP entry point but left the CLI behind: `mm session start --agent-id planner` silently lost `--agent-id` for namespace derivation and the session landed in `default`, contradicting the multi-agent contract the public page advertises. This is gap **G2** from the 2026-04-26 multi-agent surface review (memtomem-docs#13).

## Behavior change
`_start` in `cli/session_cmd.py` now follows the same priority chain as `mem_session_start`, with one step omitted by design (CLI has no process-local `app.current_namespace`):

1. explicit `--namespace` wins (escape hatch)
2. `agent-runtime:<agent-id>` when `--agent-id` is non-default
3. `default`

`mm session start` also echoes the resolved `Namespace:` alongside session id / agent / title, so users can verify the derivation before continuing — same shape as MCP `mem_session_start` already returns.

## Why this is mechanical
- Mirrors the post-PR-#475 docstring contract verbatim (`server/tools/session.py:76–95`); no design choice needed.
- Independent of G1 (the `mem_add` agent-scope flow RFC, `mem-add-agent-scope-flow-rfc.md`) — G2 only fixes the *session-record* namespace; whether subsequent `mem_add` calls inherit the agent scope is a separate (still RFC-stage) question.
- Safe against the deferred grouping decision (2026-05-09): only argument resolution changes, no surface added or removed.

## Test plan
- [x] `uv run pytest packages/memtomem/tests/test_session_cli.py` — 16 passed (13 existing + 3 new in `TestSessionStartNamespaceDerivation`: default agent → `default` ns; `--agent-id planner` → `agent-runtime:planner`; explicit `--namespace` overrides agent-id)
- [x] `uv run pytest packages/memtomem/tests/test_sessions.py packages/memtomem/tests/test_multi_agent.py packages/memtomem/tests/test_multi_agent_integration.py` — 48 passed (no regression in MCP / multi-agent paths)
- [x] `uv run ruff check` + `ruff format --check` — clean
- [ ] Manual smoke: `uv run mm session start --agent-id planner` → echoes `Namespace: agent-runtime:planner`; `mm session list --json` confirms the stored namespace

## Cross-references
- Source: `multi-agent-public-surface-review-2026-04-26.md` §3 (gap G2) + §5.1 (sequencing)
[...]

PR #491, opened 2026-04-26
fix: validate agent_id before agent-runtime namespace concat (CLI + MCP)

### LINKED ISSUE #486: Validate agent_id before agent-runtime namespace concatenation (CLI + MCP)
## Context

PR #485 review (item 2) flagged that ``f\"{AGENT_NAMESPACE_PREFIX}{agent_id}\"`` does not validate ``agent_id`` before concatenation. Hostile-shaped values like ``foo:bar``, ``../x``, or whitespace produce malformed-but-stored namespace strings (\"agent-runtime:foo:bar\", etc.) that round-trip into storage and search.

The CLI side was mirrored from the MCP path (``server/tools/session.py:96-105``) and faithfully inherits the same gap — fixing only one surface would leave the other open.

## What we want

A single ``validate_agent_id(value)`` (or ``validate_name(value, kind=\"agent-id\")``) hook applied at one boundary so both ``mm session start`` / ``mm session wrap`` and ``mem_session_start`` raise a clear error when the agent id contains:

* ``:`` (clashes with the namespace separator)
* ``/`` ``..`` (path-traversal-shaped values)
* leading/trailing whitespace
* zero-width or control characters
* anything beyond the existing canonical agent-name charset (look at ``validate_name`` callers under ``context/_names.py`` as the closest precedent)

## Acceptance

* Reject the cases above with a single, identical error message in both the CLI and MCP entry points.
* Regression test that pins ``\"agent-runtime:foo:bar\"`` *cannot* end up in storage from either surface.
* Document the validator's accepted charset alongside ``AGENT_NAMESPACE_PREFIX`` in ``constants.py``.

## Out of scope

* Renaming or migrating any agent ids users might already have stored — this is a forward gate only.
* Tightening the surface review G2 follow-up further (\"different agent_id formats per runtime\" etc.). Track separately if it comes up.

## Provenance

* Review on PR #485 (item 2 — \"agent_id is not validated before namespace concatenation\")
* MCP origin: PR #475

pandas-studio:
## Summary

- Add ``validate_agent_id`` in ``constants.py`` (thin wrapper around ``validate_name(value, kind="agent-id")``).
- Wire it into all three entry points that concatenate ``AGENT_NAMESPACE_PREFIX`` with caller input: ``mem_session_start`` (MCP), ``mm session start``, and ``mm session wrap`` (CLI).
- Pin the wiring + the regression contract in ``test_validate_agent_id``: ``"agent-runtime:foo:bar"`` cannot reach storage from either surface.

## Why

PR #485 review (item 2) flagged that ``f"{AGENT_NAMESPACE_PREFIX}{agent_id}"`` does no input validation. Hostile-shaped values (``foo:bar``, ``../x``, ``  spaces``, control chars) round-trip into storage as malformed namespace strings. The CLI mirrored the gap from the MCP path so a single-surface fix would leave the other open.

Single boundary, single charset, single error message: ``[A-Za-z0-9._-]`` (1–64 chars, no leading dash, not ``"."`` / ``".."``). MCP path raises an ``InvalidNameError`` (a ``ValueError`` subclass) which ``tool_handler`` formats as ``"Error: invalid agent-id 'foo:bar': ..."``; CLI path converts the same exception to a ``ClickException`` with the identical core text.

``mm session wrap`` validates **before** the wrapped subprocess runs so an invalid agent_id never leaves a half-set ``.current_session`` file or a rogue child.

## Out of scope

- Renaming or migrating any agent_ids users may have already stored — this is a forward gate only.
- Tightening ``mem_agent_register`` / ``mem_agent_search`` (which still ``sanitize_namespace_segment`` rather than ``validate_*``); the issue scopes this PR to the session-start surfaces, and changing the multi_agent tools' silently-mutating behavior is a separate UX call.

## Test plan

- [x] ``uv run pytest -m "not ollama"`` — 2554 passed, 46 deselected (ollama).
- [x] ``uv run ruff check`` + ``ruff format --check`` on ``src/`` and ``tests/`` — clean.
- [x] New ``test_validate_agent_id``:
  - charset acceptance + 13 hostile-input rejections at the validator
  - MCP boundary: ``mem_session_start(agent_id="foo:bar")`` returns ``"Error: invalid agent-id ..."`` and storage stays empty
  - CLI boundary: ``mm session start --agent-id "foo:bar"`` exits non-zero, storage mock confirms ``create_session`` was never called
  - ``mm session wrap`` rejects before the wrapped command can run
  - cross-surface parity: both surfaces produce the same ``"invalid agent-id 'foo:bar'"`` core text

Closes #486.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/session_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/session_cmd.py
@@ -10,7 +10,7 @@ from pathlib import Path
 
 import click
 
-from memtomem.constants import AGENT_NAMESPACE_PREFIX
+from memtomem.constants import AGENT_NAMESPACE_PREFIX, InvalidNameError, validate_agent_id
 
 logger = logging.getLogger(__name__)
 
@@ -78,6 +78,10 @@ def session() -> None:
 @click.option("--namespace", "-n", default=None, help="Namespace for session")
 def start(agent_id: str, title: str | None, namespace: str | None) -> None:
     """Start a new session and save its ID to ~/.memtomem/.current_session."""
+    try:
+        validate_agent_id(agent_id)
+    except InvalidNameError as e:
+        raise click.ClickException(str(e)) from e
     try:
         asyncio.run(_start(agent_id, title, namespace))
     except click.ClickException:
@@ -337,6 +341,11 @@ def wrap(agent_id: str, title: str | None, command: tuple[str, ...]) -> None:
     """
     import subprocess
     import sys
+
+    try:
+        validate_agent_id(agent_id)
+    except InvalidNameError as e:
+        raise click.ClickException(str(e)) from e
 
     # Start session
     session_id = str(uuid.uuid4())
--- a/packages/memtomem/src/memtomem/constants.py
+++ b/packages/memtomem/src/memtomem/constants.py
@@ -12,6 +12,9 @@ from __future__ import annotations
 
 from typing import Final
 
+from memtomem.context._names import InvalidNameError as InvalidNameError
+from memtomem.context._names import validate_name
+
 # Multi-agent buckets. The ``agent-runtime:<agent-id>`` namespace is a
 # *convenience* isolation boundary, not a security boundary — see the
 # multi-agent guide for the threat model. Provider-ingestion namespace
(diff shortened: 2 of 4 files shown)


pandas-studio:
Review fixes pushed in eaa9d3d:

* Narrowed ``validate_agent_id`` docstring to the three surfaces actually wired in this PR + an explicit "Not yet applied at" list (LangGraph adapter, ``mem_agent_register`` / ``mem_agent_search``).
* Replaced literal U+200B in ``HOSTILE_AGENT_IDS`` with ``​`` escape so the test reads as intentional.
* ``test_session_wrap_rejects_colon_before_subprocess`` now patches ``subprocess.run`` directly and asserts ``run_mock.assert_not_called()`` — pins the contract its docstring claimed.

Concern 5 (``_StubCtx`` consolidation into ``conftest.py``) deferred per "fine for now."

Follow-up issues filed per the verdict:

* #492 — LangGraph adapter ``start_agent_session`` / ``MemtomemCheckpointer.namespace_for`` parity, plus replacing the duplicated ``_AGENT_NAMESPACE_PREFIX`` literal with the import from ``constants``.
* #493 — ``mem_agent_register`` / ``mem_agent_search`` validate-vs-sanitise divergence: pick one of "switch to validate" or "document the asymmetry," but pick deliberately.

memtomem (approved):

Issue #492, opened 2026-04-26
Apply validate_agent_id at LangGraph adapter agent-runtime concat sites


pandas-studio:
## Context

PR #491 added ``validate_agent_id`` and gated the three MCP + CLI session-start surfaces (``mem_session_start``, ``mm session start``, ``mm session wrap``). Hostile-shaped ``agent_id`` values now produce a clear error rather than round-tripping into storage as malformed ``agent-runtime:foo:bar`` namespace strings.

The LangGraph adapter (``packages/memtomem/src/memtomem/integrations/langgraph.py``) builds the same namespace shape but bypasses the gate:

* ``MemtomemStore.start_agent_session`` (around line 294) only checks ``agent_id`` is non-empty, then concatenates at line 299: ``ns = namespace or f"{_AGENT_NAMESPACE_PREFIX}{agent_id}"``.
* The same prefix is also concatenated at lines 171/173/190 (``MemtomemStore``) and inside ``MemtomemCheckpointer.namespace_for``.

Today the in-process Python caller can still produce ``agent-runtime:foo:bar`` records via the LangGraph adapter even though the MCP / CLI surfaces refuse them.

A separate, smaller drift point worth resolving in the same PR: ``langgraph.py:43`` redefines ``_AGENT_NAMESPACE_PREFIX = "agent-runtime:"`` instead of importing from ``memtomem.constants``. The redefinition predates the multi-agent shipping work and was kept on the assumption that the integration shouldn't import from MCP-side modules — but ``constants.py`` is intentionally above the MCP / CLI / integration split exactly so all three can derive from it.

## What we want

* Apply ``validate_agent_id`` at every site in ``langgraph.py`` that interpolates an ``agent_id`` into the agent-runtime namespace.
* Replace the local ``_AGENT_NAMESPACE_PREFIX`` literal with ``from memtomem.constants import AGENT_NAMESPACE_PREFIX`` (and ``validate_agent_id``).
* Update the validator docstring (``constants.py``) to reflect that the LangGraph adapter is now wired in.

## Acceptance

* All three concat sites in ``langgraph.py`` raise ``InvalidNameError`` (or its public alias) before the malformed namespace ever reaches storage.
* Regression test (alongside the existing ``test_validate_agent_id`` parity tests) pinning that ``MemtomemStore.start_agent_session(agent_id="foo:bar")`` cannot land an ``agent-runtime:foo:bar`` row.
[...]

PR #494, opened 2026-04-26
fix: validate agent_id at mem_agent_register / search and `mm agent register` (#493)

### LINKED ISSUE #493: mem_agent_register / mem_agent_search: validate vs sanitize divergence
## Context

PR #491 wired ``validate_agent_id`` into the three session-start surfaces so a malformed ``agent_id`` is loudly rejected:

```
mem_session_start(agent_id="foo:bar")
→ "Error: invalid agent-id 'foo:bar': must match [A-Za-z0-9._-]+ ..."
```

The same hostile shape, sent at the multi-agent registration / search surface, takes a different path:

```
mem_agent_register(agent_id="foo:bar")
→ silently runs sanitize_namespace_segment("foo:bar")
→ stores under sanitised ``agent-runtime:foo_bar`` (or similar)
```

(See ``packages/memtomem/src/memtomem/server/tools/multi_agent.py:71-75`` and ``:118-123`` plus ``packages/memtomem/src/memtomem/cli/agent_cmd.py:108-112``.)

So the same input now produces two UX outcomes depending on which tool the caller picked first — registering an agent rewrites the id in place, but starting a session for that same agent rejects it. That asymmetry is small today (multi-agent surface review just shipped) but it'll get harder to reason about as more tools fan out.

## What we want

A deliberate decision between two options, applied consistently across ``mem_agent_register``, ``mem_agent_search``, and the matching ``mm agent ...`` CLI subcommands:

1. **Switch to validate.** Replace ``sanitize_namespace_segment`` calls with ``validate_agent_id``. Same loud-reject behavior as the session-start surfaces. Breaking for any caller that was relying on silent sanitisation, but those callers were storing under unexpected namespaces anyway.
2. **Document the divergence.** Keep the silent sanitise behavior, but add a docstring + user-guide note explaining that registration / search are forgiving and that ``mem_session_start`` is the strict gate.

Option 1 is the cleaner long-term answer; option 2 is the cheaper short-term answer. Either way, the choice should be visible in the docstring + CHANGELOG so users aren't surprised.

## Acceptance

* One of the two paths above shipped, not "we'll figure it out later."
* If option 1 (validate): regression tests pin the error parity with the session-start surfaces; CHANGELOG flags the breaking shape.
* If option 2 (document): docstrings + ``docs/guides/multi-agent-quickstart.md`` (or equivalent) note the read/write asymmetry; the validator's docstring in ``constants.py`` stays honest about which surfaces it covers.

## Out of scope

* The LangGraph adapter parity gap — tracked in its own follow-up issue.
* Adding a new "loose vs strict" config knob — that adds an axis we don't actually need until users ask for it.

## Provenance

* PR #491 review (concern 2): ``mem_agent_register`` / ``mem_agent_search`` semantic divergence with the new session-start gate.
* Same review's verdict explicitly recommends filing this separately so it doesn't drift quietly.

pandas-studio:
## Summary

Closes #493.

PR #491 wired `validate_agent_id` into the three session-start surfaces
(`mem_session_start` / `mm session start` / `mm session wrap`) so a
hostile-shaped `agent_id` like `"foo:bar"` or `"../x"` was rejected
loudly. The matching multi-agent registration / search surfaces
(`mem_agent_register`, `mem_agent_search`, `mm agent register`) still
ran `sanitize_namespace_segment`, so the same input produced two UX
outcomes — registering an agent silently rewrote the id to
`agent-runtime:foo_bar`, while `mem_session_start` for that same id
loudly rejected. This PR closes that asymmetry.

The read / write contract is now symmetric across the MCP + CLI
surfaces — an `agent_id` either works on every surface or fails on
every surface, with the same `invalid agent-id 'X'` error fragment
regardless of entry point. `sanitize_namespace_segment` itself stays
as the permissive helper for the ingest pipeline (where filesystem /
provider segments still need rewriting); only the multi-agent
surfaces stop calling it.

## Breaking change

Callers that relied on silent sanitisation will now see a hard error.
Those callers were already storing memories under unexpected
namespaces (e.g. `agent-runtime:foo_bar` from `agent_id="foo:bar"`)
so the change is the forcing function to pick a canonical id matching
`[A-Za-z0-9._-]+`. Flagged in `CHANGELOG.md` under
`[Unreleased] > Changed (BREAKING)`.

## What changed

- `packages/memtomem/src/memtomem/server/tools/multi_agent.py` —
  `mem_agent_register` and `mem_agent_search` call `validate_agent_id`
  (raises `InvalidNameError`, surfaced as `Error: ...` by
  `tool_handler`); drop the now-redundant non-empty / sanitised-empty
  branches; drop the `sanitize_namespace_segment` import.
- `packages/memtomem/src/memtomem/cli/agent_cmd.py` — `mm agent
  register` validates with the session-start `try/except InvalidNameError
  → ClickException` pattern.
- `packages/memtomem/src/memtomem/constants.py` — `validate_agent_id`
  docstring lists every MCP + CLI surface that now validates, points
  at `CHANGELOG.md` Unreleased + #493 for the breaking-change note,
  and keeps the LangGraph adapter under "Not yet applied at" (tracked
  in #492 — out of scope here).
- `CHANGELOG.md` — `Unreleased > Changed (BREAKING)` entry.

## Tests

- `packages/memtomem/tests/test_validate_agent_id.py` adds:
  - `TestMcpMultiAgentBoundary` — parametrised over the existing
    `HOSTILE_AGENT_IDS` corpus (13 shapes); pins error string + that
    `agent-runtime:foo:bar` never reaches `list_namespace_meta`; valid
    ids still succeed; `agent_id=None` for `mem_agent_search` is
    unaffected.
  - `TestCliAgentRegisterBoundary` — `mm agent register foo:bar` /
    `../etc` reject; storage `set_namespace_meta` never called on the
    rejection path; valid id still succeeds.
  - `test_session_and_multi_agent_surfaces_share_core_error_text` —
    cross-surface parity pin: all three boundaries emit
    `invalid agent-id 'foo:bar'`.
- `packages/memtomem/tests/test_agent_cmd.py::test_register_rejects_empty_agent_id`
  retargeted to the new shared error vocabulary (`invalid agent-id`)
  + storage-not-called assertion.
- `packages/memtomem/tests/test_multi_agent.py` docstring corrected —
  the file now exercises the helper itself, not its (former) use by
  the multi-agent tools.

## Test plan

- [x] `uv run pytest packages/memtomem/tests -m "not ollama"` — 2589 passed, 46 deselected
- [x] Targeted: `uv run pytest packages/memtomem/tests/test_validate_agent_id.py` — 70 passed

## Out of scope

- LangGraph adapter parity for `MemtomemStore.start_agent_session` and
  the remaining agent-runtime concat sites — tracked in #492 as a
  separate follow-up (review feedback: this PR shouldn't claim
  LangGraph validation it doesn't ship).
- `mem_agent_share(target=...)` validation — `target` is a full
  namespace string (not a bare `agent_id`), so `validate_agent_id`
  doesn't drop in directly. Currently a caller passing
  `target="agent-runtime:foo:bar"` plumbs straight into
  `_mem_add_core(namespace=target)` and stores under exactly the
  hostile namespace this PR closes for register / search. Worth a
  follow-up issue under the same #493 banner — see review thread.
- New "loose vs strict" config knob — adds an axis we don't actually
  need until users ask for it (per #493 acceptance).

## Review fixups (post-#494 review)

- `94e9422` — drop LangGraph claim from `validate_agent_id` docstring.
  The original PR docstring listed
  `integrations.langgraph.MemtomemStore.start_agent_session` under
  "Applied at", but the LangGraph wiring is not in this PR (it lives
  in the stacked-PR work for #492). Until #492 lands this PR shouldn't
  claim coverage it doesn't ship — the LangGraph mention is now under
  "Not yet applied at" with a concrete pointer to #492. The vague
  "see issue tracker history" history pointer is also replaced with
  a concrete `CHANGELOG.md` Unreleased + #493 anchor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,31 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Changed (BREAKING)
+
+- **`mem_agent_register`, `mem_agent_search`, and `mm agent register`
+  now reject malformed `agent_id` values loudly instead of silently
+  rewriting them.** PR #491 had wired `validate_agent_id` into the
+  three session-start surfaces (`mem_session_start` / `mm session
+  start` / `mm session wrap`) so a hostile shape like `"foo:bar"` or
+  `"../x"` raised
+  `Error: invalid agent-id 'foo:bar': must match [A-Za-z0-9._-]+ ...`.
+  The matching multi-agent registration / search surfaces still ran
+  `sanitize_namespace_segment`, so the same input produced two UX
+  outcomes depending on which tool the caller hit first — registering
+  an agent rewrote the id in place under `agent-runtime:foo_bar`,
+  while starting a session for that same id rejected it. The read /
+  write contract is now symmetric: an `agent_id` either works on
+  every surface or fails on every surface, with the same
+  `invalid agent-id 'X'` error fragment regardless of entry point.
+
+  **Migration:** callers passing values that were previously rewritten
+  (anything containing `:`, `/`, `..`, whitespace, or characters
+  outside `[A-Za-z0-9._-]`) will now see a hard error. Those callers
+  were already storing memories under unexpected namespaces — pick a
+  canonical id matching the documented charset and re-register. The
+  LangGraph adapter parity gap remains tracked separately. Closes #493.
+
 ## [0.1.31] — 2026-04-26
 
 ### Added
--- a/packages/memtomem/src/memtomem/cli/agent_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/agent_cmd.py
@@ -8,7 +8,12 @@ from typing import TYPE_CHECKING
 
 import click
 
-from memtomem.constants import AGENT_NAMESPACE_PREFIX, SHARED_NAMESPACE
+from memtomem.constants import (
+    AGENT_NAMESPACE_PREFIX,
+    SHARED_NAMESPACE,
+    InvalidNameError,
+    validate_agent_id,
+)
 
 if TYPE_CHECKING:
     from memtomem.storage.sqlite_backend import SqliteBackend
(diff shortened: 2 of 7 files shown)


pandas-studio:
Thanks — addressed the blocking review item:

**#1 (LangGraph overclaim) — fixed in 94e9422**

You were right, the docstring as written claimed coverage this PR doesn't ship. Moved the LangGraph mention back under `Not yet applied at` (mirroring `main` pre-#493) and pointed at #492 explicitly. Also replaced the vague "see issue tracker history" pointer with a concrete `CHANGELOG.md` Unreleased + #493 anchor. The stacked-PR work for #492 will re-promote LangGraph to `Applied at` when it lands. Updated the PR description to drop the matching "LangGraph adapter unaffected (already validates as of #491 land)" sentence — same overclaim.

**#2 (mem_agent_share target validation) — filed as #497**

`target=` is a full namespace string (not a bare `agent_id`), so `validate_agent_id` doesn't drop in directly — needs a "split prefix, validate the suffix when prefix == AGENT_NAMESPACE_PREFIX" gate. Sibling gap to #493 but worth its own scope. Acceptance criteria mirror this PR (cross-surface error parity, storage never sees malformed namespace, docstring update). Linked from this PR's `Out of scope` list.

**#3 (vague history pointer) — fixed in 94e9422 alongside #1.**

**#4 (validate return value discarded) and #5 (test error string change polish)** — leaving as-is per your "acceptable / optional" verdict; the cross-surface error vocabulary unification was the intentional outcome and the empty-id case is covered transitively by the BREAKING entry's "contains... whitespace" line.

Ready for re-review when you have a minute.

memtomem (approved):

PR #495, opened 2026-04-26
fix: validate agent_id at LangGraph adapter agent-runtime concat sites

### LINKED ISSUE #492: Apply validate_agent_id at LangGraph adapter agent-runtime concat sites

pandas-studio:
## Summary

- Closes the parity gap PR #491 review flagged: the LangGraph adapter's `MemtomemStore.start_agent_session` was the last entry point that built `agent-runtime:<agent_id>` from caller input without running `validate_agent_id`. Apply the gate so a Python caller can no longer land `"agent-runtime:foo:bar"` in storage even though the MCP / CLI surfaces refuse the same shape.
- Replace the local `_AGENT_NAMESPACE_PREFIX = "agent-runtime:"` literal with the `memtomem.constants` import (`AGENT_NAMESPACE_PREFIX`, `SHARED_NAMESPACE`, `validate_agent_id`) so the integration derives from the single source of truth — the redefinition predated the multi-agent shipping work and was kept on the assumption that integrations shouldn't import from MCP-side modules, but `constants.py` is intentionally above the MCP / CLI / integration split exactly so all three can derive from it. `langgraph.py → constants.py → context._names` stays one-way.
- Stacked on PR #494 (#493) so the `validate_agent_id` docstring there ("Applied at every surface that concatenates `AGENT_NAMESPACE_PREFIX` with caller input") becomes accurate the moment both PRs land.

The other concat sites in `langgraph.py` (`_resolve_search_namespace`, `_resolve_add_namespace`) only read `self._current_agent_id`, which is set exclusively by `start_agent_session` — gating that single entry point covers the whole agent-runtime concat path.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_langgraph.py packages/memtomem/tests/test_validate_agent_id.py packages/memtomem/tests/test_multi_agent_integration.py packages/memtomem/tests/test_multi_agent.py packages/memtomem/tests/test_agent_cmd.py -m "not ollama"` — 149 passed
- [x] `uv run ruff check` + `ruff format --check` on the two changed files
- [x] `test_empty_agent_id_raises` updated to expect `InvalidNameError` ("invalid agent-id '': empty") rather than the old "non-empty" `ValueError`
- [x] New `test_hostile_agent_id_blocked_before_storage` parametrises the same hostile shapes pinned at the MCP / CLI boundaries (`foo:bar`, `../etc`, `a/b`, `a b`, `-leading-dash`) and asserts `storage.create_session` is never awaited and `_current_agent_id` stays unset on the rejection path

Closes #492.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,19 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   canonical id matching the documented charset and re-register. The
   LangGraph adapter parity gap remains tracked separately. Closes #493.
 
+- **`MemtomemStore.start_agent_session` (LangGraph adapter) now
+  rejects the same hostile `agent_id` shapes via `validate_agent_id`
+  instead of the prior `if not agent_id` empty-only check.** Closes
+  the last in-process surface where a Python caller could land
+  `"agent-runtime:foo:bar"` in storage even though the MCP / CLI
+  surfaces refuse the same shape. The raised exception narrows from
+  the generic `ValueError("agent_id must be a non-empty string")` to
+  `InvalidNameError("invalid agent-id 'X': ...")` — `InvalidNameError`
+  is a `ValueError` subclass, so `except ValueError:` callers keep
+  working, but any code substring-matching the old message text will
+  break and should switch to matching `"invalid agent-id"`. Closes
+  #492.
+
 ## [0.1.31] — 2026-04-26
 
 ### Added
--- a/packages/memtomem/src/memtomem/constants.py
+++ b/packages/memtomem/src/memtomem/constants.py
@@ -38,10 +38,12 @@ SHARED_NAMESPACE: Final[str] = "shared"
 def validate_agent_id(value: object) -> str:
     """Return *value* unchanged if it is a valid agent identifier.
 
-    Applied at every MCP + CLI surface that concatenates
-    ``AGENT_NAMESPACE_PREFIX`` with caller input. Session-start:
-    ``mem_session_start`` (MCP), ``mm session start`` /
-    ``mm session wrap`` (CLI). Multi-agent registration / search:
+    Applied at every MCP + CLI + Python-adapter surface that
+    concatenates ``AGENT_NAMESPACE_PREFIX`` with caller input.
+    Session-start: ``mem_session_start`` (MCP), ``mm session start`` /
+    ``mm session wrap`` (CLI),
+    ``integrations.langgraph.MemtomemStore.start_agent_session``
+    (Python adapter). Multi-agent registration / search:
     ``mem_agent_register`` / ``mem_agent_search`` (MCP), ``mm agent
     register`` (CLI).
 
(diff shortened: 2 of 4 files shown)


pandas-studio:
Replaced by #498 — same 3-commit stack rebased onto current main after #494 squash-merged the base branch (auto-close per `feedback_stacked_pr_base_delete_auto_close.md`). All review feedback already addressed inline.

Issue #448, opened 2026-04-24
Windows compatibility issue due to fcntl import


powerzist:
### Problem

When running on Windows, the following error occurs:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\user\.local\bin\mm.exe\__main__.py", line 4, in <module>
    from memtomem.cli import cli
  File "C:\Users\user\AppData\Roaming\uv\tools\memtomem\Lib\site-packages\memtomem\cli\__init__.py", line 70, in <module>
    _register()
    ~~~~~~~~~^^
  File "C:\Users\user\AppData\Roaming\uv\tools\memtomem\Lib\site-packages\memtomem\cli\__init__.py", line 45, in _register
    from memtomem.cli.uninstall_cmd import uninstall
  File "C:\Users\user\AppData\Roaming\uv\tools\memtomem\Lib\site-packages\memtomem\cli\uninstall_cmd.py", line 28, in <module>
    import fcntl
ModuleNotFoundError: No module named 'fcntl'
```

### Cause

I believe this issue occurs because `fcntl` is a Unix/macOS-specific Python standard library module and is not available on Windows.

### Proposed Solution

One possible approach would be to use OS-specific branching and import different modules depending on the platform. However, I would like to suggest a cleaner approach: using a platform-independent file locking module.

From what I found, `portalocker` handles platform-specific locking internally. On macOS/Unix-like systems, it uses `fcntl`, while on Windows it uses the `LockFileEx` Win32 API. Therefore, replacing the current usages of `fcntl` with `portalocker` should resolve this issue while keeping the locking behavior cross-platform.

Please let me know if this approach sounds acceptable. If it does, I can work on the change and submit a pull request.

PR #515, opened 2026-04-28
fix(init): resolve hang when selecting Ollama on Windows

powerzist:
### Problem

There was an issue when running `mm init` on Windows.

In the `Choose setup style` step, I selected:
`[4] Advanced - full 10-step wizard (all options)`

Then, in the `1. Embedding Provider: Choose how to generate embeddings` step, I selected:
`[3] Ollama - local dense embeddings, needs GPU`

the program would sometimes stop without doing anything further.

### Cause

The `~/.memtomem` directory had been deleted, but up to the point where the program stopped, there did not appear to be any access to `~/.memtomem`. Therefore, the presence or absence of the `~/.memtomem` directory did not seem to be the cause of the issue.

I then tried one experiment.
When I performed the same steps while Ollama was not running, Ollama was started automatically, but the program then hung.
When I performed the same steps while Ollama was already running, the program proceeded normally.

My Claude Code interpreted the issue as follows:
```
The root cause of the issue is Windows pipe inheritance. The problem occurs as follows:

[1] subprocess.run(["ollama", "list"], capture_output=True, timeout=5)
→ Opens stdout/stderr pipes and starts the ollama list process

[2] ollama list runs
→ Detects that Ollama is not running
→ Automatically starts a new Ollama server process ← the tray icon appears here
→ At this point, the Ollama server inherits the pipe handles

[3] 5 seconds pass → TimeoutExpired
→ subprocess.run kills the ollama list process
→ It then calls communicate() again to read any remaining data from the pipes

[4] ← Infinite wait (hang) occurs here
The Ollama server process is still holding the pipe handles
→ The pipes never receive EOF
→ Python waits forever

This issue does not occur on macOS/Linux because process termination and pipe handle inheritance behave differently there. This is why the issue only occurs on Windows.
```

### Conclusion

Manually starting Ollama before run `mm init` resolves the issue, but I do not think that is a natural user experience. Therefore, I believe the expected behavior is for the program to automatically start Ollama when it is not running and then continue execution.

Since this issue only occurs on Windows, I added a conditional branch so that this part behaves differently on Windows only.

After modifying the code, I confirmed on Windows that the program no longer hangs.


memtomem:
Thanks for the careful diagnosis — the Windows pipe-inheritance root cause analysis is solid. A few suggestions before merge:

**Code**
- The inline comment `# 최대 ~24초 대기` should be in English to match the rest of the codebase (per `CLAUDE.md` conventions). Suggest: `# wait up to ~24s`.
- The `~24s` figure is optimistic: each `_ollama_running()` call still uses a 5s timeout, so worst case is `12 * (2s sleep + up to 5s timeout)` ≈ 84s. Consider passing a shorter timeout (1–2s) for the polling path, or update the comment to reflect the real worst case.
- Optional cleanup: instead of branching on `sys.platform` inside `_ollama_running`, you could extend `_run()` with a `capture: bool = True` parameter and have callers pass `capture=False` here. Keeps the platform check out of the helper. Not blocking.

**Tests**
- No tests added. The Windows-only branch is hard to exercise on CI, but the polling logic itself is testable: `monkeypatch sys.platform = "win32"` + a fake `_ollama_running` that returns `False` N times then `True` (and one that always returns `False`) would pin both the success and timeout messages. Worth adding.

**Minor**
- On Windows, redirecting stderr to `DEVNULL` loses debugging signal if Ollama fails for an unrelated reason. Acceptable trade-off, just noting.

Otherwise looks good — diagnosis is correct and the fix is appropriately scoped.

powerzist:
I reverted the changes I had previously made to the existing code and updated `_run()` so that it passes `capture_output=False` only on Windows.

I confirmed that this change alone resolves the related issue, ~~and I also added tests for this behavior.~~

Sorry. I'll check and retry this.

powerzist:
When I tried to `run uv run pytest` to check whether there was an issue with the newly added test code, I encountered the following error:
```
packages\memtomem\tests\test_uninstall_cmd.py:12: in <module>
    import fcntl
E   ModuleNotFoundError: No module named 'fcntl'
```
It looks like the part I previously reported(#448) and you fixed(#456) is still causing an error here. This needs to be resolved first before I can work on fix the test code locally.

memtomem:
Thanks again for the careful repro on this — and especially for flagging the `fcntl` collection failure in `test_uninstall_cmd.py`. That was a real gap from #456: the production fix landed but the test module still had a top-level `import fcntl`, so anyone trying to run `pytest` on Windows would hit `ModuleNotFoundError` before any test could even collect.

I shipped that as a separate small PR — #523 (now merged) — so it doesn't get tangled with the Ollama hang fix here. Once you rebase this branch on `main`, `uv run pytest packages/memtomem/tests/test_init_cmd.py` should collect cleanly on Windows.

When you're ready to retry, the main thing I'd love to clarify on this PR is the relationship between the source change and the new tests — the tests assert that `_ollama_running` uses `subprocess.DEVNULL` for stdout/stderr and that `_step_embedding` has a polling loop that prints "Ollama is ready." / "Ollama did not start in time.", but I don't see those changes in the current diff (only the 2-line `_run` change). Did some of the source edits get lost in the revert? Happy to help work through it either way.

No rush — take your time. 🙏

powerzist:
In the previous commit, I misunderstood the behavior of the `capture_output` parameter, so I modified the code incorrectly.

In the updated code, `_run()` now accepts an additional `output` parameter. When `output` is `False`, `capture_output` is disabled, and `stdout` and `stderr` are set to `DEVNULL` instead.

In `_ollama_running()`, the value of the `output` parameter is passed differently depending on whether `sys.platform` is `"win32"` or not.

Since `_ollama_running()` does not need `stdout`, I believe this behavior now matches the original intent.

For `_ollama_has_model()`, it uses `stdout`, and the risk of the pipe inheritance issue occurring there seems low, so I judged that keeping the existing behavior is appropriate.

Regarding the pipe inheritance issue, it appears to occur when Ollama is not running: running the `ollama list` command starts the Ollama server in the background, and the pipe handles are inherited, which causes the problem. I confirmed that when Ollama is already running, the Ollama server has already been started, so the pipe handle inheritance issue does not occur.

As for the test code, it is currently difficult to work on it in the Windows environment. It seems that many new Windows-related bugs have appeared as other parts have been updated. Because of that, I have not been able to properly work on the test code yet.

memtomem:
Thanks for the update — and totally fair, the Windows test environment has been a moving target lately. The latest source change (`_run` + `output` parameter) is exactly right and matches the original diagnosis cleanly. 👍

Two things still blocking CI, both on the test side:

1. **`test_win32_polling_succeeds_before_timeout` and `test_win32_polling_times_out_when_ollama_never_starts`** assert behavior from my earlier review suggestion (a polling loop in `_step_embedding` printing `"Ollama is ready."` / `"Ollama did not start in time."`). When you reverted that approach in favor of the cleaner `_run(output=...)` change, those two tests need to come out — they're testing code that no longer exists in the diff. The remaining test `test_ollama_running_uses_devnull_on_win32` is the right one to keep; it actually pins the fix.

2. **`ruff format`** — the new `subprocess.run(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, timeout=timeout)` line exceeds the 100-char limit. `uv run ruff format packages/memtomem/src/memtomem/cli/init_cmd.py` from a Linux/macOS shell (or even via the GitHub web editor) will reformat it.

Once those two are sorted CI should go green. No rush — happy to wait until you've got a working environment, or I can push the cleanup directly to your branch if that's easier (you've checked "Maintainers are allowed to edit"). Just let me know which you prefer.

PR #533, opened 2026-04-29
feat(indexing): parse temporal-validity frontmatter into chunk metadata

memtomem:
## Summary

- First-cut of the temporal-validity RFC Goals 1+2+3: frontmatter parser, schema migration, and indexer threading.
- Adds `valid_from` / `valid_to` parsing for `YYYY-MM-DD` (date-only) and `YYYY-QN` (quarter) formats; surfaces the parsed window on `ChunkMetadata` and persists it to two new `chunks` columns (`valid_from_unix`, `valid_to_unix`).
- No user-visible behaviour change yet — chunks without frontmatter validity stay always-valid (RFC §Goal 4 backward-compat). The pipeline `validity_filter` stage, `mem_search(as_of=...)`, the `mm search --as-of` CLI flag, and the Web UI badge land in follow-up PRs.

## Design choices worth noting

- **Inclusive-end semantics**: lower bound = unit start (00:00:00 UTC), upper bound = unit's last day end (23:59:59 UTC). Matches the user mental model in the RFC ("from Aug 15" = the whole day Aug 15 onward).
- **Liberal parser**: a malformed value on one side returns `None` for that side without poisoning the other or aborting indexing. A typo in `valid_from` should not block a file from being indexed.
- **`UPDATE` includes the new columns**: same pattern as `tags` (also frontmatter-derived). This means reindex propagates frontmatter validity edits to chunks whose `content_hash` did not change — otherwise an editor changing only the validity window would see no effect on body chunks.

## RFC reference

`memtomem-docs/memtomem/planning/temporal-validity-rfc.md` (commit `297470a` on `memtomem-docs` main). Goals 1+2+3 only.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_temporal_validity.py` — 24 / 24 pass.
- [x] Touched-module regression sweep (`test_chunking`, `test_chunking_blockquote_tags`, `test_chunkers_extended`, `test_storage`, `test_storage_extended`, `test_storage_noop`, `test_sqlite_schema`, `test_chunk_links_schema`, `test_temporal_validity`) — 143 / 143 pass.
- [x] `uv run ruff check` + `uv run ruff format --check` clean.
- [ ] CI green.

### DIFF
--- a/packages/memtomem/src/memtomem/chunking/markdown.py
+++ b/packages/memtomem/src/memtomem/chunking/markdown.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import re
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 
 from memtomem.models import Chunk, ChunkMetadata, ChunkType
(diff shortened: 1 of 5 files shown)


memtomem:
Review feedback addressed in follow-up commit `460592b`.

**🟡 (addressed)** The `bm25_search` / `dense_search` queries were SELECTing only the 13 core chunk columns and slicing `row[:13]` into `_row_to_chunk`, so the new `len(row) >= 21` validity guard never tripped — every search-derived chunk carried `valid_from_unix=None` regardless of what was stored. The same shape also silently zeroed `overlap_before` / `overlap_after` and dropped `importance_score`, a latent bug independent of this RFC.

Switched both queries to `SELECT c.*, sub.<score>` so the full row layout reaches `_row_to_chunk` and every defensive guard activates uniformly. Sliced with `row[:-1]` / `row[-1]` to stay index-agnostic if future ALTER TABLE additions extend the chunks table further.

Three new round-trip tests in `test_temporal_validity.py` lock the behaviour:
- `test_bm25_search_preserves_validity_window`
- `test_dense_search_preserves_validity_window`
- `test_bm25_search_returns_none_pair_for_unset_chunks` (confirms `c.*` did not silently fabricate values for the always-valid backward-compat default)

Regression sweep is 285/285 across storage / chunking / schema and the search-pipeline-adjacent suite (`test_context_window` included, since context expansion reads `overlap_before/after` — the side-effect restoration is safe).

**🟢 (deferred)** The remaining nits:
- Nested YAML key (e.g. `metadata: { valid_from: ... }`) matching — adding a leading-space check is not worth the extra branch right now. Frontmatter in this codebase is overwhelmingly flat.
- Inline `#` comment unsupported — better captured as a small RFC doc note in a follow-up PR than as parser code.
- `valid_from > valid_to` cross-validation — naturally lives with the pipeline filter PR (Goal 4); a reverse window is unconditionally excluded by the filter, so risk is bounded.
- Double `_FRONT_MATTER_RE.match` (in `_extract_frontmatter_tags` + `_extract_validity_window`) — intentional separation kept; cost is trivial. If frontmatter fields keep growing, a single parse → dict refactor is the right consolidation moment.

PR #535, opened 2026-04-29
feat(privacy): block secret-bearing writes at mem_add ingress

memtomem:
## Background

The LTM `mem_add` / `mem_batch_add` tools previously accepted any
content as written. Anything indexed into LTM is reachable later via
search + recall + (when STM is in front) auto-surfacing — so a single
unintended secret write keeps leaking into future agent contexts until
it's manually scrubbed. The cleanup cost on a leaked secret is
asymmetric to the cost of a false-positive block, so this PR shifts
the trust boundary one layer inward: the ingress refuses
secret-pattern matches by default.

The detector lives parent-side, not in the STM proxy, on purpose. The
STM proxy is a convenience / capability layer; an agent that bypasses
STM (direct `mem_add` over MCP) must still hit the same redaction
guard. STM-bypass != safety-bypass.

## What changed

- **New `memtomem.privacy` module.** 9 secret-class regex patterns:
  provider tokens (`sk-…`, `ghp_…`, `xox[bps]-…`, `github_pat_…`,
  `npm_…`, Stripe-style), AWS access keys (`AKIA…`/`ASIA…`), JWT-ish
  triplets, PEM private-key headers, and `(api_key|secret_key|access_token|password|passwd|pwd)` followed by `:` / `=`.
  10K-char scan window matches the STM compression-side scanner so
  the two views of "is this content sensitive" stay aligned at the
  floor.
- **Three-label outcome counter.** `blocked` (rejected), `pass` (no
  hit), `bypassed` (`force_unsafe=True` overrode a hit). Per-tool
  breakdown under `by_tool`. Surfaced through the new
  `mem_add_redaction_stats` MCP tool (registered under `crud`,
  routed via `mem_do` in core mode).
- **`force_unsafe: bool = False` kwarg** on `mem_add`,
  `mem_batch_add`, and `_mem_add_core`. Internal callers
  (`reflection`, `scratch`, `procedure`, `mem_consolidate_apply`,
  multi-agent share) inherit `False` — LLM-driven internal writes
  cannot bypass the guard without a code change.
- **`mem_batch_add` rejects the whole batch on any hit**, listing
  the offending zero-based indices in the error message.
  Partial-success semantics on a flagged batch would break the
  caller's transactional assumption.

## Behavior change

This is a **breaking-by-design** change: previously-accepted content
that matches a secret pattern now returns an error message instead of
being written. Callers can opt back in per-call with
`force_unsafe=True`. Operators tracking false-positive impact can
read the live counter via `mem_add_redaction_stats`.

The pattern set is the **secrets-only subset** of the STM proxy's
content scanner. STM matches an additional email/PII pattern
(`<word>@<word>.<tld>`) that is intentionally excluded here — block
semantics on PII would force `force_unsafe=True` on most legitimate
contact / meeting / conversation notes, defeating the audit value of
the bypass label. STM keeps the email pattern because its
compression-routing use has a benign false-positive mode (fall back
to selective compression).

## Cross-repo sync rule

`CLAUDE.md` gains an asymmetric sync rule:

- STM additions of **secret-class** patterns (provider tokens, key
  formats, etc.) require sync into `memtomem/privacy.py` + a SHA bump
  in the module docstring.
- STM additions of **PII-class** patterns (email, phone, name,
  address, etc.) do not auto-sync; including any new PII-class
  pattern requires a separate decision pass.

Module docstring pins the current STM source SHA (`a98636e`).

## Tests

- `test_privacy.py` — 24 unit tests:
  - Pattern count pin (`len == 9`) — drift signal toward PII drift.
  - Negative contract: clean prose with emails / phone numbers /
    plain text returns zero hits. Catches future PII-style additions
    directly.
  - Per-pattern positive: each of the 9 patterns hits a known sample.
  - Scan-window cap parity with STM.
  - Counter: increment, deep-copy snapshot, unknown-outcome drop +
    warning, reset.
- `test_memory_crud_redaction.py` — 7 wire-in tests on real
  components (`bm25_only_components` fixture):
  - `mem_add` blocks on hit, no file created, `blocked` counter +1.
  - `mem_add(force_unsafe=True)` writes the file, `bypassed` +1,
    `blocked` unchanged.
  - `mem_add` clean → `pass` +1.
  - `mem_batch_add` full reject lists hit indices `[1, 3]`, file not
    created, two hit items recorded as `blocked`.
  - `mem_batch_add(force_unsafe=True)` records `bypassed` per hit
    item and `pass` per clean item.
  - Clean batch records `pass` per entry.
  - `mem_add_redaction_stats` returns expected JSON shape.

Each behavior assertion is paired with a counter-snapshot assertion
so a future change that drops `record(...)` calls without altering
the visible block / pass / bypass logic fails the test instead of
silently regressing.

## Validation

- `uv run pytest packages/memtomem/tests/test_privacy.py packages/memtomem/tests/test_memory_crud_redaction.py` — 31/31 pass.
- Regression sweep on `mem_add` callers
  (`test_batch_add_tag_isolation`, `test_multi_agent_integration`,
  `test_server_degraded_mode`, `test_tools_logic`,
  `test_validation_error_messages`,
  `test_validate_namespace_architectural_guard`) — 145/145 pass.
- Full suite: 3079 passed, 46 deselected (Ollama), 47s.
- `ruff check` + `ruff format --check` clean on touched files.
- `mypy` clean on touched files (advisory).

## Test plan

- [ ] Verify `mem_add(content="api_key: sk-aaaa…")` returns the
      blocking error and does not create the file.
- [ ] Verify `mem_add(content="…", force_unsafe=True)` writes when
      content matches a pattern; check `mem_add_redaction_stats`
      shows a `bypassed` outcome.
- [ ] Verify a `mem_batch_add` with one flagged entry rejects the
      whole batch and reports the hit index.
- [ ] Verify `mem_add_redaction_stats` returns an outcomes +
      by_tool JSON snapshot.
- [ ] Verify routine prose with emails (`team@acme.io`) and phone
      numbers passes without `force_unsafe`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,6 +29,15 @@ for `memtomem` and both resolve to `memtomem.cli:cli`.
   and talks to this LTM server only through the MCP protocol. Don't
   `import memtomem_stm` from `packages/memtomem/src/`, and don't hand-roll an
   in-process STM client — cross-repo coupling is explicitly forbidden.
+- **`memtomem/privacy.py` syncs from STM, asymmetrically.** The redaction
+  patterns in `packages/memtomem/src/memtomem/privacy.py` are copied from
+  `memtomem-stm/proxy/privacy.py:DEFAULT_PATTERNS` (the SHA pin lives in the
+  module docstring). The trust boundary is here, not there — STM-bypass must
+  not be safety-bypass. When STM adds a **secret-class** pattern, sync it
+  here in the same release window and bump the SHA. **PII-class** patterns
+  (email, phone, name, address, etc.) do not auto-sync — they would force
+  `force_unsafe=True` on most legitimate prose ingress; including any new
+  PII-class pattern requires a separate decision pass.
 - **`mm` ≡ `memtomem`.** Both `project.scripts` entries in
   `packages/memtomem/pyproject.toml` resolve to `memtomem.cli:cli` — keep them
   in sync, don't diverge behavior or add flags to only one name.
--- a/packages/memtomem/src/memtomem/privacy.py
+++ b/packages/memtomem/src/memtomem/privacy.py
@@ -1,0 +1,146 @@
+"""Content redaction guard at the LTM trust boundary.
+
+The pattern set is duplicated from memtomem-stm
+``proxy/privacy.py:DEFAULT_PATTERNS`` as of commit ``a98636e``,
+**secrets-only subset**. STM's full pattern set includes patterns whose
+semantic category is PII (e.g., email addresses) rather than secret-class.
+Those are excluded by design here because PII false positives on prose
+ingress would be unworkable; redaction at the LTM ingress is the trust
+boundary where blocking semantics demand a tight false-positive profile.
+
+Sync rule (asymmetric):
+
+- STM additions of secret-class patterns (provider tokens, key formats,
+  PEM-style headers, etc.) require sync into this module + a SHA bump in
+  this docstring.
+- STM additions of PII-class patterns (email, phone, name, address, etc.)
+  do NOT auto-sync. Including any new PII-class pattern here requires a
+  separate decision pass — the false-positive profile in prose ingress is
+  fundamentally different from STM's compression-routing use, and a PII
+  block default would force ``force_unsafe=True`` on most legitimate
+  contact / meeting / conversation notes.
+
+This module is the LTM-side trust boundary. STM's content scanner is a
+routing signal only; if STM is bypassed (direct agent → LTM call), the
+redaction guard here still applies. STM-bypass is not safety-bypass.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+from collections import defaultdict
+from dataclasses import dataclass
+from functools import lru_cache
+from threading import Lock
+
+logger = logging.getLogger(__name__)
+
+# Patterns are secret-class only. See module docstring for sync rule.
+DEFAULT_PATTERNS: tuple[str, ...] = (
+    r"(?i)(api[_-]?key|secret[_-]?key|access[_-]?token)\s*[:=]",
+    r"(?i)(password|passwd|pwd)\s*[:=]",
+    # Provider-prefixed token formats. Anchored by prefix so false positives
+    # on arbitrary high-entropy strings are rare.
+    r"(?i)(sk-[a-zA-Z0-9]{20,}|ghp_[a-zA-Z0-9]{36}|xox[bps]-[0-9A-Za-z-]+)",
+    r"github_pat_[A-Za-z0-9_]{20,}",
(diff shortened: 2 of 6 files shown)


memtomem:
Thanks for the thorough review. Addressed in `dab8ac2`.

**1. Triple-decorator (clarification, no change):** This matches the existing convention for non-core tools. `mem_edit`, `mem_delete`, `mem_batch_add`, every `mem_ns_*` (namespace.py), and every `mem_policy_*` (policy.py) all use the same `@mcp.tool() @tool_handler @register(...)` stack. The mode filter at `server/__init__.py:171` only **removes** tools from `mcp._tool_manager._tools` — it never adds. A tool without `@mcp.tool()` is therefore not directly callable as an MCP tool even in `full` mode; only routed via `mem_do`. The triple stack is what makes a non-core tool exposable in `full`/`standard` (via `@mcp.tool()`) **and** routable via `mem_do` in `core`/`standard` (via `@register`). I'm reading CLAUDE.md's "MUST use `@register`" as "must include `@register`" rather than "must use *only* `@register`" because that's consistent with every existing precedent. Happy to push a CLAUDE.md tightening as a follow-up if the wording itself should be made explicit.

**2. 10K scan window in mem_add docstring:** Fixed. Both `mem_add` and `mem_batch_add` now state the per-call scan window in their docstrings, with a hint to split very long content if every region must be inspected. Module-level docstring stays as the authoritative explanation.

**3. Structured bypass log:** Done inline rather than as a TODO. Both bypass branches emit `logger.warning("redaction bypass via force_unsafe=True (tool=..., namespace=..., file=..., content_chars=...[, item_idx=...]))`. Matched bytes are intentionally not logged — the on-disk markdown the call is about to write holds the actual content, and pairing the request shape (in the log) with the persisted payload (on disk) gives the forensic correlation surface without doubling the secret-bytes blast radius.

**4. Asymmetric counter docs:** Fixed. `mem_add_redaction_stats` docstring now states that counts reflect attempted *write outcomes*, not raw scans — `blocked + pass + bypassed` equals writes that reached the guard, not entries inspected, because rejected batches contribute only `blocked` records.

**Nit (missing test):** Added `test_clean_content_with_force_unsafe_records_pass_not_bypassed`. This pins "force_unsafe + no hit → `pass`" so the `bypassed` label keeps measuring "the guard would have blocked" rather than degrading into "the kwarg was set". Total test count now 32 (24 unit + 8 wire-in), all passing.

**Skipped (taste / deferred):**
- `lru_cache(maxsize=1)` vs `maxsize=None`: kept the existing form to preserve textual parity with the STM source it was duplicated from. Both memoize once in practice; the SHA-pin sync rule benefits from byte-level diff parity.
- False-positive friction on `password:` / `api_key:` prose: agreed worth watching. The new structured bypass log + the existing counter together give the observable surface — if the `bypassed:pass` ratio gets noisy in real use, that's the trigger for a tighter prefix anchor or a per-tool overrides table.

PR #549, opened 2026-04-29
fix(web): surface "why nothing happened" on context Sync/Import

memtomem:
## Summary

PR1 of the multi-project context UI series (single-project UX gap fix —
no behavior change for users with populated `.memtomem/skills/` etc.).

User reported that `mm web` → Settings → Skills/Commands/Agents → click
**Sync** or **Import** showed a green-toast "completed" message but
nothing actually moved on disk. Root cause was a wire shape mismatch
plus missing metadata in the response, not a wiring bug. Both buttons
were correctly wired to backend handlers; the handlers correctly did
nothing because the canonical (`.memtomem/skills/`) and runtime
(`.claude/skills/`, `.gemini/skills/`, `.agents/skills/`) directories
were empty in cwd. The UI just didn't tell the user that.

## Changes

- **Core (`memtomem.context.{skills,commands,agents}`)**: skipped items
  are now `(name, reason, reason_code)` triples. New module
  `memtomem.context._skip_reasons` enumerates the codes:
  `no_canonical_root`, `unknown_runtime`, `parse_error`, `invalid_name`,
  `already_imported`, `canonical_exists`, `toml_parse_error`. Human
  `reason` strings stay (CLI/log output unchanged byte-for-byte).
- **Web routes**: sync responses gain `canonical_root: str`. Import
  responses gain `project_root: str` + `scanned_dirs: list[str]`. Each
  `skipped[]` entry carries `reason_code`. All additive.
- **Web client**: sync handler now reads `data.skipped` (was only
  reading `data.dropped`, which is empty for skills); on
  `reason_code === "no_canonical_root"` it shows
  `"No canonical {type} under {canonical}. Create one first."` info
  toast. Import handler on a 0/0 result shows
  `"No runtime {type} found in {project_root}. Scanned: {scan_dirs}."`
  info toast. Empty list state hint points at the canonical + scan
  paths instead of generic "Create one or import" copy.
- **i18n**: three new placeholder-form keys (`empty_hint`,
  `import_no_runtimes`, `sync_empty_canonical`) cover all 3 types via
  `{type}` substitution — 3 keys instead of 9. en.json + ko.json
  updated.
- **Cache-bust**: `context-gateway.js?v=3`.
- **Tests**: 3 core test suites updated for the 3-tuple shape; new web
  route assertions verify `reason_code`, `canonical_root`,
  `project_root`, `scanned_dirs` in sync/import responses for the
  empty-canonical and empty-runtime paths.

## Compatibility

**HTTP wire shape** — all changes are additive. Existing clients that
read `imported` / `generated` / `skipped[].name|runtime|reason` keep
working.

**Python API tuple shape** — breaking for external destructuring.
`SkillSyncResult.skipped`, `CommandSyncResult.skipped`,
`AgentSyncResult.skipped`, and the shared `ExtractResult.skipped` are
typed `list[tuple[str, str]]` → `list[tuple[str, str, str]]`. All
in-tree callers were updated in the same commit (CLI
`cli/context_cmd.py`, MCP `server/tools/context.py`, web routes,
tests). External Python consumers that unpack
`for name, reason in result.skipped:` would need to add a third
binding — we don't expect any (these dataclasses are not part of the
documented public API), but flagging for completeness.

CLI (`mm context skills/agents/commands ...`) and the MCP `context`
tool also unpack the triple via `_code` discard so their output is
byte-identical.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/ -m "not ollama"` — 3152
      passed.
- [x] `uv run ruff check packages/memtomem/src && uv run ruff format
      --check packages/memtomem/src` — clean.
- [x] Manual smoke: `mm web` against an empty cwd → both Sync and
      Import responses verified via curl. Sync returns `skipped` with
      `reason_code: "no_canonical_root"` + `canonical_root:
      ".memtomem/skills"`. Import returns `project_root` +
      `scanned_dirs: [".claude/skills", ".gemini/skills",
      ".agents/skills"]`.
- [ ] Browser smoke (reviewer): clear cache, click Sync on an empty
      project — see `No canonical skills under .memtomem/skills/.
      Create one first.` info toast (was: green "Sync completed"). Click
      Import — see `No runtime skills found in <cwd>. Scanned:
      .claude/skills, .gemini/skills, .agents/skills.` info toast (was:
      green "Import completed").

## Follow-ups (separate PRs per the approved RFC sequence)

- RFC `docs/rfc/multi-project-context-ui.md` — multi-project aggregate
  view design.
- PR2 — multi-project read-only (Add Project + lazy GET groups).
- PR3 — multi-project write + Claude user-tier (`~/.claude/skills/`).
- PR4 — Gemini/Codex user-tier + alias unification (after CLI
  verification).

Optional polish (not blocking PR1):
- Tighten `reason_code` to `Literal[...]` / `StrEnum` for stricter
  typing of the `code` field in dataclass + route response builders.
- Strengthen `test_import_empty_carries_meta` to assert the
  `project_root` value matches `tmp_path`, not just key presence.


memtomem:
Addressed review in [`f48b43a`](https://github.com/memtomem/memtomem/pull/549/commits/f48b43a):

- **#1 (low-medium)** — dropped the hardcoded `_CTX_EMPTY_HINT_META` JS map. GET `/api/context/{type}` now returns `canonical_root` + `scanned_dirs` (computed once at module import from `SKILL_DIRS` / `AGENT_DIRS` / `COMMAND_DIRS`), so the empty-state hint reads from the wire. The JS keeps a one-line `.memtomem/${type}` fallback for older backends but it never fires once the cache-bust takes effect.
- **#2 (cosmetic)** — skills sync response now uses the `CANONICAL_SKILL_ROOT` constant directly, matching the `CANONICAL_AGENT_ROOT` / `CANONICAL_COMMAND_ROOT` pattern in the other two routes. Drops the `_safe_rel(canonical_skills_root(...), ...)` round-trip — same string today, but won't drift if a future PR adds env/scope resolution to `canonical_skills_root()`.
- **#3 (low)** — import-no-runtimes toast renders `basename(project_root)`. Wire stays absolute (consistent with `/api/system/*`); UI keeps the toast short since `scanned_dirs` already gives orientation. New `_ctxBasename()` helper.

Cache-bust bumped to `context-gateway.js?v=3`.

Tests: `TestListSkills.test_empty` / `TestListCommands.test_empty` / `TestListAgents.test_empty` now assert the new GET fields. `pytest -m "not ollama"` 3152 passed; `ruff check` + `ruff format --check` clean.

Skipped per your guidance:
- **#4** `Final[str]` → `Literal` / `StrEnum` (mypy is advisory).
- **#5** additional positive route assertions for `invalid_name` / `already_imported` / `toml_parse_error` (route layer is a trivial dict comprehension; codes are covered at the core layer).
- **#6** Korean `<이름>` literal angle-bracket — intentional placeholder guide, not an unsubstituted variable. Confirmed.

PR #564, opened 2026-04-29
feat(web): arrow-key nav for main tablist + Sources sub-toggle

memtomem:
## Summary

Follow-up to #563 (main tab-nav ARIA) and #560 (Sources sub-toggle ARIA). Both tablists carried `role=tab/tablist` but the ARIA tabs authoring practices' keyboard nav contract — ArrowRight/Left to cycle siblings, Home/End to jump to either end — was unimplemented, leaving keyboard users without an in-tablist navigation primitive.

- Shared `_arrowNavIndex` helper computes the next index given the cycle direction; same call shape from both handlers
- **Main \`.tab-nav\` keydown**: cycles among visible main tabs (filters out \`data-ui-tier="dev"\` buttons in prod mode so focus never lands on a hidden tab), auto-activates so focus moves the panel without a separate Enter press — matches the existing click behavior
- **\`.sources-mode-toggle\` keydown**: same shape over the two Memory/General buttons. \`setSourcesMode\` now toggles \`tabindex\` alongside \`aria-selected\` so the single-tab-stop pattern (one tab in the keyboard order at a time) holds across mode switches

## Stacked on #563

Base = \`feat/main-nav-aria\`. After #563 squash-merges to main, rebase with \`git rebase --onto origin/main feat/main-nav-aria\`. If GitHub auto-closes on base delete, retarget to \`main\` and reopen.

## Test plan

- [x] \`uv run pytest packages/memtomem/tests/test_i18n.py -q\` (12 passed; markup-only change so parity is unaffected)
- [x] \`uv run ruff check packages/memtomem/src\` + \`ruff format --check\`
- [ ] Manual smoke: \`mm web\` → focus a main tab button (Tab key reaches the tablist), verify ArrowRight/Left cycle through visible tabs and activate the panel; Home/End jump to the first / last visible tab; same behavior on the Sources Memory/General sub-toggle

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -665,6 +665,37 @@ document.querySelectorAll('.tab-btn').forEach(btn => {
   btn.addEventListener('click', () => activateTab(btn.dataset.tab));
 });
 
+// ── ARIA tabs keyboard navigation ──
+//
+// ArrowRight/Left cycle through the tablist, Home/End jump to either end.
+// "Auto-activation" model — focus and activate together — matches the click
+// behavior so a keyboard user toggles the panel just by walking the tab row,
+// without an extra Enter press. The currently focused element is the anchor;
+// when focus is outside the tablist the move starts at index 0.
+function _arrowNavIndex(length, currentIdx, key) {
+  if (!length) return -1;
+  if (key === 'ArrowRight') return (currentIdx + 1) % length;
+  if (key === 'ArrowLeft') return (currentIdx - 1 + length) % length;
+  if (key === 'Home') return 0;
+  if (key === 'End') return length - 1;
+  return -1;
+}
+
+document.querySelector('.tab-nav')?.addEventListener('keydown', (e) => {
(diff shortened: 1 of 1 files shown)


memtomem:
Self-review surfaced two behaviors in \`activateTab\` that the original commit didn't handle:

1. **Panel auto-focus** — every activation moved focus into the panel's first input/button, so the user's second ArrowRight landed inside the search input (cursor move) instead of advancing the tablist
2. **History pollution** — \`history.pushState\` fired per cycle step, so 5 arrow presses needed 5 Back presses to escape

Folded into 91395b1: \`activateTab(name, { fromKeyboard: true })\` skips panel focus and uses \`replaceState\`. Click activation keeps the existing push-and-focus behavior.

CSS check: \`.tab-btn:focus-visible\` (style.css:1593) + global \`:focus-visible\` (1579) already provide a visible focus indicator, so arrow-cycling shows where focus is — no extra CSS needed.

Sources mode mapping (id-based) and JS test coverage gap remain — both lateral / structural rather than blocking, leaving as-is for now.

PR #575, opened 2026-04-30
feat(web): Index tab — segmented mode toggle for folder/upload/compose

memtomem:
## Summary

- Collapses the Index tab's three independent cards (Index Files / Upload / Add Memory) into a single card driven by a segmented mode toggle — `Folder index` / `Upload files` / `New memory` — using the same auto-activation tablist pattern as the main tab nav.
- Drops the duplicate `stat-card-mini` (chunks/sources) inside the Index card; the header pills already show the same numbers.
- Honest extension hints in the Upload panel: drop-zone label + `<input accept>` were advertising `.py/.js/.ts` even though `_ALLOWED_UPLOAD_EXTS` rejects them. Folder indexing accepts code; uploads are intentionally docs-only.

## Why

Three peer cards with equal visual weight made the Index tab feel scattered and the duplicate stats added noise. Mode toggle re-frames the three paths as "three ways to add memories" while keeping all input affordances reachable in one place.

## Notable correctness items

- Compose label: `Add Memory` / `직접 작성` → `New memory` / `새 기억` (clearer about *what* is composed).
- `index.help` collapses to a one-liner; the toggle labels already enumerate the paths.
- ARIA: each `role=tab` carries `aria-controls` pointing at its own panel (WAI-ARIA APG), panels point back via `aria-labelledby`. (Sources tab still uses container-id; tracked as follow-up.)
- `min-height: 320px` on the panel container so toggling doesn't jump card height; floor chosen to keep the folder panel's empty area modest on first visit.
- Drops dead `loadStats()` writes to `card-chunks`/`card-sources` (nodes removed).
- Bumps `style.css?v=` to 68.

## Out of scope (follow-up issues)

- Tab-external "folder indexing in-flight" indicator — progress bar is inside the hidden folder panel today, invisible after a mode switch.
- Lucide-svg icons for toggle labels — text-only here because mixed unicode glyphs (📁/⬆/✎) render as color-emoji for the first two but a text-glyph pencil, breaking visual weight.
- Sources tab `aria-controls` consistency pass.
- First-run onboarding nudge under the toggle when chunks == 0.

## Test plan

- [x] `uv run ruff check packages/memtomem/src` + `uv run ruff format --check packages/memtomem/src`
- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/test_i18n.py` (en/ko parity)
- [x] Playwright MCP smoke: toggle clicks swap panels (folder/upload/compose); ArrowRight/Left/Home/End cycles + auto-activates; localStorage persists last mode; folder/upload/compose panels each render full-width inputs; header chips remain authoritative for chunks/sources.
- [ ] Reviewer hard-reload (`Cmd+Shift+R`) once after pulling — `style.css?v=68` should bypass the prior cached `v=67`.


memtomem (commented):
done

PR #639, opened 2026-05-01
fix(web): render orphan-indexed sources in Sources tab

memtomem:
## Summary

The ``/api/sources`` response carries rows with ``memory_dir=null, kind=null`` for two paths:

1. **Index-tab uploads** land in ``~/.memtomem/uploads/`` ([system.py:upload_files](https://github.com/memtomem/memtomem/blob/main/packages/memtomem/src/memtomem/web/routes/system.py)) — not a configured ``memory_dir``.
2. **Removed-from-config dirs** — chunks survive after a dir is dropped from ``memory_dirs`` but no longer have an owning dir.

Server-side intent (pinned by [`test_orphan_source_kind_is_null`](https://github.com/memtomem/memtomem/blob/main/packages/memtomem/tests/test_web_routes.py#L573) / [`test_kind_general_filter_includes_orphans`](https://github.com/memtomem/memtomem/blob/main/packages/memtomem/tests/test_web_routes.py#L599)) is that orphans "ride along with general" so users can find and prune them. The Sources tab client renderer violated that intent by:

```js
const key = s.memory_dir || '';
// …
for (const k of Object.keys(sourcesByDir)) if (k) allDirs.add(k);  // empty key dropped
```

…leaving orphan rows unassigned to any vendor group and therefore unrendered.

## User-facing failure mode

File uploaded via the Index tab appears in ``mem_search`` and the namespaces tab, but the source link from namespaces lands on a Sources tab that doesn't contain it. Reported via ``mm web``: Index-tab upload → namespaces source-link cross-link → "있는데 없음" state.

## Fix

Client-only change in ``_renderMemorySourceTree``:

- Collect rows with ``!s.memory_dir`` into a separate ``orphanItems`` list.
- Attribute them to the ``user`` vendor (both feed paths are user-driven).
- Append a collapsed **"Other (unregistered)"** ``<details>`` block at the end of that vendor's active panel.
- Reuse ``_renderMemorySourceItem`` card shape so file-click → chunks drill-in works identically to indexed dirs.
- The user vendor's sub-tab badge and the empty-state guard count orphans too — a user with only Index-tab uploads sees a populated ``user`` tab instead of "user memory not found".

Server contract is unchanged. ``index.html`` ``app.js?v=`` bumped 79 → 81 (two-commit body changes) to bust disk cache. New i18n key ``sources.orphan_label`` added to ``en.json`` / ``ko.json``.

Follow-up commit (`55c091b`) addresses two self-review nits:
- `renderOrphanBlock` now passes the outer tree-wide `maxChunks` to `_renderMemorySourceItem` instead of a local max, so chunk-bar widths stay visually comparable across indexed groups and the orphan section.
- The multi-product render branch guards `products.children.length` before appending, so an orphan-only `user` vendor doesn't ship an empty `.source-vendor-products` wrapper before the orphan block.

## Test plan

- [x] ``test_i18n.py`` parity (en/ko key set + placeholder shape) — 12 tests pass
- [x] ``-k source`` slice (75 tests) — pass against the worktree
- [x] Server contract for orphans was already pinned (3 tests in ``test_web_routes.py``); this PR doesn't loosen it
- [x] Manual: ``mm web`` → Index tab upload a file → switch to Sources tab → confirm file appears under "Other (unregistered)" inside the User sub-tab — **verified via Playwright in an isolated `HOME=/tmp/mm-orphan-test` env**
- [x] Manual: remove a configured ``memory_dir`` that has indexed chunks → confirm those chunks now show under "Other (unregistered)" instead of disappearing — **verified via Playwright; both upload and ex-registered-dir files end up in the same orphan section, count=2, User badge=2**

See the [verification comment](https://github.com/memtomem/memtomem/pull/639#issuecomment-) below for full Playwright assertion output.

## Out of scope

- ``_active_runs`` polling-loop suspicion → filed as #640
- JS unit-test infrastructure (would have caught this regression cheaply) → filed as #641
- Heavy bge-m3 indexing UX (config-side discussion, not a bug fix)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,6 +187,27 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### Fixed
 
+- **`mm web` Sources tab now shows orphan-indexed files instead of
+  silently dropping them.** The `/api/sources` response carries rows
+  with `memory_dir=null, kind=null` for two paths — Index-tab uploads
+  saved to `~/.memtomem/uploads/` (which isn't a configured
+  `memory_dir`) and chunks whose owning dir was removed from
+  `memory_dirs` post-indexing. The server already grouped these as
+  "general" with the explicit intent that orphans "ride along" so
+  users can find and prune them, but the Sources tab's
+  `_renderMemorySourceTree` keyed grouping by `s.memory_dir || ''` and
+  then filtered out the empty key (`if (k) allDirs.add(k)`), so the
+  orphan rows ended up assigned to no vendor group at all and never
+  rendered. The user-facing failure mode: a file uploaded via Index
+  tab appeared in `mem_search` and the namespaces tab, but clicking
+  its source link from namespaces landed on a Sources tab that didn't
+  contain it. The renderer now collects orphan rows separately and
+  appends them as a collapsed "Other (unregistered)" sub-section under
+  the `user` vendor's active panel, with file rows sharing the same
+  click-to-browse-chunks behaviour as indexed dirs. The user vendor's
+  sub-tab badge and the empty-state guard count orphans too, so a
+  user with only Index-tab uploads sees a populated `user` tab
+  instead of "user memory not found".
 - **`mm web` now starts a `FileWatcher` and runs a startup backfill.**
   Two gaps fixed together because they presented as one bug ("files
   added to a `memory_dir` don't show up in Sources"):
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -2630,6 +2630,11 @@ function _renderMemorySourceTree(sources, list) {
   // chunks drill-in works identically to indexed dirs. Rendered only
   // when ``activeVendor === 'user'`` because ``orphans`` is populated
   // exclusively for that vendor in the stats pass above.
+  //
+  // Bar normalisation: passes the outer ``maxChunks`` (tree-wide max,
+  // line ~2477) rather than a local max so the chunk-bar widths stay
+  // visually comparable to the indexed groups above — same chunk_count
+  // → same bar length regardless of which sub-section a row sits in.
   const renderOrphanBlock = (items) => {
     if (!items.length) return null;
     const det = document.createElement('details');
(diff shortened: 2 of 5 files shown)


memtomem:
### Playwright verification (isolated env)

Ran both manual scenarios from the test plan against an isolated worktree (`HOME=/tmp/mm-orphan-test`, port 8090, `embedding.provider="none"` to keep the run fast — the orphan-render path is independent of the embedder).

#### Scenario 1 — Index tab upload

1. Navigate to `mm web` → 인덱스 → 파일 업로드.
2. Upload a 605-byte fixture markdown.
3. `GET /api/sources` returns the row with `memory_dir=null, kind=null` (server contract reproduced).
4. Switch to Sources tab and assert DOM:

```js
{
  orphanBlock: true,
  orphanLabel: "기타 (등록되지 않은 위치)",
  orphanCount: "1",
  userBadgeCount: "1",
  userBadgeHidden: false,
  activeTab: "user",
  sourceItems: 1
}
```

#### Scenario 2 — registered dir removed mid-flight

1. Drop a second markdown into `/tmp/mm-orphan-test/memories` (the only registered `memory_dir`); watcher indexes it with `memory_dir=/tmp/mm-orphan-test/memories, kind=memory`.
2. Add a dummy dir (`/api/memory-dirs/add`) so the "last memory_dir" guard doesn't block removal, then remove the real dir.
3. `GET /api/sources` returns **both** rows with `memory_dir=null, kind=null`.
4. Reload Sources tab and assert DOM:

```js
{
  orphanBlock: true,
  orphanCount: "2",
  userBadgeCount: "2",
  orphanItems: ["upload-fixture.md", "note-in-registered-dir.md"],
  indexedDirsLabel: ["…/mm-orphan-test/dummy-memories"],
  emptyProductsWrapper: "absent (good - skipped)",  // review nit 2 effective
  placeholderShown: false,
  sourcesListChildrenTags: [
    "details.source-group.source-group-memory",
    "details.source-vendor-orphan"
  ]
}
```

#### Notes on review nit 1 (bar scale)

Both files in this run had `chunk_count=1`, so the visual difference between local-max and tree-max bar normalization isn't strongly forced (`barWidth=100%` either way). The code change still lands — `_renderMemorySourceItem(s, maxChunks)` receives the outer tree-wide max via closure capture, the same value indexed dirs receive — so any future asymmetric chunk count distribution will render with consistent bars.

#### What this leaves uncovered

This is one-shot manual verification, not a regression guard. Tracked in #641 (JS unit-test infrastructure) so the next time this code path moves, a 10-line jsdom assertion catches the regression on CI instead of waiting for a user report.

Issue #590, opened 2026-04-30
Streaming indexing endpoint missing namespace + errors parity


memtomem:
## What

`IndexEngine.index_path_stream` (engine.py:707) and the `GET /api/index/stream` endpoint (system.py:768) lack two contract elements that the non-stream `index_path` / `POST /api/index` pair already has:

1. **`namespace` parameter** — silently dropped, so any UI/CLI/MCP caller that streams an index ignores the user's namespace selection.
2. **Error aggregation** — `_index_file` returns `result["errors"]` per file (engine.py:606, 628, 684, 752), but the stream loop sums only chunk counters (754-757), never propagating errors. The `complete` event lacks an `errors` field, so partial-failure runs (e.g. ONNX missing on a subset, file-too-large, binary-detected) appear successful in the UI.

## Fix

- **Engine**: add `namespace: str | None = None` to `index_path_stream`, pass through to `_index_file`. Aggregate `result["errors"]` into a loop accumulator and include it as `errors: tuple[str, ...]` in the `complete` event — same shape as `IndexingStats.errors` so the existing non-stream UI handler reuses verbatim. Per-file uncaught-exception branch (engine.py:752) prefixes `f"{fp.name}: {exc}"` to match non-stream's `asyncio.gather(return_exceptions=True)` branch (engine.py:394).
- **Route**: add `namespace` query param to `GET /api/index/stream`, forward to the engine call.
- **Tests**:
  - `test_index_stream_namespace_propagates` — `namespace="x"` → indexed chunks have `metadata.namespace == "x"`.
  - `test_index_stream_complete_errors_no_silent_drop` — force a file-level failure → `complete.errors` non-empty (regression guard against current silent-drop behavior; the test name documents intent so future grep finds it).

## Why now

Unblocks #582 PR-B (Prev #1 — Index tab two-button collapse) and informs #582 PR #6 (4.11 — indexing in-flight visibility), which will surface `complete.errors` via toast.

## Out of scope

- **Per-file streaming of errors** (option ii/iii from the planning discussion). `complete.errors` is sufficient for current UX. Revisit if user reports indicate large indexing runs (>100 files) where waiting for `complete` to surface partial failures is impractical.
[...]

PR #653, opened 2026-05-01
feat(indexing): per-chunk progress events for SSE indexing stream

memtomem:
## Summary

A 250-chunk file used to show `file.md` frozen for 10+ seconds because `_index_file` awaited the embedder's full `embed_texts` before any SSE event escaped. This PR threads per-batch progress through every embedder, forwards it via a per-file `asyncio.Queue` in `index_path_stream`, and surfaces it in the Web Index tab as `file.md — 100/250 chunks`.

- **Embedder Protocol**: optional keyword-only `on_progress: Callable[[int, int], None] | None = None` on `EmbeddingProvider.embed_texts`. OpenAI/Ollama fire after each batch in `_safe_embed`; ONNX switches from a single `to_thread(all_texts)` to a per-batch loop using `EmbeddingConfig.batch_size`. Noop accepts the kwarg for Protocol conformance and ignores it.
- **`EmbeddingConfig.progress_threshold`**: new field (default `32`, non-negative validator, runtime-mutable). Engine only forwards `chunk_progress` events when a file's chunk count exceeds the threshold; `0` is an explicit "always emit" debug affordance.
- **Engine wiring**: `_index_file` accepts `on_chunk_progress`. `index_path_stream` runs each file in a task with a per-file `asyncio.Queue` + `DONE` sentinel — chunk events forward in real time while exceptions still surface via `await task` (preserves the `complete.errors` contract from #590). Consumer cancel (HTTPException, client disconnect, `aclose()`) cancels the in-flight task so embedding work doesn't leak past the SSE response.
- **Web UI**: `app.js` adds a `chunk_progress` branch with 100 ms DOM throttle and a final-tick bypass (otherwise the user sees `(192/250)` right before the next file's name pops in). New i18n key `index.file_chunk_progress` in en/ko.
- **Tests**: new `test_embedding_progress.py` (per-backend contract — monotonic, ends at total, raise-swallow); appended to `test_indexing_engine.py` (event flow above threshold, strict ordering invariant, no-emit below threshold, `threshold=0` always emits, `complete.errors` regression guard, `aclose()` cancels in-flight task).

## Out of scope (follow-ups)

- Wizard `_seed_with_progress` chunk label — `item_show_func` assumes path-tail + 40-char truncate, doesn't accept a free-form chunk count without UX redesign.
- `mm index` (non-stream CLI) progress — uses `index_path`, separate stream-conversion follow-up.
- Memory Dirs reindex button chunk display — button text is too narrow for `5/12 — 100/250`.

## Test plan

- [ ] `uv run pytest -m "not ollama"` (3616 passed locally)
- [ ] `uv run ruff check packages/memtomem/src packages/memtomem/tests`
- [ ] `uv run ruff format --check packages/memtomem/src packages/memtomem/tests`
- [ ] `uv run mypy packages/memtomem/src/memtomem/embedding packages/memtomem/src/memtomem/indexing/engine.py`
- [ ] **Manual web**: index a folder with one large file (CHANGELOG-like, 100+ heading) — `file.md — N/M chunks` ticks in the UI; final tick `(N/N)` lands before file boundary.
- [ ] **Manual web**: index a folder of small files — no `chunk_progress` noise.
- [ ] **Manual web**: close browser tab mid-index — `/api/indexing/active` returns to 0, no leaked embedding work in server logs.
- [ ] **ONNX wall-clock**: `time uv run mm index <250-chunk markdown>` before/after — single `to_thread` → per-batch loop should add only thread-handoff overhead (µs vs 50–200 ms inference). Concerning if regression > 5 %.

### DIFF
--- a/packages/memtomem/src/memtomem/config.py
+++ b/packages/memtomem/src/memtomem/config.py
@@ -75,6 +75,14 @@ class EmbeddingConfig(BaseSettings):
     #     event flows at all).
     # Runtime-mutable: the gate lives in the engine, not in cached
     # embedder state, so a config change takes effect on the next file.
+    #
+    # Why 32 (vs ``batch_size`` default 64)? Files producing ≤ batch_size
+    # chunks finish in one batch — there's no mid-progress to show, so
+    # any threshold ≥ batch_size would suppress them naturally. Picking
+    # 32 = batch_size/2 means "1-batch files stay quiet, 2+-batch files
+    # start ticking", which is the natural break-point between "instant"
+    # and "user wonders if anything is happening". Adjustable per
+    # operator preference.
     progress_threshold: int = 32
 
     @field_validator("dimension")
--- a/packages/memtomem/src/memtomem/config.py
+++ b/packages/memtomem/src/memtomem/config.py
@@ -64,6 +64,18 @@ class EmbeddingConfig(BaseSettings):
     # would not take effect until restart anyway. Same precedent as
     # ``rerank.provider``/``model``/``api_key`` below.
     threads: int = 4
+    # Per-chunk progress emission gate for the SSE indexing stream.
+    # The index engine only forwards ``chunk_progress`` events to the
+    # stream when a single file produces more than this many chunks —
+    # avoids spamming the UI with one event per tiny file. Semantics:
+    #   * ``> 0`` (default): emit only when ``chunks_total > threshold``.
+    #   * ``0``: always emit (debug affordance — useful when validating
+    #     the SSE plumbing on small fixtures, or when a user reports
+    #     "I never see chunk_progress" and you want to confirm the
+    #     event flows at all).
+    # Runtime-mutable: the gate lives in the engine, not in cached
+    # embedder state, so a config change takes effect on the next file.
+    progress_threshold: int = 32
 
     @field_validator("dimension")
     @classmethod
@@ -84,6 +96,13 @@ class EmbeddingConfig(BaseSettings):
     def threads_non_negative(cls, v: int) -> int:
         if v < 0:
             raise ValueError("threads must be non-negative (0 = ORT default)")
+        return v
+
+    @field_validator("progress_threshold")
+    @classmethod
+    def progress_threshold_non_negative(cls, v: int) -> int:
+        if v < 0:
+            raise ValueError("progress_threshold must be non-negative (0 = always emit)")
         return v
 
 
@@ -765,7 +784,7 @@ MUTABLE_FIELDS: dict[str, set[str]] = {
         "auto_discover",
         "supported_extensions",
     },
-    "embedding": {"batch_size"},
+    "embedding": {"batch_size", "progress_threshold"},
     "decay": {"enabled", "half_life_days"},
     "mmr": {"enabled", "lambda_param"},
     "namespace": {"default_namespace", "enable_auto_ns", "rules"},
@@ -795,6 +814,7 @@ FIELD_CONSTRAINTS: dict[str, dict] = {
     },
     "indexing.auto_discover": {"type": bool},
     "embedding.batch_size": {"type": int, "min": 1, "max": 1024},
+    "embedding.progress_threshold": {"type": int, "min": 0, "max": 100000},
     "decay.enabled": {"type": bool},
     "decay.half_life_days": {"type": float, "min": 0.1},
     "mmr.enabled": {"type": bool},
--- a/packages/memtomem/src/memtomem/embedding/base.py
+++ b/packages/memtomem/src/memtomem/embedding/base.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from typing import Protocol, Sequence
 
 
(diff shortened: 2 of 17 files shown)


memtomem:
## Review follow-up (commit 8f53335)

Addressed two of the three review-recommended items:

### 1. ONNX batch-shape numerical parity test (was: "throughput 보다는 numerical drift 가 더 신경 쓰임")

Added `test_onnx_numerical_parity_across_batch_sizes` (`test_embedding_progress.py`). Embeds a mixed-length text set with `batch_size=2` and `batch_size=64` against real fastembed (`all-MiniLM-L6-v2`) and asserts `np.allclose(rtol=1e-5)`. **It passes locally** — direct evidence that the single-call → batched-loop refactor preserves embeddings within ORT floating-point determinism. Skipped via `importorskip("fastembed")` so it only runs in fastembed-bearing environments (incl. the `golden-path (ONNX bge-m3)` CI job).

This closes the "silent embedding drift after reindex" risk that would otherwise be invisible until search quality regressed across the DB.

### 2. `progress_threshold = 32` rationale in docstring

`EmbeddingConfig.progress_threshold` docstring now explains why 32 (= `batch_size` default 64 / 2): "1-batch files stay quiet, 2+-batch files start ticking" — the natural break-point between "instant" and "user wonders if anything is happening". Saves the next `git blame`.

### Still TODO (manual / out-of-band)

- **Wall-clock measurement** on a 250-chunk file (`time uv run mm index <large.md>`) — will post numbers as a follow-up comment after I run it locally. Numerical parity already eliminates the worst risk; throughput is now just an SLO question.
- **GIF / before-after screenshot** of the Index tab — agreed that the value of this PR is hard to feel from code diff alone. Will capture and attach once I've spun up the manual web check.

Re: closure half-defense (`runner` 의 `queue`/`DONE` late-binding) and `sleep(60)` test style — leaving as-is for this PR; the loop is structurally serial so the closure capture is safe today, and the sleep does cancel-propagate within milliseconds in the happy path. Will revisit if either becomes a real bug source.

memtomem:
## Wall-clock + SSE verify (commit d129eab)

### ONNX wall-clock (the test plan item)

The test-plan item caught a **real +20% regression** on a 250-text run with the original Python-side batching. Diagnosis: `model.embed(texts)` defaults to `batch_size=256` internally → 1 ORT `session.run` for 250 texts. Forcing `batch_size=64` (whether via Python chunking or the kwarg) split that into 4 ORT runs, and per-call ORT setup was the dominant cost.

| variant                                          | median (3 runs) | delta vs baseline |
| ------------------------------------------------ | --------------- | ----------------- |
| A: `model.embed(texts)` default bs=256 (old)     | 0.901s          | baseline          |
| B: `model.embed(texts, batch_size=64)`           | 1.082s          | +20.1%            |
| C: `for batch: model.embed(batch)` (prev. fix)   | 1.099s          | +22.0%            |
| D: stream-iter single `model.embed(texts)` (now) | 0.922s          | **+2.3%**         |

After commit `d129eab`, ONNX uses variant D — single `model.embed()` call, stream-iterate the generator inside `_embed_sync`, fire `on_progress` per yield. Worker-thread callbacks wrap with `loop.call_soon_threadsafe` and throttle to ~20 ticks/file via `total // 20` with mandatory final-tick emission. Re-bench with the actual `OnnxEmbedder` API: **+1.69%** vs no-progress fast path, 21 ticks per 250-text run (target ~20).

This also closes a subtle thread-safety gap — earlier the callback (often `queue.put_nowait` into the SSE event queue) fired directly from the worker thread; `call_soon_threadsafe` makes the cross-thread contract explicit.

### Manual SSE verify (chunk_progress event flow)

Stood up `mm web` in an isolated worktree (`/tmp/mm-pr653-verify`, isolated `HOME` and DB). Indexed two real markdown fixtures via `GET /api/index/stream`:

**Small file (3 chunks, threshold=1)** — exercises the always-emit path:
```
chunk_progress: 1/3
chunk_progress: 2/3
chunk_progress: 3/3   ← final tick == total
progress: files_done=1
complete: indexed=3
```

**Large file (60 chunks, threshold=32)** — exercises the throttled path:
- 20 `chunk_progress` events (= `60 // 3`-step throttle, matches `total // 20` heuristic)
- Final tick lands at 60/60 before the file's `progress` event
- `complete: indexed_chunks=60, errors=[], duration_ms=267.2`

UI screenshot/GIF (small → large in one frame, per the review suggestion) is the one item I'm leaving for direct capture — Playwright MCP wasn't available in this session and SSE event-flow already proves the data contract end-to-end. Will attach as a follow-up comment when I do the visual check.

CI test matrix should be green by now; other than the visual capture this PR is functionally verified.

memtomem:
## Follow-up tracking

Three issues filed for the items deferred from this PR's scope:

- #654 — Memory Dirs reindex button chunk display (needs layout design first)
- #655 — Wizard `_seed_with_progress` chunk label (needs `item_show_func` redesign)
- #656 — `mm index` (non-stream CLI) → stream conversion (likely bundles with #655)

A 2-week automated review of the `progress_threshold = 32` default is also scheduled (one-shot remote agent, fires 2026-05-15 ~10:00 KST, report-only — will post a single comment here with operator-signal findings or "no signal, default fine").

That closes the follow-up surface for this PR. Functional verification (CI, ONNX wall-clock, SSE event flow) is in the comments above; the only remaining manual item is the UI screenshot/GIF, which I'll attach when I do the visual check.

Issue #647, opened 2026-05-01
Windows: Path canonicalization in indexing — `kind` detection + memory_dir comparisons


memtomem:
## Background

Indexing-related tests fail on Windows where path comparisons across different canonicalization layers (drive letter case, short-name vs long-name, separator) don't match. Symptoms include \`kind=None\` where \`'memory'\` was expected, chunks not getting deleted because their source paths don't compare equal to \`memory_dirs\`, etc.

## Symptoms

\`\`\`
FAILED test_web_routes.py::TestSources::test_kind_set_when_source_under_memory_dir
  AssertionError: assert None == 'memory'

FAILED test_web_routes.py::TestRemoveMemoryDirChunkCleanup::test_delete_chunks_true_removes_matching_source_files
  assert 0 == 4

FAILED test_web_routes.py::TestUploadsUsage::test_populated
  assert 0 == 2
\`\`\`

The pattern: chunks have a source path stored at index time; later code compares against a configured \`memory_dir\`. On POSIX both go through \`Path.resolve()\` consistently. On Windows, the resolution may differ (drive case, \`\\\\?\\\\\` long-path prefix, short \`8.3\` form) so equality fails.

## Affected files (estimated)

- \`packages/memtomem/tests/test_indexing_engine.py\` — 11 failures
- \`packages/memtomem/tests/test_web_routes.py::TestSources\` / \`TestRemoveMemoryDirChunkCleanup\` / \`TestUploadsUsage\`
- Possibly: \`test_dirty.py\`, \`test_index_debounce.py\`

## Fix direction

1. Audit the production code that compares \`chunk.metadata.source_file\` against \`memory_dirs\` — confirm both go through identical canonicalization (\`Path.resolve()\` after a consistent normalization).
2. Consider storing canonical paths at index time so later comparison is just \`==\`, no resolve-on-read race.
3. Test fixtures may need to use \`tmp_path.resolve()\` consistently to avoid \`8.3\` vs long-path mismatches.

## Refs

- #634 — Windows + macOS CI matrix expansion (umbrella)
- PR #638 — matrix infrastructure (Windows currently informational)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #646, opened 2026-05-01
Windows: `'home'` fs alias + tilde expansion in web fs-list endpoint


memtomem:
## Background

\`test_web_routes.py::TestFsList\` and related fs-listing tests assume the user's home directory surfaces as the alias \`'home'\` and that \`~\` resolves consistently. On Windows the runner's home is \`C:\\Users\\runneradmin\\\`, and the fs-list endpoint surfaces the literal username instead.

## Symptoms

\`\`\`
FAILED test_roots_no_path_or_empty - assert 'runneradmin' == 'home'
FAILED test_path_param_tilde_expansion - {\"detail\":\"not_found\"} (404 instead of 200)
FAILED test_outside_allow_list_422 - 200 instead of 422 (path canonicalization differs)
\`\`\`

## Affected files

- \`packages/memtomem/tests/test_web_routes.py::TestFsList\` — multiple tests
- Possibly: \`test_web_exclude_guard.py\`

## Root cause hypothesis

The endpoint computes the alias from \`Path.home()\` and exposes \`{name: \"home\"}\` as a stable label on POSIX. On Windows, \`Path.home().name\` is \`runneradmin\` (the actual username), not the literal string \`\"home\"\`. The production code likely does \`alias = path.name\` which works on POSIX (because home is always \`/home/<user>\` on Linux and \`/Users/<user>\` on macOS, with \`.name = <user>\`) but the test asserted the alias is exactly \`\"home\"\`.

Need to inspect the endpoint to confirm whether the alias is hard-coded or derived. Either:
- production code intends a stable label and should hard-code \`\"home\"\` for the home root regardless of platform, OR
- tests should derive the expected name dynamically from \`Path.home().name\`

## Refs

- #634 — Windows + macOS CI matrix expansion (umbrella)
- PR #638 — matrix infrastructure (Windows currently informational)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

memtomem:
Resolved by PR #714 — closing.

The three failures listed here were all symptoms of #644 (HOME monkeypatch silently ignored on Windows because `Path.home()` reads `USERPROFILE` first). Once `set_home(monkeypatch, path)` started setting both `HOME` and `USERPROFILE`:

- `test_roots_no_path_or_empty` — `Path.home()` now resolves to `tmp_path/home`, whose `.name == "home"` matches the assertion.
[...]

Issue #645, opened 2026-05-01
Windows: config_signature unchanged after write — NTFS mtime resolution / cache


memtomem:
## Background

\`test_web_hot_reload.py\` has 17 failures on Windows — the entire file. The recurring symptom is the config signature (\`(path, mtime_ns)\` tuple) not changing across two writes that should have updated it.

## Symptoms

\`\`\`
FAILED TestSignature::test_signature_changes_on_config_write
  AssertionError: assert (('C:\\Users\\...\\config.json', 1777609014603566900), ...)
                      != (('C:\\Users\\...\\config.json', 1777609014603566900), ...)

FAILED test_concurrent_patches_are_serialised_by_lock
  assert 10 in (42, 99)

FAILED test_reload_if_stale_cas_yields_to_concurrent_writer_bump
  AssertionError: writer's signature was overwritten by the reader's tail
\`\`\`

The \`mtime_ns\` value is identical across the two snapshots — meaning the second \`write_text\` didn't bump the modification time. Several possibilities:

1. **NTFS lazy timestamp**: NTFS may defer mtime updates briefly; \`os.fsync\` doesn't always flush metadata immediately on Windows.
2. **Test timing**: writes in the same millisecond report the same \`mtime_ns\` (Windows mtime resolution is finer than that, but the OS may round to FILE_ATTRIBUTE_TIMESTAMP granularity).
3. **Cache layer**: \`config_signature\` might be cached at a level that doesn't invalidate on Windows (e.g., \`stat\` result caching).

## Affected files

- \`packages/memtomem/tests/test_web_hot_reload.py\` — 17 failures (entire file's relevant tests)
- Possibly: \`test_storage.py\` (1 failure), other mtime-sensitive tests

## Investigation needed

1. Confirm whether the issue is (a) production code's mtime read missing a flush, (b) test timing assumption (writes-too-fast-to-distinguish), or (c) Windows file-attribute cache not invalidating.
2. If (a): fix \`config_signature\` to read mtime via \`os.stat(path)\` *after* \`fsync\` in the writer, or via \`Path.stat()\` with \`follow_symlinks=False\` consistently.
3. If (b): tests should sleep \`time.sleep(0.01)\` between writes, or use a monotonic counter instead of mtime for tests.
4. If (c): may need \`FlushFileBuffers\` (via \`ctypes\` to \`kernel32\`) on Windows, but that's a heavy hammer.

## Refs

[...]

Issue #644, opened 2026-05-01
Windows: HOME monkeypatch doesn't reach `Path.home()` on Windows (uses USERPROFILE)


memtomem:
## Background

Many CLI tests use \`monkeypatch.setenv(\"HOME\", str(tmp_path))\` to isolate \`~/.memtomem/\` writes from the developer's real home. On Windows, \`Path.home()\` consults \`USERPROFILE\` (then \`HOMEDRIVE\`+\`HOMEPATH\`), so the HOME monkeypatch is silently ignored — tests write into / read from the actual runner home (\`C:\\Users\\runneradmin\\.memtomem\\\`).

This appears to be the largest single category of Windows failures from PR #638's matrix run (~30+ tests across multiple files).

## Symptoms

- Tests that assert state file existence under \`tmp_path / \".memtomem\"\` fail because the file landed at the real home
- Cross-test bleed: state from \`test_A\` (writing to real home) breaks \`test_B\`'s \"no config\" precondition (e.g., \`TestRequireConfigured\` 409→200 mismatches in \`test_web_routes.py\`)
- Path mismatches in error messages

## Affected files (estimated)

- \`packages/memtomem/tests/test_init_cmd.py\` — 23 failures
- \`packages/memtomem/tests/test_uninstall_cmd.py\` — 12 failures
- \`packages/memtomem/tests/test_upgrade_cmd.py\` — 4 failures
- \`packages/memtomem/tests/test_web_routes.py::TestRequireConfigured\` — config-gate failures
- Subset of: \`test_config_overrides.py\`, \`test_context_status.py\`, \`test_dirty.py\`

## Fix patterns

**Per-test**: monkeypatch both \`HOME\` and \`USERPROFILE\` (or whichever \`Path.home()\` reads on the current platform):

\`\`\`python
home_var = \"USERPROFILE\" if sys.platform == \"win32\" else \"HOME\"
monkeypatch.setenv(home_var, str(tmp_path))
\`\`\`

**Centralized**: introduce a \`monkeypatched_home(monkeypatch, tmp_path)\` helper in \`tests/helpers.py\` so the platform branching lives in one place and tests just call \`monkeypatched_home(monkeypatch, tmp_path)\`. Audit all current \`monkeypatch.setenv(\"HOME\", ...)\` sites and migrate.

**Production code**: confirm \`Path.home()\` is the only consumer; if any CLI code reads \`os.environ[\"HOME\"]\` directly, that path is also broken on Windows.

## Refs

[...]

PR #658, opened 2026-05-01
feat(web): meta row chunk progress for Memory Dirs reindex (#654)

### LINKED ISSUE #654: feat(web): show chunk count on Memory Dirs reindex button
## Context

PR #653 added `chunk_progress` SSE events to the indexing stream and surfaced them in the **Index** tab as `file.md — 100/250 chunks`. The **Sources → Memory Dirs** reindex flow consumes the same stream (`sources-memory-dirs.js` `mdReindexOne`, line 1093-1177), but PR #653 deliberately left it untouched because the per-button text already reads `5/12` and adding `5/12 — 100/250` would crowd or wrap the button.

This issue tracks the design-and-implement work for chunk-level visibility in that flow.

## Why this needs design, not just code

The Memory Dirs button is a small, fixed-width control inside a row of vendor cards. Naive concatenation breaks the layout. Options that need a decision:

1. **Two-line button** — file count on top, chunk count below. Costs vertical space across the entire dirs list.
2. **Tooltip / aria-label** — hover/focus shows `5/12 files, 100/250 chunks`. Keeps button text untouched but accessibility-only.
3. **Replace button text mid-flight** — show only `100/250 chunks` while a single file is in-flight, snap back to `N/M` once the file completes. Most informative but visually noisy.
4. **Inline progress bar under the button** — full bar like the Index tab. Most consistent UX but the largest layout change.

## Acceptance criteria

- During a Memory Dirs reindex, the user can see **per-chunk progress** for the file currently being embedded — no requirement on the exact form.
- Existing `5/12` file counter behavior preserved (or knowingly replaced).
- Layout does not wrap or shift adjacent dirs cards on any of: 1280px, 1024px, narrow mobile.
- Throttle (~100ms) and final-tick (`done == total`) bypass match the Index tab's behavior — copy/paste from `app.js:runIndexStream` is fine.
- No new SSE event types needed; consume the existing `chunk_progress` event from the stream.

## Files likely touched

- `packages/memtomem/src/memtomem/web/static/sources-memory-dirs.js` (`mdReindexOne`, line 1093-1177)
- `packages/memtomem/src/memtomem/web/static/style.css` (if option 1 or 4)
- `packages/memtomem/src/memtomem/web/static/locales/en.json` + `ko.json` (new label key)

## References

- PR #653 (parent feature)
- `feedback_layout_primitives_first.md` pattern — check existing `.dir-button` / `.reindex-btn` overflow rules before adding new CSS.

memtomem:
## Summary

Closes #654. Surfaces per-chunk progress (`file.md — 100/250 chunks`) during a Memory Dirs reindex by **reusing the existing `.memory-dirs-item-meta` slot** — no new CSS, no new i18n key, no button-width churn.

The Memory Dirs reindex flow already consumed the same SSE stream as the Index tab (PR #653), but only handled `progress` events. While a single 250-chunk file embedded, the button counter sat silently on `5/12` for tens of seconds. Widening the button text was rejected at plan time (button polo polo, vertical alignment jank); the meta row already carries mutable counts (`12 files · 1,847 chunks`), already has the nowrap+ellipsis CSS guards (`style.css:1214-1222`), and is the same one-line semantic as `app.js:runIndexStream`'s `fileEl` — so the user's mental model is consistent across tabs.

### Implementation

- **`mdReindexOne`** (`sources-memory-dirs.js`): cache `metaEl` + `_origMeta` next to `_origText`; add `chunk_progress` branch (100ms throttle + final-tick bypass + `isConnected` guard, mirroring `app.js:3739`); `progress` branch resets the throttle and snaps the meta row back to stats via a flag.
- **`progress` flag-gated reset** closes the **[big, small, small, ...] residue case** explicitly: server-side `progress_threshold` gates `chunk_progress` events, so a 250-chunk file followed by small notes would otherwise leave the meta row stuck on `"CHANGELOG.md — 250/250"` through the rest of the run. The flag also avoids a sub-100ms flash on big→big transitions because the next `chunk_progress` overwrites in-place.
- **i18n key reused** (`index.file_chunk_progress` in `en.json:168` / `ko.json:168`) — namespace mismatch (`index.*` used in Sources tab) is a knowingly accepted debt; helper-extraction follow-up issue covers the rename.
- **`index.html` cache-buster** bumped (`/sources-memory-dirs.js?v=9` → `?v=10`).

### Out of scope (follow-ups)

- Helper extraction sharing the `_lastChunkRender` + 100ms-throttle + final-tick-bypass pattern between `app.js:runIndexStream` and `sources-memory-dirs.js:mdReindexOne` (rule-of-three: defer until a third call-site appears; the helper PR will also rename the i18n key to `common.*`). — separate issue.
- JS DOM interaction test infrastructure (Playwright / Vitest+jsdom). — separate issue.
- Wizard `_seed_with_progress` chunk label / `mm index` CLI progress (PR #653 follow-ups, unchanged here).

## Test plan

- [ ] **Threshold-below**: small markdown dir → reindex → meta row stays on stats, button ticks `N/M`. No `chunk_progress` noise.
- [ ] **Threshold-above (single big file)**: dir with `CHANGELOG.md`-style 100+ heading file → button ticks `5/12` etc., meta row ticks `CHANGELOG.md — 100/250 chunks` at ~100ms gap, final `(N/N)` lands before the next file boundary.
- [ ] **Mixed [big, small, ...]**: dir with one 100+-heading file + 8 small notes → after the big file's `(N/N)`, the meta row snaps back to stats *immediately* and stays there through the small tail. (Critical regression check — without the flag-gated reset, the meta row would freeze on `250/250` through the back half.)
- [ ] **Layout (1280 / 1024 / mobile)**: chunk label ellipsis-truncates, no row height/width shift on adjacent dirs cards.
- [ ] **Cancel mid-flight**: close browser tab during reindex → `/api/indexing/active` returns to 0 (PR #653 behavior reused).
- [ ] **i18n toggle (KO ⇄ EN)**: locale switch mid-reindex → next `chunk_progress` event reflects the new locale string.
- [ ] `uv run ruff check packages/memtomem/src && uv run ruff format --check packages/memtomem/src` — green.

## References

- Parent feature: PR #653.
- Layout-primitives-first rationale (Memory: `feedback_layout_primitives_first.md`) — meta row's existing CSS guards make this a zero-CSS change.
- Plan: `~/.claude/plans/go-velvety-penguin.md` (local).

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/index.html
+++ b/packages/memtomem/src/memtomem/web/static/index.html
@@ -1297,7 +1297,7 @@
   <script src="/i18n.js?v=3"></script>
   <script src="/app.js?v=82"></script>
   <script src="/settings-config.js?v=9"></script>
-  <script src="/sources-memory-dirs.js?v=9"></script>
+  <script src="/sources-memory-dirs.js?v=10"></script>
   <script src="/settings-maintenance.js?v=1"></script>
   <script src="/settings-namespaces.js?v=5"></script>
   <script src="/settings-harness.js?v=1"></script>
--- a/packages/memtomem/src/memtomem/web/static/sources-memory-dirs.js
+++ b/packages/memtomem/src/memtomem/web/static/sources-memory-dirs.js
@@ -1096,6 +1096,20 @@ async function mdReindexOne(path, btn) {
   // Cache the original button label so we can restore it after the run —
   // ``btnLoading`` only toggles disabled+spinner, it doesn't snapshot text.
   const _origText = btn ? btn.textContent : '';
+  // Slot-reuse for chunk_progress (#654): the meta row already carries
+  // mutable count text ("12 files · 1,847 chunks"), already has the
+  // nowrap+ellipsis CSS guards, and is the only single-line slot near
+  // the button that won't shift adjacent dirs cards when its content
+  // grows. We borrow it for in-flight ``file.md — N/M chunks`` ticks
+  // and snap back via ``_origMeta`` on file boundary / cleanup.
+  const _item = btn ? btn.closest('.memory-dirs-item') : null;
+  const metaEl = _item ? _item.querySelector('.memory-dirs-item-meta') : null;
+  const _origMeta = metaEl ? metaEl.textContent : '';
+  // Throttle state — function-local so two concurrent reindex clicks
+  // (one per dir) don't share each other's clocks. Mirrors the
+  // ``runIndexStream`` closure scope in ``app.js``.
+  let _lastChunkRender = 0;
+  let _metaIsChunkLabel = false;
   showToast(t('toast.memory_dir.reindex_started', { path }), 'info');
 
   const _cleanup = () => {
@@ -1103,6 +1117,7 @@ async function mdReindexOne(path, btn) {
       btn.textContent = _origText;
       btnLoading(btn, false);
     }
+    if (metaEl) metaEl.textContent = _origMeta;
     if (typeof _indexingEnd === 'function') _indexingEnd();
     if (typeof loadSources === 'function') loadSources();
   };
@@ -1140,9 +1155,40 @@ async function mdReindexOne(path, btn) {
     }
     _sseFailCount = 0;
     if (event.type === 'progress') {
+      // File boundary: reset chunk throttle so the next file's first
+      // chunk renders immediately, and close the [big, small, ...]
+      // residue case (#654) — once a 250-chunk file ends, a stream of
+      // small files won't emit ``chunk_progress`` (server-side
+      // ``progress_threshold`` gate), which would otherwise leave the
+      // meta row stuck on "CHANGELOG.md — 250/250" through the rest
+      // of the run. The flag avoids a sub-100ms flash on big→big
+      // transitions because the next ``chunk_progress`` overwrites.
+      _lastChunkRender = 0;
+      if (metaEl && _metaIsChunkLabel) {
+        metaEl.textContent = _origMeta;
+        _metaIsChunkLabel = false;
+      }
       if (btn) {
         btn.textContent = `${event.files_done}/${event.files_total}`;
       }
+    } else if (event.type === 'chunk_progress') {
+      // Mirrors ``app.js:runIndexStream`` — 100ms throttle on
+      // intermediate ticks, final tick (done >= total) bypasses so
+      // ``(N/N)`` lands before the next file boundary. ``isConnected``
+      // guards against late events arriving after ``loadSources()``
+      // detached the row (silent no-op today, but cheap insurance).
+      if (!metaEl || !metaEl.isConnected) return;
+      const now = (typeof performance !== 'undefined' && performance.now)
+        ? performance.now() : Date.now();
+      const isFinal = event.chunks_done >= event.chunks_total;
+      if (!isFinal && now - _lastChunkRender < 100) return;
+      _lastChunkRender = now;
+      metaEl.textContent = t('index.file_chunk_progress', {
+        file: basename(event.file),
+        done: event.chunks_done,
+        total: event.chunks_total,
+      });
+      _metaIsChunkLabel = true;
     } else if (event.type === 'complete') {
       _completed = true;
       es.close();


memtomem:
## Overview

Closes #654. Wires the existing `chunk_progress` SSE events (PR #653) into the Memory Dirs reindex flow by repurposing the `.memory-dirs-item-meta` slot for in-flight ticks. Two files, +47/-1.

CI: lint, typecheck, test (ubuntu / macos), golden-path, notebooks, CLA all green; test (windows-latest) was IN_PROGRESS at review time.

## Strengths

- **Layout-primitives-first** — reuses the meta row's existing `white-space: nowrap; overflow: hidden; text-overflow: ellipsis` (`style.css:1214-1222`) so this is a zero-CSS change. Row stays one line at every viewport.
- **Same semantic as the Index tab** — `metaEl.textContent` mirrors `fileEl.textContent`. Mental model stays consistent across tabs.
- **[big, small, ...] residue case explicitly closed** — `_metaIsChunkLabel` flag + `progress`-branch reset means a 250-chunk file followed by small notes (below the engine's `progress_threshold` gate) doesn't leave the meta row stuck on `250/250` through the back half. Big→big transitions are overwritten by the next `chunk_progress` so there's no sub-100ms flash.
- **Function-local throttle state** — two concurrent reindex clicks don't share clocks (matches `runIndexStream`'s closure scope).
- **`isConnected` guard** — defends against late events landing on a row already detached by `loadSources()`. Silent no-op today, cheap insurance for future MutationObserver / reactive wrappers.

## Nits

### 1. `_item` vs `metaEl` naming inconsistency

The leading underscore on `_item` isn't a marker the rest of the file uses; same scope, same kind (DOM node cache), different convention. Recommend inlining — `_item` is a throwaway:

```diff
-  const _item = btn ? btn.closest('.memory-dirs-item') : null;
-  const metaEl = _item ? _item.querySelector('.memory-dirs-item-meta') : null;
+  const metaEl = btn?.closest('.memory-dirs-item')
+    ?.querySelector('.memory-dirs-item-meta') ?? null;
```

One fewer local, narrower scope.

### 2. Comment density (~47% of added lines)

22 of the 47 added lines are comments. `CLAUDE.md` says "default to no comments — only when WHY is non-obvious." Each comment here has a defensible WHY, but most of the same content already lives in the commit body / PR description and is reachable via `git blame`. Suggest a follow-up doc-cleanup commit (or compress before merge) to bring it down to roughly:

- one line on the slot-reuse rationale (`#654 — meta row borrowed for chunk_progress`)
- one line on the [big, small, ...] reset (`flag-gated reset closes residue case`)
- drop the "mirrors `app.js:runIndexStream`" line on the `chunk_progress` branch — `basename` + `t('index.file_chunk_progress')` already make the parallel obvious.

Letting comment density drift up unchallenged raises the codebase's baseline silently.

### 3. Missing client-side note on the `progress` ↔ `chunk_progress` ordering invariant

The reset logic depends on the server emitting `progress` (file boundary) and `chunk_progress` events in a specific order — codified in `test_indexing_engine.py`'s strict-ordering invariant from PR #653. The dependency is implicit in the JS today; one comment makes the implicit explicit so a future server-contract change pages this code:

```js
// Depends on the server invariant that chunk_progress events for a file
// arrive between its enclosing progress events (see
// test_indexing_engine.py::test_indexing_engine_chunk_progress_ordering).
```

This is exactly the kind of comment `CLAUDE.md` endorses — external invariant dependency, not behavior narration. Worth adding even given the density concern in nit 2.

## Observations

- **Auto-test absence is a known gap, tracked.** This PR adds another flow (Memory Dirs reindex happy-path + the [big, small, ...] residue scenario) that #660 should pick up as one of its first migration candidates.
- **Trust boundary worth noting.** SSE source is same-origin, authenticated local user — fine for `mm web`. If multi-user / reverse-proxy exposure (cf. #649) ever lands, revisit whether `event.file` / `event.chunks_*` need additional validation before reaching `t()` / `textContent`. Out of scope here.
- **`progress` event semantics** — code assumes `event.file` is the *currently active* file (matches `runIndexStream`'s use). If the server ever flips `progress` to mark *completion* with the just-finished file's path, the `_metaIsChunkLabel` reset fires one beat early; the next `chunk_progress` overwrites it so behavior stays correct. The comment in nit 3 is the canary.

## Edge case verified

`_cleanup()` → `loadSources()` → row replaced → user immediately re-clicks reindex → new row's button has a new click handler → previous invocation's `metaEl` is detached → `isConnected` guard makes any late `chunk_progress` a no-op. ✓

## Verdict

**Approve.** Three nits, all non-blocking. Highest-value follow-ups are nit 3 (ordering-invariant comment) and nit 2 (compress comment density). CI green → admin-merge ready once Windows job finishes.

memtomem:
## Second pass — re-review after CI completion

`test (windows-latest)` finished as **FAILURE** since the first review (was IN_PROGRESS). Investigated whether this PR introduced the regression or whether it's pre-existing infrastructure noise.

### Verdict: pre-existing, unrelated to this PR

- **This PR's diff is JS-only** (`sources-memory-dirs.js` + `index.html` cache-buster). No Python touched, no Python test imports either file.
- **`main` shows the identical failure pattern** on the last 5 CI runs (e.g. `25210401150`, `25209722171`, `25208059888`) — Windows job consistently FAILURE while overall workflow conclusion is `success` (Windows is non-blocking on this repo's required-checks config).
- **Failure clusters cleanly into already-tracked Windows-only issues** opened 2026-05-01:
    - #643 — POSIX path-separator assertions in tests (`/` vs `\`) — covers `test_wiki_cmd_override.py::test_wiki_*_stdout_contract` (`agents\demo\overrides\codex.toml` ≠ `agents/demo/overrides/codex.toml`).
    - #644 — `Path.home()` doesn't pick up `HOME` monkeypatch on Windows (uses `USERPROFILE`) — covers `TestFsList::test_roots_no_path_or_empty` (`runneradmin` ≠ `home`) and the `~` expansion failures.
    - #645 — `config_signature` unchanged after write (NTFS mtime resolution / cache) — covers the entire `test_web_hot_reload.py` cluster (`signature_changes_on_*`, `reload_if_stale*`, etc.).
    - #646 — `'home'` fs alias + tilde expansion — covers `TestFsList::test_path_param_tilde_expansion` and `test_outside_allow_list_422`.
    - #647 — Path canonicalization for `kind` detection + memory_dir comparisons — covers `test_kind_set_when_source_under_memory_dir`, the `TestRequireConfigured::*[index/stream|reindex|upload|add]` 409→200/403/422 mismatches, etc.

None of those touch the SSE event-handling path that this PR modifies. The 116-failure count is alarming on first read but resolves to "Windows infra debt already in flight, not introduced here."

### No change to merge readiness

- Required checks (lint, typecheck, test ubuntu/macos, golden-path, notebooks, CLA) all green.
- Verdict from the first pass holds — **Approve**, three nits non-blocking. Admin-merge clear once a reviewer signs off.

### Adjacent suggestion (out of scope here)

Five separate Windows-only issues in the tracker is a strong signal for a meta-tracker (umbrella issue or label) so newcomers don't open #648, #650, #651 for the same root infrastructure. Per `feedback_umbrella_issue_close_pattern.md` precedent — but tangential to this PR.

PR #661, opened 2026-05-01
feat(wizard): explicit guardrails for failed init steps (closes #626)

### LINKED ISSUE #626: [feature]: add guardrails for failed steps in the `mm init` wizard
### Problem

It might be helpful to add some guardrails to the `mm init` wizard.

Currently, even if something goes wrong during the wizard, the setup flow may continue to the next steps. Continuing while silently ignoring a failed or missing intermediate step could potentially lead to confusing behavior or later issues.

To prevent this, it may be better if the wizard explicitly informs the user when a step fails, and then either forces the user back to the previous step or aborts the entire `mm init` wizard as failed.

In this particular case, the root cause appears to be unrelated to `memtomem`: the Ollama server is not working correctly. I have not yet identified why the Ollama server fails to start, but it seems clear that the failure itself is caused by something outside of `memtomem`.

```
1. Embedding Provider
  (b: back, q: quit)
  Choose how to generate embeddings:
    [1] Quick start — keyword search only, no setup needed (recommended)
    [2] Local ONNX — dense search, no server needed (pip install memtomem[onnx])
    [3] Ollama — local dense embeddings, needs GPU
    [4] OpenAI — cloud dense embeddings, needs API key
  Select [1]: 3

  Ollama not running. Starting 'ollama serve'...
  Run 'ollama serve' in another terminal if this fails.

  Available models:
    [1] nomic-embed-text — English, fast (768d)
    [2] bge-m3 — multilingual, higher accuracy (1024d)
  Select [1]: 2
  Model 'bge-m3' not found. Pull now? [Y/n]: Y
  Pulling bge-m3... (this may take a few minutes)
  Pull failed: 2026/05/01 03:37:58 WARN Failed to rotate log older=C:\Users\user\AppData\Local\Ollama\app-1.log newer=C:\Users\user\AppData\Local\Ollama\app.log error="rename C:\\Users\\user\\AppData\\Local\\Ollama\\app.log C:\\Users\\user\\AppData\\Local\\Ollama\\app-1.log: The process cannot access the file because it is being used by another process."
time=2026-05-01T03:37:58.901+09:00 level=INFO source=app_windows.go:282 msg="starting Ollama" app=C:\Users\user\AppData\Local\Programs\Ollama version=0.22.0 OS=Windows/10.0.19045
time=2026-05-01T03:37:58.902+09:00 level=INFO source=eventloop.go:309 msg="existing instance found, not focusing due to startHidden"
time=2026-05-01T03:37:58.902+09:00 level=INFO source=app_windows.go:79 msg="existing instance found, exiting"
Error: timed out waiting for server to start
  You can pull manually: ollama pull bge-m3

2. Reranker (optional)
  (b: back, q: quit)
  Cross-encoder reranking sharpens search relevance after BM25+dense fusion.
  Runs locally via fastembed ONNX — no API key, no server.
  Enable reranker? [y/N]:
```

Looking at part of the wizard flow, during the “Available models” selection step, `ollama list` appears to be executed internally to retrieve the list of installed models. However, because the Ollama server is not working properly, the actual output is: `Error: timed out waiting for server to start`

The wizard seems to receive this output and interpret it as meaning that the selected model is not installed, so it then attempts to pull the model. (Even though the model is already pulled and installed.)

Although the root cause and the resulting issue are relatively minor, the current flow feels somewhat awkward because the wizard continues to the next step despite the failed operation. It would be better to make the flow stricter by enforcing a clear failure path for the user.

### Proposed solution

I think the user experience would be better if guardrails were added so that, when a step fails during the wizard, the user is either returned to the previous step or the entire wizard is treated as failed and aborted.

### Alternatives considered

_No response_

### Additional context

_No response_

memtomem:
## Summary

- Adds `StepRetry` exception and `fail_step()` helper to `wizard.py` so a failed wizard step prompts the user to **retry / back / quit** instead of silently advancing.
- Wires the new mechanism into `_step_embedding`'s Ollama branch — server reachability check, model verification, and `ollama pull` are now guarded.
- `_ollama_has_model` now returns `bool | None` (tri-state) so callers distinguish "server unreachable / errored" from "model absent". The original `False`-on-error conflation is the root cause of the #626 misread (a server timeout was being interpreted as "model not installed", triggering a doomed pull and silent fall-through to the Reranker step).
- Drops the misleading "Ollama not running. Starting 'ollama serve'..." line — the wizard never actually started the server, only printed the line.

## Reproduction (the #626 scenario)

```
[3] Ollama → bge-m3 → "model not found" → pull fails → wizard moves on as if nothing happened
```

After this PR the user sees:

```
✗ Pull failed: Error: timed out waiting for server to start
Retry, back, or quit? [R/b/q]:
```

`r` re-runs only the pull (model selection is preserved). `b` returns to the previous step. `q` cancels the wizard via the same path as the `q` shortcut.

## Incidental scope

The OpenAI branch's "API key invalid + decline to continue" path used `raise SystemExit(1)`; switched to `WizardCancel` for uniform cancel UX. Exit code on this path changes from 1 to 0 — this is a deliberate cancel by the user, not an error. No tests pinned the previous exit code.

## Out of scope (follow-up)

- Silent-continue audit of other steps (`_step_memory_dir` dir creation, `_step_settings` writes, etc.). The mechanism added here can be reused per step in a follow-up PR.
- Automatic backoff retry on transient failures.
- Actually starting `ollama serve` from the wizard.

## Test plan

- [x] `uv run ruff check + format --check packages/memtomem/{src,tests}` clean
- [x] `uv run pytest -m "not ollama"` — 3873 passed
- [x] New unit tests pin the `StepRetry` / `fail_step` choice→exception mapping (12 cases) and the `_ollama_has_model` tri-state contract (5 cases)
- [x] New integration tests pin the #626 invariant: `_step_embedding` does NOT silently advance to the next step on Ollama failure (sentinel next-step check)
- [x] Manual pty-driven smoke against `mm init --advanced` with a fake `ollama` binary, in 5 scenarios:
  - **A** (server unreachable + `q`) — red `✗ Ollama server is not reachable.` prompt fires, wizard cancels (exit 0). The `Starting 'ollama serve'...` lie is gone.
  - **B** (server unreachable + `b`) — falls back to `(already at first step)`, re-runs `_step_embedding` from the top. Wizard never advances past it.
  - **C** (server unreachable + `r` with server still broken + `q`) — outer retry loop re-runs the server check only; the embedding-provider menu does NOT re-render.
  - **D** (server up, model missing, pull fails + `r` + still fails + `q`) — inner retry loop re-runs only the pull action; the model-selection menu is NOT re-prompted.
  - **E** (pull fails + `b`) — wizard returns to the embedding choice; `_step_reranker` (next step) never runs. This is the core #626 invariant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -173,6 +173,24 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### Changed
 
+- **`mm init` wizard now prompts on failed steps instead of silently
+  advancing (#626).** When `_step_embedding`'s Ollama branch hits a
+  failure (server unreachable, model status indeterminate, `ollama pull`
+  errored), the wizard surfaces a red `✗` line and a
+  `Retry, back, or quit? [R/b/q]` prompt — pre-fix it printed a warning
+  and walked into the next step (Reranker), which masked the real
+  problem. Mechanism is reusable: new `StepRetry` exception +
+  `fail_step()` helper in `memtomem.cli.wizard`. Tri-state
+  `_ollama_has_model` returns `bool | None` so server-error is no longer
+  conflated with "model absent". **Behavior change**: the OpenAI
+  `--provider openai` "API key invalid + decline to continue" path now
+  exits **0** (cancelled by user) instead of **1** (error). Scripts
+  parsing the exit code to detect a bad-key cancel must switch to
+  parsing stdout for `Wizard cancelled.` or use `--yes` mode (which
+  retains the non-zero exit on missing extras). Other wizard steps
+  still use the pre-#626 silent-continue pattern; an audit and follow-up
+  PR will migrate them onto `fail_step` next.
+
 - **`embedding.threads` default flipped from `0` (= ORT default = all
   physical cores) to `4`** so a bulk reindex doesn't pin every core and
   starve the web server / other apps. Live diagnosis of #640 confirmed
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -590,8 +590,6 @@ def _step_embedding(state: dict) -> None:
         else:
             click.secho("  API key test failed. Check your key and try again.", fg="red")
             if not nav_confirm("  Continue anyway?", default=False):
-                # Was ``raise SystemExit(1)`` — switched to ``WizardCancel``
-                # so the wizard exits via the same path as ``q`` (#626).
                 raise WizardCancel()
 
     state["provider"] = provider
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -17,7 +17,16 @@ from typing import Literal
 import click
 
 from memtomem.cli.init_presets import PRESETS, _VALID_PRESETS, get_preset
-from memtomem.cli.wizard import nav_confirm, nav_prompt, run_steps, silent_step, step_header
+from memtomem.cli.wizard import (
+    StepRetry,
+    WizardCancel,
+    fail_step,
+    nav_confirm,
+    nav_prompt,
+    run_steps,
+    silent_step,
+    step_header,
+)
 
 InstallType = Literal["source", "project", "tool", "uvx"]
 CwdInstallType = Literal["source", "project", "pypi"]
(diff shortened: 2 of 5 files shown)


memtomem:
Admin-merging despite `test (macos-latest)` red.

**Evidence the failure is pre-existing and unrelated to this PR:**
- 3 consecutive macOS failures, all in `packages/memtomem/tests/test_embedding_progress.py` (introduced in #653):
  1. `test_onnx_streams_progress_per_yield[asyncio]` — `expected 5 calls, got []`
  2. `test_onnx_streams_progress_per_yield[asyncio]` — same assertion (rerun #1)
  3. `test_onnx_throttles_thread_hops_for_large_input[asyncio]` — `expected ~20 throttled ticks for n=200, got 0`
- The same test family failed on the `e034d10` main run (the very PR that introduced these tests). The pattern is `on_progress` callbacks not firing on the macOS GitHub runner with Python 3.14.
- This PR's diff is scoped to `cli/wizard.py`, `cli/init_cmd.py`, the corresponding tests, and `CHANGELOG.md` — nothing touching ONNX, async progress, or embedding internals.
- Local pytest on macOS passes the entire `test_embedding_progress.py` file (12/12).
- All other required checks green: `lint`, `notebooks`, `golden-path (ONNX bge-m3)`, `test (ubuntu-latest)`, `typecheck`. Windows is not in the required list.

The flake is upstream of this PR and should be triaged separately.

Issue #672, opened 2026-05-01
_searchByTag pollutes search-input with tag text


memtomem:
## What

`_searchByTag(tag)` (`packages/memtomem/src/memtomem/web/static/app.js:3631-3640`) sets **both** `search-input` (full-text query) and `tag-filter` to the clicked tag value:

```js
qs('search-input').value = tag;   // ← full-text query also gets the tag string
qs('tag-filter').value = tag;     // ← intended tag filter
```

Clicking a tag pill in the Tags tab therefore runs a search where the BM25 query *and* the tag filter both equal the tag string.

## Why this surfaced

While auditing the same axis-mismatch pattern that produced #672 (vendor sub-tab follow-through on cross-tab navigation), this came up as an out-of-scope sibling. The Sources family fixes are about a *new* axis being ignored by older consumers; `_searchByTag` instead has the *same axis duplication* that has shipped since `v0.1.0` (commit `82bdc54`). Filing separately to keep #672 scoped per `CONTRIBUTING.md` "one focused change per PR".

## Two possibilities — needs a decision

1. **Intentional union**: BM25 ranks documents that mention the tag string in their body higher, on top of the tag filter's exact-match constraint. Result set = documents tagged with the tag, ordered by how often they also reference the tag in prose. Some users might rely on this implicit behaviour.
2. **Design flaw**: clicking a tag should narrow to that tag, not add an unrelated full-text constraint. Filling `search-input` makes the search bar look like the user typed it, which is misleading; clearing the tag filter later still leaves the query string behind.

The user-visible behaviour change rules out a silent bug-fix patch — whichever way it lands needs a `CHANGELOG.md` entry.

## Repro

1. Tags tab → Cloud or List view.
2. Click any tag pill.
3. Observe: Search tab activates; search bar shows the tag text; tag-filter dropdown shows the tag; results combine both.

## Suggested resolution path

- Decide intent (likely a UX call — link to similar UX patterns in mem0 / Zep / Letta if useful).
- If "design flaw": drop the `qs('search-input').value = tag` line, keep `tag-filter` only, add a CHANGELOG entry under "Changed".
- If "intentional union": rename to make it obvious (`_searchTextAndTag`?) and add a tag-only entry-point for users who want strict tag scoping.

No regression — `v0.1.0+` behaviour, not a recent ADR-0007 / #604 / #618 follow-up.

PR #587, opened 2026-04-30
feat(web): Index tab — surface hard-to-discover behaviors (static, 7-item bundle) (#579)

memtomem:
Closes #579.

## Summary

Bundles the 7 static-surfacing items from issue #579 into one PR — text/CSS/placeholder/cursor changes only, no policy decisions, no schema breaking. PR #575 unified the Index tab's 3 modes behind a segmented toggle but left several backend behaviors (one-shot vs persistent, where files land, what the Extensions hint means) silent. This bundle fills those gaps.

One commit per item so reviewers can step through; squash merge produces one release-notes line.

## Commits

1. **`feat(web): Index folder panel — surface one-shot vs persistent`** — inline hint at the top of the folder panel ("One-shot scan. For automatic re-indexing, use Sources → Add path.") + matching action button on the success toast that navigates to Sources tab and focuses the Add path button.
2. **`feat(web): Compose — surface target file`** — placeholder shows the actual default save location with today's date (`~/.memtomem/memories/2026-04-30.md (leave blank for auto)`); add success toast echoes the saved path. Removes deprecated `toast.saved_n_indexed` key.
3. **`feat(web): Upload — echo saved path in success toast`** — adds `path: str | None` to `UploadFileResult` (additive, default None — no breaking change), upload route returns the resolved `dest` (post-collision-rename), toast echoes it.
4. **`feat(web): header sys-info chip jumps to Settings → Config`** — `provider/model · backend` chip becomes clickable (cursor + hover + focus-visible), Enter/Space activates, lands on Settings → Config.
5. **`style(web): Compose textarea — double the default height`** — `min-height: 120 → 240px`.
6. **`style(web): Index Extensions hint — drop callout chrome, add ⓘ glyph`** — `#index-config-hint` ID-specific override stripping background/border/padding; ⓘ via CSS `::before` (single source of truth in stylesheet, regardless of writer). Other `.config-hint` users (`#home-config-info`, `#ns-config-info`) untouched.
7. **`copy(web): Compose target file label — clarify "save location"`** — `Target File` → `Target file (save location)` / `대상 파일 (저장 위치)`.

## Risk handling

These were flagged before coding; addressed in the commits above.

- **i18n.applyDOM rewrites textContent** on every lang toggle and would strip a nested `<a>` from a single `data-i18n` paragraph. Commit 1 splits the hint into sibling `<span>` + `<a>` + `<span>` so the link element survives toggles.
[...]

PR #673, opened 2026-05-01
fix(web): vendor sub-tab follow-through on cross-tab source navigation

memtomem:
## Summary

Two cross-tab navigation entry-points into the Sources panel set one axis (namespace filter or source path) but did not align the panel's vendor sub-tab axis (user / claude / openai). Because the panel only renders the active vendor's tree (#570), the user-visible chip / active-source state suggested the link succeeded while the rendered content was an unrelated vendor — reported as "Sources 화면으로 넘어가긴 하는데 직접 링크가 안 된다."

Audit (3 parallel Explore agents) found `_navigateToSource` is the same shape of bug; both fixes share the helper and live next to each other in `app.js`, so they ship together as one focused change.

### `navigateToSourcesByNs` (Settings → Namespaces → Sources button)

`packages/memtomem/src/memtomem/web/static/app.js` — `_renderMemorySourceTree`:
- `filterActive` now also returns true when `STATE.sourcesNsFilter` is set, not just when the path-text input is non-empty. Empty memory_dir headers (28 of 29 in the repro for a `claude:…` namespace) are now hidden under the existing `visibleCatsRaw` guard.
- After computing per-vendor totals, if the current vendor has zero matches and another vendor has any, the active sub-tab shifts to the vendor with the most. Not persisted via `_writeActiveSourcesVendor` — this is a navigation hint, not the user's stored preference, so the next manual click still wins.

### `_navigateToSource` (Home recent source, command palette)

Same root cause, narrower entry-point. Resolves the target vendor from the source's `memory_dir → provider` (looked up via `STATE.allSources` + `STATE.memoryStatusByPath`) before `activateTab('sources')`, so a Claude/OpenAI source clicked from Home doesn't land on whichever sub-tab the user last viewed and silently miss the `.source-item[title=path]` scroll-into-view that follows. Falls through to current behaviour when `STATE.allSources` isn't populated yet (cold load) — no regression.

### Cache bust

`index.html` `app.js?v=82 → v=84` so the JS body change is actually picked up by browsers (per [`feedback_static_asset_cache_bust.md`](https://github.com/memtomem/memtomem/issues/587), unbumped query strings get served stale from disk cache).

## Verification

- `uv run ruff check packages/memtomem/src` — clean.
- `uv run pytest -m "not ollama" -k "namespace or sources or web"` — 986 passed.
- Playwright (against `mm web`):
  - localStorage forced to `user` vendor → click `claude:-Users-pdstudio-Work-agent-harness-memtomem` namespace's Sources button → vendor tab shifts to `claude`, exactly 1 dir group with 284 file items, chip shows the namespace.
  - Clear chip → 29 dir groups + 356 files reappear under Claude.
  - Click `default` namespace from `user` vendor → vendor stays `user`, 2 dirs + 22 files (auto-switch correctly inert when current vendor has matches).
  - `_navigateToSource(<claude path>)` from `user` vendor → vendor switches to `claude`, target source becomes `.source-item.active` and gets `scrollIntoView`.
  - `_navigateToSource(<user path>)` from `user` vendor → no vendor switch, target source becomes active (regression check).

## Out-of-scope sibling

`_searchByTag(tag)` (`app.js:3631-3640`) populates both `search-input` and `tag-filter` with the tag string — same axis-mismatch shape but `v0.1.0`-since behaviour, not a recent regression. Filed as a separate issue (#672) for an intent-vs-design-flaw decision; user-visible behaviour change so it needs its own CHANGELOG entry.

## Test plan

- [x] `ruff check` clean
- [x] Pytest namespace/sources/web suite green
- [x] Playwright: cross-vendor namespace card → auto-switch to matching vendor sub-tab
- [x] Playwright: chip-clear restores full vendor tree
- [x] Playwright: same-vendor namespace card preserves current vendor
- [x] Playwright: `_navigateToSource` cross-vendor + same-vendor regression
- [ ] Manual: refresh in real browser to confirm `?v=84` bust takes effect

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -1268,6 +1268,29 @@ function _renderStorageHealth(config, sources, embStatus) {
 }
 
 function _navigateToSource(path) {
+  // Resolve target vendor before switching so a Claude/OpenAI source
+  // clicked from Home recent or the command palette doesn't land on
+  // whichever vendor sub-tab the user last viewed — the
+  // ``.source-item[title=path]`` lookup below would silently miss
+  // because non-active vendors aren't rendered into the DOM. Same
+  // root cause as the navigateToSourcesByNs auto-switch in
+  // ``_renderMemorySourceTree``: a navigation entry-point sets one
+  // axis (path) without aligning the vendor sub-tab axis. Fall through
+  // when the source isn't in ``allSources`` yet (cold load) — current
+  // behaviour preserved, no regression.
+  const src = (STATE.allSources || []).find(s => s.path === path);
+  const status = src && src.memory_dir
+    ? (STATE.memoryStatusByPath || {})[src.memory_dir]
+    : null;
+  const provider = (status && typeof _SOURCES_VENDORS !== 'undefined'
+    && _SOURCES_VENDORS.includes(status.provider))
+    ? status.provider : null;
+  if (provider && STATE.sourcesActiveVendor !== provider) {
+    STATE.sourcesActiveVendor = provider;
+    if (typeof _syncSourcesVendorTabs === 'function') {
+      _syncSourcesVendorTabs(provider);
+    }
+  }
   activateTab('sources');
   setTimeout(() => {
     document.querySelectorAll('.source-item').forEach(el => {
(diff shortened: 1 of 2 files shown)


memtomem:
Review responses pushed in `9b4ae1d`:

**Concern 1 (stats badge one-tick stale) — confirmed visible, fixed.**
Reproduced with Playwright before the fix: clicking the `claude:…` namespace's Sources button while sitting on the User sub-tab landed `activeVendor=claude`, `fileItems=284` (Claude), but `statsText="22 files · 318 chunks"` — the User vendor totals sitting above a Claude tree. Until the next `renderSourceTree` call, the caption claimed a different vendor than the tree.

Fix: extracted the stats compute into `_renderSourcesStats(vendor)` and moved the call inside `_renderMemorySourceTree` right after the NS-filter auto-switch resolves `activeVendor`. `renderSourceTree` is now a thin delegator, so the caption can't race the sub-tab. Verified all three vendor branches:
- chip-clear → vendor=`claude`, stats=`357 files · 1,084 chunks` (matches tree)
- switch to User → vendor=`user`, stats=`22 files · 318 chunks`
- switch to OpenAI (empty) → stats `hidden=true`

**Concern 2 (tie-break sort stability rationale) — documented.**
Added a comment naming the policy: stable sort + `PROVIDER_ORDER = ['user', 'claude', 'openai']` means equal NS-match totals favour `user` because it's the conventional primary surface. Avoids the next reader wondering whether the asymmetry was intentional.

**Concern 3 (cold-load silent miss in `_navigateToSource`) — deferred per review verdict ("후속 개선 여지").**
Existing inline comment names the trade-off. If users hit it (unlikely — both call sites are post-dashboard) a follow-up can wire a `loadSources()` retry.

`v=84 → v=85` cache-bust, `ruff` clean, namespace/sources/web tests still 986 passed.

PR #674, opened 2026-05-01
fix(web): activate the clicked source on Home recent navigation

memtomem:
## Summary

Clicking a Home recent source switched to the Sources tab but the file the user actually clicked rarely became active — the user reported it as "그냥 소스탭으로만 이동해서 어색해." Three cold-load factors stacked:

1. **`loadDashboard` didn't mirror its fetches into `STATE`.** `/api/sources` was stored only in a local variable, so `STATE.allSources` and `STATE.memoryStatusByPath` stayed empty until the user opened Sources manually. `_navigateToSource`'s eager vendor resolve had no data to read.
2. **Stale `localStorage` vendor pointer.** Without an eager switch, Sources opened on whatever vendor sub-tab was last persisted (e.g. `openai` left over from a previous session). The clicked source's `.source-item` was on a different vendor and not in the DOM.
3. **`setTimeout(300)` race.** The fallback queried `.source-item[title=path]` 300 ms after `activateTab`, which raced `loadSources`'s fetch + render. On either miss the user saw "Sources tab opened but nothing happened."

This is the cold-load case PR #673's review flagged as "후속 개선 여지" (Concern #3). Filing as a separate fix per `feedback_one_change_per_pr.md`.

## Changes

`packages/memtomem/src/memtomem/web/static/app.js`:

- **`loadDashboard` mirrors `/api/sources` + `/api/memory-dirs/status` into `STATE`.** One extra `fail-soft` request added to the existing `Promise.all`. `loadSources` still owns these fields when the Sources tab opens — the dashboard's snapshot is just a fast path, never a source of truth.
- **`STATE.pendingActivatePath`** field added. `_navigateToSource` sets it and triggers `activateTab('sources')`. `setTimeout(300)` removed.
- **`_renderMemorySourceTree` consumes the pending path** at the end of every render:
  - If `STATE.sourcesNsFilter` is set, drop it first (path-based navigation is an explicit user-intent override; without this guard the NS auto-switch fights the pending vendor resolve in a re-render loop).
  - Re-resolve the vendor from fresh STATE. If the eager guess in `_navigateToSource` was wrong because data hadn't landed yet, switch sub-tab and re-render once.
  - Query the rendered list for `.source-item[title=path]`, add `.active`, `scrollIntoView`, fire `browseSource`. Clear `STATE.pendingActivatePath = ''` immediately so a follow-up filter/sort/reindex render doesn't re-scrollIntoView a source the user already navigated past.

`index.html`: `app.js?v=85 → v=86` cache-bust (per `feedback_static_asset_cache_bust.md`).

## Verification

End-to-end with Playwright in an **isolated worktree fixture** built via `scripts/setup-worktree.sh` + isolated `HOME` (not touching real `~/.memtomem/`). Six files seeded across all three vendors (user, claude, openai) so each sub-tab has matchable content.

Three sub-scenarios, each forcing a *stale* `localStorage` sub-tab pointing at a different vendor than the clicked source:

| stale `localStorage` | click target               | expected sub-tab | result |
| -------------------- | -------------------------- | ---------------- | ------ |
| `openai`             | `notes_alpha.md` (user)    | `user`           | ✅ pass |
| `user`               | `feedback_gamma.md` (claude) | `claude`         | ✅ pass |
| `claude`             | `codex_delta.md` (codex)   | `openai`         | ✅ pass |

In each case: vendor sub-tab switched, the clicked file became `.source-item.active`, chunks browser loaded the file content (`chunksLoaded: true`).

NS-override path verified separately on the working tree's live server: chip set via Settings → Namespaces → "Sources" button on a `claude:…` namespace; subsequent Home recent click on a different `claude:…` source dropped the chip and activated the target source.

`uv run ruff check packages/memtomem/src` clean. `uv run pytest -m "not ollama" -k "namespace or sources or web"` 986 passed.

## Out-of-scope follow-up surfaced during verification

`/api/sources` returns the source's `memory_dir` as a raw path (e.g. `/tmp/...`) while `/api/memory-dirs/status` returns the resolved path (`/private/tmp/...` on macOS). PR #668 normalized the latter; the former still needs the same treatment for client-side `STATE.memoryStatusByPath[s.memory_dir]` lookups to work when a memory_dir lives under a symlinked prefix. Doesn't affect the typical real-`HOME` deployment (`/Users/...` has no symlink resolution); only surfaces with `/tmp` (or other symlinked) memory_dir paths. Filing separately as a #668 follow-up — not blocking this fix.

## Test plan

- [x] `ruff check` clean
- [x] Pytest namespace/sources/web suite green
- [x] Playwright cold-load × 3 vendor scenarios in isolated fixture
- [x] NS-override path verified on live working-tree server
- [x] Verified the patch served at `?v=86` (cache bust)

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -994,9 +994,16 @@ async function loadDashboard() {
     // the target vendor sub-tab before activating the Sources tab —
     // otherwise ``_navigateToSource`` falls through to a stale
     // localStorage vendor on the cold-load path. ``loadSources`` will
-    // overwrite both fields with a fresh fetch when the Sources tab
-    // is actually opened, so the dashboard's snapshot is just a fast
-    // path, never a source of truth.
+    // overwrite both fields with a fresh fetch (``?limit=10000``) when
+    // the Sources tab is actually opened, so the dashboard's snapshot
+    // is just a fast path, never a source of truth — the
+    // ``/api/sources`` call here uses the server's default limit, so
+    // on instances with more sources than that default the mirror is
+    // a partial set. Eager resolve in ``_navigateToSource`` will miss
+    // the unseen tail and fall through to the cold-load branch in
+    // ``_renderMemorySourceTree``, which re-resolves vendor against
+    // the post-``loadSources`` STATE — so partiality stays a perf
+    // hint, not a correctness issue.
     STATE.allSources = allSources;
     const _memStatusByPath = {};
     for (const entry of (memDirsResp && memDirsResp.dirs) || []) {
@@ -2876,6 +2883,14 @@ function _renderMemorySourceTree(sources, list) {
       target.classList.add('active');
       target.scrollIntoView({ behavior: 'smooth', block: 'nearest' });
       if (typeof browseSource === 'function') browseSource(path);
+    } else if (STATE.uiMode === 'dev') {
+      // Path made it through the vendor + filter resolves but the
+      // ``.source-item`` isn't in the rendered list — most likely an
+      // edge case the cold-load fix didn't anticipate (orphan source,
+      // path normalisation drift like #675, etc.). Surface it in dev
+      // so the next bug report has a breadcrumb instead of "click did
+      // nothing"; prod stays silent because users can't act on it.
+      console.warn('[memtomem] pendingActivatePath found no .source-item:', path);
     }
     STATE.pendingActivatePath = '';
   }
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -25,8 +25,16 @@ const STATE = {
   detailViewSource: '',
   detailViewMode: 'view',
   allSources: [],
+  memoryStatusByPath: {},
   sourcesSortBy: 'name',
   sourcesNsFilter: '',
+  // Path the next ``_renderMemorySourceTree`` should focus + browse
+  // after rendering the tree. ``_navigateToSource`` sets this and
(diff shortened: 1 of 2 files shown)


memtomem:
Review responses pushed in `a9f30de`:

**Concern 1 (`/api/sources` limit asymmetry) — comment expanded.**
`loadDashboard`'s mirror block now spells out the contract: dashboard uses the server-default limit, `loadSources` uses `?limit=10000`, partiality stays a perf hint because the cold-load branch in `_renderMemorySourceTree` re-resolves against the post-`loadSources` STATE. No behaviour change — the asymmetry was already covered by the existing fall-through, just not documented as such.

**Concern 3 (silent `pendingActivatePath` drop) — dev `console.warn`.**
When the pending path makes it past the vendor + NS-filter guards but `querySelector('.source-item[title=…]')` returns null, dev sessions now get a `console.warn('[memtomem] pendingActivatePath found no .source-item:', path)` breadcrumb. Gated on `STATE.uiMode === 'dev'` so prod stays silent (users can't act on it). The `STATE.pendingActivatePath = ''` clear still happens — no leak.

**Concern 2 (automated regression coverage) — separate scope.**
Per the review verdict ("follow-up 으로 분리해도 무방"), keeping this PR focused. The repo doesn't have a Playwright e2e harness yet — adding one for `pendingActivatePath` cleared / vendor-switch-once invariants belongs in its own infrastructure PR rather than wedged onto this fix.

**Concern 4 (`/api/sources` `memory_dir` not resolved) — already filed as #675.**
Linked in the PR body.

`v=86 → v=87` cache-bust, `ruff` clean, namespace/sources/web tests still 986 passed.

PR #703, opened 2026-05-02
feat(web): surface fastembed model load state via banner (#696)

### LINKED ISSUE #696: First search is silent for 30s+ while embedding/reranker models download — surface progress
## What

The first search after `mm web` boot can take 30s to several minutes with no
UI feedback while fastembed downloads and loads the configured ONNX model
(and reranker, if enabled). The Search button shows the standard spinner;
the user can't tell whether the request is hung, the model is loading from
cache, or a multi-GB download is in progress.

The issue is most painful for users who selected the multilingual stack at
`mm init` time:

- `BAAI/bge-m3` — ~2.3 GB ONNX bundle.
- `jinaai/jina-reranker-v2-base-multilingual` — ~1.1 GB.

These are explicit choices in the `mm init` wizard
(`packages/memtomem/src/memtomem/cli/init_cmd.py:441,615`), but the
download then happens silently on the *first* search — minutes after the
choice was made — so users often think the UI is broken and reload or kill
the server.

## Where it bites

- `OnnxEmbedder._get_model()`
  (`packages/memtomem/src/memtomem/embedding/onnx.py:72`) lazily instantiates
  `fastembed.TextEmbedding(...)` on first use. fastembed's constructor is
  what triggers the HuggingFace snapshot download, and it does **not**
  expose progress events the caller can subscribe to — it just blocks until
  the files are on disk.
- A similar story holds for the cross-encoder reranker if enabled.
- The Web UI's search request goes through `pipeline.search`
  (`packages/memtomem/src/memtomem/search/pipeline.py:357`) →
  `embedder.embed_query` → `_get_model()` → fastembed download. From the
  UI's perspective it's a single long-running fetch with no intermediate
  signal.

`/api/embedding-status` exists already
(`packages/memtomem/src/memtomem/web/routes/system.py:728`) but is about
config-mismatch detection, not load state — it does not solve this.

## Design options (to discuss before PR)

Three deliberate cuts, smallest first:

### A. Generic "warming up" hint (zero-backend)

Detect when the first search exceeds, say, 3 seconds and replace the
spinner with a translated string like **"First search after launch may
take a moment — loading models…"**. Pure frontend; no server changes; no
download-vs-load distinction.

* Pros: ~1-day PR, no API surface, nothing to maintain.
* Cons: Still silent for the multi-minute download case. User does not
  learn how long to wait or what's happening.

### B. New readiness endpoint + banner (recommended starting point)

Add `GET /api/system/embedding-readiness` returning something like:

```json
{
  "state": "downloading" | "loading" | "ready" | "error",
  "model": "BAAI/bge-m3",
  "approx_size_mb": 2300,
  "cache_present": true,
  "reranker": { "state": "ready", "model": "..." }
}
```

`state` derivation:

- `cache_present` → check the resolved fastembed cache dir
  (`fastembed_cache.resolve_fastembed_cache_dir()`) for the model's
  snapshot directory.
- `state="downloading"` when the snapshot dir is missing and a download
  task is in flight.
- `state="loading"` when the snapshot dir is present but `_get_model()`
  hasn't returned yet.
- `state="ready"` after the first successful `_get_model()`.

Wire a small banner in the SPA header that polls this endpoint while
non-`ready` and shows e.g. **"Downloading bge-m3 (~2.3 GB). First search
will start when ready."**

* Pros: Concrete user signal; survives the multi-minute download case.
  No streaming or new dependencies.
* Cons: No actual progress percentage — fastembed doesn't expose it.
  States are coarse-grained.

### C. Streamed download progress (overkill for v0.1.x)

Wrap `huggingface_hub.snapshot_download` directly to capture per-file
progress and stream it via SSE. Bypasses fastembed's blocking constructor.

* Pros: Real percent-complete bar.
* Cons: Significant code surface — duplicates fastembed's resolution
  logic, has to track schema changes, and the value-add over (B) is
  marginal once the user knows roughly how long to wait.

## Open questions

- Should the banner also show **before the first search** (e.g. on
  `mm web` boot if the cache is empty), or only when blocking a request?
  Up-front warning is friendlier but adds a new always-present UI element.
- Does the reranker get its own banner or share one? They can download in
  parallel and finish at different times.
- Translations — banner copy needs `en` + `ko` and probably a model-name
  + size template. New `home.banner.embedding_*` keys.

## Severity

Medium UX. Not data loss; not a regression in the strict sense (the
download has always been silent). But it is the single most common "first
launch is broken" perception, and the multilingual stack — which is the
default Korean / multilingual choice in `mm init` — exposes it most.

## Suggested next step

Pick option (B), draft a small RFC or PR for the endpoint + banner with
explicit answers to the open questions above.

memtomem:
Closes #696 (option B).

## Summary

- Adds `GET /api/system/model-readiness` reporting per-component (embedder + reranker) load state derived from `_loading` / `_load_error` flags newly attached to `OnnxEmbedder` and `FastEmbedReranker`, plus a filesystem probe of the fastembed cache.
- Wires a header banner that polls the endpoint and renders `Downloading bge-m3 (~2300 MB)…` / `Loading model…` / `Model failed to load — check Settings` so the first search after a cold-cache boot no longer feels like a hung UI.
- Boot hydrate, `visibilitychange` re-hydrate, and a `doSearch()` pre-flight cover the three entry points.

## Why option B

`fastembed` does not expose snapshot-download progress events, so a real percent bar (option C) would mean wrapping `huggingface_hub.snapshot_download` ourselves and replicating fastembed's resolution logic. Option B is the smallest change that converts "frozen Search button" into "I see what's happening" without that complexity. See the issue body for the option A/B/C tradeoffs.

## Commits

Split for review; one PR for atomic ship:

1. `feat(embedding): track loading state on lazy fastembed loaders`
   – `_loading` / `_load_error` flags + `embedding/readiness.py` cache probe + size map.
2. `feat(web): add /api/system/model-readiness endpoint`
   – Schema, route, endpoint tests. No UI.
3. `feat(web): show model-loading banner with polling`
   – Banner DOM + CSS, polling logic, i18n keys, cache busters.

## State machine

| condition | state |
| --- | --- |
| provider not in `{onnx, fastembed}` OR component disabled | `skipped` |
| `_load_error` set | `error` |
| `_model is not None` | `ready` |
| `_loading=True` and cache absent | `downloading` |
| `_loading=True` and cache present | `loading` |
| otherwise | `cold` |

`cold` is intentionally non-terminal for `doSearch()`-initiated polls — the first tick can race the request and observe `cold` before the backend's lazy loader flips `_loading=True`. The boot hydrate does treat `cold` as terminal so a fully-warm install doesn't run a needless 4 s/tick background loop.

## Out of scope (deliberate)

- Real per-file download progress (option C). Documented in #696.
- Pre-warming models at server boot. Adds startup latency; revisit if banner alone proves insufficient.
- Provider-specific readiness for Ollama / Cohere. Their connection-based readiness is a separate concern; both are reported as `state="skipped"` here.

## Cache-buster note

`app.js?v=94→96` and `style.css?v=76→77`. The `v=96` jump leapfrogs an in-flight `v=95` from sibling PR #694 (namespace-tooltip). Whichever lands first, the second will need a rebase + bump.

## Test plan

Backend:

- [x] `uv run pytest packages/memtomem/tests/test_embedding_readiness.py packages/memtomem/tests/test_web_model_readiness.py` — 20 tests cover the cache probe + state machine.
- [x] `uv run pytest -m "not ollama"` — 3908 passed locally.
- [x] `uv run ruff check packages/memtomem/src && uv run ruff format --check packages/memtomem/src packages/memtomem/tests` clean.

Manual UI (cold-cache reproduction):

```bash
mv ~/.memtomem/cache/fastembed ~/.memtomem/cache/fastembed.bak
uv run mm web --host 127.0.0.1 --port 8765 &
# Open the URL — banner is hidden because no search is in flight.
# Trigger a search — banner appears as "Downloading bge-m3 (~2300 MB)…"
# (assuming the user's config selects bge-m3). Banner persists through
# the load and disappears on completion. Search results render.
kill %1
mv ~/.memtomem/cache/fastembed.bak ~/.memtomem/cache/fastembed
```

Korean toggle: switch language and re-run; banner copy must come from `ko.json` strings.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -435,10 +435,21 @@ def _step_embedding(state: dict) -> None:
             click.echo()
             state.setdefault("_extras_warned_inline", set()).add("onnx")
 
+        # Sizes are pulled from ``embedding/aliases.py`` so this wizard
+        # text and the runtime "Downloading {model} (~{size} MB)…" banner
+        # surfaced by ``GET /api/system/model-readiness`` agree exactly.
+        from memtomem.embedding.aliases import ONNX_EMBEDDER_MODELS, format_size
+
+        _all_minilm_size = format_size(ONNX_EMBEDDER_MODELS["all-MiniLM-L6-v2"][2])
+        _bge_small_size = format_size(ONNX_EMBEDDER_MODELS["bge-small-en-v1.5"][2])
+        _bge_m3_size = format_size(ONNX_EMBEDDER_MODELS["bge-m3"][2])
+
         click.echo("  Available models:")
-        click.echo("    [1] all-MiniLM-L6-v2 — English, fast, tiny (~22 MB, 384d)")
-        click.echo("    [2] bge-small-en-v1.5 — English, better accuracy (~33 MB, 384d)")
-        click.echo("    [3] bge-m3 — multilingual KR/EN/JP/CN (~1.2 GB, 1024d)")
+        click.echo(f"    [1] all-MiniLM-L6-v2 — English, fast, tiny (~{_all_minilm_size}, 384d)")
+        click.echo(
+            f"    [2] bge-small-en-v1.5 — English, better accuracy (~{_bge_small_size}, 384d)"
+        )
+        click.echo(f"    [3] bge-m3 — multilingual KR/EN/JP/CN (~{_bge_m3_size}, 1024d)")
         model_choice = nav_prompt("  Select", type=click.IntRange(1, 3), default=1)
 
         models = {
@@ -450,7 +461,7 @@ def _step_embedding(state: dict) -> None:
 
         if model_choice == 3:
             click.secho(
-                "  Note: bge-m3 is ~1.2 GB (similar to Ollama models). "
+                f"  Note: bge-m3 is ~{_bge_m3_size} (substantial download). "
                 "For lightweight EN-only search, choose option 1 or 2.",
                 fg="yellow",
             )
@@ -609,10 +620,17 @@ def _step_reranker(state: dict) -> None:
         click.echo()
         return
 
+    # Sizes pulled from ``embedding/aliases.py:FASTEMBED_RERANKER_SIZES``
+    # so the wizard and the readiness banner agree.
+    from memtomem.embedding.aliases import FASTEMBED_RERANKER_SIZES, format_size
+
+    _en_size = format_size(FASTEMBED_RERANKER_SIZES["Xenova/ms-marco-MiniLM-L-6-v2"])
+    _multi_size = format_size(FASTEMBED_RERANKER_SIZES["jinaai/jina-reranker-v2-base-multilingual"])
+
     click.echo()
     click.echo("  Available models:")
-    click.echo("    [1] English (Xenova/ms-marco-MiniLM-L-6-v2) — 80 MB")
-    click.echo("    [2] Multilingual (jinaai/jina-reranker-v2-base-multilingual) — 1.1 GB")
+    click.echo(f"    [1] English (Xenova/ms-marco-MiniLM-L-6-v2) — {_en_size}")
+    click.echo(f"    [2] Multilingual (jinaai/jina-reranker-v2-base-multilingual) — {_multi_size}")
     click.echo("        Recommended for Korean/Chinese/Japanese/mixed content.")
     choice = nav_prompt("  Select", type=click.IntRange(1, 2), default=1)
 
--- a/packages/memtomem/src/memtomem/embedding/aliases.py
+++ b/packages/memtomem/src/memtomem/embedding/aliases.py
@@ -1,0 +1,95 @@
+"""Single source of truth for fastembed model aliases and approximate sizes.
+
+Used by every surface that references a model name or has to tell the
+user how big the download is:
+
+* ``OnnxEmbedder._get_model`` — short-name → fastembed-id resolution.
+* ``GET /api/system/model-readiness`` — banner annotation
+  ("Downloading bge-m3 (~2300 MB)…").
+* ``cli init`` wizard — size text shown to the user.
+
+Sizes come from fastembed's own ``list_supported_models()`` /
+``size_in_GB`` field (rounded to MB). Custom-registered models — only
+``BAAI/bge-m3`` today, see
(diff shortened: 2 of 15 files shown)


memtomem:
Addressed both blocking review issues:

**#1 Size discrepancy** — Used fastembed's `list_supported_models()` `size_in_GB` (and `add_custom_model(size_in_gb=2.3)` for bge-m3) as the source of truth. Corrected several wrong numbers in both directions:

| model | before | after |
| --- | --- | --- |
| `BAAI/bge-m3` (init wizard) | 1.2 GB | **2.3 GB** |
| `BAAI/bge-small-en-v1.5` (init wizard) | 33 MB | **67 MB** |
| `BAAI/bge-small-en-v1.5` (readiness map) | 130 MB | **67 MB** |
| `sentence-transformers/all-MiniLM-L6-v2` (init wizard) | 22 MB | **90 MB** |
| `nomic-ai/nomic-embed-text-v1.5` (readiness map) | 280 MB | **520 MB** |
| `jinaai/jina-reranker-v2-base-multilingual` (readiness map) | 1100 MB | **1110 MB** |

The wizard's "bge-m3 is ~1.2 GB (similar to Ollama models)" copy was also adjusted to "~2.3 GB (substantial download)" — calling 2.3 GB "similar to Ollama" was misleading.

**#2 Alias map dedup** — Added `embedding/aliases.py` as the single source of truth. `OnnxEmbedder._resolve_model`, `_resolve_fastembed_model_id` in routes, and the `_APPROX_SIZE_MB` map in readiness.py all gone. Init wizard reads sizes via `aliases.format_size` so wizard text and banner agree by construction.

Drift guard: `tests/test_embedding_aliases.py::test_resolve_embedder_id_matches_legacy_resolver` snapshots the short-name contract — a future refactor that reintroduces a private duplicate fails the test instead of silently drifting.

**Remaining minor notes** are deliberately out of scope here:
- Raw `_load_error` in the API response — local-only tool, low risk; would muddy the schema if I started truncating.
- Polling cap silent stop at 13 min — the cap is already a worst-case safeguard (visibilitychange re-arms); bumping it without changing behavior felt like noise.
- Cache-presence probe doesn't verify blob completeness — the brief `loading`-instead-of-`downloading` UX wobble during the actual download isn't worth complicating the probe.

Tests: 3923 passed (added 15 alias/dedup tests on top of the 3908 baseline). `ruff check + format --check` clean.

PR #715, opened 2026-05-02
fix: extend FTS5 special-char regex to cover dot, slash, backslash, angle brackets, tilde

### LINKED ISSUE #697: BM25 FTS5 query sanitization misses `.` `/` `\` `<` `>` (silent degradation)
## Summary

User search queries containing punctuation common in markdown / YAML / URLs / filesystem paths can crash BM25 with `fts5: syntax error near "."`. The error is caught and logged but the HTTP response stays `200`, so users see degraded (dense-only) results without any UI signal.

Found by Playwright UX review of v0.1.34 prod (2026-05-02). See `docs/reports/mm-web-prod-v0.1.34-playwright-review.md` (P1 — BM25 search can fail on raw markdown/YAML-like queries).

## Evidence

**Sanitization gap** — `packages/memtomem/src/memtomem/storage/fts_tokenizer.py:18`

```python
_FTS5_SPECIAL_RE = re.compile(r'[*"()\-+^:]')
```

Missing characters that FTS5 treats as syntactically meaningful: `.` `/` `\` `<` `>`. URLs (`https://example.com`), filesystem paths (`a/b/c`), dotted filenames (`file.name.ext`), and YAML/frontmatter (`key: value`, `---`) all flow through unquoted and trip the parser.

**Silent degradation** — `packages/memtomem/src/memtomem/search/pipeline.py:468-474`

```python
if use_bm25:
    try:
        bm25_results = await bm25_task
    except Exception as exc:
        logger.warning("BM25 search failed: %s", exc)
        bm25_results = []
        bm25_error = str(exc)
```

`bm25_error` lands in `RetrievalStats` but the web UI does not surface it. Users get dense-only results and have no signal that keyword search broke.

**Tests** — no FTS5-syntax-in-query regression cases in `packages/memtomem/tests/`. Queries with frontmatter, code spans, URLs, paths, punctuation are uncovered.

## Suggested fix

1. Extend `_FTS5_SPECIAL_RE` to include `.` `/` `\` `<` `>` (and re-audit the rest of the FTS5 special set against the sqlite docs).
2. Add regression cases to `tests/` that pass each of: frontmatter (`---\nkey: value`), URL (`https://example.com/path`), dotted filename (`file.name.ext`), unix path (`a/b/c`), and a code-span fragment.
3. Surface `bm25_error` in the web UI as a non-blocking warning (e.g., "Keyword search degraded — using vector results only") rather than swallowing it. This matches the repo's loud-vs-silent invariant.

(1)+(2) is a self-contained `good first issue`-sized change. (3) touches the web layer and is a separate follow-up — keep it out of this PR unless requested.

## References

- Tracking umbrella: TBD (linked once opened)

Giggitycountless:
## Summary

`_FTS5_SPECIAL_RE` was missing 6 FTS5 special characters: `.`, `/`, `\`, `<`, `>`, `~`. Tokens containing these characters — URLs (`https://example.com`), filesystem paths (`a/b/c`), dotted filenames (`file.name.ext`), YAML frontmatter (`key: value`), and proximity queries (`word~n`) — leaked into the FTS5 query parser unquoted and triggered `fts5: syntax error`.

The error was caught silently (logged at WARNING, HTTP 200, dense-only fallback), so users had no indication that keyword search had failed.

## Changes

**`fts_tokenizer.py`**: Extend `_FTS5_SPECIAL_RE` to the full unicode61 token-class special set. Words containing any special character are now wrapped in double quotes (literal phrase match) instead of receiving a `*` prefix wildcard.

**`test_fts_tokenizer.py`** (new): 45 regression tests covering:
- Regex matches every FTS5 special character
- `_apply_prefix_wildcard` quotes URLs, paths, dotted names, code spans, tilde
- `tokenize_for_fts` end-to-end query path for special-character tokens
- Plain words still get wildcarded, empty input handled

## How to test

```
uv run pytest packages/memtomem/tests/test_fts_tokenizer.py -v
```

Fixes #697

### DIFF
--- a/packages/memtomem/src/memtomem/storage/fts_tokenizer.py
+++ b/packages/memtomem/src/memtomem/storage/fts_tokenizer.py
@@ -14,8 +14,11 @@ import re
 
 _log = logging.getLogger(__name__)
 
-# FTS5 special characters that must not appear in prefix-match tokens
-_FTS5_SPECIAL_RE = re.compile(r'[*"()\-+^:]')
+# FTS5 special characters that must not appear in prefix-match tokens.
+# Includes all unicode61 token-class characters that FTS5 treats as
+# operators or token separators.  Missing any of these causes
+# "fts5: syntax error" when the token reaches the query parser unquoted.
+_FTS5_SPECIAL_RE = re.compile(r'[-*\"()+^:\\./<>~`!@#%&=\'{}\[\];?]')
 
 # Active tokenizer backend: "unicode61" or "kiwipiepy"
 _active_tokenizer: str = "unicode61"
--- a/packages/memtomem/tests/test_fts_tokenizer.py
+++ b/packages/memtomem/tests/test_fts_tokenizer.py
@@ -1,0 +1,119 @@
+"""Regression tests for FTS5 query sanitization.
+
+Covers tokens containing characters that FTS5 treats as operators or
+token separators (`.`, `/`, `\\`, `<`, `>`, `~`, etc.).  Without the
+full ``_FTS5_SPECIAL_RE`` character set these tokens leak into the
+query parser unquoted and trigger ``fts5: syntax error``.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from memtomem.storage.fts_tokenizer import (
+    _FTS5_SPECIAL_RE,
+    _apply_prefix_wildcard,
+    tokenize_for_fts,
+)
+
+
+# ---------------------------------------------------------------------------
+# _FTS5_SPECIAL_RE: character-class coverage
+# ---------------------------------------------------------------------------
+
+class TestFTS5SpecialRegex:
+    """Every FTS5 special character must be caught by the regex."""
+
+    @pytest.mark.parametrize(
+        "char",
+        list('-*"()+^:\\./<>~`!@#%&=\'{}[];?'),
+        ids=lambda c: f"char_{repr(c)}",
+    )
+    def test_special_char_detected(self, char: str) -> None:
+        assert _FTS5_SPECIAL_RE.search(char), f"Character {char!r} not matched"
+
+    def test_plain_word_not_matched(self) -> None:
+        assert _FTS5_SPECIAL_RE.search("normal_word") is None
+
+    def test_alphanumeric_not_matched(self) -> None:
+        assert _FTS5_SPECIAL_RE.search("hello123") is None
+
+
+# ---------------------------------------------------------------------------
+# _apply_prefix_wildcard: quoting vs prefix-wildcard
+# ---------------------------------------------------------------------------
+
+class TestApplyPrefixWildcard:
+    """Words with special chars must be quoted, not wildcarded."""
+
+    def test_url_is_quoted(self) -> None:
+        result = _apply_prefix_wildcard("https://example.com")
+        assert result == '"https://example.com"', f"URL not quoted: {result!r}"
+
+    def test_filesystem_path_is_quoted(self) -> None:
+        result = _apply_prefix_wildcard("a/b/c")
+        assert result == '"a/b/c"', f"Path not quoted: {result!r}"
+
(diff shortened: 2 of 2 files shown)


memtomem:
Thanks for the careful fix and the thorough test file — really appreciate the parametrized regex coverage plus the end-to-end cases.

I ran the tests locally on the PR branch:

- `pytest test_fts_tokenizer.py` → **45/45 pass**
- pre-existing `test_usability_fixes.py::TestFTS5HyphenQuoting` → **4/4 still pass**
- broader `pytest -k "fts or tokeniz or bm25 or search"` → **218/218 pass**, no adjacent regressions
- mutation check (revert the regex to old, re-run new tests) → **25 of 45 fail** — confirms the tests genuinely pin the new behavior

### Required before merge

- [ ] **Sign the CLA** — the bot comment up-thread has the exact text to copy.
- [ ] **`uv run ruff format`** — `ruff format --check` currently fails on three spots, all auto-fixable in one command:
  1. `fts_tokenizer.py:18` — single-quoted regex string → ruff prefers double-quoted (`r"..."`).
  2. `test_fts_tokenizer.py` — class definitions need two blank lines before them (PEP 8) at lines 24, 46, 95.
  3. `test_fts_tokenizer.py:90` — trailing whitespace on the `'say* """hello"""'` assertion line.

  Note: `ruff check` already passes — it's only `ruff format --check` that flags these. Both are gating in CI.

### Suggestion (non-blocking)

Could you add a `### Fixed` line under `[Unreleased]` in the top-level `CHANGELOG.md`? Something like:

> - `mem_search` BM25 leg now handles queries containing `.` `/` `\` `<` `>` `~` (and the rest of the unicode61 separator set) — previously silently fell back to dense-only with a WARNING log. (#715, fixes #697)

### Out of scope (just noting for future, not for this PR)

- The kiwipiepy backend at `fts_tokenizer.py:101` *drops* tokens matching the regex rather than quoting them. That pre-dates this PR, but the silent-drop surface widens as the regex grows — worth a follow-up.
- The `if safe:` branch at `fts_tokenizer.py:123` is dead (`text.split()` never yields empty, `.replace` only grows). Pre-existing.

### Approve

Once the format pass + CLA land, happy to approve and merge. Thanks again for picking this up!

memtomem:
Hey @Giggitycountless — friendly check-in on this one. The fix and the test file still look great on my end; the only things keeping it from merging are on the contributor side:

1. **CLA signature** — copy the exact line from the bot comment above (`I have read the CLA Document and I hereby sign the CLA`) into a new comment on this PR. One time per GitHub account.
2. **`uv run ruff format`** — the three auto-fixable spots from my earlier comment (regex string quoting, blank-lines-before-class, trailing whitespace). One command will handle all of them.

No rush — totally fine if you've gotten busy and want to come back to this later. If you'd rather hand it off, just leave a comment and I can take it from here without losing your authorship credit. Otherwise I'll keep this open and revisit in a week or two.

Thanks again for the careful work on the regex coverage and the mutation-pinned tests.

tsdata:
Thanks for this, @Giggitycountless — and for the thorough 45-case regression suite. 🙏

This PR has been open for a while, and in the meantime the underlying issue #697 was resolved by a parallel change. #963 (`400274d`, merged 2026-05-12) took a slightly different approach: instead of enumerating the FTS5 special characters, it broadened `_FTS5_SPECIAL_RE` to `re.compile(r"[^\w]", re.UNICODE)` — i.e. *any* non-word character is treated as special.

That already covers every character this PR adds (`.` `/` `\` `<` `>` `~`, plus the rest of your set) and more, so the change here is now redundant against `main`. #963 also added live `:memory:` MATCH coverage and a "Keyword degraded" UI signal for the silent-fallback problem you called out.

Given the elapsed time and that the issue is already fixed on `main`, I'm closing this as superseded — not a reflection on the patch, which correctly diagnosed the bug. Really appreciate the contribution, and hope to see you on another issue.

Issue #643, opened 2026-05-01
Windows: replace POSIX path-separator assertions in tests (`/` vs `\\`)


memtomem:
## Background

Tests assert against path strings using POSIX \`/\` separators, breaking on Windows where \`str(Path(...))\` produces \`\\\` separators.

Surfaces in PR #638's first Windows matrix run.

## Symptoms

\`\`\`
FAILED test_wiki_cmd_override.py::test_wiki_skill_override_invokes_editor
  assert False
  + where False = '...\\\\skills\\\\hello\\\\overrides\\\\claude.md'.endswith('/skills/hello/overrides/claude.md')

FAILED test_wiki_cmd_override.py::test_wiki_skill_override_stdout_contract
  assert 'Seeded skills/hello/overrides/codex.md' in 'Seeded skills\\\\hello\\\\overrides\\\\codex.md\\n...'
\`\`\`

## Affected files (8+ tests known)

- \`packages/memtomem/tests/test_wiki_cmd_override.py\` — 8 failures (all endswith / substring of POSIX paths)
- Likely scattered in: \`test_uninstall_cmd.py\`, \`test_web_routes.py\`, \`test_init_cmd.py\` (subset of their failures)

## Fix pattern

Three options per call site:
- \`endswith(os.path.join(\"skills\", ...))\` — uses native sep
- \`Path(actual).as_posix().endswith(\"/...\")\` — normalize actual to POSIX
- \`actual.replace(os.sep, \"/\")\` then compare — cheapest at scale

\`as_posix()\` is usually cleanest for assertion targets that already exist as Paths. Production code (CLI stdout) should be audited too — if \`mm wiki skill override\` prints \`Seeded skills\\hello\\overrides\\codex.md\` on Windows, that's arguably a UX fix rather than a test-only fix.

## Refs

- #634 — Windows + macOS CI matrix expansion (umbrella)
- PR #638 — matrix infrastructure (Windows currently informational)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

memtomem:
## Windows compat sweep — complete

`main` (d5b1fe3) Windows CI is now fully green. The sweep that ran out of this issue lands as follows:

| Cluster | Symptom | PR | Issue |
|---|---|---|---|
[...]

PR #726, opened 2026-05-02
fix(config): use as_posix() for tilde-expanded path_glob (#643 cluster E)

memtomem:
## Summary

**Real production bug**, surfaced by Cluster E of the #643 Windows sweep.

`NamespacePolicyRule._expand_and_validate_glob` (`config.py:341-347`) stored expanded path_globs via `str(Path(v).expanduser())`. On Windows that yields backslashes (`C:\Users\runneradmin\some\path\**`). The downstream consumer is `pathspec.GitIgnoreSpec` (`engine.py:_build_exclude_spec`) which interprets gitignore patterns — gitignore syntax is POSIX-only and treats `\` as an escape character. A backslash-bearing pattern **silently never matches any file** on Windows.

### Pre-fix Windows behavior

Any namespace rule with a leading `~/` in its glob — a common config shape (`~/.claude/projects/**`, `~/Documents/notes/**`, etc.) — silently failed to match. Files that should have landed in a project-specific namespace fell through to `default_namespace`, with no warning.

### Fix

`Path(v).expanduser().as_posix()` instead of `str(Path(v).expanduser())`:

- Keeps the tilde expansion (still absolute, still platform-correct via `Path.home()`).
- Forces `/` separators in the stored pattern.
- Matches gitignore semantics on every platform.
- POSIX no-op for absolute paths.

## Test plan

- [x] macOS: `uv run pytest packages/memtomem/tests/test_indexing_engine.py packages/memtomem/tests/test_config_overrides.py packages/memtomem/tests/test_validate_namespace.py` — **339/339 pass**.
- [x] Lint: `ruff check` + `ruff format --check` clean.
- [ ] Windows CI: `test_home_tilde_in_path_glob_expanded` should pass; tilde-prefixed namespace rules should now actually match.

## Why production fix (not test fix)

The failing test (`test_home_tilde_in_path_glob_expanded`) asserted `rule.path_glob.endswith("/some/path/**")` — this **already encoded the POSIX-glob invariant** the rule downstream depends on. The test was right; production was lying about the invariant on Windows.

## Out of scope

Remaining: A (9 — dirty state, production change), H (2 — misc) = **11 fails** after this lands. Cluster B already shipping via parallel work (`fix/720-source-filter-windows` merged as `88c3e81`).

### DIFF
--- a/packages/memtomem/src/memtomem/config.py
+++ b/packages/memtomem/src/memtomem/config.py
@@ -343,7 +343,12 @@ class NamespacePolicyRule(BaseSettings):
         if not v:
             raise ValueError("path_glob must be non-empty")
         if v == "~" or v.startswith("~/"):
-            v = str(Path(v).expanduser())
+            # ``as_posix()`` not ``str()``: ``pathspec.GitIgnoreSpec`` is the
+            # downstream consumer (engine.py:_build_exclude_spec) and gitignore
+            # patterns require ``/`` separators. On Windows ``str(Path(...))``
+            # would emit backslashes, which pathspec treats as escape characters
+            # and silently mismatches every file.
+            v = Path(v).expanduser().as_posix()
         return v
 
     @field_validator("namespace")
--- a/packages/memtomem/src/memtomem/config.py
+++ b/packages/memtomem/src/memtomem/config.py
@@ -349,6 +349,14 @@ class NamespacePolicyRule(BaseSettings):
             # would emit backslashes, which pathspec treats as escape characters
             # and silently mismatches every file.
             v = Path(v).expanduser().as_posix()
+        elif Path(v).is_absolute():
+            # Same normalization for already-absolute paths supplied directly
+            # (e.g. a config fragment that pre-expanded ``~/`` to ``C:\...``).
+            # Without this, two fragments declaring the equivalent rule via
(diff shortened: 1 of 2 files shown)


memtomem:
Heads-up: pushed a follow-up commit `b815684` to address a Windows CI regression surfaced by the previous commit `7974a26`.

`test_config_d_namespace_rules_dedup_after_home_expansion` failed on Windows because the validator normalized only the `~/` form — a fragment that supplied an already-expanded absolute path (`C:\Users\foo\**`) went through unchanged with backslashes, defeating dedup against the normalized `~/` form (`C:/Users/foo/**`).

The follow-up symmetrizes the normalization:

- `~/...` → `Path(v).expanduser().as_posix()` (existing)
- `is_absolute()` → `Path(v).as_posix()` (new)
- relative patterns → unchanged (gitignore escapes can legitimately use `\` in non-path contexts)

Pre-#726 the test passed on Windows by accident — both paths leaked backslashes and matched textually, while neither would have actually matched any file due to the gitignore-escape bug `7974a26` documents. Dedup of two non-matching rules was moot.

Test updated: `abs_form` now built via `as_posix()` to assert against the new (correct) stored form. 339/339 macOS pass.

PR #735, opened 2026-05-02
fix: return real source total in /api/chunks so Load All button renders

### LINKED ISSUE #731: [bug]: /api/chunks total equals limit, hiding "Load All" button
## Summary
The "Load All" button in Sources never appears even when a source file has >100 chunks.

## Root cause
`/api/chunks` returns `total = len(out)` (capped to the limit) instead of the real total for that source. The UI checks `data.chunks.length < data.total`, so the button never renders.

### Surface

Web UI (`mm web`)

## Steps to reproduce
1. Create a large file:
   ```bash
   python3 - <<'PY'
   from pathlib import Path
   p = Path.home() / "memories/load_all_test.md"
   p.parent.mkdir(parents=True, exist_ok=True)
   parts = []
   for i in range(1, 140):
       parts.append(f"# Section {i}\n")
       parts.append(("Lorem ipsum " * 120).strip() + "\n\n")
   p.write_text("".join(parts))
   print(p)
   PY

2. Run `mm web`.
3. Index `~/memories` from the Index tab.
4. Go to Sources and open `load_all_test.md`




### Expected behavior

Header shows `100 of N shown` (real total), and the `Load All` button appears.

### Actual behavior

Header shows `100 of 100 shown`, and the `Load All` button is missing.

### memtomem version

0.1.34

### Install method

uv tool install

### Python version

3.12.13

### Operating system

macOS

### MCP client (if applicable)

None

### Additional context


## Suggested fix
In `packages/memtomem/src/memtomem/web/routes/chunks.py`, return the real total count (e.g., COUNT(*)) while still limiting the returned list.

Discovered while working on #727. Happy to take this after that PR lands.

mahendra-shah:
Closes #731

## Problem
`/api/chunks` returned `total = len(out)` (capped by the limit), so
`data.chunks.length < data.total` was always false and the “Load All”
button never appeared.

## Fix
Add `count_chunks_by_source` to storage, implement it in sqlite, and
use it to return the real total in `/api/chunks` while keeping the list
limited.

## Testing
- ./.venv/bin/python -m pytest packages/memtomem/tests/test_web_routes.py -k chunks
- Manual: header shows “100 of 139 shown”, and “Load All” appears/disappears after loading


memtomem:
LGTM 👍

- CI all green (lint/typecheck/tests on 3 OS, golden-path, notebooks)
- Minimal, focused change with regression test (`test_list_chunks_total_reflects_source_count`)
- `norm_path` applied for path-normalization consistency with other mutators
- Confirmed `SqliteBackend` is the only `StorageBackend` implementation (factory hardcodes it), so adding the abstract method to the Protocol won't break anything

Merging.

memtomem (approved):
LGTM — see comment above.

PR #736, opened 2026-05-03
fix(web): hide Settings Runtime group header in prod (closes #701)

### LINKED ISSUE #701: Settings Runtime group header visible in prod despite all dev-tier children (empty expansion)
## Summary

In More / Settings on the prod surface, the **Runtime** group header expands to nothing: all four child entries are marked `data-ui-tier="dev"` and hidden in prod, but the group header itself has no tier marker so it stays clickable. Users see the arrow flip from collapsed to expanded with no children, and the content panel keeps showing whatever was previously selected.

Found by Playwright UX review of v0.1.34 prod (2026-05-02). See `docs/reports/mm-web-prod-v0.1.34-playwright-review.md` (P2 — Settings navigation expands sections without visible child content).

## Evidence

`packages/memtomem/src/memtomem/web/static/index.html`:

```html
<button class="settings-nav-group" type="button" data-group="runtime" aria-expanded="false">  <!-- 645 -->
  <span data-i18n="settings.nav.runtime">Runtime</span>
</button>
<!-- All 4 children are data-ui-tier="dev": -->
<button class="settings-nav-btn" data-ui-tier="dev" data-section="harness-sessions" ...>  <!-- 649 -->
<button class="settings-nav-btn" data-ui-tier="dev" data-section="harness-scratch" ...>   <!-- 656 -->
<button class="settings-nav-btn" data-ui-tier="dev" data-section="harness-procedures" ...> <!-- 663 -->
<button class="settings-nav-btn" data-ui-tier="dev" data-section="harness-health" ...>    <!-- 670 -->
```

The dev-tier hide pass (`app.js:160`) only walks `[data-ui-tier="dev"]`, so the group header itself never gets hidden.

## Suggested fix (two valid approaches)

1. **Mark the header as dev-tier too** — add `data-ui-tier="dev"` to the `settings-nav-group` button at line 645. Smallest diff. Acceptable if Runtime is *always* dev-only.

2. **Compute group visibility from children** — in `app.js:160` (or wherever the dev-tier pass runs), for each `.settings-nav-group`, hide the header when every child `.settings-nav-btn` in the same `data-group` is dev-tier and STATE.uiMode is prod. More robust if other groups acquire dev-only tabs later.

Either fix should also include an empty-state path: if a future group ends up with zero visible children at runtime (e.g., feature-flagged), render a "no items" message instead of leaving expansion silent.

## References

- Tracking umbrella: TBD

memtomem:
## Summary

- The **Runtime** group header in Settings nav had no `data-ui-tier` marker, so `_applyUiModeFilter` (`app.js:170`) walked past it. In prod the header stayed clickable and expanded to **nothing** — caret flipped, no children appeared, content panel kept showing the previously-selected section.
- Mark the group header `data-ui-tier="dev"` so it shares the same hide pass as its four already-dev-tier children (`harness-sessions`, `harness-scratch`, `harness-procedures`, `harness-health`).
- Picked option 1 (smallest diff) from the issue. Option 2 (compute visibility from children) would be more robust if a future runtime child becomes prod-tier, but all four are dev-only by design today, so option 1 is the right scope.

Closes #701.

## Test plan

- [ ] In prod mode (default), open Settings — confirm Runtime group is no longer visible in the nav column.
- [ ] In dev mode (`?ui_mode=dev` or env toggle), confirm Runtime group still expands and shows all four children (Sessions / Working Memory / Procedures / Health Report).
- [ ] Toggle dev → prod → dev round-trip without page reload via the dev-mode banner; confirm Runtime header visibility flips with it (covered by `_applyUiModeFilter` re-run).

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/index.html
+++ b/packages/memtomem/src/memtomem/web/static/index.html
@@ -647,7 +647,7 @@
           </span>
         </button>
 
-        <button class="settings-nav-group" type="button" data-group="runtime" aria-expanded="false">
+        <button class="settings-nav-group" type="button" data-ui-tier="dev" data-group="runtime" aria-expanded="false">
           <span class="nav-group-caret" aria-hidden="true">▸</span>
           <span data-i18n="settings.nav.runtime">Runtime</span>
         </button>


pandas-studio (approved):
Hide mechanism is doubly safe — `_applyUiModeFilter` (app.js:170) sets both `hidden=true` and inline `display:none`, plus the global `[hidden] { display: none !important }` rule at style.css:6 backstops it. Click-handler / collapse-state passes that still iterate the now-hidden runtime group in prod are harmless (wasted work, not a bug).

Trade-off vs. option 2 (compute group visibility from children) correctly justified — all four runtime children are dev-only by design today, so the smaller diff is the right scope. If a future contributor adds a prod-tier child to runtime, breakage is obvious in dev preview.

Approving.

PR #737, opened 2026-05-03
fix(web): re-apply hidden=true on tab deactivation in activateTab() (closes #699)

### LINKED ISSUE #699: Tab `hidden` attribute not re-applied on deactivation in `activateTab()`
## Summary

`activateTab()` at `packages/memtomem/src/memtomem/web/static/app.js:680-694` removes the `.active` class from previously-active panels but never re-applies `hidden = true`. After the first tab activation any panel that has been visited stays `hidden = false` for the rest of the session, relying solely on CSS `display: none` (from the absent `.active` class) to hide content.

Found by Playwright UX review of v0.1.34 prod (2026-05-02). See `docs/reports/mm-web-prod-v0.1.34-playwright-review.md` (originally classified as P0 cross-tab a11y leak; verification reclassified to P2 — modern browsers do exclude `display:none` from the a11y tree, but the DOM-state hygiene gap is real and Playwright snapshots do dump the leaked nodes).

## Evidence

`packages/memtomem/src/memtomem/web/static/app.js:680-694`:

```js
// Hide all panels
document.querySelectorAll('.tab-panel').forEach(p => p.classList.remove('active'));   // 681

// ...

// Show panel
const panel = qs(`tab-${tabName}`);
if (panel) {
  panel.hidden = false;          // 694 — set on activate, never undone elsewhere
  panel.classList.add('active');
  // ...
}
```

The HTML (`packages/memtomem/src/memtomem/web/static/index.html`) starts with `hidden` on each inactive panel, but once `activateTab()` runs for a panel it strips `hidden` and never reapplies it on tab switch.

## Why this is worth fixing even though `display:none` works

- DOM state accumulates; over a session every visited panel ends with `hidden=false`. Test automation tools (Playwright `browser_snapshot`) walk the DOM and report stale content.
- Future CSS edits or a switch from `display:none` to e.g. `visibility:hidden` would silently regress a11y. Belt-and-braces (`hidden` attr + class) is cheap.
- `inert` could be added at the same time to also remove focus / pointer events from inactive panels.

## Suggested fix

In `activateTab()`, change the deactivation pass to:

```js
document.querySelectorAll('.tab-panel').forEach(p => {
  p.classList.remove('active');
  p.hidden = true;          // belt-and-braces with the CSS-driven display:none
});
```

…and the activation block continues to set `panel.hidden = false`.

Optionally also toggle `inert` for keyboard / pointer isolation.

## References

- Tracking umbrella: TBD

memtomem:
## Summary

- `activateTab()` removed `.active` from all panels but never set `hidden = true` back. After the first activation every visited panel ended the session with `hidden = false`, relying only on the CSS `display:none` from the absent `.active` class to stay hidden.
- Modern browsers do exclude `display:none` from the a11y tree, so this isn't a hard a11y bug — but Playwright `browser_snapshot` walks the DOM and dumps the leaked nodes (this is exactly how #699 surfaced, in the v0.1.34 prod UX review). A future CSS-only refactor that drops `display:none` would also silently regress hiding.
- Pair `classList.remove('active')` with `p.hidden = true` in the deactivation forEach. Belt-and-braces with the existing CSS rule.
- Bump `app.js` cache-buster v97 → v98.

Closes #699.

## Test plan

- [ ] Click through all main tabs once (Home → Sources → Search → Index → Tags → Timeline → Settings) then back to Home; in DevTools check every other `.tab-panel` shows `hidden` attribute set, only the active panel has `hidden = false`.
- [ ] Playwright `browser_snapshot` after the round-trip: only one `tab-panel` should appear unhidden in the snapshot tree.
- [ ] Confirm no regression: panels still render content correctly when re-activated (focus restoration, history pushState, tab-specific loaders all still fire).

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -803,10 +803,8 @@ function activateTab(tabName, opts = {}) {
     b.setAttribute('tabindex', '-1');
   });
 
-  // Belt-and-braces with `display:none` from the absent `.active` class —
-  // without re-applying `hidden` here, visited panels leak into a11y /
-  // Playwright snapshots and a future CSS-only refactor would silently
-  // regress hiding.
+  // Re-apply `hidden` so DOM state matches CSS — without this, visited panels
+  // leak into a11y / Playwright snapshots (#699).
   document.querySelectorAll('.tab-panel').forEach(p => {
     p.classList.remove('active');
     p.hidden = true;
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -803,8 +803,14 @@ function activateTab(tabName, opts = {}) {
     b.setAttribute('tabindex', '-1');
   });
 
-  // Hide all panels
-  document.querySelectorAll('.tab-panel').forEach(p => p.classList.remove('active'));
+  // Belt-and-braces with `display:none` from the absent `.active` class —
+  // without re-applying `hidden` here, visited panels leak into a11y /
+  // Playwright snapshots and a future CSS-only refactor would silently
+  // regress hiding.
+  document.querySelectorAll('.tab-panel').forEach(p => {
+    p.classList.remove('active');
+    p.hidden = true;
+  });
 
   // Activate the correct button
   const btn = document.querySelector(`.tab-btn[data-tab="${tabName}"]`);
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
Fix is correct and scope-appropriate — the change is purely DOM hygiene (both `[hidden] { display: none !important }` at style.css:6 and the base `.tab-panel { display: none }` at style.css:216 already hide inactive panels visually), so this is the snapshot-pollution / future-proofing fix #699 advertised, not a hidden visual bug.

Verified:
- `.tab-panel` selector is scoped to exactly 7 top-level containers, no sub-component reuse.
- Initial HTML ships with `hidden` on inactive panels (60/413/481/588), so the post-fix invariant matches the static markup.
- Comment trimmed to two lines (just pushed) to comply with CLAUDE.md "one short line max" convention.

Approving.

PR #713, opened 2026-05-02
test: skipif POSIX-only assertions on Windows (mode bits, fcntl, signal) — Windows CI PR 2/N

memtomem:
PR 2 of N in the Windows-CI green-up roadmap. (PR 1 was #711.)

## Summary

Adds `@pytest.mark.skipif(sys.platform == "win32", reason="…")` to **9 tests across 5 files** that exercise features which either don't exist on Windows or take a deliberately-different code path that's already covered by an existing `test_windows_skips_kill`-style sibling. Single uniform change axis — no assertion logic touched, no production source modified.

## Cross-check table

Triaged against the latest Windows fail log (`gh run view 25244446944 --log-failed --job=74026052967`, 113 failures total).

| Test | Class | Why POSIX-only |
|------|-------|---------------|
| `test_context_atomic.py::test_atomic_write_explicit_mode_0o600` | mode bits | NTFS ignores `stat.S_IMODE` |
| `test_context_atomic.py::test_atomic_write_respects_custom_mode` | mode bits | same |
| `test_context_install.py::test_install_files_written_with_default_mode` | mode bits | same |
| `test_context_settings.py::TestClaudeSettingsAtomicWrite::test_mode_is_0o600` | mode bits | same |
| `test_server_sigterm.py::test_legacy_lock_sh_allows_multiple_holders` | fcntl | `ModuleNotFoundError: No module named 'fcntl'` at body-import time |
| `test_upgrade_cmd.py::test_no_running_server_just_reinstalls` | signal/path | Windows takes the "Detected Windows; skipping process termination" branch (covered by `test_windows_skips_kill`) |
| `test_upgrade_cmd.py::test_running_server_sigterm_path` | signal | exercises SIGTERM kill path; Windows skips it |
| `test_upgrade_cmd.py::test_running_server_escalates_to_sigkill` | signal | SIGKILL escalation path |
| `test_upgrade_cmd.py::test_pid_file_unlink_skipped_if_respawned` | signal | SIGKILL respawn-detection path |

**No Windows coverage is lost for the upgrade flow** — `test_windows_skips_kill` (line 124 of `test_upgrade_cmd.py`) explicitly tests the Windows branch and continues to run.

## Why skipif and not test rewrite

For the file-mode and `fcntl` tests, the property under test is meaningful only on POSIX. NTFS doesn't honour `stat.S_IMODE` and Python on Windows doesn't ship `fcntl` — there is nothing equivalent to assert.

[...]

PR #711, opened 2026-05-02
test: normalize path separators in 7 cross-platform assertions (#643)

memtomem:
Closes #643.

## Summary

The Windows CI job runs as `informational` (`continue-on-error: true` at `.github/workflows/ci.yml:45`). On the latest main run it accumulated **117 test failures**, falling into 6 distinct classes. This PR addresses the smallest, lowest-risk class — **path-separator literals in test assertions** — so the rest can be tackled cleanly in follow-up PRs.

7 tests across 3 files compared production output (a `Path` whose `str()` uses the platform separator) against hardcoded forward-slash literals. Wrapping the LHS in `Path(...).as_posix()` makes the comparison portable without changing what the test actually exercises.

**Test-only PR.** No production source touched (notably `norm_path()` in `storage/sqlite_helpers.py:24-38` is left as-is — its `str(p.resolve())` behaviour is on-disk-key-shaped and changing it would force a migration of every existing chunk row).

## Cross-check table

Triage was done against `gh run view 25244100712 --log-failed --job=74025071567`. Of the 14 candidates flagged by exploration, only the 7 below were genuinely path-separator failures; the others were misclassified and belong to other failure classes (see *Out of scope* below).

| Test | Path-sep? | Pattern | Notes |
|------|:---------:|:-------:|-------|
| `test_config_overrides.py::test_env_var_wins_over_config_json_scalar` | ✅ | 1 | `'\\from\\env.db' == '/from/env.db'` |
| `test_config_overrides.py::test_env_var_wins_over_config_json_list` | ✅ | 1 (list) | `['\\from\\env'] == ['/from/env']` |
| `test_config_overrides.py::test_env_and_config_coexist_on_different_fields` | ✅ | 1 | same scalar shape |
| `test_config_overrides.py::test_regression_pr247_mcp_json_env_block` | ✅ | 1 (×2) | scalar + list, tilde-prefixed |
| `test_config_overrides.py::test_config_d_env_wins_over_fragments` | ✅ | 1 | env-over-fragment |
| `test_session_summary_rescue.py::TestBoostSourcesHelper::test_above_threshold_walks_chunk_links_to_source_files` | ✅ | 1 (set comp) | `{'\\tmp\\src\\…'} == {'/tmp/src/…'}` |
[...]

Issue #634, opened 2026-05-01
ci: add Windows runner to CI matrix


memtomem:
## Background

Windows isn't currently CI-tested or formally documented as supported.
This issue proposes promoting Windows from *implicitly unsupported* to
*CI-tested*, motivated by PR #633 — which fixed 5 tests (one of which
propagates to a 14-test class via its fixture) that hit
`OSError: [WinError 1314] (client does not hold the required privilege)`
because they call `os.symlink` / `Path.symlink_to`. CI never noticed
because every job in `.github/workflows/ci.yml` runs on `ubuntu-latest`.

There's no broken cross-platform promise to fix here — `README.md`,
`packages/memtomem/README.md`, and `CONTRIBUTING.md` don't claim Windows
support today. The relevant prior art is the implicit POSIX-only stance
of the `mm: command not found` README tip (which mentions `~/.local/bin`
on macOS/Linux only). What this issue proposes is a deliberate step
*toward* supporting Windows by adding the CI signal first.

There are likely other latent Windows-specific issues hiding behind the
ubuntu-only blind spot — see "Likely first-run failures" below.

## Proposal

Add `windows-latest` (and probably `macos-latest` while we're at it) to
selected jobs in `.github/workflows/ci.yml`. The repo currently has 6
jobs; per-job recommendations:

| Job | OS matrix recommendation | Why |
| --- | --- | --- |
| `lint` | ubuntu only | Ruff output is identical across OSes; matrix would just slow PRs |
| `typecheck` | ubuntu only | mypy is platform-agnostic (advisory anyway per `CLAUDE.md`) |
| `test` | **`[ubuntu-latest, windows-latest, macos-latest]`** | The actual surface this issue is about |
| `golden-path (ONNX bge-m3)` | matrix | Cache key already uses `${{ runner.os }}` — prepared for it; ONNX/fastembed has known cross-platform behavior worth pinning |
| `llm-nightly (optional)` | ubuntu only | Opt-in nightly, not on PR fast path |
| `notebooks` | matrix (low priority) | Jupyter on Windows has its own quirks; defer until base test job is green |

Concretely:

```yaml
test:
  strategy:
    fail-fast: false
    matrix:
      os: [ubuntu-latest, windows-latest, macos-latest]
  runs-on: ${{ matrix.os }}
```

### Notes on cost / latency

Actions minutes are **free and unlimited for public repos** (this repo
is public), so the 2x/10x billing multipliers for Windows/macOS *don't
[...]

Issue #316, opened 2026-04-20
Windows: wizard fails to auto-detect Claude/Codex memory dirs (backslash path regex)


memtomem:
## Symptom

Windows users running `mm init` don't get Claude Code / Codex memory directories auto-detected by the wizard's "Provider memory folders" step. Any such path they add manually is classified under the generic `user` category instead of `claude-memory` / `claude-plans` / `codex`, so the preset namespace rules added in #312 (`claude:{ancestor:1}`, `claude-plans`, `codex`) don't get applied either.

Silent failure: no error, no warning — the wizard just doesn't find the directories it would find on macOS/Linux.

## Root cause

`categorize_memory_dir` at `packages/memtomem/src/memtomem/config.py:1040-1064` uses regex patterns that match forward slashes only:

```python
_PROVIDER_CATEGORY_PATTERNS = (
    ("claude-memory", re.compile(r"/\.claude/projects/[^/]+/memory/?$")),
    ("claude-plans",  re.compile(r"/\.claude/plans/?$")),
    ("codex",         re.compile(r"/\.codex/memories/?$")),
)
```

On Windows, `str(Path(...))` returns backslash-separated paths (`C:\Users\foo\.claude\projects\bar\memory`), so none of the patterns match and the function falls through to `return "user"` (line 1064). The docstring at lines 1056-1058 already acknowledges this gap:

> Uses forward-slash regex, so on Windows a backslash-normalised path will fall through to `"user"` until a future path-sep-agnostic pass lands.

This issue is that "future pass".

## Scope

- Normalise the path-separator before matching (e.g., `Path(path).as_posix()` or `str(path).replace("\\", "/")`).
- Preserve existing POSIX-path behaviour — no regression for macOS/Linux users.
- Independent of RFC #304 vocabulary decisions. The category names and the axis stay exactly as they are today; only the matching mechanism changes.
- The vocabulary lock from #313 (`_VALID_PROVIDER_CATEGORIES`) stays untouched — this fix does not modify the locked set.

## Testing notes

CI (`.github/workflows/ci.yml`) runs on `ubuntu-latest` only — no Windows runner. A fix therefore needs platform-agnostic tests that exercise the backslash code path without a Windows host.

Preferred fixture pattern: **hand-constructed backslash strings** passed directly to `categorize_memory_dir`, e.g.

```python
assert categorize_memory_dir(r"C:\Users\foo\.claude\projects\abc\memory") == "claude-memory"
```

[...]

PR #738, opened 2026-05-03
fix(runtime-paths): make Windows runtime dir resolution NTFS-aware (closes #637)

### LINKED ISSUE #637: Windows: replace POSIX file-mode XDG_RUNTIME_DIR tests with NTFS-aware equivalents
## Background

The Windows + macOS CI matrix expansion (#634) skips three test classes
in \`packages/memtomem/tests/test_runtime_paths.py\` on Windows:

- \`TestRuntimeDir\`
- \`TestEnsureRuntimeDir\`
- \`TestServerPidPath\`

These classes assume POSIX semantics that don't translate cleanly to
NTFS:

- The shared \`_make_safe_xdg\` fixture does \`os.chmod(xdg, 0o700)\` to
  set up the XDG base. NTFS reports synthesized POSIX mode bits
  (typically \`0o666\` or \`0o777\`) that don't match \`0o700\`, so
  \`_is_safe_dir\` (which checks \`S_IMODE(st.st_mode) & 0o077\`) would
  reject the directory and the XDG branch silently falls through to the
  tempdir form — breaking the tests that assert the XDG path is used.
- \`TestEnsureRuntimeDir.test_creates_directory_with_owner_only_mode\`
  asserts \`S_IMODE(d.stat().st_mode) == 0o700\`, which is POSIX-only.
- \`TestEnsureRuntimeDir.test_explicit_chmod_survives_wild_umask\` uses
  \`os.umask(0o177)\`, which Windows ignores.
- \`TestEnsureRuntimeDir.test_refuses_existing_loose_mode\` depends on
  the \`0o755\` vs \`0o700\` distinction.

The runtime code itself
(\`packages/memtomem/src/memtomem/_runtime_paths.py\`) already has a
Windows-aware path through \`tempfile.gettempdir()\` + \`uid=0\` (the
\`os.geteuid()\` calls are \`hasattr\`-guarded), but its **behavior on
NTFS is unverified** today — that's what this issue tracks.

## Scope

1. Verify \`runtime_dir()\` and \`ensure_runtime_dir()\` behave correctly
   on Windows: pid file lands at \`%LOCALAPPDATA%\Temp\memtomem-0\`
   (or whatever \`tempfile.gettempdir()\` returns), \`ensure_runtime_dir()\`
   doesn't raise on a fresh runner, and validation doesn't no-op away
   the security guarantees we care about.
2. Decide the NTFS security contract. Options:
   - Relax \`_is_safe_dir\` on Windows (accept the kernel default).
   - Use \`os.access(path, os.W_OK)\` + owner-SID checks via stdlib
     \`os.stat()\` + \`stat.FILE_ATTRIBUTE_*\` (or \`pywin32\` if needed).
   - Document Windows as best-effort and skip mode-bit validation.
3. Either rewrite the three skipped classes to be Windows-aware, or
   add a separate \`TestRuntimeDir{Windows,Posix}\` split with native
   assertions per platform.

Out of scope: changing the runtime code's POSIX contract. The pid-file
location and ownership semantics on Windows are an open design
question that this issue should decide before relaxing anything.

## References

- #634 — Windows + macOS CI matrix expansion
- \`packages/memtomem/src/memtomem/_runtime_paths.py:63-145\` — runtime
  resolver + \`ensure_runtime_dir\`
- \`packages/memtomem/tests/test_runtime_paths.py\` — three test classes
  currently skipped on Windows

pandas-studio:
## Summary

- Pre-fix, `ensure_runtime_dir()` was broken on Windows from the **second** invocation onward: NTFS synthesizes POSIX mode bits as `0o666`/`0o777`, so `stat.S_IMODE(st.st_mode) & 0o077` was always non-zero and the validator raised `PermissionError("unsafe permissions")` against a directory it had just created.
- Make the Windows path explicit: `runtime_dir()` skips the `XDG_RUNTIME_DIR` branch on Windows (it's a Linux/systemd convention, and honoring it without the POSIX mode-bit gate is half-baked); `ensure_runtime_dir()` keeps the symlink + non-directory rejections on every platform but skips the mode-bit + owner-uid gates on Windows (`%LOCALAPPDATA%\Temp\` is already per-user, and proper NTFS owner-SID checks via `GetSecurityInfo` are out of scope).
- Module docstring documents the asymmetric trust posture so a future contributor doesn't re-enable a check that has no NTFS semantics.
- Replace the three `skipif(not hasattr(os, "geteuid"))` guards on the POSIX-only test classes with `skipif(os.name == "nt")` — same effect today but parallel to the production guard.
- Add `TestWindowsRuntimeDir` covering the NTFS-equivalent contract: tempdir-only resolution with `uid=0` suffix, XDG ignored when set, **idempotent** `ensure_runtime_dir` (regression guard for the mode-bit failure), and the symlink + non-directory rejections that survive the NTFS relaxation. Symlink test uses the existing `requires_symlinks` marker so it auto-skips when the Windows runner doesn't have Developer Mode/admin.

## What changed

| File | Change |
|---|---|
| `packages/memtomem/src/memtomem/_runtime_paths.py` | XDG branch gated on `os.name != "nt"`; mode-bit + owner-uid gates in `ensure_runtime_dir` gated on `os.name != "nt"`; module docstring rewritten to call out the asymmetric posture |
| `packages/memtomem/tests/test_runtime_paths.py` | 3× skipif guard reworded; new `TestWindowsRuntimeDir` class (7 tests) |

POSIX behaviour is untouched. `_is_safe_dir` is now POSIX-only by contract — its docstring says so.

## Why these gates can be relaxed on Windows

- `tempfile.gettempdir()` on Windows resolves to `%LOCALAPPDATA%\Temp\` — already a per-user directory. The cross-user threat that motivates `0o700` on shared `/tmp` does not apply.
- POSIX mode bits on NTFS are synthesized (`0o666`/`0o777`) and do not reflect the real ACL — checking `& 0o077` is theatre and false-positives against any directory we ourselves just created.
- Symlinks on Windows exist (require Developer Mode/admin to create) and are exploitable identically to POSIX symlinks once present, so the symlink rejection survives. Non-directory rejection survives for the same reason.
- Proper NTFS owner-SID validation would need `pywin32` / `ctypes` calls into `GetSecurityInfo`. Heavyweight optional dependency for marginal value when the temp base is already per-user — explicitly out of scope.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_runtime_paths.py -v` — 18 passed, 7 skipped on macOS
- [x] `uv run pytest packages/memtomem/tests/test_uninstall_cmd.py packages/memtomem/tests/test_runtime_paths.py` — 46 passed, 7 skipped (downstream `runtime_dir()` consumer)
- [x] `uv run pytest packages/memtomem/tests/test_web_routes_context.py packages/memtomem/tests/test_context_skills.py` — 86 passed (other consumers)
- [x] `uv run mypy packages/memtomem/src/memtomem/_runtime_paths.py` — no issues
- [ ] Windows CI matrix run will exercise the new `TestWindowsRuntimeDir` class — currently informational, should pass in this PR's run

## References

- Closes #637
- Builds on #634 (CI matrix expansion) and #638 (matrix infra)
- Sibling Windows fixes: #714 (HOME → USERPROFILE sweep, refs #644), #716 (#316 backslash regex), #713 (POSIX skipifs), #711 (path separator normalization)

### DIFF
--- a/packages/memtomem/src/memtomem/_runtime_paths.py
+++ b/packages/memtomem/src/memtomem/_runtime_paths.py
@@ -11,23 +11,38 @@ Resolution order:
    exports the var). Per-user, ``tmpfs``-backed, kernel-managed lifecycle.
    Accepted only if the base is a real directory (not a symlink), owned
    by the effective uid, and has mode ``0o700`` (no group/world bits).
+   Skipped on Windows: ``XDG_RUNTIME_DIR`` is a Linux/systemd convention,
+   and the POSIX-mode-bit gate that protects it is meaningless against
+   NTFS ACLs.
 2. ``{tempfile.gettempdir()}/memtomem-{uid}`` — macOS (where
-   ``gettempdir()`` resolves to a per-user ``/var/folders/.../T/`` already)
-   and Linux without systemd. ``uid`` suffix disambiguates a shared
-   ``/tmp``; ``mode=0o700`` at creation keeps it private to the user.
+   ``gettempdir()`` resolves to a per-user ``/var/folders/.../T/`` already),
+   Linux without systemd, and Windows (where ``gettempdir()`` returns
+   ``%LOCALAPPDATA%\\Temp\\`` — already per-user). On Windows ``uid``
+   collapses to ``0`` since there is no ``geteuid``; the cross-user
+   collision risk that motivated the suffix on shared ``/tmp`` does not
+   apply on Windows because the temp base is per-user already.
 
 Security posture for the runtime directory itself:
 
 - Never follow a symlink — an attacker on a shared ``/tmp`` could
   pre-create ``memtomem-{uid}`` as a link into the user's home, and a
   naive ``mkdir`` would silently no-op through it. ``os.stat`` with
-  ``follow_symlinks=False`` catches this.
-- Refuse any pre-existing directory not owned by the effective uid or
-  not at mode ``0o700``. This trades convenience for a predictable
-  contract: a ``root``-owned leftover from ``sudo mm …`` or a
-  mode-degraded dir raises :class:`PermissionError` with a remediation
+  ``follow_symlinks=False`` catches this. Enforced on every platform
+  (Windows symlinks require Developer Mode/admin to create but exist
+  and are exploitable the same way once present).
+- POSIX only — refuse any pre-existing directory not owned by the
+  effective uid or not at mode ``0o700``. This trades convenience for a
+  predictable contract: a ``root``-owned leftover from ``sudo mm …`` or
+  a mode-degraded dir raises :class:`PermissionError` with a remediation
   hint instead of silently writing the pid file into a world-readable
   directory. The only fix is ``rm -rf`` the runtime dir and retry.
+- Windows — skip the mode-bit + owner-uid gates. NTFS synthesizes POSIX
+  mode bits as ``0o666``/``0o777`` (the values do not reflect the real
+  ACL), so ``& 0o077`` would always trip and ``ensure_runtime_dir`` would
+  refuse the second invocation against its own previously-created dir.
+  Proper NTFS owner-SID validation needs ``pywin32`` / ``ctypes`` calls
+  into ``GetSecurityInfo`` and is out of scope; we rely on
+  ``%LOCALAPPDATA%\\Temp\\`` already being per-user.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


memtomem (approved):
Fix is correct and well-scoped — the idempotency regression guard pins the exact failure mode #637 reported, and the trust-posture trade-off (drop synthesized-mode-bit checks, keep symlink + non-directory rejection) is documented in the module docstring so a future contributor can't silently re-enable a check that has no NTFS semantics. Approving.

Optional follow-ups (non-blocking):
- Multi-line test docstrings in `TestWindowsRuntimeDir` violate the repo's "one short line max" convention from CLAUDE.md — class docstring already covers the why.
- Dead-code nit: `hasattr(os, "geteuid")` inside `if os.name != "nt":` (`_runtime_paths.py:136`) can drop.
- `requires_symlinks` will auto-skip on default Windows runners, leaving the only remaining safety gate effectively untested in CI most of the time. Worth a mock-based unit test to exercise the symlink rejection platform-independently.

Issue #732, opened 2026-05-02
Windows: capture installed_at from FS mtime with us ceiling, not wall clock


memtomem:
## Background

Discovered as a follow-up to the umbrella #634 (Windows + macOS CI matrix) and the PR-D C2a / #630 invariant.

A single dirty-detection cluster in `test_dirty.py` + `test_context_status.py` fails on Windows-CI matching the `'dirty' == 'clean' / 'ok' / 'behind' / 'stale-pin'` shape. The root cause is **not** content / line-ending / permission related — `dirty.py:_iter_files` already skips symlinks, and CRLF / file-mode bits never enter the comparison (the rule is strictly `mtime`). The bug is two stacked clock / precision defects in how `installed_at` is captured.

### 1. Capture domain

`installed_at = utcnow_iso8601_z()` reads Python's wall clock. NTFS records `st_mtime` from a different timer (`FILETIME`). Under `dirty.py:148`'s strict `mtime > installed_at_epoch` check, just-written files can land *strictly later* than the captured timestamp. The current code was written under the POSIX assumption that both timers share a source — true on POSIX, false on Windows.

### 2. µs round-trip truncation

Even if capture moves to the FS, `st_mtime` (float seconds) → ISO-8601Z `%H:%M:%S.%fZ` (microsecond format) → `datetime.fromisoformat().timestamp()` **truncates** the NTFS 100-ns residual. The just-written file's own `st_mtime_ns` still carries the full nano-tick, so even self-comparison can violate `mtime ≤ installed_at_epoch` by ≤1µs. This is the actual failure surface — moving capture to FS without addressing precision leaves the same shape of bug.

## Fix

`installed_at_from_dest(dst)`:

1. Read `st_mtime_ns` (int, lossless) from the FS using the shared installed-file walker.
2. Ceiling-divide ns → µs before formatting.

The formatted `installed_at` round-trips to a value `>=` every walked file's actual `st_mtime` — byte-identical to today on POSIX (where ordinary writes give `st_mtime_ns % 1000 == 0`; ceil is a no-op) and with a `<= 1µs` safety margin on NTFS.

Locked decision #5 in PR-D C2 (`strict >`) is preserved — we strengthen the underlying capture, not the comparison.

## Validation

- Pin tests in `TestInstalledAtFromDest` (`test_dirty.py`) cover empty fallback, single/multi mtime, skip rules, and the load-bearing µs residual.
[...]

PR #739, opened 2026-05-03
test(windows): force distinguishable mtime in concurrent-merge abort test (closes #645)

### LINKED ISSUE #645: Windows: config_signature unchanged after write — NTFS mtime resolution / cache

pandas-studio:
## Summary

- `test_aborts_on_mtime_change` is the last residual Windows CI failure from #645. The test simulates a concurrent writer by patching `_read_with_mtime` to write to the target between the read and the production code's mtime re-check. On Windows the two `write_text` calls can land inside the same system-clock tick, so `st_mtime_ns` reports the same value and the abort branch never fires.
- Force a distinguishable mtime via `os.utime(..., ns=(atime, mtime + 1ms))` after the simulated write so the test deterministically exercises the production guard regardless of OS timer granularity. A real-world second writer (separate `mm context sync` run, user editing in editor) is naturally well above the threshold — this only covers the test-induced race.
- Test-only change, no production impact.

## Why mtime collides on Windows

NTFS native FILETIME resolution is 100ns, but `WriteFile`-induced metadata updates use the *system clock*, which advances at ~15.6ms on typical Windows hosts (default timer interrupt). Two writes within that tick get the same mtime even though the second strictly happened after the first. The fix bumps mtime by 1ms (1_000_000 ns) — well above the synthetic resolution, well below any real-world inter-process gap.

## Status of the broader #645 cluster

The issue body listed 17 `test_web_hot_reload.py` failures. The latest Windows CI run on `main` (run [25264999636](https://github.com/memtomem/memtomem/actions/runs/25264999636), 2026-05-02) reports **0** hot-reload failures — those have been cleared by previous Windows CI work (PRs #711, #713, #714, #716, #732). Only this single `test_context_settings.py::test_aborts_on_mtime_change` failure remained, and this PR closes it.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_context_settings.py -v` — all 35 tests pass on macOS
- [ ] Windows CI matrix — should drop from 1 failure to 0

## References

- Closes #645
- Sibling Windows fixes: #714 (HOME → USERPROFILE), #716 (#316 backslash regex), #713 (POSIX skipifs), #711 (path separators), #732 (installed_at us ceiling), #738 (#637 NTFS runtime dir)

### DIFF
--- a/packages/memtomem/tests/test_context_settings.py
+++ b/packages/memtomem/tests/test_context_settings.py
@@ -259,6 +259,18 @@ class TestClaudeSettingsMergeConcurrent:
                 target.write_text(
                     json.dumps({"hooks": {}, "_bumped": True}) + "\n", encoding="utf-8"
                 )
+                # Two writes inside the same system-clock tick can report the
+                # same ``st_mtime_ns`` on Windows (NTFS native resolution is
+                # 100ns but ``WriteFile``-induced metadata updates use the
+                # system clock, which often advances at ~15.6ms). Bump
+                # explicitly so the simulated concurrent-writer is
+                # distinguishable regardless of OS timer granularity — a
+                # real-world second writer is naturally well above this
+                # threshold, so this only covers the test-induced race.
+                import os as _os
+
+                st = target.stat()
+                _os.utime(target, ns=(st.st_atime_ns, st.st_mtime_ns + 1_000_000))
             return result
 
         import unittest.mock


memtomem (approved):
Test-only 12-line fix scoped exactly to #645 — explicit `os.utime` bump (`+1ms`) ensures the mocked second writer has a distinguishable `st_mtime_ns` regardless of OS timer granularity (NTFS via `WriteFile` reports system-clock resolution ~15.6ms). Comment correctly notes a real-world second writer is naturally well above this threshold so this only covers the test-induced race. Approving.

Issue #659, opened 2026-05-01
refactor(web): extract shared chunk_progress throttle helper (defer to rule-of-three)


memtomem:
## Context

After PR #658 (Memory Dirs reindex meta-row chunk progress, #654), the same throttle + final-tick-bypass pattern lives in two call-sites:

1. \`packages/memtomem/src/memtomem/web/static/app.js:3716-3751\` — Index tab \`runIndexStream\`, \`fileEl\` target.
2. \`packages/memtomem/src/memtomem/web/static/sources-memory-dirs.js:1108-1191\` — Memory Dirs \`mdReindexOne\`, meta-row target.

Both share:

- \`let _lastChunkRender = 0;\` function-local state (closure-scoped to isolate concurrent calls).
- 100ms gap on intermediate \`chunk_progress\` ticks.
- Final-tick bypass (\`done >= total\`) so \`(N/N)\` lands before the next file boundary.
- Reset to \`0\` on \`progress\` events (file boundary).
- Same i18n key (\`index.file_chunk_progress\`) for the rendered label.

## Why defer (rule-of-three)

Two call-sites is the danger zone for premature abstraction — a helper extracted now will be shaped by exactly these two consumers and bend awkwardly when a third arrives. Wait until the third use materializes (likely candidates: \`mm index\` CLI progress if it ever switches to the streaming endpoint, or wizard \`_seed_with_progress\`), then extract with the third call-site's needs informing the API.

## Scope when this fires

- Extract a helper (e.g. \`makeChunkProgressRenderer({ targetEl, formatKey })\` returning \`{ onChunk, onProgressBoundary, onCleanup }\`) into a shared utility module loaded before \`app.js\` and \`sources-memory-dirs.js\`.
- Migrate both existing call-sites verbatim (no behavior change).
- **Rename the i18n key** \`index.file_chunk_progress\` → \`common.file_chunk_progress\` (or wherever the helper lives) since it's no longer Index-tab-specific. Update both \`en.json\` and \`ko.json\` plus all \`t()\` call-sites in the same PR.
- Bump the cache-busters on the touched script tags (\`?v=N\` → \`?v=N+1\`).

## Acceptance criteria

- Three or more call-sites share the helper.
- Behavior identical to today (manual test plan from PR #658 and PR #653 still passes).
- i18n key namespace matches the helper's location, both locales updated together.
- No new \`chunk_progress\` SSE event types.

## References

- PR #653 (parent feature — Index tab).
[...]

Issue #656, opened 2026-05-01
feat(cli): convert mm index to stream + click.progressbar


memtomem:
## Context

`mm index <path>` (`packages/memtomem/src/memtomem/cli/indexing.py:107` and `:194`) calls `comp.index_engine.index_path()` — the **non-stream** path. So even after PR #653 added per-chunk SSE progress, direct CLI users see exactly the same thing they did before: nothing until the run finishes, then a single summary block.

The wizard already uses `index_path_stream` (`init_cmd.py:_seed_with_progress`) and gets per-file progress. Folding `mm index` into the same pattern would unify the CLI feedback story.

## Proposal

Switch `mm index` to consume `index_path_stream` and render a `click.progressbar` matching the wizard's shape. Likely overlap with #655 (wizard chunk display) — both surfaces share the same `click.progressbar` constraints, so it's natural to either:

- **Bundle**: solve #655's design once, reuse the bar + `item_show_func` here.
- **Sequence**: ship #655 first (wizard already uses the bar), then this one becomes a 5-line refactor pointing the CLI at the same helper.

## Acceptance criteria

- `mm index <large-dir>` shows live per-file progress (and per-chunk if #655 lands first).
- `Ctrl-C` mid-run still prints the same resume hint as today.
- `--force` and `--namespace` flags preserved.
- Exit code stays the same on success / partial failure / hard failure.
- The non-stream `index_path` API doesn't need to be removed — internal callers (tests, MCP `mem_index`) may still want it.

## Files likely touched

- `packages/memtomem/src/memtomem/cli/indexing.py` (the two `index_path` call sites)
- Possibly extracting a shared `_run_indexing_with_bar` helper between this file and `init_cmd.py:_seed_with_progress` (depends on #655's outcome).

## References

- PR #653 — feature parent.
- #655 — wizard chunk progress (closely related; consider bundling).

PR #740, opened 2026-05-03
feat(cli): show chunk progress in mm init wizard's _seed_with_progress (closes #655)

### LINKED ISSUE #655: feat(cli): show chunk progress in mm init wizard's _seed_with_progress
## Context

PR #653 added per-chunk SSE progress and surfaced it in the web Index tab. The `mm init` wizard's seed-index path (`packages/memtomem/src/memtomem/cli/init_cmd.py:_seed_with_progress`, line 1496-) consumes the same `index_path_stream` but ignores the new `chunk_progress` events. PR #653 left this out deliberately — see PR description "Out of scope" and the docstring note in the code.

## Why this needs design

`_seed_with_progress` uses `click.progressbar` configured for **file-unit length** with this `item_show_func`:

```python
click.progressbar(
    length=expected_total,                            # files
    label="  Seeding",
    item_show_func=lambda item: (item or "").rsplit("/", 1)[-1][:40],
)
```

Two problems:

1. The `item_show_func` extracts the **path tail** and truncates to **40 chars**. Passing a free-form `"file (32/250)"` works for "/"-less inputs but the chunk count gets eaten by the truncate.
2. The bar's length is in files, so `chunk_progress` events have no natural advance unit. Just calling `bar.update(0, label)` works (advance=0, label refresh) but the wizard's UX tone is "minimal CLI noise" — fast-firing chunk labels may feel busy.

## Options

1. **Refresh sub-label without advance** — `bar.update(0, f"{file} {done}/{total}")` and a wider `item_show_func` that handles both formats. Simplest. Risk: noisy on fast files.
2. **Switch bar length to chunk units** — `length=expected_chunks` instead of `expected_total` files; advance per chunk. Requires pre-counting chunks (currently only files are pre-counted). Cost-heavy.
3. **Two-bar layout** — outer file-bar, inner per-file chunk-bar shown only when `chunks_total > N`. Most informative but `click.progressbar` doesn't compose well.

## Acceptance criteria

- Wizard seed run shows chunk-level progress for files producing more than `progress_threshold` chunks (default 32).
- Small files (under threshold) stay quiet — match the web Index tab behavior.
- KeyboardInterrupt cleanup still works (test: ^C mid-seed prints the resume hint, `_close_bar` runs).
- No regression in the multi-path aggregation summary line at the end.

## Files likely touched

- `packages/memtomem/src/memtomem/cli/init_cmd.py` (`_seed_with_progress`)
- Possibly `packages/memtomem/src/memtomem/cli/indexing.py` if `mm index` (non-stream) is folded in at the same time — see #656 for the related conversion.

## References

- PR #653 — feature parent. PR description "Out of scope" section explains why this was deferred.

pandas-studio:
## Summary

The wizard's `_seed_with_progress` consumed `chunk_progress` SSE events from `index_path_stream` (added in #653) but ignored them, so multi-minute embedder runs from `mm init` looked hung — exactly the UX issue `chunk_progress` was introduced to fix, just on a different surface. This wires the wizard up to the same chunk-level signal the web Index tab uses.

## Design

Option 1 from #655: refresh the bar's sub-label without advancing.

- Bar length stays in **file units** (no chunk pre-count).
- `chunk_progress` -> `bar.update(0, (file, done, total))` re-renders the label only.
- `item_show_func` dispatches on `tuple` vs `str` so the existing `progress`-event path keeps the basename label.
- Throttle/render rules mirror the web Index tab (`app.js` ~L4219-4256): 100ms gap between intermediate renders via `time.monotonic()`, final tick (`chunks_done >= total`) bypasses throttle, clock resets to 0 on every `progress` event so the next file's first chunk renders immediately.
- `_ensure_bar()` extracted so the bar can lazily appear on either the first `chunk_progress` or `progress` event.
- Drive-by: switched basename render from `rsplit("/", 1)` to `Path(...).name` so Windows backslash paths display correctly.

Server-side gating in `indexing/engine.py:786` already filters small files (default `progress_threshold=32`), so the CLI doesn't need to threshold itself — small files stay quiet, matching the web Index tab.

## Invariants preserved

- KeyboardInterrupt cleanup unchanged (`_close_bar()` + `try/except` block intact).
- Multi-path aggregation summary line unchanged (only the `complete` event feeds it).
- No changes to `index_path_stream` / `engine.py` / any web static file.
- `#659` (shared throttle helper between CLI and JS sites) deferred to rule-of-three; becomes natural after `#656` lands. This is the second of the three call-sites it tracks.

## Test plan

- [x] Two new tests in `TestInitialSeedThreshold` pin the no-advance contract (chunk events leave `bar.pos == file_count`) and the no-regression contract (zero `chunk_progress` events == today's behavior).
- [x] `uv run pytest packages/memtomem/tests/test_init_cmd.py -m "not ollama" -q` -> 264 passed.
- [x] `uv run ruff check` + `ruff format --check` clean.
- [x] `uv run mypy packages/memtomem/src/memtomem/cli/init_cmd.py` clean.

Closes #655

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -1609,6 +1609,41 @@ def _seed_with_progress(paths: list[Path]) -> bool:
     # the first one. Stays accurate across serial iteration.
     expected_total = sum(_collect_seed_scale(p)[0] for p in paths)
 
+    # Throttle clock for ``chunk_progress`` label refreshes. Mirrors the web
+    # Index tab (``web/static/app.js`` ~L4219-4256): 100ms gap between
+    # intermediate renders, final tick (chunks_done >= chunks_total) bypasses
+    # the throttle so ``(N/N)`` lands before the next file boundary, and the
+    # clock resets to 0 on every ``progress`` event so the next file's first
+    # chunk renders immediately. ``time.monotonic()`` (not ``time.time()``)
+    # so a wall-clock jump can't stall the bar. Issue #659 tracks extracting
+    # this into a shared helper once a third call-site appears
+    # (rule-of-three) — keep this implementation self-contained for now.
+    throttle_state: dict = {"last_render": 0.0}
+
+    def _format_item(item: object) -> str:
+        if not item:
+            return ""
+        if isinstance(item, tuple):
+            file, done, total = item
+            return f"{Path(file).name} ({done}/{total})"[:60]
+        # Legacy str case: the existing ``progress`` branch passes a path str.
+        # ``Path(...).name`` (not ``rsplit("/", 1)``) handles Windows
+        # backslash paths correctly.
+        return Path(str(item)).name[:40]
+
+    def _ensure_bar() -> None:
+        # Lazy creation: the bar can come into existence on EITHER the first
+        # ``chunk_progress`` OR the first ``progress`` event, whichever
+        # arrives first. ``chunk_progress`` for a given file is emitted before
+        # that file's ``progress`` summary, so for large files the bar appears
+        # at chunk-level rather than waiting for the first file to finish.
+        if bar_state["bar"] is None:
+            bar_state["bar"] = click.progressbar(
+                length=expected_total,
+                label="  Seeding",
+                item_show_func=_format_item,
+            ).__enter__()
+
     def _close_bar() -> None:
         if bar_state["bar"] is not None:
             try:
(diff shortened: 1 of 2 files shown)


memtomem (approved):
Mirrors the web Index tab's chunk_progress pattern (app.js): 100ms throttle, final-tick bypass, reset on file boundary, lazy bar creation on first event of either type. Helper kept self-contained per #659 rule-of-three deferral — correct call, the third call-site (#656 mm index streaming) is the natural extraction trigger. Solid regression test for the 'label refresh, no advance' contract (chunk events must not double-count file-unit bar length). Approving.

PR #741, opened 2026-05-03
feat(cli): convert mm index to stream + click.progressbar (closes #656)

### LINKED ISSUE #656: feat(cli): convert mm index to stream + click.progressbar

pandas-studio:
Stacked on #740 (rebases onto main once #740 merges).

## Motivation

Direct CLI users running `mm index <large-dir>` previously saw nothing until
the run finished — `index_path()` blocks for the entire duration. On
multi-minute embedder runs (bge-m3 / CPU, large corpus) this looked hung.
Switch `_index()` to consume `index_path_stream()` and render the same
throttled progress bar that #740 just polished for the wizard's seed flow.

## Design

Lifted the streaming + bar lifecycle into a new module
`memtomem.cli._index_progress` so both call sites (`_seed_with_progress` and
`mm index`'s `_index`) drive identical UX (throttled chunk-progress label,
lazy bar creation, file-unit length). `_collect_seed_scale` moved with it,
re-exported from `init_cmd` so existing test imports keep working.

Only `_index()` is converted. The debounce closure at `indexing.py:194`
keeps the non-stream API since hook-driven drains (`--flush` /
`--debounce-window`) don't want progress noise. The non-stream
`index_path()` stays for tests, MCP `mem_index`, and the closure — issue
text covers this explicitly.

## Invariants preserved

- Summary line shape `Indexed N file(s): N new, N unchanged, N deleted (Nms)` — stable interface scripts may grep.
- `--force`, `--namespace`, `--recursive` forward to `index_path_stream`.
- Ctrl-C prints `Cancelled. Resume with: mm index <path>` (idempotent re-run via content-hash dedup).
- Exit code unchanged on success / partial failure / hard failure.
- `_seed_with_progress` keeps its zero-chunks warning, multi-path web hint, graceful-skip — verified by the 264 existing wizard tests (no regressions).

## Test plan

- [x] 5 new tests in `tests/test_indexing_cli.py`: stream-event pass-through with legacy summary-line pin, error-line rendering, Ctrl-C resume hint, `--namespace`/`--force` kwarg propagation, default-no-namespace stub-engine compat.
- [x] Full wizard suite (264 tests in `test_init_cmd.py`) green, including #740's two new chunk-progress regression tests.
- [x] `test_cli_index_noop_e2e.py` (inline + subprocess) green.
- [x] Full sweep: 3991 passed, 46 deselected.
- [x] `ruff check` and `ruff format --check` clean.
- [x] `mypy` clean on `_index_progress.py`, `indexing.py`, `init_cmd.py`.

Closes #656

### DIFF
--- a/packages/memtomem/src/memtomem/cli/_index_progress.py
+++ b/packages/memtomem/src/memtomem/cli/_index_progress.py
@@ -1,0 +1,230 @@
+"""Shared streaming progress runner for ``mm index`` and the ``mm init``
+seed flow.
+
+Both call sites stream :meth:`IndexEngine.index_path_stream` and render the
+same ``click.progressbar`` shape: file-unit length pre-computed up front,
+``progress`` events advance the bar by one and reset the chunk-throttle clock,
+``chunk_progress`` events refresh the sub-label only (no advance), throttled
+to a 100 ms gap with a forced final-tick render so ``(N/N)`` lands before the
+next file boundary. Issue #659 tracks extracting the throttle into a helper
+shared with the web Index tab (``web/static/app.js``); that's deferred to
+rule-of-three on the JS side. The CLI side fires now (two callers: the
+wizard's :func:`_seed_with_progress` and ``mm index``'s ``_index``).
+
+Caller responsibilities (deliberately split out so the two surfaces can keep
+their distinct UX):
+
+* Compute ``expected_total`` (file-unit bar length) — the wizard sums it
+  across multiple paths via :func:`_collect_seed_scale`; ``mm index`` passes
+  the single-path count.
+* Catch :class:`KeyboardInterrupt` for the resume hint (``mm index <path>``
+  vs ``mm web`` Reindex All — different copy per surface).
+* Print the final summary line — ``mm index`` mirrors the legacy
+  ``Indexed N file(s): …`` shape (stable for scripts that grep the output);
+  the wizard prints a green "Seeded initial index" line plus zero-chunks
+  warning. Different shapes, both want the same aggregate counters.
+
+Helper guarantees: bar is always closed on exit (including raise), stream
+runs serially over the supplied ``paths``, returned aggregate dict has
+stable keys ``total_files``, ``indexed``, ``skipped``, ``deleted``,
+``total_chunks``, ``duration_ms``, ``errors``."""
+
+from __future__ import annotations
+
+import time
+from collections.abc import Sequence
+from pathlib import Path
+from typing import Any
+
+import click
+
+
+def _collect_seed_scale(memory_dir: Path) -> tuple[int, int]:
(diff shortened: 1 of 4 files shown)


memtomem (approved):
Solid extraction — bar lifecycle is sound (helper's `finally: _close_bar()` covers KbInt cleanup before propagation), chunk-event no-advance contract is pinned by test (`bar.pos == 1` after 4 chunk + 1 progress event), summary-line shape preserved verbatim for grep-stable scripts.

Two follow-ups filed (separate issues) — non-blocking:
1. `namespace` conditional kwarg shim in `run_with_progress` is a stub-compat workaround for wizard test fakes that predate the kwarg. Should be cleaned up by updating the fakes to accept it.
2. `_collect_seed_scale` only counts `.md` files for bar length — fine for the wizard (markdown memos), but `mm index ./src/` or `mm index ./docs/api.json` gets no bar at all. Engine-provided count would fix both this and the duplicated FS walk.

Borderline rule-of-three (two Python callers, JS as referenced third) is the right judgment call — both implementations would otherwise land identical in this cycle and the contract is well-specified by the chunk_progress event shape.

Approving.

PR #745, opened 2026-05-03
fix(init): wrap _step_memory_dir / _step_settings filesystem ops in fail_step (#664)

### LINKED ISSUE #664: Wizard step audit: `_step_memory_dir` and `_step_settings` crash on filesystem failure (refs #626)
## Summary

Follow-up to #626. The Ollama branch of `_step_embedding` was migrated to the new `fail_step()` mechanism in #661, but two other interactive steps still perform filesystem side effects without any try/except. They don't silently advance (the original #626 bug pattern) — they **crash with an uncaught stack trace**, which is the symmetric failure mode and equally surprising for the user.

### 1. `_step_memory_dir` (`packages/memtomem/src/memtomem/cli/init_cmd.py:638`)

```python
if not memory_path.exists():
    create = nav_confirm(f"  '{memory_dir}' doesn't exist. Create it?", default=True)
    if create:
        memory_path.mkdir(parents=True, exist_ok=True)   # <-- uncaught
        click.secho(f"  Created {memory_path}", fg="green")
```

If `mkdir` fails (PermissionError, OSError, ENOSPC), the wizard exits with a Python stack trace. The user has to read the trace, fix the underlying issue, then re-run `mm init` from the top.

**Suggested fix**: wrap the `mkdir` in a `while True: try / except (PermissionError, OSError) as exc: fail_step(f"Could not create '{memory_path}': {exc}", retryable=True)` block so the user can fix permissions / pick a different path / quit.

### 2. `_step_settings` (`packages/memtomem/src/memtomem/cli/init_cmd.py:889-894`)

```python
canonical = project_root / CANONICAL_SETTINGS_FILE
canonical.parent.mkdir(parents=True, exist_ok=True)   # <-- uncaught
if not canonical.exists():
    canonical.write_text(                              # <-- uncaught
        json.dumps({"hooks": {}}, indent=2) + "\n",
        encoding="utf-8",
    )
```

Same pattern: `mkdir` and `write_text` can both raise. Crashes on read-only project root, full disk, etc.

**Suggested fix**: same shape as above — try/except + `fail_step` with retryable=True.

## Out of scope (intentionally)

The "warn-and-save" patterns in `_step_embedding` (ONNX/Ollama/OpenAI client missing, kiwipiepy missing in `_step_language`) are documented UX — the wizard saves the config so the user can install the extra later. Those are not bugs.

## Why a separate PR

Per the project's "one focused PR per change" convention, the Ollama branch fix in #661 deliberately did not touch other steps. Once this PR lands, the `fail_step` adoption is consistent across every step that performs a side effect.

## Acceptance

- `mkdir` / `write_text` failures in `_step_memory_dir` and `_step_settings` surface a `fail_step` prompt instead of a stack trace.
- Tests pin: simulate `mkdir` raising `PermissionError` → assert the prompt appears, `r` retries the action, `b` returns to the previous step, `q` cancels via `WizardCancel` (exit 0).

pandas-studio:
## Summary

- Both `_step_memory_dir` and `_step_settings` performed filesystem side effects (`mkdir`, `write_text`) without try/except. A read-only project root, full disk, or `PermissionError` would crash `mm init` with an uncaught stack trace and force the user to restart from the top.
- Wraps the failing ops in the standard `fail_step` retry/back/quit pattern adopted by the Ollama branch in #661 — symmetric counterpart to that fix.
- Pins the new behavior with six tests (three per step) shaped after the existing `TestStepEmbeddingOllamaGuards`: `q` cancels via `WizardCancel`, `r` retries the inner op without re-prompting earlier inputs, and a failed op does NOT silently advance to the next step.

The nested try is required because `fail_step` raises `StepRetry` from inside the `except OSError` handler — a sibling `except StepRetry` on the same statement would not catch it. Comment in code explains this.

Closes #664. Out-of-scope per the issue body: the warn-and-save patterns in `_step_embedding` (ONNX/Ollama/OpenAI client missing) and `_step_language` (kiwipiepy missing) are documented UX, not bugs.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_init_cmd.py -k "TestStepMemoryDirFsGuards or TestStepSettingsFsGuards"` — 6 passed
- [x] `uv run pytest -m "not ollama"` — 3997 passed, 7 skipped
- [x] `uv run mypy packages/memtomem/src/memtomem/cli/init_cmd.py` — no issues

### DIFF
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -659,8 +659,28 @@ def _step_memory_dir(state: dict) -> None:
     if not memory_path.exists():
         create = nav_confirm(f"  '{memory_dir}' doesn't exist. Create it?", default=True)
         if create:
-            memory_path.mkdir(parents=True, exist_ok=True)
-            click.secho(f"  Created {memory_path}", fg="green")
+            # Symmetric counterpart to the Ollama-branch fail_step adoption
+            # in ``_step_embedding`` (#626 / #661): an uncaught
+            # PermissionError / OSError here would crash the wizard with a
+            # raw stack trace and force ``mm init`` from the top. Wrap the
+            # mkdir so the user gets the same retry/back/quit recovery as
+            # the rest of the wizard. The nested try is required because
+            # ``fail_step`` raises ``StepRetry`` from inside the
+            # ``except OSError`` handler — a sibling ``except StepRetry``
+            # on the same statement would not catch it. (#664)
+            while True:
+                try:
+                    try:
+                        memory_path.mkdir(parents=True, exist_ok=True)
+                    except OSError as exc:
+                        fail_step(
+                            f"Could not create '{memory_path}': {exc}",
+                            retryable=True,
+                        )
+                    click.secho(f"  Created {memory_path}", fg="green")
+                    break
+                except StepRetry:
+                    continue
     state["memory_dir"] = memory_dir
     click.echo()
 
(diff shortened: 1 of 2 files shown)


memtomem (approved):
LGTM

PR #746, opened 2026-05-03
fix(web): resolve memory_dir paths in /api/sources (#675)

### LINKED ISSUE #675: /api/sources memory_dir not resolved (mismatches /api/memory-dirs/status — #668 follow-up)
## What

`/api/sources` returns each source's `memory_dir` as the **raw configured path**, while `/api/memory-dirs/status` returns the **resolved path** (after `Path.resolve()`).

Repro on macOS with a memory_dir registered under `/tmp` (which symlinks to `/private/tmp`):

```
$ curl -s http://127.0.0.1:8090/api/sources?limit=1 | jq -r '.sources[0].memory_dir'
/tmp/mm-cold-load-home/.claude/projects/.../memory

$ curl -s http://127.0.0.1:8090/api/memory-dirs/status | jq -r '.dirs[0].path'
/private/tmp/mm-cold-load-home/.claude/projects/.../memory
```

The two prefixes (`/tmp` vs `/private/tmp`) come from the same on-disk directory but read as distinct strings, breaking client-side dictionary lookups.

## Where it bites

`packages/memtomem/src/memtomem/web/static/app.js` builds `STATE.memoryStatusByPath` keyed off `/api/memory-dirs/status` `path` values, then looks up `STATE.memoryStatusByPath[source.memory_dir]` to derive the vendor (`status.provider`) for a given source. With the prefix mismatch the lookup misses, so vendor inference fails — fallbacks downstream end up rendering the source under whatever sub-tab is currently active.

This was caught in #674's verification fixture (which uses `/tmp/` as the isolated `HOME`); switching the fixture to `/private/tmp/` made every test pass. The typical real-`HOME` deployment (`/Users/<name>/...` on macOS, `/home/<name>/...` on Linux) doesn't trigger any symlink resolution, so the bug stays invisible — but a user who registers a memory_dir under any symlinked prefix would hit it silently.

## Why this is a #668 follow-up

#668 normalized `/api/memory-dirs/status` to resolve `path`. The same treatment never reached `/api/sources`. This issue tracks bringing the two endpoints into agreement so both surfaces use post-resolve paths.

## Suggested fix

Apply the same `path.resolve()` (or the equivalent helper #668 introduced) to the `memory_dir` field in the `/api/sources` response. Add a parity test: take a memory_dir under a symlinked prefix and assert the two endpoints' paths match.

## Severity

Low — invisible on the dominant real-`HOME` deployments. Surfaces in CI fixtures, Docker mounts, or any user who registers a memory_dir via a symlinked path. Not blocking — listed as out-of-scope in #674's PR body.

pandas-studio:
## Summary

- Symmetric one-line fix to #668: `/api/sources` was returning `memory_dir` as the raw configured path, while `/api/memory-dirs/status` returned the resolved form. The two endpoints feed the same frontend render pass (`STATE.memoryStatusByPath` keyed by status `path`, looked up against sources `memory_dir`), so the prefix mismatch — `/tmp` vs `/private/tmp` on macOS, any Docker bind mount, any user-registered symlink — silently broke vendor inference and rendered sources under whichever sub-tab happened to be active.
- Wizard-written config never goes through `/api/memory-dirs/add` (which already resolves before persisting), so the bug was invisible on real-HOME deployments and only surfaced through CI fixtures / PR #674's `/tmp`-rooted verification harness.
- Three regression tests in `TestSources` mirror the #668 test shape: symlink-resolve, tilde-resolve, and a cross-endpoint parity guard asserting `/api/sources` `memory_dir` == `/api/memory-dirs/status` `path`.

The fixture's `_Indexing`-singleton reset is extended to also reset `memory_dirs` (it already reset `exclude_patterns` for the same reason). Without that, the new TestSources mutations leak into TestIndex downstream and trip its path-inside-memory_dirs gate (403). The existing TestMemoryDirsStatus tests didn't expose this because they sit *after* TestIndex in collection order; my new tests sit *before* it.

Closes #675.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_web_routes.py::TestSources` — 9 passed (3 new)
- [x] `uv run pytest packages/memtomem/tests/test_web_routes.py` — 118 passed
- [x] `uv run pytest -m "not ollama"` — 3994 passed, 7 skipped
- [x] `uv run ruff check` + `ruff format --check` on changed files
- [x] `uv run mypy packages/memtomem/src/memtomem/web/routes/sources.py` — no issues

### DIFF
--- a/packages/memtomem/src/memtomem/web/routes/sources.py
+++ b/packages/memtomem/src/memtomem/web/routes/sources.py
@@ -44,9 +44,17 @@ async def list_sources(
     # first match in the inner loop is the longest-prefix-wins one —
     # matches :func:`resolve_owning_memory_dir`'s tie-break rule for
     # nested configured dirs without repeating the comparison logic.
+    #
+    # ``dir_path`` is resolved (not just expanded) so the response
+    # ``memory_dir`` matches sibling ``/api/memory-dirs/status`` paths
+    # — both produce ``Path(d).expanduser().resolve()`` strings. Without
+    # this, frontend ``STATE.memoryStatusByPath[source.memory_dir]``
+    # lookups miss whenever a memory_dir is registered under a symlinked
+    # prefix (macOS ``/tmp`` → ``/private/tmp``, Docker bind mounts).
+    # Same one-line treatment as #668 / engine.py:memory_dir_stats. (#675)
     indexed_dirs: list[tuple[str, Path, MemoryDirKind]] = sorted(
         (
-            (norm_dir_prefix(d), Path(d).expanduser(), memory_dir_kind(d))
+            (norm_dir_prefix(d), Path(d).expanduser().resolve(), memory_dir_kind(d))
             for d in config.indexing.memory_dirs
         ),
         key=lambda t: -len(t[0]),
--- a/packages/memtomem/tests/test_web_routes.py
+++ b/packages/memtomem/tests/test_web_routes.py
@@ -202,8 +202,14 @@ def app():
     application.state.index_engine = index_engine
     cfg = FakeConfig()
     # _Indexing is a class-level singleton — reset mutable fields so tests that
-    # mutate exclude_patterns don't leak into later tests.
+    # mutate exclude_patterns or memory_dirs don't leak into later tests.
+    # The memory_dirs reset matters for any test that reassigns the list to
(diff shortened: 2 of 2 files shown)


memtomem (approved):
LGTM

memtomem (approved):
LGTM

PR #747, opened 2026-05-03
test(embedding): wait on Event for ONNX progress flush instead of single sleep(0) (#663)

### LINKED ISSUE #663: Flaky macOS CI: `test_embedding_progress.py` ONNX progress callbacks intermittently fire 0 calls
## Summary

Tests in `packages/memtomem/tests/test_embedding_progress.py` (introduced in #653) flake intermittently on the macOS GitHub runner with Python 3.14. The `on_progress` callbacks expected during streaming ONNX embedding occasionally fire **0 times**, causing assertions like `expected 5 calls, got []` and `expected ~20 throttled ticks for n=200, got 0`. The same tests pass reliably on `ubuntu-latest` and on local macOS.

## Receipts

- **PR #661 (3 consecutive macOS failures):**
  - Run 1 — `test_onnx_streams_progress_per_yield[asyncio]` — `expected 5 calls, got []` (job 73922637345)
  - Run 2 (same job, rerun) — `test_onnx_streams_progress_per_yield[asyncio]` — same assertion (job 73923291156)
  - Run 3 (rerun again) — `test_onnx_throttles_thread_hops_for_large_input[asyncio]` — `expected ~20 throttled ticks for n=200, got 0: []` (job 73923719020)
- **`main` branch run on `e034d10`** (the very PR that introduced these tests, #653) — same family of macOS failures including `test_onnx_throttles_thread_hops_for_large_input[asyncio]`.
- Subsequent main runs (`c685a3a`, `46e1a1a`) passed — confirming the issue is intermittent rather than a hard regression.

## Hypothesis

The ONNX streaming embedder uses a `ThreadPoolExecutor` with periodic `asyncio.sleep(0)` thread hops to dispatch progress callbacks. On the macOS GH runner with Python 3.14, the asyncio scheduler appears to drain the thread's yield batches before any callback dispatches, so every `on_progress` call lands after the test's collection phase has already returned. The test sees `len(calls) == 0` even though the embedder produced the expected output.

This is consistent with:
- macOS-only — ubuntu does not flake
- Python 3.14-only on CI (local Python 3.13 macOS is fine)
- Two different tests in the same file fail with the same `0 calls` shape

## Impact

PRs whose diff has nothing to do with embedding/progress code get blocked by a required check. PR #661 (wizard guardrails) was admin-merged with a bypass-justification comment after three consecutive failures of this test family. Without a fix, every PR will need the same bypass dance.

## Suggested next step

Either:
1. Tighten the test's wait-for-callback semantics (await an `asyncio.Event` set inside the callback rather than asserting on `len(calls)` after a `gather` returns), or
2. Pin `fastembed` / ONNX runtime to a known-good combination on macOS, or
3. Add a retry decorator scoped to this test file so the GH runner's scheduler has more chances.

Probably (1) — the assertion is racing the callback dispatch loop. cc @ author of #653 for context.

pandas-studio:
## Summary

- `test_onnx_streams_progress_per_yield` and `test_onnx_throttles_thread_hops_for_large_input` were flaking on the macOS GH runner with Python 3.14 — `on_progress` callbacks intermittently registered as **0 calls** (issue body's three consecutive failures on PR #661 reruns + the same shape on `main` at `e034d10`).
- The existing flush `await asyncio.sleep(0)` yields exactly **one** loop iteration. That was enough on Linux and on macOS Py3.13 but not on Py3.14's macOS runner — the `loop.call_soon_threadsafe` items the embedding worker scheduled were still pending in the loop's ready queue when `await asyncio.to_thread` resumed and the assertion ran.
- Replace with awaiting an `asyncio.Event` set inside the callback when `done == total`. FIFO ordering of `call_soon_threadsafe` guarantees that once the final-tick callback runs, every earlier callback has already run — count assertions no longer race the queue regardless of how many loop iterations the drain takes. This is the issue body's suggestion (1).

Diagnostic-friendly failure mode: if no callback ever fires (the worst-case version of the original flake), `wait_for(timeout=5.0)` raises `TimeoutError` instead of silently failing the count assertion at 0.

Production behavior unchanged — `OnnxEmbedder.embed_texts`'s `_thread_cb` / `call_soon_threadsafe` pipeline is fire-and-forget and untouched. This PR only tightens the test harness's wait semantics. Closes #663.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_embedding_progress.py -v` — 12 passed
- [x] 30× stress loop on the two ONNX tests — 30/30 clean (local macOS Py3.13)
- [x] `uv run pytest -m "not ollama"` — 3991 passed, 7 skipped
- [x] `uv run ruff check` + `ruff format --check` on the test file
- [ ] Validation on the macOS Py3.14 GH runner once CI runs — the actual flake host

### DIFF
--- a/packages/memtomem/tests/test_embedding_progress.py
+++ b/packages/memtomem/tests/test_embedding_progress.py
@@ -59,6 +59,33 @@ def _record() -> tuple[list[tuple[int, int]], callable]:
     return calls, cb
 
 
+def _record_with_done_event() -> tuple[list[tuple[int, int]], callable, asyncio.Event]:
+    """Like :func:`_record`, but also returns an ``asyncio.Event`` set
+    when a callback fires with ``done == total`` — i.e. the final tick.
+
+    Used by the ONNX tests to wait for trailing
+    ``loop.call_soon_threadsafe`` callbacks to drain off the loop's
+    ready queue before asserting on call count. ``asyncio.to_thread``
+    can resume its awaiter before the worker thread's queued
+    callbacks dispatch — a single ``await asyncio.sleep(0)`` yields
+    only one loop iteration and the macOS GH runner under Python 3.14
+    has been observed to need more (#663). Awaiting an Event set by
+    the final callback is robust regardless of how many iterations
+    the loop needs to fully drain the FIFO queue: once the
+    ``done == total`` callback runs we know every earlier callback
+    has already run (FIFO ordering of ``call_soon_threadsafe``).
+    """
+    calls: list[tuple[int, int]] = []
+    done_event = asyncio.Event()
+
+    def cb(done: int, total: int) -> None:
+        calls.append((done, total))
+        if done == total:
+            done_event.set()
+
+    return calls, cb, done_event
+
+
 def _assert_progress_contract(calls: list[tuple[int, int]], total: int, expected_calls: int):
     assert len(calls) == expected_calls, f"expected {expected_calls} calls, got {calls}"
     # All totals match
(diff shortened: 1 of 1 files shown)


memtomem (approved):
LGTM

memtomem (approved):
LTMG

PR #754, opened 2026-05-03
fix(tests): unblock Windows pytest cleanup of components fixture (#206)

### LINKED ISSUE #206: Windows: test components fixture leaves SQLite/embedder handles open, blocking pytest cleanup
## Context

Discovered by @powerzist during work on #162 (Windows compatibility). Filing here per their request so the underlying handle leak gets tracked separately from the PR-level scope split.

## Symptom

On Windows, several tests fail with `FileExistsError` when trying to create directories under `tmp_path`:

```
self = WindowsPath('C:/Users/user/AppData/Local/Temp/pytest-of-user/pytest-30/test_happy_path_indexes_codex_0/memories')
mode = 511, parents = False, exist_ok = False

    def mkdir(self, mode=0o777, parents=False, exist_ok=False):
        try:
>           os.mkdir(self, mode)
E           FileExistsError: [WinError 183] 파일이 이미 있으므로 만들 수 없습니다:
            'C:\\Users\\user\\AppData\\Local\\Temp\\pytest-of-user\\pytest-30\\
             test_happy_path_indexes_codex_0\\memories'
```

Reproduced at `packages/memtomem/tests/test_cli_ingest.py:551` and `:581`.

Environment from the report:
- Windows 10/11
- `cpython-3.13.5-windows-x86_64` (uv-managed)

## Powerzist's observation

> The temporary directory system creates up to three directories, then starts reusing them. When reusing a directory, it attempts to delete the existing contents before creating it again. However, on Windows, some files inside the directory may still be locked, which can cause the deletion step to fail. As a result, the directory remains in place, and a `FileExistsError` is raised when attempting to recreate it.
>
> In my environment, I consistently observed that only up to three directories were present under the temporary path, which appears to align with this reuse behavior.

This matches pytest's default `tmp_path_retention_count=3` — pytest reuses tmp directory roots across sessions, and if the previous session's `rmtree` cleanup fails, the next run sees the leftover. On Linux/macOS `rmtree` generally succeeds even while files are still open; Windows blocks on any open handle.

## Hypothesis (root cause upstream of `mkdir`)

The `components` fixture (`packages/memtomem/tests/conftest.py:41-71`) tears down via `close_components()`, which only closes `storage`, `embedder`, `search_pipeline`, and `llm`. If any of those retain an OS-level handle after `await …close()` returns, pytest's subsequent cleanup of the tmp dir fails on Windows, and the leftover `memories/` directory is the downstream symptom.

Likely culprits to investigate first:

- **SQLite WAL/SHM sidecars** — `test.db-wal` / `test.db-shm` may still be mapped after the connection closes if WAL checkpoint hasn't run.
- **ONNX Runtime mmap of the model file** — the embedder may keep an mmap handle on the ONNX session that isn't released by `await embedder.close()`.

A blanket `exist_ok=True` at the call sites (the path #162 originally took) suppresses the symptom but hides the leak — and the leftover directory may carry a stale `test.db` / `test.db-wal` from the previous run, which could cause quieter flakes later that are much harder to bisect.

## Suggested investigation path

1. Verify on Windows: after `close_components()` returns, are there still open handles on `storage._db`, `embedder.session`, etc.? (Windows `handle.exe` from Sysinternals works for this.)
2. Audit `Storage.close()`, `Embedder.close()`, `LLMProvider.close()` for missing finalization (WAL checkpoint, ONNX session.release, file handle close).
3. If a third-party library holds a C-level lock that can't be released, fall back to **site-targeted** `exist_ok=True` on only the demonstrated paths (`conftest.py:48`, `test_cli_ingest.py:551`, `test_cli_ingest.py:581`) with a comment pointing at the upstream cause — not a blanket apply across sites that don't hold handles.

## Acceptance

- Tests under the `components` fixture run cleanly on Windows across consecutive sessions (no `FileExistsError` from leftover `memories/` dirs).
- If a fix is applied at the source (`close_components()` or one of the `.close()` methods), no `exist_ok=True` is needed at call sites.
- If a site-targeted fallback is used, the comment at each site explains *why* (handle leak in component X) and not just *what* (allow reuse).

## Related

- PR #162 — original Windows-compat PR; scope reduced to `encoding="utf-8"` only after this issue was filed.

Credit: investigation and traceback by @powerzist.

memtomem:
## Summary

- `OnnxEmbedder.close()` and `FastEmbedReranker.close()` now `gc.collect()` after dropping `self._model`, forcing the underlying ORT `InferenceSession` (mmap on the model file + thread-local arenas) to release promptly. Best-effort handle hygiene.
- Three `components`-style fixtures (`conftest.py`) and two `test_cli_ingest` Codex tests now `mkdir(exist_ok=True)` (with `shutil.rmtree(..., ignore_errors=True)` first where `.md` content gets written), so reused `tmp_path` roots from a prior failed-cleanup session no longer trip `FileExistsError [WinError 183]`. Each site has an inline `#206` reference.

`Storage.close()` already runs `PRAGMA wal_checkpoint(TRUNCATE)` and closes the read-pool — no change needed there.

Closes #206. Original report by @powerzist during #162.

## Why both source-level and site-targeted

The issue's acceptance bullets allow either approach:
- If source fix is sufficient: no `exist_ok=True` needed at sites.
- If site-targeted: each site needs a comment pointing at upstream cause.

Neither is fully reliable on its own under Windows file-locking semantics (handles get released when the previous Python process dies, but Windows takes a moment; AV/indexing services can also briefly hold scans). The site-targeted change is the load-bearing one — without it, any future regression in a `.close()` method silently re-introduces the failure. The source-level change makes it less likely the leftover ever happens in the first place. Defense in depth, with comments at every site so neither layer becomes invisible.

## Test plan

- [x] `uv run pytest -m "not ollama" packages/memtomem/tests` — 4012 passed, 7 skipped, 46 deselected
- [ ] Windows verification — pinging @powerzist who originally reported the leak in their environment, since this repo doesn't run a Windows CI matrix (#634)

### DIFF
--- a/packages/memtomem/src/memtomem/embedding/onnx.py
+++ b/packages/memtomem/src/memtomem/embedding/onnx.py
@@ -205,4 +205,12 @@ class OnnxEmbedder:
         return embeddings[0]
 
     async def close(self) -> None:
+        # Force-collect after dropping the fastembed reference: the underlying
+        # ORT InferenceSession holds an mmap on the model file plus thread-local
+        # arenas, and on Windows that handle can outlive ``self._model = None``
+        # long enough for pytest's tmp_path rmtree to fail with WinError 183
+        # on the next session. See #206.
+        import gc
+
         self._model = None
+        gc.collect()
--- a/packages/memtomem/src/memtomem/search/reranker/fastembed.py
+++ b/packages/memtomem/src/memtomem/search/reranker/fastembed.py
@@ -109,4 +109,10 @@ class FastEmbedReranker:
         ]
 
     async def close(self) -> None:
+        # Same shape as ``OnnxEmbedder.close``: force-collect so the underlying
+        # ORT InferenceSession releases its mmap and thread-local arenas before
+        # pytest cleans up tmp_path on Windows. See #206.
+        import gc
+
         self._model = None
+        gc.collect()
--- a/packages/memtomem/tests/conftest.py
+++ b/packages/memtomem/tests/conftest.py
@@ -101,7 +101,10 @@ async def components(tmp_path):
 
     db_path = str(tmp_path / "test.db")
     mem_dir = str(tmp_path / "memories")
-    (tmp_path / "memories").mkdir()
+    # exist_ok: pytest reuses tmp_path roots; on Windows a previous session's
+    # SQLite/ONNX handle leak can leave this dir behind and trip WinError 183.
+    # See #206.
+    (tmp_path / "memories").mkdir(exist_ok=True)
 
     os.environ["MEMTOMEM_STORAGE__SQLITE_PATH"] = db_path
     os.environ["MEMTOMEM_INDEXING__MEMORY_DIRS"] = json.dumps([mem_dir])
@@ -170,7 +173,8 @@ async def bm25_only_components(tmp_path, monkeypatch):
     """
     db_path = tmp_path / "bm25.db"
     mem_dir = tmp_path / "memories"
-    mem_dir.mkdir()
+    # exist_ok: see ``components`` fixture above — Windows tmp_path reuse. #206.
+    mem_dir.mkdir(exist_ok=True)
 
     from helpers import isolate_memtomem_env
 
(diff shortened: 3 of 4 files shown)


memtomem:
@powerzist — could you give this a try in your Windows environment when you have a chance? You filed the original report in #206, and since the repo doesn't run a Windows CI matrix yet (#634), your reproduction is the only way to confirm the `FileExistsError [WinError 183]` actually goes away across consecutive pytest sessions. The relevant tests are `test_cli_ingest.py::TestCodexIngestIntegration` (skipped without Ollama, but the `components` fixture itself is exercised by most of the suite). No rush — happy to wait.

PR #807, opened 2026-05-05
fix(web): show real chunk count on search drag-zone upload toast

memtomem:
## Summary

- The search-tab drag-zone success toast at `app.js:5725` aggregated chunks via `(data.results || []).reduce(...)`, but `/api/upload` returns `{files, total_indexed}` and has never exposed a `results` field (see `UploadResponse` in `web/routes/system.py`). The fallback to `[]` silently rendered "Indexed N files → 0 chunks" since v0.1.0 even when the backend reported a non-zero indexed count.
- Fix: read from `data.files` — same shape the upload-tab caller already uses for its merged response.
- Caught while reviewing PR #805 — the diff didn't introduce the bug but did wrap the affected branch in a new partial-suppression guard, which made the stale field name visible. Splitting into its own PR because #805 had already landed and the scope is unrelated to the narrowed-retry change there.
- Cache-bust `app.js?v=109 → v=110` so cached SPAs reload the fixed toast (per `feedback_static_asset_cache_bust.md`).

## Test plan

- [x] New vitest spec `packages/memtomem/tests-js/search-drag-zone-toast-chunks.test.mjs` — boots the JS under JSDOM, dispatches a `drop` on `#tab-search` with two stub files (3 + 4 chunks), asserts the success toast reads "Indexed 2 files → 7 chunks" not 0.
- [x] **Mutation-validated**: reverting the fix to `(data.results || [])` fails the spec with the literal `"Indexed 2 files → 0 chunks"` text — this is the toast that users have actually been seeing since v0.1.0.
- [x] `cd packages/memtomem/tests-js && npm test` — 19/19 vitest specs green (1 new + 18 existing).
- [x] `uv run ruff check packages/memtomem/src && uv run ruff format --check packages/memtomem/src` — clean.

## Notes

- No i18n change — the locale string `toast.indexed_files_chunks` is unchanged; only the JS that fills its `{chunks}` placeholder.
- The upload-tab caller (Index → Upload mode) was always correct because it uses `data.total_indexed` directly. Only the search-tab drag-zone path was affected.
- Net diff: +98 / −2 (mostly the new test file).

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -5722,7 +5722,11 @@ qs('group-toggle').addEventListener('click', () => {
         // followed by a contradicting "indexed N files" message.
         const partial = upload.blockedFileCount > 0 && !upload.bypassed;
         if (!partial) {
-          const chunks = (data.results || []).reduce((s, r) => s + (r.indexed_chunks || 0), 0);
+          // ``/api/upload`` returns ``{files, total_indexed}`` — never a
+          // ``results`` field. The original ``data.results`` read fell
+          // back to ``[]`` so this toast has been showing "indexed N
+          // files (0 chunks)" since v0.1.0.
+          const chunks = (data.files || []).reduce((s, r) => s + (r.indexed_chunks || 0), 0);
           showToast(t('toast.indexed_files_chunks', { files: files.length, chunks }), 'success');
         }
       }
--- a/packages/memtomem/src/memtomem/web/static/index.html
+++ b/packages/memtomem/src/memtomem/web/static/index.html
@@ -1335,7 +1335,7 @@
   <script src="/vendor/prism-bash.min.js?v=1"></script>
   <script src="/vendor/prism-yaml.min.js?v=1"></script>
   <script src="/i18n.js?v=3"></script>
-  <script src="/app.js?v=109"></script>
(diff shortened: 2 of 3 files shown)


memtomem:
Self-review pass on the two P2 cancel-refresh notes — both look like they describe pre-#805 behavior that's already on `main`. Documenting disposition so reviewers don't chase ghosts:

**P2.1 — Upload-tab cancel refresh (`app.js:4159-4161`)**
Lines 4159-4166 are the explanatory comment, not the bug site. The cancel branch at 4167-4181 has no `return`; control falls through to `_markDataStale(); loadSourceFilter(); loadStats(); loadUploadUsage();` at 4205-4208. PR #805 (`505589e`) removed the old `if (upload.cancelled) return;` and reshaped this into the unified-refresh structure deliberately.

**P2.2 — Search-tab drag cancel refresh (`app.js:5684`)**
Line 5684 is the dragenter early-return for non-file drags, unrelated to upload cancel. The drop handler's cancel branch (5719-5721) also falls through to `_markDataStale(); loadSourceFilter(); loadStats();` at 5736-5738 — same #805 hunk (`@@ -5598,10 +5706,26 @@`) rewrote both paths in one go.

If I'm misreading either path, point me at the specific line that early-returns and I'll re-investigate. Otherwise this PR's scope stays narrowed to the chunk-count toast fix.

PR #811, opened 2026-05-05
fix(web/a11y): role=button + aria-label on search/source/timeline rows

### LINKED ISSUE #700: List-row a11y: search / sources / timeline rows lack role / aria-label
## Summary

Search results, source-detail chunk cards, and Timeline entries are rendered as bare `<div>` containers without semantic roles or accessible names. Keyboard / screen-reader users encounter them as anonymous regions. The only list-row in the app that exposes a button role is `home-source-item` on the Home tab.

Found by Playwright UX review of v0.1.34 prod (2026-05-02). See `docs/reports/mm-web-prod-v0.1.34-playwright-review.md` (P2 — Accessibility labels are weak for result/list items).

## Evidence

Across `packages/memtomem/src/memtomem/web/static/app.js` and `index.html`:

- The only `role="button"` applied to a list-row is on `home-source-item` (`index.html:1310`):
  ```html
  <div class="home-source-item home-recent-item" tabindex="0" role="button" ...>
  ```
- `role="list"` and `role="listitem"` are not used anywhere in the static surface.
- Search results (`.result-item`), source detail chunk cards, and Timeline rows are constructed as plain `<div>` containers with no `role`, `aria-label`, or `aria-labelledby`.

## Why this matters

- Screen-reader users land on selectable rows that announce as "group" or generic landmarks instead of "button: <source>:<line range>".
- Keyboard users can't tell from the focus ring whether a row is interactive.
- Browser automation (and future a11y tests) has to lean on brittle text matching instead of semantic queries.

## Suggested fix (incremental — does not have to be one PR)

For each list surface, decide between two patterns:

1. **`role="list"` + `role="listitem"`** when the row is informational only. Add `aria-label` describing what each row contains (e.g. `aria-label="${basename}, lines ${start}-${end}, ${namespace}"`).
2. **`role="button"` + `tabindex="0"`** when the entire row is clickable / activates an action. Match the existing `home-source-item` pattern.

Targets:

- Search results list (`#results-list .result-item`)
- Source detail chunk cards (find via grep for chunk card construction in `app.js`)
- Timeline entries (Timeline tab list rows)

Generate accessible names from the same source/line/namespace data that's already on the row.

## References

- Tracking umbrella: TBD

memtomem:
## Summary

Closes #700. Brings list-row a11y semantics to the rest of the static surface — `home-source-item` was the only row exposing `role=button` before this PR.

Five row classes now ship with `role="button"`, `tabindex="0"`, `aria-label`, and an Enter/Space keydown handler that mirrors the existing click handler:

| Surface | File | Notes |
| --- | --- | --- |
| `.result-item` | `app.js` `_buildResultItem` | role + aria-label new; keydown was already there |
| `.chunk-card` | `app.js` `browseSource` | aria-label always; role/tabindex/keydown only on the collapsible branch (`scrollHeight > 120`) |
| `.timeline-item` | `timeline.js` `renderChunkView` | full quad |
| `.timeline-file-item` | `timeline.js` `renderFileView` outer | full quad; keydown ignores Enter inside `.tl-file-open-source` / `.tl-file-chunk-item` so nested controls keep their behavior |
| `.tl-file-chunk-item` | `timeline.js` `renderFileView` inner | full quad; click and keydown both `e.stopPropagation()` to avoid bubbling into the outer expand toggle |
| `.tl-heatmap-col` | `timeline.js` `renderTimeline` | bar-chart columns get role/tabindex/aria-label + keydown — the chart was previously opaque to assistive tech |

`aria-label` strings are built from data the row already exposes (filename, line range, namespace, time, count). No new strings to translate.

## Tests

`packages/memtomem/tests/test_web_a11y.py` — 15 source-scan pin tests anchored on the construction site of each row class. Source-scan rather than browser tests because the static surface has no JS test harness and the regression class to catch is "a future contributor deletes an attribute," which a literal-string scan handles directly.

Mutation-validated: `git checkout` of `app.js` + `timeline.js` to `origin/main` then re-running the file fails 14 of 15. The surviving pass is `test_result_item_keydown_handler` because the keydown handler already existed on `result-item` pre-PR.

## How to test

```
cd /tmp/mm-i700
uv run pytest packages/memtomem/tests/test_web_a11y.py -v
uv run ruff format --check packages/memtomem/src packages/memtomem/tests
```

A manual sanity pass in `mm web`:
- Tab to a search result → screen-reader should announce `<filename>, lines X-Y, ...`
- Tab to a long chunk-card → Enter/Space should expand/collapse
- Tab to a Timeline row in either chunks or files view → Enter/Space mirrors click
- Tab to a heatmap bar → Enter/Space scrolls to the matching date

## Verification adjustment to issue #700

The issue body claimed `home-source-item` was the only list-row exposing `role=button` and that `result-item` was a bare `<div>`. The first part was correct on `origin/main`; the second was effectively correct (`tabindex` was set but neither `role=button` nor `aria-label` was). Both gaps are closed in this PR.

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -1999,6 +1999,7 @@ function _buildResultItem(r) {
   item.className = 'result-item';
   item.dataset.id = r.chunk.id;
   item.setAttribute('tabindex', '0');
+  item.setAttribute('role', 'button');
 
   const checkLabel = document.createElement('label');
   checkLabel.className = 'result-check-wrap';
@@ -2017,6 +2018,14 @@ function _buildResultItem(r) {
   const fname = basename(r.chunk.source_file || '');
   const dir = shortDir((r.chunk.source_file || '').split('/').slice(0, -1).join('/') || '/');
   const age = relativeTime(r.chunk.created_at);
+  const lineRange = (r.chunk.start_line && r.chunk.end_line)
+    ? `lines ${r.chunk.start_line}-${r.chunk.end_line}`
+    : null;
+  const nsLabel = r.chunk.namespace && r.chunk.namespace !== 'default'
+    ? `namespace ${r.chunk.namespace}`
+    : null;
+  const ariaParts = [fname, lineRange, nsLabel, age].filter(Boolean);
+  item.setAttribute('aria-label', ariaParts.join(', '));
   const nsBadge = r.chunk.namespace && r.chunk.namespace !== 'default'
     ? ` <span class="badge badge-ns">${escapeHtml(r.chunk.namespace)}</span>` : '';
   const validityBadge = _validityBadgeHtml(r.chunk.valid_from_unix, r.chunk.valid_to_unix);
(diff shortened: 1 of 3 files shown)


memtomem:
Superseded by the three split PRs (#808 #809 #810) created earlier today by a parallel session. The split PRs are CI-green and better-scoped per feedback_one_change_per_pr — keeping those, closing this consolidated version. No content was unique to this PR.

PR #824, opened 2026-05-07
fix(web/ctx): Q-PR1 — Context Gateway dashboard i18n correctness bundle

memtomem:
## Summary

Bundle five audit findings from the PR #816 Context Gateway dashboard
review into one i18n-correctness PR. All five live in `context-gateway.js`
+ `locales/{en,ko}.json` + `index.html`; collapsing them keeps the
review surface focused on a single axis instead of fragmenting into
five follow-ups.

| # | Severity | Item |
|---|---|---|
| Bug-1 | P0 | langchange leaves overview cards in stale locale (inline `t()` in innerHTML; `applyDOM` can't re-walk) — and the multi-toggle race that needs a sequence guard |
| Bug-2 | P0 | `0/0 synced` + green badge on a fresh / un-imported tile (false-OK) |
| Bug-3 | P1 | raw `'Error'` literal in the `d.error` overview branch (narrow path; piggybacked) |
| Drift-2 | P1 | "Detect" button name vs "Refresh" i18n value vs `detection_complete` toast key — drift across button id / i18n / handler / toast |
| Drift-4 | P1 | `parse error` runtime status had no i18n label key — fell through to raw English |

Audit catalog: `~/.claude/plans/context-gateway-ui-hazy-biscuit.md`
(local). Out of scope and intentionally deferred: Drift-1 (sidebar
KO) + Drift-3 (dev-tier desc) → Q-PR2; Visual-1/2/4 → Q-PR3 (Visual-1
needs API shape decisions); Info-1/2/3/4/5 → ADR + 5 issues.

### Bug-1 — langchange + multi-toggle race
- `langchange` listener now calls `loadCtxOverview()` when the section
  is mounted. The cards are inline-templated via `t()` in innerHTML;
  `I18N.applyDOM` only walks `data-i18n*` attributes and can't re-write
  the rendered text on toggle.
- New module-level `_ctxOverviewSeq` counter, captured at the top of
  `loadCtxOverview`. Both the success path (just before
  `el.innerHTML = html`) and the error path (before the empty-state
  overlay) bail when the captured seq is stale, so a slow fetch from
  an earlier toggle cannot clobber the cards rendered by a later one.

### Bug-2 — `0/0 synced` false-OK + Bug-3 — raw `'Error'`
- New `isEmpty = typ.key !== 'settings' && total === 0 && !d.error && !hasIssue`
  branch above the count fallthrough. Picks `badge-gray` and the new
  `settings.ctx.badge_empty` key (`Empty` / `비어 있음`).
- The Settings tile is gated out by `typ.key !== 'settings'` because
  it has its own `_SETTINGS_STATUS_I18N`-driven branch above the count
  ladder; the new isEmpty branch must not intercept it.
- The `d.error` branch is now routed through the existing
  `settings.hooks.badge_error` key (`error` / `오류`); the raw English
  `'Error'` literal is gone.

### Drift-2 — Detect → Refresh
The handler always called `loadCtxOverview()`; "detect" was an alias.
Renamed across every surface so id / i18n / handler / toast match:

| Surface | Before | After |
|---|---|---|
| `index.html:777-780` | `id="ctx-detect-btn"` + `Detect` fallback + `settings.ctx.detect{,_tooltip,_aria}` | `id="ctx-refresh-btn"` + `Refresh` fallback + `settings.ctx.refresh{,_tooltip,_aria}` |
| `context-gateway.js:309-318` | `getElementById('ctx-detect-btn')` + `t('toast.detection_complete')` | `getElementById('ctx-refresh-btn')` + `t('toast.refresh_complete')` |
| `locales/{en,ko}.json` | `detect{,_tooltip,_aria}`, `toast.detection_complete` | `refresh{,_tooltip,_aria}`, `toast.refresh_complete` |
| `tests/test_i18n.py:231` | `"toast.detection_complete"` (required allow-list) | `"toast.refresh_complete"` |

Translated values are unchanged (already "Refresh" / "새로고침" /
"Refresh complete" / "새로고침 완료") — only the key names shift to
match what the button does.

### Drift-4 — parse error i18n
- `_ctxStatusLabel` adds `'parse error': 'settings.ctx.status_parse_error'`.
- New `settings.ctx.status_parse_error` key (`Parse error` / `파싱 오류`).
- Distinct from the existing `settings.ctx.badge_parse_error` ("parse error"
  / "파싱 오류") which is the count-prefix label on the overview tile;
  `status_*` keys are the runtime-badge labels.

### Tests added (10 new)

`packages/memtomem/tests/test_i18n.py` — file-scan regressions:
1. `test_q_pr1_required_keys_present` — both new keys + four renamed
   keys present in en/ko.
2. `test_q_pr1_no_legacy_detect_keys` — old `detect*` keys absent.
3. `test_q_pr1_no_legacy_detect_in_ui_assets` — `ctx-detect-btn` /
   `toast.detection_complete` / `settings.ctx.detect*` absent in
   HTML + JS.
4. `test_q_pr1_overview_badge_routes_error_through_t` — Bug-3 pin:
   `d.error` branch uses `t(...badge_error)`.
5. `test_q_pr1_overview_has_sequence_guard` — `_ctxOverviewSeq`
   counter exists, captured at entry, and the bail check appears in
   both success + catch paths.
6. `test_q_pr1_langchange_listener_reloads_overview` — listener body
   contains `loadCtxOverview()`.
7. `test_q_pr1_status_parse_error_mapped` — `_ctxStatusLabel` maps
   `'parse error'` to `settings.ctx.status_parse_error` (extracts the
   block first so it doesn't confuse with `_ctxStatusCls`).

`packages/memtomem/tests/web/test_context_gateway_overview.py` — 3
Playwright specs:
1. `test_zero_total_renders_empty_badge_not_green_synced` — Bug-2
   pin with positive (`Empty` + `badge-gray`) + symmetric negative
   (no `badge-success`, no `synced` literal). Mutation-validated:
   forcing `isEmpty = false` makes the test fail with
   `'0/0 synced' != 'Empty'`.
2. `test_langchange_rerenders_card_label_text` — Bug-1 single-toggle
   pin (`Skills` → `스킬` after `setLang('ko')`).
3. `test_multi_toggle_race_keeps_newest_locale_on_cards` — Bug-1
   multi-toggle pin: with KO response delayed in the route handler,
   final card text must be the EN locale set by the second toggle.

Limitation: the multi-toggle race window is hard to reproduce
deterministically with sync Playwright route handlers (handlers
serialize on the dispatcher thread, so a delay on one fetch also
delays the next). The static `test_q_pr1_overview_has_sequence_guard`
covers the seq-guard mechanism itself; the Playwright spec covers the
multi-toggle outcome.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_i18n.py` — 31 pass
- [x] `uv run pytest packages/memtomem/tests/web/test_context_gateway_overview.py` — 3 pass (chromium)
- [x] `uv run pytest packages/memtomem/tests -k "web or i18n or context_gateway" -m "not ollama"` — 625 pass, 0 fail
- [x] en/ko key parity (`jq` diff) — empty
- [x] Mutation-validated Bug-2: `isEmpty = false` → spec fails as expected
- [ ] Manual `uv run mm web --port 8089` smoke for Bug-1 (EN/KO toggle on cards) — recommended for reviewer


memtomem:
Follow-up commit 2e042a2 addresses the P2 active-section gate finding. #settings-ctx-overview is always present in the DOM so the prior `qs('ctx-overview-content')` truthy gate fired the overview fetch on every language toggle from any page. The listener now checks `classList.contains('active')` matching `switchSettingsSection`'s contract (app.js:1191). Mutation-validated with two new tests: (a) static guard requires the active-class check; (b) new Playwright spec `test_langchange_off_overview_does_not_refetch_overview` toggles language from the Search tab and asserts zero `/api/context/overview` requests. Bumped `?v=22` → `?v=23`. The other minor follow-ups in your review (panelLoading-on-langchange UX flash + audit of other inline-`t()` regions) deserve a tracking issue rather than scope creep on this PR — happy to file if useful.

memtomem:
Second-pass follow-up commit 79c052f addresses the new P2 finding. Right — `activateTab` only toggles at the tab-panel level (app.js:1046-1049), so leaving Settings for Search keeps `#settings-ctx-overview.active` set and the section-only gate still let off-Settings toggles refetch.

Listener now requires *both* `#tab-settings.active` AND `#settings-ctx-overview.active`. Mutation-validated: forcing `settingsVisible = true` (section-only fallback) makes the new spec fail with one extra `/api/context/overview` per `setLang` invocation.

New Playwright spec `test_langchange_after_tab_switch_does_not_refetch_overview` mounts the dashboard, switches to Search main tab, sanity-pins the precondition (settings tab inactive, section still .active — the exact state the bug depends on), toggles language twice, asserts no refetch. `?v=23` → `?v=24`.

memtomem:
Pass-3 follow-up commit 825e17e addresses both action items:

**Optional (cross-namespace):** Added `settings.ctx.badge_error` (en="error", ko="오류") and switched the d.error branch to it. Symmetric guard now requires the new key AND asserts `settings.hooks.badge_error` absent in the d.error branch — a future "simplify" refactor can't quietly reach back across the namespace. Mutation-validated.

**Required (manual mm web smoke):** Ran end-to-end against `uv run mm web --port 8089` via Playwright MCP. Observations on a real prod-mode dashboard:
- Bug-2: skills/agents tiles render `Empty` (EN) / `비어 있음` (KO) + `badge-gray` for `total === 0` payloads.
- Bug-1: KO toggle re-renders all cards to 스킬 / 서브에이전트 / 훅 + 비어 있음 / 동기화됨.
- Drift-2: `ctx-refresh-btn` exists, `ctx-detect-btn` does not, click fires `새로고침 완료` toast.
- Pass-2 gate: switching to Search main tab leaves `{settingsTabActive: false, sectionActive: true}` (the precondition the bug depended on); subsequent KO/EN toggles produce **0** `/api/context/overview` calls. The pass-1 (section-only) gate would have produced 2.

`?v=24` → `?v=25`.

Re the other minor follow-ups (panelLoading flash on langchange + audit of other inline-`t()` regions in this file): happy to file tracking issues if they're worth pursuing — not in scope for this PR.

memtomem:
Filed two follow-up tracking issues for the deferred minor items from review:

* **#825** — skip `panelLoading` on langchange-driven overview reload (UX flicker only, correctness fine)
* **#826** — audit other inline-`t()` regions in context-gateway.js for the same staleness pattern (currently scoped narrowly to the dashboard cards)

Also pushed commit 7372c32 to make `test_langchange_after_tab_switch_does_not_refetch_overview` robust to mount-time call-count variance — the spec was failing on CI's chromium because boot fired two `/api/context/overview` calls during `_open_context_gateway` (vs one locally), but the spec's actual invariant is "no *additional* refetch after the main-tab switch". Now captures the post-mount baseline and asserts post-toggle equality. Mutation re-validated: dropping the main-tab gate still flips the spec to fail.

PR #838, opened 2026-05-08
feat(cli): mm sync-doctor — read-only validator for private-repo multi-device sync

memtomem:
## Summary

- Implements Phase 2 of the multi-device sync RFC ([memtomem-docs#34](https://github.com/memtomem/memtomem-docs/pull/34)).
- New `mm sync-doctor` — read-only validator that catches the common private-repo sync footguns Phase 0 documents.
- Adds a small (~22 LOC) cloud-mount path-prefix helper alongside the command (RFC §Phase 2 ≤30 LOC budget).

## What it does

Run from inside the private synced repo's working tree. Six checks, exits non-zero on any `✗`:

```
$ mm sync-doctor
✓ no *.db files staged
✓ config.json absent from worktree
! ~/.memtomem/config.d/ does not exist on this machine
  bridge synced fragments via symlink/copy per RFC §Design
✓ ~/.claude/projects/ slug matches synced layout
✓ memory_dir paths resolve under $HOME
✓ no memory_dir under known cloud-sync mounts
```

| Check | Maps to RFC |
|---|---|
| `*.db` family staged via `git ls-files` | §What syncs, what does not |
| `config.json` staged anywhere in tree | §Anti-patterns #4 |
| `~/.memtomem/config.d/*.json` fragment count | §The convention bridge |
| `~/.claude/projects/<encoded-cwd>` exists | §Auto-memory sub-section |
| `memory_dirs[]` resolve under `$HOME` | portability hint |
| `memory_dirs[]` cloud-mount detection | §Phase 2 helper |

Out of scope, explicitly: no `push`, no `pull`, no auto-fix (RFC §Non-goals).

## Helper design

`cloud_mount_prefix(path) -> str | None` is a pure path-prefix match against a fixed list of expanded `$HOME` prefixes — no fs probing, no platform branching:

```
~/Library/CloudStorage/    macOS sync clients
~/Library/Mobile Documents/com~apple~CloudDocs/   iCloud Drive
~/Dropbox/                                         Dropbox legacy
~/OneDrive*/                                       OneDrive (any variant)
```

The OneDrive matcher accepts `OneDrive`, `OneDrive-Personal`, `OneDrive - Foo` but rejects substring false positives like `OneDriveStuff`.

## Open question for review

Per RFC §Open question 3: **what does cloud-mount detection scan?**

- This PR scans **all `memory_dirs[]`** (entries both inside and outside the synced repo's worktree). Matches the §Phase 2 example output.
- Alternative: only `memory_dirs[]` entries that sit inside the synced repo's worktree (narrower surface, cleaner boundary).

Defaulted to the broader scan because most users hit cloud-mount issues with paths *outside* the sync repo (per `feedback_gdrive_watcher_gap.md`). Happy to flip if review prefers the tighter boundary.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_sync_doctor_cli.py` — 29 cases (helper variants, each check, CLI exit codes)
- [x] `uv run ruff check packages/memtomem/src` + `ruff format --check` — clean
- [x] `uv run mypy packages/memtomem/src/memtomem/cli/sync_doctor_cmd.py` — clean
- [x] `uv run mm sync-doctor` smoke from this checkout — exits 0, output matches RFC example shape
- [x] `mm sync-doctor --help` registered under `mm --help`
- [ ] Maintainer: verify boundary decision for `memory_dirs` scope (above)

### DIFF
--- a/packages/memtomem/src/memtomem/cli/__init__.py
+++ b/packages/memtomem/src/memtomem/cli/__init__.py
@@ -56,6 +56,7 @@ def _register() -> None:
     from memtomem.cli.session_cmd import activity, session
     from memtomem.cli.shell import shell
     from memtomem.cli.status_cmd import status
+    from memtomem.cli.sync_doctor_cmd import sync_doctor
     from memtomem.cli.uninstall_cmd import uninstall
     from memtomem.cli.upgrade_cmd import upgrade
     from memtomem.cli.watchdog_cmd import watchdog
@@ -76,6 +77,7 @@ def _register() -> None:
     cli.add_command(session)
     cli.add_command(activity)
     cli.add_command(status)
+    cli.add_command(sync_doctor)
     cli.add_command(watchdog)
     cli.add_command(schedule)
     cli.add_command(web)
--- a/packages/memtomem/src/memtomem/cli/sync_doctor_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/sync_doctor_cmd.py
@@ -28,35 +28,51 @@ import click
 
 # ---- Cloud-mount path-prefix helper ----------------------------------------
 #
-# Pure path-prefix match against the user's expanded ``$HOME``. No fs probing,
-# no platform branching — the prefix list is fixed (RFC §Phase 2). Intended to
-# be applied to each entry in ``memory_dirs[]`` after ``Path.expanduser()``.
-
-CLOUD_MOUNT_PREFIXES: tuple[str, ...] = (
-    "~/Library/CloudStorage/",
-    "~/Library/Mobile Documents/com~apple~CloudDocs/",
-    "~/Dropbox/",
-    "~/OneDrive",  # matches OneDrive, OneDrive-Personal, "OneDrive - Foo", etc.
-)
+# Pure path comparison against a fixed list of cloud-sync mounts under
+# ``$HOME`` (RFC §Phase 2). No fs probing. ``Path.relative_to`` keeps the
+# match separator-safe on Windows (``\\``) as well as POSIX (``/``).
+#
+# Matched prefixes (display form):
+#   ~/Library/CloudStorage/                          (macOS sync clients)
+#   ~/Library/Mobile Documents/com~apple~CloudDocs/  (iCloud Drive on macOS)
+#   ~/Dropbox/                                       (Dropbox legacy fallback)
+#   ~/OneDrive*/                                     (OneDrive, any suffix variant)
 
 
 def cloud_mount_prefix(path: Path, *, home: Path | None = None) -> str | None:
     """Return matched cloud-mount prefix (display form), or ``None`` if none.
 
     ``path`` should already be expanded (``Path.expanduser()``). ``home`` is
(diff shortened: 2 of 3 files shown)


memtomem:
Good catch — fixed in b579c84. Two related issues, addressed both:

**1. HOME / USERPROFILE (the one you flagged)** — added `_set_home(monkeypatch, home)` helper that sets both `HOME` and `USERPROFILE`, and replaced every `monkeypatch.setenv("HOME", ...)` call site with it. `os.path.expanduser("~")` on Windows reads `USERPROFILE` first, so the prior fixtures left `Path.home()` pointing at the real user profile under Windows CI.

**2. `cloud_mount_prefix` impl was also Windows-broken** — separately from the env-var issue, the helper used `str.startswith` against forward-slash literals. On Windows, `tmp_path` is backslash-separated, so even the tests that pass `home=tmp_path` directly (bypassing `Path.home()`) would have mismatched. Refactored to use `Path.relative_to` for separator-safe component-aware comparison; the OneDrive variant matcher now inspects `rel.parts[0]` rather than slicing strings. Dropped the unused `CLOUD_MOUNT_PREFIXES` constant; the matched prefixes live in the module-level comment block now.

29/29 tests still pass on macOS. Helper is still ≤ 30 LOC.

memtomem:
All three addressed in 50765a6.

**1. Read-only invariant** — replaced `load_config_overrides` with a new private `_apply_memory_dirs_override_no_write` helper that mirrors only the `indexing.memory_dirs` slice we need (preserves env precedence) and never calls `_migrate_auto_discover_once`. New regression test `test_does_not_rewrite_legacy_config_json` writes a legacy `auto_discover=True` config.json, runs the doctor, and asserts byte-identity before/after.

**2. Repo-root scan** — extracted `_repo_top_level(start)` helper; `_git_ls_files` now resolves `git rev-parse --show-toplevel` first and lists from there, so subdir invocations cover the entire repo. Applied the same anchor to the Claude slug check too — the slug Claude Code records is for the dir `claude` was launched from (the repo root, normally), so a subdir invocation shouldn't see `slug differs` when the root is mapped correctly. Two new tests cover this: `test_subdir_invocation_sees_root_files` (unit) and `test_subdir_invocation_finds_root_db` + `test_subdir_invocation_anchors_slug_at_repo_root` (CLI).

**3. Windows slug encoding** — replaced the one-shot `str(cwd).replace("/", "-")` with:
- **Fast path**: candidate set covering POSIX (`/` → `-`) and two Windows variants (`\` → `-` with drive `:` dropped or `:` → `-`). Candidates that still contain `/` or `\` after replacement are filtered (wrong-platform encoding) so `projects / cand` doesn't silently fall through to `cand` itself when `cand` is absolute.
- **Slow path**: iterate real entries, decode each (`-` → `/`), `samefile`-compare to cwd. Picks up cases where the encoded form isn't predictable.

POSIX still works (the original literal-dash case the existing memtomem decoder is fragile around is handled by the fast path); Windows valid layouts now match via the drive-colon-dropped or drive-colon-as-`-` candidate.

Total: 33 tests green (was 29 + 4 regression tests for the three issues).

PR #853, opened 2026-05-08
docs(guides): document Obsidian as the editor layer on git transport

tsdata:
## Summary

Docs-only enrichment of the multi-device sync guide: a focused "Obsidian as editor on top of git transport" section that explains how to reuse the existing markdown-as-SoT + git-as-transport layout with Obsidian as a pure editor layer. No code change.

Other transports (Obsidian Sync paid SaaS, iCloud, Dropbox) are deliberately out of scope — their fs-watcher / consistency characteristics differ from git's and would invalidate the layout / anti-patterns the rest of the guide depends on.

## Files

| File | Change |
|---|---|
| `docs/guides/multi-device-sync.md` | New section "Obsidian as editor on top of git transport" inserted between "Auto-memory (Claude Code)" and `mm sync-doctor`. ~95 lines. |
| `docs/guides/configuration.md` | Adds `**/.obsidian/**` to the canonical `exclude_patterns` example block so the rule is discoverable from the configuration reference, not only the sync guide. |
| `docs/guides/reference.md` | One-line see-also under `### Importing from Obsidian` pointing at the new live-sync section, explicitly framed as "rather than one-shot ingest" so the existing import flow is not read as deprecated. |
| `CHANGELOG.md` | New `### Documentation` subsection in `[Unreleased]` (none existed yet) describing the change. |

## What the new section covers

1. **Premise.** Obsidian = editor only. Transport stays git. Other transports excluded.
2. **Vault layout.** Two arrangements — (a) vault root = synced repo root, (b) vault contains a `memories/` sub-folder — both reusing the existing namespace tree.
3. **Required `.obsidian/` exclude.** Vault metadata is JSON-heavy and memtomem indexes `.json` by default, so an explicit `**/.obsidian/**` entry in `exclude_patterns` is needed. Provided as a `~/.memtomem/config.d/30-obsidian.json` fragment so the rule itself is portable.
4. **`.gitignore` for vault-local state.** Suggests gitignoring `.obsidian/workspace*.json` and `.obsidian/cache/`; treats the rest as Obsidian-side preferences rather than memtomem concerns.
5. **Plugin-generated markdown.** `.canvas` is ignored automatically; `.excalidraw.md` matches the `.md` filter and can be excluded with `**/*.excalidraw.md`.
6. **Workflow.** Save in Obsidian → watcher handles temp-rename → incremental re-embed. Post-pull workflow applies unchanged.
7. **Distinction from one-shot import.** Cross-link to `mem_do(action="import_obsidian", …)` with explicit "different use case" framing.

## Cross-link / anchor audit

All anchors verified against existing headings:

- `#the-layout--namespace-aligned-directory-tree` → `## The layout — namespace-aligned directory tree` (em-dash collapses, double dash from preserved spaces)
- `#what-syncs-what-does-not` → `## What syncs, what does not`
- `#post-pull-workflow` → `## Post-pull workflow`
- `#obsidian-as-editor-on-top-of-git-transport` → new heading at line 225
- `configuration.md#exclude-patterns` → `### Exclude patterns` (line 325, unchanged)
- `reference.md#importing-from-obsidian` → `### Importing from Obsidian` (line 617, unchanged)

## Source-of-truth verification

Per `feedback_docs_as_tests.md` — every claim about memtomem behavior in the new section was checked against current `main`:

- `.json` is default-indexed → confirmed in `packages/memtomem/src/memtomem/config.py` default `supported_extensions`.
- watchdog handles temp-rename atomic writes → confirmed via `packages/memtomem/src/memtomem/indexing/watcher.py:_MarkdownEventHandler`.
- `.canvas` not in defaults; `.excalidraw.md` matched by `.md` → confirmed by inspection of the same default extension list.
- `exclude_patterns` semantics (root-relative, non-retroactive, layered on built-in denylists) → consistent with `configuration.md:325-357`.

## Test plan

- [x] `uv run ruff check packages/memtomem/src` — clean (no Python touched)
- [x] `uv run ruff format --check packages/memtomem/src` — clean
- [x] All cross-links manually resolved against actual heading slugs (table above)
- [ ] Reviewer: spot-check the new section reads coherently relative to the surrounding "Auto-memory (Claude Code)" and `mm sync-doctor` sections
- [ ] Reviewer: confirm the `.gitignore` recommendations (`workspace*.json`, `cache/`) match Obsidian's actual non-portable state — softened to "often useful to sync" / "Obsidian-side preference" rather than prescriptive

## Out of scope

- **Adding `.obsidian` to default `_EXCLUDED_DIRS`** — would let users skip the manual `exclude_patterns` step entirely. Tracked separately as a possible follow-up; not bundled here per `feedback_one_change_per_pr.md`.
- **Notion** — proprietary block tree, lossy markdown round-trip; SoT contract incompatible with this guide. Not documented here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,20 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   `_write_config_and_summary`) now invoke it before
   `_atomic_write_json`.
 
+### Documentation
+
+- **Obsidian as editor on top of git transport.** New section in the
+  multi-device sync guide (`docs/guides/multi-device-sync.md`) describing
+  vault layout (vault root = synced repo vs. vault contains a `memories/`
+  sub-folder), the required `**/.obsidian/**` exclude_patterns step (vault
+  metadata includes JSON which memtomem indexes by default), `.gitignore`
+  guidance for vault-local state, and plugin-generated formats (`.canvas`,
+  `.excalidraw.md`). Cross-references to the existing one-shot
+  `mem_do(action="import_obsidian", …)` flow in `reference.md` clarify
+  that they cover different use cases (live sync vs. one-shot ingest).
+  Also adds an `**/.obsidian/**` example to the
+  `configuration.md#exclude-patterns` block.
+
 ## [0.1.36] — 2026-05-06
 
 ### Added
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -337,8 +337,10 @@ extend them but cannot override them.
     "exclude_patterns": [
       "**/subagents/**",         // Claude Code subagent metadata
       "**/antigravity-browser-profile/**",
-      "**/.gemini/**/*.json"     // defensive — only relevant if you manually
+      "**/.gemini/**/*.json",    // defensive — only relevant if you manually
                                  // add ~/.gemini/ to memory_dirs
+      "**/.obsidian/**"          // Obsidian vault metadata (workspace.json,
+                                 // plugin state) when a vault is a memory_dir
     ]
   }
 }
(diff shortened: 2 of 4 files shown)


memtomem (approved):
Fix verified: JSON fragment is now copy-paste valid; loader strict json.loads constraint called out in prose. Cross-link audit in PR body checks out.

PR #863, opened 2026-05-08
fix(web/i18n): drop duplicate hooks empty-state message + finish hooks i18n parity

tsdata:
## Summary

Sibling fix to #858 / de1ca36 — when `.memtomem/settings.json` is
missing, the Hooks detail panel printed the same line twice: status
badge at the top said "No .memtomem/settings.json found", and the
body below echoed the exact same string because `loadHooksSync()`
threaded `badge.text` into `emptyState()`'s message slot. #858 fixed
only the layout (icon/message/hint vertical stack); the duplication
itself stayed.

- Body now carries the **actionable hint** as its message and drops
  the redundant badge text. Status badge above already names the
  condition, body tells the user what to do.
- Pulled three adjacent hardcoded English strings through `t()` —
  `no_hooks_defined` (line 103), `load_failed` (line 145),
  `error_hint` (line 50). KO users now see Korean on those paths
  instead of falling back to English. Same i18n cleanup category as
  de1ca36.
- Four new locale keys in `settings.hooks.*`: `no_source_hint`,
  `error_hint`, `no_hooks_defined`, `load_failed`. en/ko symmetric.
- Cache-bust `settings-hooks-watchdog.js?v=3` → `?v=4` (only file
  with body changes; locale JSON is fetched `no-store` so no bump
  needed there).

## Before / after

| state | before | after |
|---|---|---|
| `no_source` | badge: `No .memtomem/settings.json found` + path; body: same string + hint | badge: same; body: hint only (single actionable line) |
| `error` | badge: `<data.error>`; body: `<data.error>` again + "Fix the JSON file ..." | badge: same; body: localized "Fix the JSON file ..." only |

## Test plan

- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/test_i18n.py` — 44/44 passing (parity guard catches the 4 new keys in both locales)
- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/test_web_mode.py packages/memtomem/tests/test_init_cmd.py` — 306/306 passing
- [x] `uv run ruff check packages/memtomem/src` + `ruff format --check` clean
- [x] Manual UI smoke (`uv run mm web`, navigate Settings → Hooks): single body line on no_source, locale toggle KO swaps body string, badge stays in its own row.


memtomem (approved):
LGTM

PR #860, opened 2026-05-08
fix(web/i18n): repair Korean empty-state hint (scan_dirs, josa, button names)

tsdata:
## Summary

Fix three i18n bugs in the context-gateway empty-state hint, surfaced now
that PR #858 made the hint actually readable (instead of overlapping the
title).

| Bug | Surface | Fix |
|---|---|---|
| `{scan_dirs}` replaced with empty string | KO: `이 프로젝트의 [공백] 에 이미 있는`, EN: `from  within this project.` (double space) | Thread `data.scanned_dirs` into `_ctxRenderItemsHtml`, reuse the same `(scannedDirs \|\| []).join(', ') \|\| `.${type}/`` fallback already used at `context-gateway.js:732` |
| `{type}을(를)` placeholder leaks ("skills을(를)") | KO empty hint | Sentence rewrite — attach josa to the fixed Korean noun "폴더" instead of the dynamic `{type}`. All four current types (`agents`/`skills`/`commands`/`hooks`) read uniformly, no josa picker needed |
| Inline button names stayed English | KO: `Sync 를 누르거나` / `Import 로 가져올 수` while the actual buttons render as 동기화 / 가져오기 | Updated KO copy to match the on-screen labels |

## Files modified

| File | Lines | Change |
|---|---|---|
| `packages/memtomem/src/memtomem/web/static/context-gateway.js` | 620, 622-626, 679-685 | Add `scannedDirs` parameter to `_ctxRenderItemsHtml`, compute scan-list with the line-732 fallback, pass `data.scanned_dirs \|\| []` from the only caller |
| `packages/memtomem/src/memtomem/web/static/locales/ko.json` | 334 | Sentence rewrite: drop `을(를)` placeholder, localize Sync/Import to 동기화/가져오기, switch `{scan_dirs} 에` to `{scan_dirs} 에서` |
| `packages/memtomem/src/memtomem/web/static/index.html` | 1391 | Bump `context-gateway.js?v=27` → `?v=28` |

`en.json` is unchanged: with `{scan_dirs}` correctly interpolated the
existing English template reads grammatically (e.g. `pull existing skills
from .skills/ within this project.`).

`i18n.js` already fetches `/locales/{lang}.json` with `cache: 'no-store'`,
so the ko.json edit needs no version bump.

## Verification (Playwright MCP, real DB)

`uv run mm web` against the existing `~/.memtomem/`. Hard-reload to
serve `?v=28`. Programmatic DOM check on the More tab → 서브에이전트:

[...]

PR #864, opened 2026-05-08
fix(web): formalize emptyState CTA and fix namespaces layout regression

### LINKED ISSUE #859: QA: namespaces empty-state + CTA layout (post-#858 follow-up)
## Summary

Follow-up to PR #858 (`fix(web): wrap empty-state spans …`).

After #858 makes `emptyState()` emit its own `<div class="empty-state">…</div>`
wrapper, the only callsite that combines `emptyState()` with extra content
in the same outer wrapper is `settings-namespaces.js:193`:

```js
const cta = `<a class="empty-state-cta" …>${escapeHtml(t('settings.ns.empty.cta'))} →</a>`;
list.innerHTML = `<div class="empty-state">${emptyState('🗂', …)}${cta}</div>`;
```

Post-#858 the rendered DOM becomes:

```html
<div class="empty-state">
  <div class="empty-state">             <!-- inner, owned by emptyState() -->
    <span class="empty-state-icon">🗂</span>
    <span>{title}</span>
    <span class="empty-state-hint">{body}</span>
  </div>
  <a class="empty-state-cta" …>…</a>     <!-- sibling of inner block -->
</div>
```

Both the outer and inner `.empty-state` are flex-column with `gap: 6px`
(`style.css:1941`), so the title/body stack inside the inner box and the
CTA stacks below the inner box in the outer box. **In theory** spacing is
consistent.

## What needs verifying

This couldn't be eyeballed during #858 QA because the local DB has 5
namespaces — the empty state never rendered. We need a deliberate test
on a fresh DB. Use a throwaway `HOME` so the real `~/.memtomem/` is not
touched:

```bash
# Fresh, empty memtomem instance (no namespaces yet)
export FAKE_HOME=/tmp/mm-issue-859
mkdir -p "$FAKE_HOME"
HOME="$FAKE_HOME" uv run mm init -y --mcp skip --preset minimal
HOME="$FAKE_HOME" uv run mm web    # http://127.0.0.1:8080
```

Then:

1. Open the More tab → 네임스페이스 (Namespaces).
2. Confirm:
   - Vertical gap between title (`settings.ns.empty.title`) and body
     (`settings.ns.empty.body`) inside the inner `.empty-state` is ~6px.
   - Vertical gap between the inner `.empty-state` block and the CTA `<a>`
     (siblings in the outer `.empty-state`) is also ~6px (outer's
     `gap: 6px`).
   - Inner `.empty-state` does NOT vertically over-stretch via its
     `height: 100%` (`style.css:810`); CTA stays directly under the
     inner block, not pushed to viewport bottom.

Cleanup after QA: `rm -rf /tmp/mm-issue-859`.

## If a regression is found

Two simple fixes:

1. Remove the outer wrap from `settings-namespaces.js:193` and either
   - render the CTA as a sibling of the inner `.empty-state` (outside the
     function output, but in the same parent container), or
   - add a 4th `cta` parameter to `emptyState()` so the function emits
     `<div class="empty-state">…<a class="empty-state-cta">…</a></div>`
     in one shot. This is the "formalize CTA" follow-up already promised
     in #858's PR body.

## Related

- PR #858 — the fix this is following up on
- `feedback_layout_primitives_first.md` — the principle this preserves
  (don't introduce a new modifier class until the existing primitive
  proves insufficient)

Husam-Abdulraheem:
Addresses issue #859 by adding dedicated ctaHtml support to emptyState() helper. This removes the need for redundant outer wrappers in settings-namespaces.js and app.js, preventing nested .empty-state divs and vertical over-stretching.
Fixes #859

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -973,10 +973,14 @@ function panelLoading(container) {
 
 // ── A5: Empty State ──
 // Owns its `.empty-state` wrapper — callers must NOT add their own.
-function emptyState(icon, message, hint) {
+// cta is structured as { href, label } for safe escaping.
+function emptyState(icon, message, hint, cta) {
   const i = icon ? `<span class="empty-state-icon">${icon}</span>` : '';
   const h = hint ? `<span class="empty-state-hint">${escapeHtml(hint)}</span>` : '';
-  return `<div class="empty-state">${i}<span>${escapeHtml(message)}</span>${h}</div>`;
+  const c = cta
+    ? `<a class="empty-state-cta" href="${escapeHtml(cta.href)}" target="_blank" rel="noopener">${escapeHtml(cta.label)} →</a>`
+    : '';
+  return `<div class="empty-state">${i}<span>${escapeHtml(message)}</span>${h}${c}</div>`;
 }
 
 // ── A1: Toast Notifications ──
@@ -1764,7 +1768,7 @@ function _renderNsChart(namespaces) {
 function _renderHomeRecent(allSources) {
   const recentList = qs('home-recent-list');
   if (!allSources.length) {
-    recentList.innerHTML = '<div class="empty-state">' + emptyState('📁', 'No sources indexed yet', 'Add files from the Index tab') + '</div>';
+    recentList.innerHTML = emptyState('📁', 'No sources indexed yet', 'Add files from the Index tab');
     return;
   }
 
@@ -3197,7 +3201,7 @@ function _renderSourcesStats(activeVendor) {
 function hideBrowser() {
   hide(qs('chunks-browser-content'));
   const browser = qs('chunks-browser');
-  browser.innerHTML = '<div class="empty-state">' + emptyState('📄', 'Select a source to browse its chunks') + '</div>';
+  browser.innerHTML = emptyState('📄', 'Select a source to browse its chunks');
 }
 
 // ── Memory-mode source tree ─────────────────────────────────────────
--- a/packages/memtomem/src/memtomem/web/static/index.html
+++ b/packages/memtomem/src/memtomem/web/static/index.html
@@ -1381,11 +1381,11 @@
   <script src="/vendor/prism-bash.min.js?v=1"></script>
   <script src="/vendor/prism-yaml.min.js?v=1"></script>
   <script src="/i18n.js?v=3"></script>
-  <script src="/app.js?v=114"></script>
+  <script src="/app.js?v=115"></script>
   <script src="/settings-config.js?v=9"></script>
   <script src="/sources-memory-dirs.js?v=11"></script>
   <script src="/settings-maintenance.js?v=2"></script>
-  <script src="/settings-namespaces.js?v=5"></script>
+  <script src="/settings-namespaces.js?v=6"></script>
   <script src="/settings-harness.js?v=2"></script>
   <script src="/settings-hooks-watchdog.js?v=5"></script>
   <script src="/timeline.js?v=3"></script>
--- a/packages/memtomem/src/memtomem/web/static/settings-namespaces.js
+++ b/packages/memtomem/src/memtomem/web/static/settings-namespaces.js
@@ -189,8 +189,10 @@ async function loadNamespacesTab() {
     const data = await api('GET', '/api/namespaces');
     const namespaces = data.namespaces || [];
     if (!namespaces.length) {
-      const cta = `<a class="empty-state-cta" href="https://github.com/memtomem/memtomem/blob/main/docs/guides/configuration.md#namespace" target="_blank" rel="noopener">${escapeHtml(t('settings.ns.empty.cta'))} →</a>`;
-      list.innerHTML = `<div class="empty-state">${emptyState('🗂', t('settings.ns.empty.title'), t('settings.ns.empty.body'))}${cta}</div>`;
+      list.innerHTML = emptyState('🗂', t('settings.ns.empty.title'), t('settings.ns.empty.body'), {
+        href: 'https://github.com/memtomem/memtomem/blob/main/docs/guides/configuration.md#namespace',
+        label: t('settings.ns.empty.cta'),
+      });
       return;
     }
     list.innerHTML = '';


Husam-Abdulraheem:

tsdata (commented):
Thanks for picking this up — the `ctaHtml` shape matches what #859's
"If a regression is found §1.2" prescribed, and the un-wrappings at
`app.js:1686, 3000, 3372-3374` are correct (mechanically equivalent
post-#858, since both layers were already flex-column).

One blocker and a couple of nits.

### 🔴 `index.html` cache-bust bump missing

`packages/memtomem/src/memtomem/web/static/index.html:1383, 1387`
still pin `app.js?v=111` and `settings-namespaces.js?v=5`, but both
file bodies changed. This repo bumps `?v=` on every JS body change
(see #858: `?v=110 → ?v=111`, #860: `?v=27 → ?v=28`).

User-visible failure mode if we ship as-is: a returning user with a
cached `app.js?v=111` (post-#858, 3-param `emptyState`) plus a fresh
`settings-namespaces.js?v=5` (now passing 4 args) gets the old cached
function — which silently drops the 4th arg. The CTA disappears from
the namespaces empty state until the user hard-reloads. Exactly the
partial-cache regression `?v=` is supposed to prevent.

```suggestion
  <script src="/app.js?v=112"></script>
```

(and `settings-namespaces.js?v=5 → ?v=6` on the corresponding line).

### 🟡 PR body doesn't mention the `app.js` cleanups

Title and description only call out `settings-namespaces.js`, but the
diff also un-wraps three pattern-B sites in `app.js` (lines 1686, 3000,
3372–3374). They're correct and on-scope for the "formalize CTA"
follow-up #858 promised — just worth listing in the description so
reviewers know the call paths covered. ~10 more pattern-B sites
remain in `settings-harness.js` etc. (out of scope here, fine to
leave for a follow-up).

### 🟡 Test plan / verification

#858 and #860 both included a short Playwright/manual confirmation
of the rendered DOM. For #864 the QA scenario from #859 already
spells it out — fresh `HOME=/tmp/mm-issue-859`, empty namespaces tab,
confirm title + body + CTA in three rows with ~6px gaps. A line in
the PR body confirming this passes would close the loop on the
issue this PR is resolving.

### Tiny

- `ctaHtml` is passed verbatim (unescaped), symmetric with `icon`.
  Reasonable, but the `// Owns its '.empty-state' wrapper — callers
  must NOT add their own.` header comment could add "ctaHtml is
  inserted as raw HTML; caller is responsible for escaping". One
  line, not a blocker.

Once the `?v=` bump lands this is good to go from my side.

memtomem:
Thanks for the fix — the nested `.empty-state` regression from #859 is correctly resolved, and the call-site cleanup (the four `<div class="empty-state">${emptyState(...)}` outer wraps) lines up with the helper's existing "owns its wrapper" contract.

One concern on the new `ctaHtml` parameter before merge:

**`packages/memtomem/src/memtomem/web/static/app.js:899`** — `ctaHtml` is appended directly into the `innerHTML` template string, so the helper's existing safe-by-default contract (where `message` and `hint` are escaped internally via `escapeHtml`) is weakened: every future caller has to remember to escape all CTA content correctly. The current `settings-namespaces.js` caller does escape its label, so this PR is safe today — but a year from now a second CTA caller pasting unescaped state would silently introduce an XSS vector.

Suggested shape change: take a structured CTA parameter so escaping stays inside the helper. Something like:

```js
function emptyState(icon, message, hint, cta) {
  const i = icon ? `<span class="empty-state-icon">${icon}</span>` : '';
  const h = hint ? `<span class="empty-state-hint">${escapeHtml(hint)}</span>` : '';
  const c = cta
    ? `<a class="empty-state-cta" href="${escapeHtml(cta.href)}" target="_blank" rel="noopener">${escapeHtml(cta.label)} →</a>`
    : '';
  return `<div class="empty-state">${i}<span>${escapeHtml(message)}</span>${h}${c}</div>`;
}
```

…and the `settings-namespaces.js` call site becomes:

```js
emptyState('🗂', t('settings.ns.empty.title'), t('settings.ns.empty.body'), {
  href: 'https://github.com/memtomem/memtomem/blob/main/docs/guides/configuration.md#namespace',
  label: t('settings.ns.empty.cta'),
})
```

If a future caller genuinely needs raw HTML, name that parameter explicitly (e.g. `ctaHtmlUnsafe`) and document the escaping requirement at the call site — that keeps the structured path safe-by-default and makes the unsafe path opt-in/auditable.

Happy to fold the change myself if it's easier, or leave it for a follow-up commit on your branch — your call.

Husam-Abdulraheem:
Fixed as suggested

memtomem:
Hey @Husam-Abdulraheem — thanks again for picking #859 up and for the structured-`cta` refactor; that XSS hardening is a much nicer shape than the original `ctaHtml`, and it lines up exactly with the safe-by-default contract the helper has for `message` / `hint`.

I went ahead and folded two small remaining items directly (using "Allow edits from maintainers") so we can land this without another round-trip on your side:

1. **`app.js?v=114 → ?v=115`** in `index.html` — `?v=114` is what current `main` ships, and your branch now changes `app.js` body, so without bumping we'd hit the exact partial-cache failure mode the earlier review flagged.
2. **Restored `!plan.discoveredCount` guard + comment** in `_renderMemorySourceTree` (around `app.js:3587`). I'm 90% sure this was an accidental conflict-resolution casualty from the 2026-05-11 `Merge main into fix branch and resolve conflicts` pass — that hunk landed via #896 (`fix(web): keep Discovered dirs visible when filter is active`) between when you opened the PR and when you re-merged main, and it sits right next to the empty-state code you were editing. Without it, vendors with discovered dirs but zero indexed files get silently wiped to "No matches for that filter" while a filter is active. Definitely not on you — these mid-merge regressions are nasty to spot when the surrounding diff is unrelated.

I also rewrote the branch as a single clean commit on top of current `main` to drop the noise from the intermediate merge commits — the final tree is identical to what you had plus the two items above. Authorship is preserved (commit author = you, with `memtomem` as committer / co-author).

CLA is re-running on the new SHA; once it goes green I'll squash-merge. Really appreciate the patience on this one — the layout fix is a clean improvement and the structured-cta API is going to make the remaining ~10 pattern-B call sites much easier to clean up in a follow-up.

memtomem (approved):
Approving on behalf of the cache-bust + #896 guard restore that just landed via maintainer-edit force-push (see comment above). Structured-cta refactor LGTM, single-commit branch is clean on top of current main.

Issue #759, opened 2026-05-04
[bug]: Errors encountered during `uv run pytest` on Windows


powerzist:
### Summary

Hello,

I encountered the following errors during `uv run pytest` and wanted to report them with the relevant traceback for reference.

I ran it in Command Prompt as an **administrator**.

## ===short test summary info===
```
FAILED packages/memtomem/tests/test_cli_index_noop_e2e.py::TestFreshNoopIndexSubprocess::test_init_index_search_via_subprocess - TypeError: argument of type 'NoneType' is not iterable
FAILED packages/memtomem/tests/test_index_debounce.py::TestPersistenceAndConcurrency::test_concurrent_enqueue_does_not_lose_entries - AssertionError: assert 11 == 20
FAILED packages/memtomem/tests/test_uninstall_cmd.py::TestRuntimePidCleanedWithOther::test_runtime_pid_deleted_and_subdir_pruned - AssertionError: empty runtime subdir must be pruned
FAILED packages/memtomem/tests/test_wiki_store.py::TestIsDirty::test_dirty_with_modified_tracked_file - UnicodeDecodeError: 'cp949' codec can't decode byte 0xe2 in position 196: illegal multibyte sequence
```

## ===FAILURES===
```
____________________________________________________________________________________ TestFreshNoopIndexSubprocess.test_init_index_search_via_subprocess ____________________________________________________________________________________

self = <tests.test_cli_index_noop_e2e.TestFreshNoopIndexSubprocess object at 0x0000020DDF615F90>, tmp_path = WindowsPath('C:/Users/user/AppData/Local/Temp/pytest-of-user/pytest-46/test_init_index_search_via_sub0')

    def test_init_index_search_via_subprocess(self, tmp_path):
        """Out-of-process variant: catches regressions that only manifest
        across the HOME / XDG boundary (e.g. config.json path resolution).

        Uses the ``mm`` script installed by ``uv pip install -e`` (co-located
        with ``sys.executable``) rather than ``python -m memtomem`` — the
        package has no ``__main__`` module, and the installed entry point is
        what real users hit.
        """
        # ``shutil.which`` adds the platform-correct suffix (``.exe`` on
        # Windows via PATHEXT, none on POSIX), so the same lookup works
        # against both ``.venv/bin/mm`` and ``.venv/Scripts/mm.exe``.
        bin_dir = os.path.dirname(sys.executable)
        mm_bin = shutil.which("mm", path=bin_dir)
[...]

PR #875, opened 2026-05-09
test(context): add subprocess + inline e2e for mm context flow

pandas-studio:
## Summary

- Add `packages/memtomem/tests/test_context_cli_subprocess_e2e.py` with inline + subprocess round-trip pins for `mm context init → generate → detect → status` on an empty project.
- Both classes assert the host-write invariant: without `--include=settings`, `mm context generate` must not touch the fake HOME's `~/.claude/settings.json` or `~/.codex/agents/`.
- Subprocess class catches regressions that only surface across the process boundary (`HOME` / `USERPROFILE` resolution, module-level path constants bound at import time, lazy imports) — same class of bug as #759 in `mm index`.

Pattern mirrors `test_cli_index_noop_e2e.py` (`TestFreshNoopIndexInline` + `TestFreshNoopIndexSubprocess`) exactly: same `shutil.which("mm", path=bin_dir)` lookup, same `encoding="utf-8"` / `errors="replace"` cp949-safety guard, same `set_home(monkeypatch, ...)` + `_bootstrap._CONFIG_PATH` patching for the inline variant.

## Why

Most context-gateway coverage today runs in-process via `CliRunner` (`test_context_settings.py`, `test_context_agents.py`, `test_context_install.py`, etc.). That misses regressions that only manifest across the process boundary — e.g. a `Path.home()` call that runs before the `HOME` env override, or an import-time bound `_CONFIG_PATH` that points at the developer's real home. #759 was exactly that for `mm index`; the `mm context` family deserves the same guard.

The host-write invariant is the loud signal we want if `mm context generate` ever flips to running settings sync by default — that would be a critical regression and the assertion here would catch it before it ships.

## Test plan

- [x] `uv run ruff check packages/memtomem/tests/test_context_cli_subprocess_e2e.py` — passes
- [x] `uv run ruff format --check packages/memtomem/tests/test_context_cli_subprocess_e2e.py` — passes
- [x] `uv run pytest packages/memtomem/tests/test_context_cli_subprocess_e2e.py -v` — 2 passed in 0.51s locally
- [ ] CI verifies on Linux / macOS / Windows (subprocess path on Windows is the cp949 cross-check)

### DIFF
--- a/packages/memtomem/tests/test_context_cli_subprocess_e2e.py
+++ b/packages/memtomem/tests/test_context_cli_subprocess_e2e.py
@@ -34,16 +34,14 @@ import pytest
 from click.testing import CliRunner
 
 from memtomem.cli import cli
+from memtomem.context.generator import GENERATORS
 
-from .helpers import set_home
+from .helpers import _MEMTOMEM_ENV_VARS, set_home
 
-_GENERATED_PROJECT_FILES = (
-    "CLAUDE.md",
-    ".cursorrules",
-    "GEMINI.md",
-    "AGENTS.md",
-    ".github/copilot-instructions.md",
-)
+# Derived from the runtime registry rather than hand-curated so a new
+# runtime added to ``GENERATORS`` (cline, aider, etc.) is automatically
+# covered without this list silently under-asserting.
+_GENERATED_PROJECT_FILES = tuple(g.output_path for g in GENERATORS.values())
 
 
 def _seed_project(root: Path) -> None:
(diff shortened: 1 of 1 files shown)


tsdata (approved):
LGTM — subprocess + inline e2e for mm context flow, host-write invariant pinned, registry-derived assertion list. All CI green (lint/typecheck/CLA/3-OS test/browser/golden-path/notebooks/js-unit).

PR #874, opened 2026-05-09
feat(context): detect duplicate hooks across settings tiers (ADR-0010 §4)

pandas-studio:
## Summary

Implements ADR-0010 §4: detect memtomem-managed hooks duplicated across
the three settings tiers (`~/.claude/settings.json`,
`<project>/.claude/settings.json`,
`<project>/.claude/settings.local.json`). Closes #871.

Claude Code 2.x merges hook entries across tiers additively, so a user
who flips `hooks.target_scope` (added in #870) and runs `mm context sync`
would silently leave hook entries in the previous tier — duplicates
fire once per tier.

## Changes

- **Shared detector** (`memtomem.context.settings_doctor`): canonical
  signature matching (event + matcher + command shape, whitespace
  normalized), symlink-aware tier dedup. Detection-only — migration is
  the future #872.
- **Sync-time warning** (primary surface): `_print_settings_generate`
  / `_print_settings_diff` emit non-blocking yellow stderr warnings
  before write, naming the offending tier and pointing at #872. Covers
  `mm context sync --include=settings`, `generate`, and `diff`.
- **`mm context settings-doctor`**: new subcommand with `--json` and
  `--scope=…` override. Exit `0` clean / `1` duplicates. Mirrors the
  existing `@context.command("migrate")` flat shape.
- **Web `GET/POST /api/settings-sync`**: responses gain a
  `duplicate_tier_warnings` list with the same payload shape as the
  CLI `--json` output. Frontend banner deferred — data layer only.
- **Doc**: short "Detecting duplicate hooks across tiers" subsection in
  `docs/guides/integrations/claude-code.md`.
- **Tests**: 25 new cases — detector unit behavior (whitespace
  variants, missing/empty matcher, malformed JSON, active-scope
  exclusion, symlink dedup), CLI exit codes + `--json` schema, end-to-
  end sync warning via `CliRunner`, Web response shape. Mutation-
  validated.

## Stacked on #873

This PR depends on `hooks.target_scope` from #870 (#873). The base of
this PR is `feat/issue-870-target-scope`; once #873 squash-merges,
this branch will need `git rebase --onto origin/main`.

## Out of scope (per issue body / ADR-0010 §4)

- Auto-fix / move logic — that's #872.
- `mm sync-doctor` extension — explicitly excluded; that command stays
  scoped to multi-device sync hygiene.
- Web UI scope picker — gated on the future project-config ADR.
- Migration of non-memtomem-managed hook entries.

## Test plan

[...]

Issue #870, opened 2026-05-09
Implement hooks.target_scope config field + _claude_target / target_file plumbing (ADR-0010 §3)


pandas-studio:
## Summary

ADR-0010 (merged in #869) decided that settings-hooks fan-out gets a
configurable `hooks.target_scope` (`user` / `project_shared` /
`project_local`) with `user` as the v1 default. This issue lands the
foundation: the config field, its default, the two call sites that
resolve scope to a `Path`, and the per-invocation CLI override flag.

This is the **first** of the three follow-ups enumerated in
ADR-0010's Consequences section. Issue 2 (sync-time warning + scoped
doctor) and Issue 3 (settings-migrate subcommand) depend on this
landing first.

## Scope (do this)

- Add `hooks.target_scope: Literal["user", "project_shared",
  "project_local"]` to the config schema at
  `packages/memtomem/src/memtomem/config.py` (alongside existing
  sections). **Default: `"user"`** — zero behavior change for existing
  installs.
- Update `_claude_target()` at
  `packages/memtomem/src/memtomem/web/routes/settings_sync.py:28-29`
  to take `project_root` + resolved scope and return the resolved Path
  (`~/.claude/settings.json` for `user`, `<project>/.claude/settings.json`
  for `project_shared`, `<project>/.claude/settings.local.json` for
  `project_local`).
- Update `ClaudeSettingsGenerator.target_file()` at
  `packages/memtomem/src/memtomem/context/settings.py:122-123` (same
  signature change).
- Loosen `ClaudeSettingsGenerator.is_available()` at `:119-120` to
  "any of the three scopes resolves" so a user with only project-local
  settings doesn't see the tile vanish.
- Add a one-shot `--scope=<user|project_shared|project_local>` flag on
  `mm context sync` for per-invocation override (e.g., a user whose
  default is `user` but wants `project_local` for one specific repo).
- Tests: 3 scope values × 2 call sites = 6 path resolutions covered;
  add a default-preservation test that pins behavior when the field
  is absent.

## Out of scope (resist drift)

- **Sync-time duplicate-hook warning** — separate ticket (Issue 2).
- **Settings-migrate subcommand** — separate ticket (Issue 3).
- **Web UI scope display + duplicate-tier banner** — gated on this
  ticket, separate follow-up.
- **`<project>/.memtomem/config.json` project-level config layer** —
  ADR-0010 §3 explicitly defers this to a future ADR. Resist
  introducing it as part of this PR; the user-level layer at
[...]

PR #876, opened 2026-05-09
feat(context): mm context settings-migrate (ADR-0010 §4, #872)

### LINKED ISSUE #872: Implement settings-migrate subcommand (ADR-0010 §4)
## Summary

ADR-0010 references a future settings-migrate subcommand for users
to move memtomem-managed hooks from one tier to another. This is the
**third** of the three follow-ups enumerated in ADR-0010's
Consequences section, and the third gate on §5's default-flip
trigger (the `mm init` default cannot move from `user` to
`project_local` until users have a documented migration path).

**Depends on**: #870 (config field), #871 (canonical-signature
detection logic — reused for source/target identification).

## Scope (do this)

- New subcommand for migrating memtomem-managed hooks between tiers.
  Suggested shape: `mm context settings-migrate --from <scope> --to
  <scope>` (mirrors the flat-subcommand convention from
  `@context.command("migrate")` at
  `packages/memtomem/src/memtomem/cli/context_cmd.py:1383`).
- Behavior:
  1. Read memtomem-managed hooks from the source tier (canonical-
     signature match — reuse the helper from #871).
  2. Write them to the target tier using the existing additive merge
     in `ClaudeSettingsGenerator.merge()` at
     `packages/memtomem/src/memtomem/context/settings.py:128-185`.
  3. Remove them from the source tier.
- `--dry-run` flag emits the planned changes without touching disk.
- Idempotent: re-running with no changes is a no-op (signature match
  finds nothing in source after a successful migration).
- Tests: round-trip migration `user → project_local`, dry-run
  preview, idempotent re-run, partial-overlap handling (some entries
  already at target).

## Out of scope (resist drift)

- **Auto-detect best target** — the user picks `--to` explicitly.
  Auto-detection conflates user intent with config state.
- **Migration of non-memtomem-managed hook entries** — they stay
  where they are. Only entries matching the canonical signature of
  bundled or canonical-source hooks move.
- **Cross-project migration** — operates on the active project's
  scope tiers only. Users with multiple projects re-run per project.
- **Auto-fire on default flip** — the §5 default flip is a separate
  decision; this command stays user-invoked.

## Acceptance

- Round-trip: a `user → project_local` migration moves the entries
  cleanly, `~/.claude/settings.json` no longer contains the
  signature-matched entries, `<project>/.claude/settings.local.json`
  does.
- `--dry-run` prints the same plan without disk writes.
- Re-running after a successful migration is a clean no-op.
- Lint + tests + typecheck green.

## References

- ADR-0010 §4 (Migration), §5 (Default-flip trigger criterion 3) —
  `docs/adr/0010-settings-hooks-target-scope.md`.
- #870 — config field foundation.
- #871 — canonical-signature detection logic (reused here).
- `packages/memtomem/src/memtomem/context/settings.py:128-185` —
  existing additive merge to reuse for the target write.

pandas-studio:
## Summary

Third and final follow-up from ADR-0010. With #870 (target_scope field) and
#874 (canonical-signature detection) shipped, this adds the action surface
that moves memtomem-managed hook entries between settings tiers — closing
the §5 default-flip path.

CLI:

```bash
mm context settings-migrate --from=user --to=project_local           # dry-run
mm context settings-migrate --from=user --to=project_local --apply   # mutate
mm context settings-migrate --from=user --to=project_local --json    # CI
```

- Operates at **inner-hook granularity** — a user's hand-authored entry
  sharing a matcher with a memtomem-managed entry stays put; only
  canonical-signature-matched inners move.
- Target receives the **canonical rule** from `.memtomem/settings.json`,
  not the source-tier whitespace variant, so target ends up byte-clean.
- Write order is **target first, then source**. A crash between the two
  leaves a transient cross-tier duplicate that the next migrate heals
  idempotently — target write is a no-op (already_at_target), source
  clean-up is the only step left.
- Default is a dry-run; `--apply` to mutate; `--yes` skips the
  host-write confirmation when the target / source resolves outside the
  project root (e.g. `--from=user`); `--json` for scripting.

CLI shape mirrors the existing `mm context migrate` (default dry-run +
`--apply`) rather than the issue's literal "--dry-run" wording, on the
grounds that this command is destructive (rewrites two settings files)
and the conservative default matches the established repo idiom.

This unblocks ADR-0010 §5's third default-flip criterion. The flip
itself remains a separate ADR-light commit per §5.

## ADR-0010 §4 acceptance

- ✅ Round-trip: `user → project_local` migration moves entries
  cleanly, `~/.claude/settings.json` no longer carries them, the
  project's `settings.local.json` does (`test_round_trip_user_to_project_local`).
- ✅ `--dry-run` (the default) prints the plan without disk writes
  (`test_dry_run_no_changes`).
- ✅ Idempotent re-run: second invocation finds nothing to move
  (`test_apply_idempotent_rerun`, `test_idempotent_rerun`).
- ✅ Partial overlap (some entries already at target): target write is
  skipped, source clean-up still runs
  (`test_partial_overlap_already_at_target`, `test_partial_overlap_idempotent`).
- ✅ Lint + tests + typecheck green locally.

## Test plan

- [x] `uv run ruff check packages/memtomem/src` — passes
- [x] `uv run ruff format --check packages/memtomem/src` — passes
- [x] `uv run mypy` on touched files — passes
- [x] `uv run pytest -m "not ollama"` — 4409 passed, 10 skipped
- [ ] CI green
- [ ] Manual smoke: dry-run + `--apply --yes` user → project_local with
  bundled hook in a fresh checkout

## Closes

Closes #872. Builds on #870, #874, ADR-0010 §4.

### DIFF
--- a/docs/guides/integrations/claude-code.md
+++ b/docs/guides/integrations/claude-code.md
@@ -261,9 +261,24 @@ mm context settings-doctor --scope=project_local   # one-shot scope override
 
 The match is by canonical signature (event + matcher + command shape, with
 whitespace normalized) so a hand-edited variant of a memtomem-managed entry
-still classifies. Detection is non-mutating; the `mm context settings-migrate`
-subcommand (see [issue #872](https://github.com/memtomem/memtomem/issues/872))
-will be the action surface that moves duplicate entries into the active scope.
+still classifies. Detection is non-mutating; the action surface is the
+companion `mm context settings-migrate` subcommand:
+
+```bash
+mm context settings-migrate --from=user --to=project_local           # dry-run
+mm context settings-migrate --from=user --to=project_local --apply   # mutate
+mm context settings-migrate --from=user --to=project_local --json    # CI / scripting
+```
+
+Default is a dry-run preview; pass `--apply` to mutate disk. The migrator
+takes the canonical rule from `.memtomem/settings.json` (so the target
+ends up byte-clean rather than carrying the source-side whitespace
+variant) and removes the matched inner-hook entries from the source tier.
+Idempotent — re-running after a clean migration finds nothing to move.
+
+When the source or target lives outside the project root (e.g.
+`--from=user`, which resolves to `~/.claude/settings.json`), `--apply`
+prompts for confirmation; pass `--yes` to skip the prompt in scripts.
 
 ---
 
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -69,6 +69,11 @@ from memtomem.context.settings_doctor import (
     detect_duplicate_tiers,
     format_warning,
 )
+from memtomem.context.settings_migrate import (
+    apply_migration,
+    format_plan_summary,
+    plan_migration,
+)
 from memtomem.context.skills import (
     diff_skills,
     extract_skills_to_canonical,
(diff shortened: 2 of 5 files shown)


pandas-studio:
Addressed both [P2] findings in b71d234:

**1. Target same-matcher / different inner now classified as conflict**
Replaced the signature-only `_target_already_has` with `_classify_target` (three-state: missing / exact / conflict). `exact` now requires the inner to be **byte-equal** to the canonical inner; anything else under the same `(event, matcher)` is a conflict. Apply leaves both source and target untouched on conflict, and the CLI surfaces the per-move reason and exits 1.

**2. Same-command / drifted timeout no longer silently cleans source**
Same fix: byte-equal check rejects drift cases that the signature match was tolerating. The user sees the conflict reason and resolves before re-running.

Also fixed the all-conflict apply path: `is_noop` is True when every move is a conflict (`applicable_moves` is empty), so the previous early-return swallowed the exit-1. Now `is_noop and not conflicts` (genuine no-op) is differentiated from `is_noop and conflicts` (drift to surface).

Four new tests pin the behavior:
- `test_target_same_matcher_different_inner_is_conflict` — finding 1
- `test_target_near_match_with_drifted_timeout_is_conflict` — finding 2
- `test_conflict_leaves_both_source_and_target_untouched` — apply-level guarantee
- `test_apply_with_target_conflict_exits_one` — CLI exit code

Plus `test_target_byte_equal_canonical_inner_is_already_at_target` to pin that the `exact` classification still maps to the already-at-target fast path. 25 tests pass; lint + format + mypy clean.

pandas-studio:
Folded the optional cleanups into bbb58d3 before merge:

**1. `is_noop` docstring/contract drift** — rewrote the docstring to spell out the two situations that land here ("no moves" / "every move is a conflict") and pointed callers at the conflict-pair check the CLI uses for the exit-1 gate. Implementation also switched to `not self.applicable_moves` so the property reads as the negation of the field it's tied to (instead of the obscure double-negation). Semantically equivalent — just less footgun-prone for future readers.

**2. Unused `scope` kwarg on `_print_migrate_plan_human`** — dropped. `plan.source_scope` covers it.

**3. JSON `needs_confirmation` test gap** — added `test_json_needs_confirmation_payload`. Pins that `--apply --json` without `--yes` against host targets refuses with `status="needs_confirmation"`, surfaces the offending paths in `host_writes`, and leaves disk untouched.

**4. `format_plan_summary` count-conservation invariant** — added `TestFormatPlanSummary` class with five tests, including `test_conservation_across_three_categories` that pins `fresh + already + conflict == len(moves)`. Per `feedback_count_conservation_invariant_pin.md` — a future status-class addition that forgets to plumb through the summary now fails fast.

Skipped: the `_safe_load_json_dict` dup with `settings_doctor` and the cross-module `_normalize_*` import. The third-caller threshold the reviewer flagged isn't tripped yet, so promoting to `_signatures.py` is premature; happy to do it the moment that pressure shows up.

826 targeted tests pass, lint + format + mypy clean.

PR #877, opened 2026-05-09
test(cli): strip MEMTOMEM_* env in subprocess e2e (sibling cleanup)

pandas-studio:
## Summary

- `packages/memtomem/tests/test_cli_index_noop_e2e.py` 의 subprocess test (`TestFreshNoopIndexSubprocess`) 의 env setup 에 `MEMTOMEM_*` strip + `USERPROFILE` 추가.
- Same fix as PR #875 fix-up commit `aa31531` (which applied the pattern to `test_context_cli_subprocess_e2e.py`); inline class is unchanged because `set_home` + `_CONFIG_PATH` patching already covers it.

## Why

PR #875 review note 2 follow-up commitment. Fix-up commit `aa31531` on #875 wrote: "The sibling `test_cli_index_noop_e2e.py` has the same gap; that one is left for a follow-up so this PR stays scoped to its own file." This is that follow-up.

The risk this closes: `HOME` override only isolates `~/.memtomem/config.json` reads. pydantic-settings still applies env-var overrides from the parent shell (e.g. `MEMTOMEM_INDEXING__MEMORY_DIRS` pointing at the developer's real memory dir) into the subprocess — un-hermeticizing the test. `helpers.isolate_memtomem_env` closes the gap for in-process tests; this PR reuses the same `_MEMTOMEM_ENV_VARS` constant for the subprocess path so any new top-level config section's env binding flows automatically.

`USERPROFILE` is added alongside `HOME` for Windows parity — `Path.home()` reads `USERPROFILE` first there, so a bare `HOME` override is silently ignored on Windows runners (already the lesson encoded in `helpers.set_home`).

## Test plan

- [x] `uv run ruff check packages/memtomem/tests/test_cli_index_noop_e2e.py` — passes
- [x] `uv run ruff format --check ...` — passes (no formatting change needed)
- [x] `uv run pytest packages/memtomem/tests/test_cli_index_noop_e2e.py -v` — 2 passed in 1.62s locally
- [ ] CI verifies on Linux / macOS / Windows (Windows is the meaningful cross-check for `USERPROFILE`)

### DIFF
--- a/packages/memtomem/tests/test_cli_index_noop_e2e.py
+++ b/packages/memtomem/tests/test_cli_index_noop_e2e.py
@@ -26,7 +26,8 @@ import pytest
 from click.testing import CliRunner
 
 from memtomem.cli import cli
-from .helpers import set_home
+
+from .helpers import _MEMTOMEM_ENV_VARS, set_home
 
 
 def _make_memory_dir(home: str) -> str:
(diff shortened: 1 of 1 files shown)


tsdata (approved):
LGTM — broadened MEMTOMEM_* strip to full prefix in both inline + subprocess classes per second-pass review (P2). Hostile env probe (MEMTOMEM_SEARCH__ENABLE_BM25=false + custom MEMTOMEM_INDEXING__MEMORY_DIRS) confirms hermeticity in both cases.

PR #881, opened 2026-05-09
test(web): cover Sync All severity branches

### LINKED ISSUE #880: test(web): cover non-aborted severity branches in Sync All + resolve !r.ok branch
## Context

PR #878 (test-only PR closing audit gap 3) pinned the **aborted** envelope on three flows because that was the audit P0 surface. The reviewer (suggestion 2) noted that the rest of the severity ladder is still unpinned — three branches that would benefit from regression nets even though they were not P0:

| Site | Branch | Spec gap |
|---|---|---|
| `static/context-gateway.js:492-493` | settings POST `error` → `toast.sync_failed` error | Add a 4th case to `test_context_gateway_sync_all.py` |
| `static/context-gateway.js:496-506` | settings POST `needs_confirmation` → info toast + Open Settings action button | Add a 5th case to the same spec |
| `static/settings-hooks-watchdog.js:136-139` | resolve POST `!r.ok` (HTTP 4xx/5xx) → toast.error, no `loadHooksSync` | Add a 3rd case to `test_settings_hooks_resolve_inline.py` |

The first two cases share the existing fixture shape from `test_sync_all_happy_path_emits_success_toast` / `test_sync_all_settings_aborted_emits_mtime_conflict_warning` — only the response payload changes. The third mirrors `test_resolve_aborted_envelope_keeps_card_and_emits_error_toast` but with HTTP 500 and the `err.detail` path.

## Suggested approach

- Same module-level i18n constants (`toast.sync_failed` template needs the `{error}` placeholder substituted in the assertion).
- Same symmetric pin discipline: positive on the toast text + class, negative on the success branch not firing.
- The `needs_confirmation` case also has an inline action button (`onClick: () => switchSettingsSection('hooks-sync')`) — pin that the click navigates to the Hooks Sync section.

## Out of scope

- Adding scenarios for non-overview / non-hooks Web routes (install flows etc.).
- ADR-0003 Status: Proposed → Accepted promotion (separate chore).

## Acceptance

- [ ] Three new test cases (one each in the two existing spec files).
- [ ] Mutation-validated.
- [ ] All `tests/web/` specs still green.

References: PR #878 review thread.

looooown2006:
## Summary

Adds targeted browser regressions for the remaining non-aborted severity branches requested in #880:

- Sync All settings `error` result emits the `toast.sync_failed` error toast and does not emit a success toast.
- Sync All settings `needs_confirmation` result emits the partial-info toast and its Open Settings action navigates to Hooks Sync.
- Hooks Sync resolve HTTP 500 surfaces `err.detail`, keeps the unresolved conflict card, and does not call `loadHooksSync`.

## Tests

- `git diff --check`
- `uv run ruff check packages/memtomem/tests/web/test_context_gateway_sync_all.py packages/memtomem/tests/web/test_settings_hooks_resolve_inline.py`
- `uv run pytest packages/memtomem/tests/web/test_context_gateway_sync_all.py packages/memtomem/tests/web/test_settings_hooks_resolve_inline.py -m browser`

Closes #880.

### DIFF
--- a/packages/memtomem/tests/web/test_context_gateway_sync_all.py
+++ b/packages/memtomem/tests/web/test_context_gateway_sync_all.py
@@ -39,6 +39,10 @@ pytestmark = pytest.mark.browser
 SYNC_ALL_TITLE = "Sync All"  # settings.ctx.sync_all
 SYNC_SUCCESS_TOAST = "Sync completed"  # settings.ctx.sync_success
 MTIME_CONFLICT_TOAST = "File was modified externally. Reloading..."  # settings.ctx.mtime_conflict
+SYNC_PARTIAL_NEEDS_CONFIRMATION_TOAST = (
+    "Sync All complete except Settings — confirm host writes in the Settings panel."
+)
+SYNC_FAILED_TEMPLATE = "Sync failed: {error}"  # toast.sync_failed
 
 
 _HEALTHY_OVERVIEW = {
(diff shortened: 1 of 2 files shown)


looooown2006:

pandas-studio (approved):
Thanks for the follow-up on #880 — these three regressions close exactly the gaps I was hoping to see covered.

I cross-checked each assertion against the production branches:
- `error` → `context-gateway.js:492-493` (`toast.sync_failed` template) ✅
- `needs_confirmation` → `context-gateway.js:496-506` (info toast + `Open Settings` action calling `switchSettingsSection('hooks-sync')`) ✅
- HTTP 500 resolve → `settings-hooks-watchdog.js:136-140` (`err.detail` toast, no `loadHooksSync` reload) ✅

Toast literals match `locales/en.json:677` (`toast.sync_partial_settings_needs_confirmation`) and `:681` (`toast.sync_failed`). Symmetric assertions throughout — positive (specific toast text + severity class) paired with negative (no success toast / preserved conflict card / unchanged GET counter) — exactly the pin-and-invert shape we want. `expect_request` synchronization on the resolve POST avoids the dispatcher race we've been bitten by elsewhere.

Approving as-is. Three non-blocking notes for possible future follow-ups (no need to address in this PR):

1. **Toast string parity isn't enforced by the tests.** `SYNC_PARTIAL_NEEDS_CONFIRMATION_TOAST` and `SYNC_FAILED_TEMPLATE` are inline duplicates of `en.json` strings; if the JSON changes, the assertion silently pins the old value. A future refactor that loads `en.json` once and references `toast.sync_partial_settings_needs_confirmation` / `toast.sync_failed` by key would close that drift surface. Same shape already exists for `SYNC_SUCCESS_TOAST` / `MTIME_CONFLICT_TOAST`, so this is pre-existing — not new debt introduced here.

2. **The `Open Settings` action label isn't asserted.** The `needs_confirmation` test verifies click → navigation but not that the button text resolves to `t('toast.open_settings_action')`. Navigation is the load-bearing behavior so this is cosmetic, but a label assertion would close the gap symmetrically with the toast-text assertion.

3. **Stubbed GET shape comment.** In `test_sync_all_settings_needs_confirmation_opens_hooks_sync`, the `**/api/settings-sync` stub returns `status: "conflicts"` with empty hook arrays — enough to render the section but no conflict cards. The `wait_for_function(... !classList.contains('hidden'))` is the right scope; a one-line comment naming the intent ("we only verify the navigation handler, not that conflicts render") would help future readers.

Squashing once branch protection is satisfied. Thanks again!

PR #888, opened 2026-05-10
feat(web): per-source AI summary in Source tab with heuristic fallback

pandas-studio:
## Summary

- New **Source tab preview** layer: every row now shows a heuristic title (first heading) + body (first paragraph). LLM-free, derived at read time from existing chunks.
- Opt-in **AI prose summary** (`indexing.auto_summarize=true` + `llm.enabled=true`): 2-3 sentence summary per source, cached in `_memtomem_meta`, refreshed only when the chunk content actually changes (SHA256 over per-chunk `content_hash`). LLM failure → silent fallback to heuristic, indexing never blocks.
- Opt-in **bulk regenerate** flow when `summary_language` changes: language drift surfaces as a Source-tab banner with [Regenerate all] / [Later]; clicking runs a singleton background task with a `/api/sources/regenerate-status` polling endpoint.

## Why

Filename + chunk count alone don't tell users what an indexed file is. The first 200 chars of a markdown file are often frontmatter / boilerplate, so a heuristic-only path under-delivers on long-tail files; a fully-LLM path is expensive across thousands of files. Splitting the two — heuristic always-on, LLM opt-in with content-aware caching — keeps cost proportional to actual content change while raising the floor of source identifiability.

## Design choices worth highlighting

- **`auto_summarize` defaults to false** (staged-default-flip pattern). LLM is gated by both `llm.enabled` and `auto_summarize` so existing users aren't auto-billed when they upgrade.
- **Signature is over chunk `content_hash`, not file mtime** — survives `git checkout` / `touch` / re-format passes that change mtime without changing meaning.
- **Language drift is *not* an auto-regen trigger** — flipping `summary_language` could otherwise silently re-summarise the whole corpus on next reindex. The banner makes the cost explicit.
- **Three-tier visual hierarchy**: filename bold mono (anchor), heading title sans muted (subtitle), AI prose ✨ + accent-tinted (rich preview) OR heuristic excerpt 78% opacity (fallback). Korean fonts synthesise italic poorly, so subtitle differentiation is via weight/size/color only.
- **Fail-soft everywhere**: LLM raise → log warning + return; orphan summary → cleared on `delete_by_source` even when no chunks remain.

## Test plan

- [x] `uv run ruff format --check ...` — clean
- [x] `uv run pytest -m "not ollama"` — **4576 passed, 10 skipped**
- [x] New test coverage: 16 summarizer (signature stability, prompt language, skip rules, fail-soft, bulk override), 11 storage (CRUD + drift queries + delete cleanup including orphan cache), 13 web routes (heuristic preview + AI summary + drift banner conditional + regenerate endpoints incl. idempotency + auto-disable rejection).
- [x] e2e against live Ollama (qwen2.5:0.5b): first index emits 1 `/api/generate` per file, unchanged reindex emits 0 (1 ms), single-file edit emits exactly 1, language flip + bulk regenerate clears banner + retags both records.
- [x] Playwright UI smoke: banner renders i18n-localised text, [Regenerate all] triggers polling and disappears on completion, ✨ AI prose vs heuristic excerpt rendered with `data-ai="true"` distinction.
- [x] AST architectural guards: new `POST /api/sources/regenerate-summaries` registered in CSRF + redaction registries (no user-supplied content payload — body-less trigger).

## Out of scope (follow-ups)

- Per-row "regenerate" button (signature-based auto-regen + bulk flow cover the common cases).
- `summary_language: auto` (per-source language detection).
- `auto_summarize` default flip — wait one or two release cycles + observe.


pandas-studio:
Addressed in `fb89a1a` — `reset_all()` now explicitly deletes `_memtomem_meta` rows under the `ai_summary:` prefix, so the "Delete ALL data" contract holds. Embedding-config rows (`embedding_dimension`, `embedding_provider`, `embedding_model`) are unaffected — that's pinned by a sibling regression test (`test_reset_all_preserves_embedding_meta_after_summary_clear`) that runs the path *with* a summary present so the LIKE filter is actually exercised.

The deletion count is surfaced in the `reset_all()` return dict as `ai_summaries: <n>` so reset receipts include it.

pandas-studio:
Addressed in `643ae97`. ``delete_chunks`` now captures each affected source's normalised path during its existing rowid lookup, and after the delete drops the ``ai_summary:`` row for any source whose remaining chunk count is zero. The chunk-id batch path (web chunk-delete fallback, dedup, decay) was the gap — full-source deletion already went through ``delete_by_source`` and was clean.

Partial deletions are intentionally left alone — the content-hash signature drifts on the next reindex and ``maybe_update_ai_summary`` regenerates cleanly, so clearing on every partial shave would just burn LLM budget on dedup/decay churn.

Three regression tests pin the matrix:
- full-empty source clears,
- partial deletion preserves,
- a mixed batch (one source emptied, another partially deleted) handles each correctly so a future shortcut like "clear all affected sources" can't silently regress.

pandas-studio:
Both addressed in `b420e6d`.

**P2-A (`delete_chunks` mid-transaction)**: the cleanup branch is now gated on `not self._in_transaction`. Reindex (`IndexingEngine._index_file`) wraps `delete_chunks` + `upsert_chunks` in one `storage.transaction()`, so the cache row stays put while the source is briefly empty between delete and upsert; the post-transaction `maybe_update_ai_summary` either refreshes it (cache hit on signature → no-op, signature drift → regenerate) or, if the LLM call fails / `auto_summarize` is off, leaves the previous prose intact — fail-soft. Standalone `delete_chunks` callers (web chunk-delete, dedup, decay) still get the eager cleanup.

**P2-B (rebuild path drops LLM)**: `_revert_to_stored` now threads `llm=app.llm_provider` into the rebuilt `IndexEngine`. Co-located comment in the call site so a future rebuild that copies this template can't drop the kwarg unnoticed. Sibling test pins both axes: engine swapped *and* LLM identity preserved.

Three regression tests added:
- in-transaction rewrite preserves the cache (delete + upsert in one transaction → summary survives),
- in-transaction delete-only leaves the cache to the outer scope (documents the deferred-cleanup contract so a future inversion flags design intent first),
- `revert_to_stored` keeps the LLM on the rebuilt engine (sentinel-LLM identity check, both axes).

90 targeted tests + ruff green.

pandas-studio:
Both addressed in `c54b8f0`. ``maybe_update_ai_summary`` now actively clears the stale cache on both invalidation paths instead of leaving it intact:

**Zero-chunk reindex**: when a previously-summarised file reindexes into zero chunks (emptied, became unchunkable, exclusion pattern moved), the row is evicted. The deferred-cleanup contract from the earlier fix means this hook is the *only* place that catches a genuine "now empty" reindex — the engine's transactional ``delete_chunks`` skips it on purpose to support rewrites.

**LLM failure on signature drift**: when the chunk signature changes (content edited) but the LLM call fails or returns empty, the row is evicted so ``/api/sources`` falls back to the heuristic excerpt instead of rendering prior prose against the new chunk set. Pre-fix users could see a ✨ AI summary describing content their file no longer contains — a silent correctness/privacy bug. The eviction is gated on ``cached is not None`` so the initial-pass failure mode (no prior cache, transient Ollama error) doesn't take a write lock per file.

A small ``storage.delete_ai_summary(path)`` primitive backs the new path — separate from ``delete_by_source`` so the summarizer can clear prose without touching chunks. Idempotent.

Six new tests pin the matrix: zero-chunk no-cache (no-op), zero-chunk with-cache (clears), exception+drift (clears), empty-output+drift (clears), exception+no-cache (no defensive delete — pinned so a future "always clear" simplification flags the cost trade-off first), and the storage primitive's round-trip + missing-row idempotence.

Full pytest: 4589 passed.

pandas-studio:
Addressed in `6f095fd`. ``update_chunks_scope_for_source`` now renames the AI summary cache row in the same transaction as the chunk-path rewrite. ``mm context memory migrate`` (and any other in-tree path relocation that reuses this method) preserves the cached prose at the new path and removes the orphan at the old path so it can't keep contributing to ``count_language_drift`` / ``get_all_ai_summaries``.

Renaming (rather than dropping) is the correct semantic — a path migration doesn't change what the file is *about*, only where it lives, so the prose stays valid and the content_hash signature still matches the migrated chunks. ``INSERT OR REPLACE`` against the new key handles the pathological crisscrossed-migration case where the destination already had a cache row; the source-of-truth is the row keyed by ``old`` since that's the one whose signature matched the chunks just rewritten. The old key is then explicitly deleted so the orphan can't drift back in.

Two regression tests pin the matrix:
- migration with cached summary: old key gone, new key carries the same record verbatim (summary, signature, language all preserved),
- migration with no cached summary: chunks move, no spurious meta rows appear at destination (defends the ``IF EXISTS``-style branch against a future "always insert" simplification).

Full pytest: 4591 passed.

memtomem (commented):
LGTM

PR #892, opened 2026-05-10
refactor(context): NO_FANOUT skip instead of raise in generate_all_/diff_ (#891)

### LINKED ISSUE #891: ADR-0011 E3/E4 follow-up: branch on scope before raising in generate_all/diff (PR #890 carry-over)
## Background

ADR-0011 PR-E follow-up (#890) converted six bare `assert x is not None` sites in `generate_all_*` and `diff_*` (across `agents.py` / `skills.py` / `commands.py`) to explicit `if x is None: raise RuntimeError(...)` so the `target_file` / `target_dir` contract survives `python -O`. The branches are **contractually unreachable on the default `scope="project_shared"`** because the runtime fan-out table never returns `NO_FANOUT` for that tuple.

E-series follow-ups (E3 — `mm context sync --scope`, E4 — `mm context migrate <kind> <name>`) will thread an explicit `scope` argument through these surfaces. When that lands, some `(artifact, runtime, scope)` tuples will legitimately resolve to `None` (e.g. every `project_local` entry per ADR-0011 §3, plus per-runtime exceptions like Codex commands). The current six raise sites would then fire as **false positives** for those legitimate cases.

The fail-loud RuntimeError message (`"… returned None for default project_shared scope"`) names the scope explicitly, so this would be a loud failure rather than a silent regression — but the message would be wrong (the caller's actual scope wouldn't be project_shared).

## Action — when E3/E4 lands

The six sites need to evolve from "single contract, raise on None" to "branch on scope before checking None":

| File | Line (post-#890) | Function |
|---|---|---|
| `context/agents.py` | ~624 | `generate_all_agents` |
| `context/agents.py` | ~905 | `diff_agents` |
| `context/commands.py` | ~419 | `generate_all_commands` |
| `context/commands.py` | ~777 | `diff_commands` |
| `context/skills.py` | ~252 | `generate_all_skills` |
| `context/skills.py` | ~497 | `diff_skills` |

Suggested shape (per site):

```python
target = gen.target_file(project_root, name, scope=resolved_scope)
if target is None:
    if resolved_scope in ("project_local",):  # or whatever NO_FANOUT scopes apply
        # legitimate — emit NO_PROJECT_FANOUT_FOR_RUNTIME skip and continue
        ...
    else:
        raise RuntimeError(...)  # contract violation for this scope
```

The `RUNTIME_FANOUT_TABLE` in `_runtime_targets.py` already has `NO_FANOUT` markers for the legitimate cases — the fix is at the consumer side (these six functions), not the table.

## Out of scope

- This issue does **not** request a PR until E3 or E4 starts. Filing now to avoid the contract drift surfacing as a green CI surprise mid-E3.
- The `extract_*_to_canonical` paths already handle `None` correctly via `runtime_fanout_root` short-circuits + `NO_PROJECT_FANOUT_FOR_RUNTIME` skip codes — only the `generate_all_*` / `diff_*` consumers are affected.

## References

- PR #890 — N1 commit (`e3b29d4`) introduced the six raise sites.
- ADR-0011 §3 — `project_local` has no runtime fan-out by design.
- `~/.claude/plans/delegated-wandering-lollipop.md` — master plan for E3/E4 (local).
- Codex review of #890 (delta round) — concern #4: "project_local legitimate-None path 의 cross-reference 부재".

pandas-studio:
## Summary

Convert six `raise RuntimeError(...)` sites in `generate_all_*` / `diff_*` (agents/skills/commands) to graceful NO_FANOUT skips so they don't false-positive once E3 (`mm context sync --scope`) wires non-default scope through these callers.

The fan-out table (`_runtime_targets.RUNTIME_FANOUT_TABLE`) is the contract source-of-truth — `KeyError` on missing tuple is the real fail-loud surface. The generator-level raise was a redundant defense-in-depth that becomes a wrong-message regression as soon as E3 starts threading scope through.

### Behavior change

- **`generate_all_*`**: emit a typed `NO_PROJECT_FANOUT_FOR_RUNTIME` skip `(artifact_name, "no fan-out for runtime <runtime> at this scope", code)` — slot 1 = artifact name (parallel to `PARSE_ERROR`), slot 2 = human-readable reason mentioning the runtime (per `_skip_reasons.py:6-9` contract).
- **`diff_*`**: probe `gen.target_*(project_root, "__probe_891__")` at the top of the per-generator loop and `continue` silently when `None` — so the realistic NO_FANOUT shape (canonical present + runtime side absent) returns zero rows for that runtime instead of leaking `"missing target"` rows from the canonical-only branch. A defensive per-name `None` fallback skip remains.

### Tests

- Rename `tests/test_context_runtime_table_none_raises.py` → `tests/test_context_runtime_table_none_skips.py`.
- Invert symmetric assertions per `feedback_pin_invert_symmetric_assertion.md`:
  - `generate_all_*`: positive marker (skip emitted with correct slots) + negative marker (no entry in `generated` for forced-None runtime).
  - `diff_*`: silent-skip marker (zero rows for forced-None runtime) — both with-runtime and canonical-only variants.
- New `test_diff_runtime_no_fanout_canonical_only_emits_zero_rows` variant uses `with_runtime=False` to cover the realistic NO_FANOUT shape (Codex review fold).
- Two paired mutations validated locally per `feedback_pin_test_mutation_validation.md`: drop `skipped.append(...)` (positive marker fires) + substitute `UNKNOWN_RUNTIME` for `NO_PROJECT_FANOUT_FOR_RUNTIME` (code-typed marker fires). Both fail for the right reason.

### Out of scope (deferred to E3)

- No `scope=` kwarg on `generate_all_*` / `diff_*`. Today's callers (CLI helpers + web routes) all use default `project_shared`; adding the kwarg without threading it through `_runtime_*_names` would create an incoherent partial implementation.
- No changes to the `_runtime_agent_names` / `_runtime_command_names` helpers (agents/commands) or the **inline** runtime-root code in `diff_skills` (skills.py:471-477). The helper-vs-inline asymmetry across the three artifacts is deliberately preserved here and queued for E3 unification onto `runtime_fanout_root`.
- Vendor override `_override.resolve(scope=...)` pins at agents.py:636 / skills.py:261 / commands.py:429 stay at `"project_shared"`. E3 may widen these per-scope.
- Full E3 diff semantics: this PR fixes the specific `"missing target"` leak Codex identified (canonical-only NO_FANOUT shape), but a user-facing `"no fan-out"` status string for `mm context diff --scope` is an E3 UX concern.
- No web route changes — PR-F handles UI scope selection.

### Verification

| Check | Result |
|---|---|
| `uv run pytest packages/memtomem/tests/test_context_runtime_table_none_skips.py -v` | 9 passed (3 generate × 1 + 3 diff × 2 variants) |
| `uv run pytest -m "not ollama" packages/memtomem/tests/test_context_*.py` | 581 passed |
| `uv run ruff check packages/memtomem/src` | clean |
| `uv run ruff format --check packages/memtomem/src` + new test file | clean |
| Codex pre-commit review (skip-tuple shape parity) | first round NEEDS-FIX (caught the `"missing target"` leak); fold round SHIP — zero blockers/majors/minors |
| Future-readiness probe (monkeypatch `target_* = None` against canonical fixtures for all 3 kinds) | all return `SkipResult` with `NO_PROJECT_FANOUT_FOR_RUNTIME`, no traceback |

No CHANGELOG entry — internal-only refactor with zero user-visible behavior change for default scope.

Closes #891.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_context_runtime_table_none_skips.py -v` — 9 passed
- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/test_context_*.py` — 581 passed
- [x] `uv run ruff check packages/memtomem/src` + `uv run ruff format --check packages/memtomem/src` — clean
- [x] Mutation validation (mutations 1 + 2 per pin docstring) — both fail for right reason, reverted
- [x] Codex review SHIP after fold


memtomem (approved):
All 11 CI green; Codex SHIP both rounds; mutation-validated. Admin-merging per feedback_green_ci_admin_merge.md.

PR #894, opened 2026-05-11
feat(gc): mm gc orphan-projects for deleted project_root cleanup (#884)

### LINKED ISSUE #884: [ADR-0011 follow-up] Orphan project chunk GC after directory deletion
**Tracked from:** ADR-0011 §"Open questions for the implementation issues"
([`docs/adr/0011-canonical-artifact-scope-hierarchy.md`](../blob/main/docs/adr/0011-canonical-artifact-scope-hierarchy.md))
— third bullet, "Orphan project chunk garbage collection."

> If a user indexes `/tmp/foo/.memtomem/memories/` then deletes the directory,
> stale rows remain in the user-local DB. Watcher prunes file-level deletions
> within tracked roots, but root-removal is a new case. Document as a v1
> known limitation in PR-B's release notes; file a follow-up issue.

**Status (post-#882):** v1 known limitation. Memory schema (PR-B) and
write surface (PR-D) shipped 2026-05-09 via #882; the orphan-row case
above is **not** addressed there. The watcher prunes file-level
deletions within tracked roots but the root-removal case (the project
directory itself goes away) leaves rows behind in `~/.memtomem/memtomem.db`
with `scope IN ('project_shared', 'project_local')` and a now-vanished
`project_root`.

### Symptom

```
$ mm add "team note" --scope project_shared --confirm-project-shared --yes  # in /tmp/foo/
$ rm -rf /tmp/foo
$ sqlite3 ~/.memtomem/memtomem.db "SELECT count(*) FROM chunks WHERE project_root LIKE '/tmp/foo%';"
# → still > 0
```

The rows are unreachable from default search inside any other project
(cross-project leak prevention in §6 filters by `project_root = <X>`),
but they bloat the DB and surface under explicit
`--scope=project_shared` cross-project queries.

### Acceptance criteria

- A sweep path (job, lazy-on-search, or `mm mem gc` subcommand —
  decision deferred to design phase) detects rows whose `project_root`
  no longer exists on disk and offers to remove them.
- The sweep is **opt-in** — no automatic deletion. ADR-0011 §8 ("no
  default flip, ever" for memory authorship) implies user-authored
  memory cannot disappear without explicit consent.
- Deleted-but-still-reachable case (mounted FS, removable disk) must
  not trigger false positives. Probably needs a confirmation step or a
  configurable timeout (e.g., "missing for ≥N days") before removal.
- A regression test pins the sweep behavior on a temp directory that
  is created → indexed → deleted → swept.

### Out of scope

- Cross-DB migration GC (separate follow-up — see chunk-link lineage
  preservation issue).
- Automatic GC on every `mm` invocation (would surprise users who
  unmount drives).

### References

- ADR-0011 §"Open questions" item 3
- PR #882 (PR-B/C/D bundle) — does not address this case
- Watcher implementation: `packages/memtomem/src/memtomem/indexing/watcher.py`

pandas-studio:
Closes #884.

## Summary

- Adds `mm gc orphan-projects` — finds chunks whose recorded `project_root` no longer exists on disk and, with `--apply`, removes them. Default is dry-run; `--apply` prompts per root; `--apply --yes` is the non-interactive form.
- Storage helpers live in a new module (`storage/orphan_gc.py`) so the four-surface cleanup (chunks + chunks_fts + chunks_vec + `_memtomem_meta` `ai_summary:*`) can be unit-tested against a real `sqlite3.Connection` without booting the Components stack.
- `SqliteBackend` gains thin async wrappers that thread `_has_vec_table` through to the helper.

## Why `mm gc` and not `mm context gc`

`mm context` operates on the Context Gateway artifact tree (agents / commands / skills under `.memtomem/`). This GC operates on user-local storage rows. Pulled the design review feedback into `mm gc` as the top-level home for storage-maintenance commands — the placement is explicit so future GC variants don't accidentally get reframed as Context Gateway work.

## Safety boundaries

- **`scope` predicate is strict:** the DELETE is `scope IN ('project_shared', 'project_local') AND project_root = ?`. User-scope rows that happen to carry a stale `project_root` value are deliberately untouched (ADR-0011 §8 — that is a separate authorship class).
- **Opt-in everywhere:** dry-run default; `--apply` requires interactive confirm; `--apply --yes` is the explicit non-interactive form; `--yes` without `--apply` is a usage error so the safety gate cannot be silently degraded.
- **AI-summary cache is defensive:** only dropped when no chunks remain for the source. A scope reclassification that left a user-scope row behind keeps its summary.
- **Unmounted FS / removable disks** are reported as orphaned — the per-root confirm prompt is the user-side check. `min-missing-days` was considered and dropped: `chunks.updated_at` means "chunk last update", not "missing since", so a real timeout gate would need a separate observed-missing state.

## Tests

- `tests/test_orphan_gc.py::TestStorageHelpers` — real-SQLite backend (`dimension=0` so the vec path is also covered by parameter, with the rowid-count gate on the column). Pins:
  - find returns only roots whose path is missing
  - find ignores user-scope rows with stray `project_root`
  - sweep cleans chunks + chunks_fts + ai_summary (all four surfaces)
  - sweep does not touch other project roots
  - sweep does not touch user-scope rows even when `project_root` matches
  - sweep keeps the ai_summary cache when other (user-scope) chunks still reference the source
  - sweep on an unknown root is a no-op
  - `SqliteBackend.find_orphan_project_roots()` / `sweep_orphan_project_root()` async wrappers
- `tests/test_orphan_gc.py::TestCli` — Click `CliRunner` with a mocked Components object pins dry-run-default / `--apply` confirm / `--apply --yes` non-interactive / `--yes` without `--apply` is a usage error.

## Test plan

- [x] `uv run ruff check packages/memtomem/src ...` clean
- [x] `uv run ruff format --check ...` clean
- [x] `uv run pytest packages/memtomem/tests -m "not ollama"` — 4733 passed, 10 skipped
- [x] `uv run mypy ...` clean on changed files
- [x] `uv run mm gc orphan-projects --help` renders

## Notes

- Branch was cut from `origin/main` (not from `feat/adr-0011-pr-e4`/#893) per design review. #893 also touches `cli/__init__.py:_register` to add `mm context migrate`, so a rebase after #893 merges may need to resolve the import + `cli.add_command` lines — both PRs add a new line in alphabetical order, conflict is mechanical.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/__init__.py
+++ b/packages/memtomem/src/memtomem/cli/__init__.py
@@ -45,6 +45,7 @@ def _register() -> None:
     from memtomem.cli.config_cmd import config
     from memtomem.cli.context_cmd import context
     from memtomem.cli.embedding_cmd import embedding_reset
+    from memtomem.cli.gc_cmd import gc
     from memtomem.cli.indexing import index
     from memtomem.cli.ingest_cmd import ingest
     from memtomem.cli.memory import add, recall
@@ -72,6 +73,7 @@ def _register() -> None:
     cli.add_command(config)
     cli.add_command(context)
     cli.add_command(embedding_reset)
+    cli.add_command(gc)
     cli.add_command(purge)
     cli.add_command(reset)
     cli.add_command(session)
--- a/packages/memtomem/src/memtomem/cli/gc_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/gc_cmd.py
@@ -1,0 +1,124 @@
+"""CLI: ``mm gc`` — storage maintenance commands (ADR-0011 follow-up #884).
+
+Currently exposes one subcommand:
+
+* ``mm gc orphan-projects`` — find chunks whose recorded ``project_root``
+  no longer exists on disk and, with ``--apply``, delete them. Default
+  is dry-run; ``--apply`` requires interactive confirmation;
+  ``--apply --yes`` is the non-interactive form.
+
+Placed at the top level (not under ``mm context``) because the rows it
+cleans live in user-local storage, not the Context Gateway artifact
+tree. The ``gc`` group is the natural home for any future storage
+maintenance subcommands (#884 review point 1).
+"""
+
+from __future__ import annotations
+
+import asyncio
+
+import click
+
+
+@click.group("gc")
+def gc() -> None:
+    """Storage maintenance: garbage-collect stale rows in the user-local DB."""
+
+
+@gc.command("orphan-projects")
+@click.option(
+    "--apply",
+    "apply_",
+    is_flag=True,
+    help="Actually delete. Without this flag, prints what would be deleted (dry-run).",
+)
+@click.option(
+    "--yes",
+    "assume_yes",
+    is_flag=True,
+    help="Skip the per-root confirmation prompt. Requires --apply.",
+)
+def orphan_projects(apply_: bool, assume_yes: bool) -> None:
+    """Remove chunks whose ``project_root`` directory no longer exists.
+
+    When you index a folder under ``--scope project_shared`` or
+    ``project_local`` and later delete the folder, the rows survive in
+    ``~/.memtomem/memtomem.db`` with a now-vanished ``project_root``.
+    They are hidden by the default in-project search filter but bloat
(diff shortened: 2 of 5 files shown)


memtomem (approved):
LGTM

PR #896, opened 2026-05-11
fix(web): keep Discovered dirs visible when filter is active

pandas-studio:
## Summary

- The Sources tab's `_renderMemorySourceTree` hides all Discovered dirs
  (chunks=0, files>0) whenever `filterActive` is true. For a user with
  30+ auto-discovered `~/.claude/projects/*/memory` entries, setting the
  path-filter input or clicking an NS card (which sets
  `STATE.sourcesNsFilter`) collapses the entire vendor down to the one
  indexed dir.
- The original `Claude (0)` guard was scoped to indexed-empty rows, not
  Discovered ones. Restore the Discovered carve-out so the filter only
  narrows indexed dirs; Discovered dirs always render alongside.
- Follow-up commit (codex review): the later empty-state guard tripped
  on `!plan.totalFiles` too, so a discovered-only vendor or a filter
  with zero indexed matches still got replaced with "No matches."
  Track `plan.discoveredCount` separately and gate the fallback on
  it so the carve-out survives the empty-state check.

## Files

- `app.js:3245-3253` — filter predicate gains `|| isDiscovered(d)`.
- `app.js:3270-3290` — `plan.discoveredCount` field; empty-state guard
  consults it alongside `totalFiles` and `isEmptyVendor`. `totalFiles`
  keeps its "indexed + orphans" shape so the sub-tab badge meaning is
  unchanged.
- `index.html` — `app.js` cache-bust bumped `v=112 → v=114`.
- `tests/test_web_sources_tree.py` — new JS source-grep file with three
  pins (carve-out, `isDiscovered` predicate shape, empty-state guard).

## Repro (before fix)

1. Config with one user-tier dir plus many auto-discovered
   `~/.claude/projects/*/memory` dirs (chunks=0, files>0).
2. Open Sources → Claude sub-tab. Discovered group lists every project
   dir.
3. Index a single dir (or click an NS card so
   `STATE.sourcesNsFilter` is set).
4. All other Discovered dirs vanish — only the indexed one remains.

After fix: Discovered dirs stay visible regardless of filter state,
and a vendor with *only* Discovered dirs no longer falls through to
the "No matches for that filter" placeholder.

## Test plan

- [x] `pytest packages/memtomem/tests/test_web_sources_tree.py` — 3 PASS.
- [x] Mutation check: dropping `|| isDiscovered(d)` from the filter
      predicate fails the carve-out pin; dropping `!plan.discoveredCount`
      from the empty-state guard fails the new follow-up pin
      (per `feedback_pin_test_mutation_validation.md`).
- [ ] Manual UI smoke (`uv run mm web` → Claude sub-tab → type into
      filter input, click NS card): Discovered list stays populated.

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -3275,12 +3275,21 @@ function _renderMemorySourceTree(sources, list) {
       (sum, [, indexed]) => sum + indexed.reduce((s, d) => s + (sourcesByDir[d] || []).length, 0),
       0,
     ) + vendorOrphans.length;
+    // Tracked separately from ``totalFiles`` so the sub-tab badge keeps its
+    // "indexed + orphans" meaning while the empty-state guard further down
+    // can still see discovered dirs (codex review on #896: a vendor with
+    // only Discovered dirs would otherwise hit the "No matches" fallback
+    // because ``totalFiles`` is zero, hiding the section the carve-out in
+    // ``visibleCatsRaw`` just preserved).
+    const discoveredCount = visibleCats.reduce(
+      (sum, [, , discovered]) => sum + discovered.length, 0,
+    );
     const visibleIndexedCats = visibleCats.filter(([, indexed]) => indexed.length);
     const isSingleLeaf = visibleIndexedCats.length === 1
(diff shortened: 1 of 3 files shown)


memtomem (approved):
LGTM

PR #898, opened 2026-05-11
feat(status): dense-vector coverage in mem_status + /api/embedding-status

pandas-studio:
## Summary

When an embedder fails to initialize (ONNX model download fails,
Ollama unreachable, NoopEmbedder fallback) the indexing pipeline
quietly falls through to a BM25-only path: chunks land in `chunks` +
`chunks_fts` but never get an embedding row in `chunks_vec`. Semantic
search returns nothing while keyword search keeps working, and no
existing surface tells operators why. `mem_status` only reports
configured provider/dimension, not what's actually in the database.

This PR adds one coverage signal that both the MCP/CLI status report
and the Web UI status endpoint consume.

## What changes

- **`SqliteBackend.get_dense_coverage`** returns `{"total": N,
  "with_dense": M}`. Falls back to `with_dense=0` when `chunks_vec`
  is absent (post `embedding-reset --mode purge` or pre-first-index).
- **`GET /api/embedding-status`** grows a `coverage` field shaped
  `{total, with_dense, percent}`. `percent` rounds to one decimal
  and is `0.0` for empty databases so the SPA can render the value
  with no divide-by-zero guard.
- **`mem_status` / `mm status`** add a `Dense vectors: M/N (P%)`
  line:
  - `with_dense == 0 && total > 0` → `(BM25-only — dense retrieval
    will return nothing)` hint.
  - `0 < with_dense < total` → `(partial dense coverage — some
    chunks BM25-only)` hint.
  - Full coverage stays quiet so the happy path doesn't acquire
    noise.

Both surfaces gate the new field on `hasattr(storage,
"get_dense_coverage")`, so older test doubles (and any future
storage backend that hasn't grown the method) leave `coverage` at
`None` / omit the line rather than 500ing the endpoint.

## Scope (intentional)

Stage 1 of the embedding-coverage thread. Telemetry only — no change
to how indexing decides whether to embed. Stage 2 (separate PR) will
make the underlying init failure loud (engine-level abort instead of
silent BM25-only fallback). Splitting keeps the review surface scoped
to "expose what's there" before "change what runs."

## Test plan

- [x] `pytest tests/test_storage_extended.py` — three new tests
      covering empty DB, full coverage after upsert, and the
      reset-vec-table-then-chunks-survive shape that mirrors the
      production failure mode.
- [x] `pytest tests/test_web_routes.py::TestEmbeddingStatus` — full,
      BM25-only, partial-rounding, and empty-DB cases on the API.
- [x] `pytest tests/test_cli_status.py::TestStatusOutput` — four new
      cases for the CLI/MCP text shape (full, BM25-only with hint,
      partial with hint, skipped-when-method-missing).
- [x] `pytest tests/test_storage_extended.py tests/test_cli_status.py
      tests/test_web_routes.py -q` — 216 PASS.

## Out of scope

- The actual init-failure abort lives in a follow-up PR (Stage 2).
- No Sources tab badge UI yet — operator-facing telemetry first;
  per-chunk UI badge deferred until we see how the rollup reads.

### DIFF
--- a/packages/memtomem/src/memtomem/server/tools/status_config.py
+++ b/packages/memtomem/src/memtomem/server/tools/status_config.py
@@ -107,6 +107,31 @@ async def format_status_report(app: AppContext) -> str:
     except Exception:
         logger.debug("Orphan detection failed", exc_info=True)
 
+    # Dense-vector coverage line. The hint at the end surfaces the
+    # BM25-only run case loudly: an embedder that crashed mid-init or
+    # fell back to NoopEmbedder will still index chunks into ``chunks``
+    # + ``chunks_fts`` but skip ``chunks_vec`` entirely, so semantic
+    # search returns nothing while keyword search keeps working. The
+    # check is gated on ``hasattr`` so older storage doubles that
+    # haven't grown the method don't break the report.
+    if hasattr(app.storage, "get_dense_coverage"):
+        try:
+            cov = await app.storage.get_dense_coverage()
+            total = int(cov["total"])
+            with_dense = int(cov["with_dense"])
+            if total > 0:
+                pct = round((with_dense / total) * 100, 1)
+                hint = ""
+                if with_dense == 0:
+                    hint = "  (BM25-only — dense retrieval will return nothing)"
+                elif with_dense < total:
+                    hint = "  (partial dense coverage — some chunks BM25-only)"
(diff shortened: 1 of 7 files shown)


memtomem (approved):
LGTM

memtomem (approved):
Approving — Codex SHIP and CI is all green.

PR #903, opened 2026-05-11
fix(indexing): abort when embedder dim=0 with a real model name

pandas-studio:
## Summary

Stage 2 of the embedding-coverage thread (#898 was Stage 1 telemetry).

`IndexEngine._index_file` previously had a single `dimension > 0`
gate around the embedding call. When the gate was false the engine
silently dropped to a BM25-only path: chunks landed in `chunks` +
`chunks_fts` while `chunks_vec` stayed empty, with no log line
explaining why semantic search would return nothing afterwards.

The gate was originally meant for the explicit BM25-only mode
(`embedding.provider == "none"` → `NoopEmbedder` reports `dim=0`)
but also caught the misconfigured-embedder case as collateral
damage: an ONNX / Ollama provider whose init crashed could end up
returning `dim=0` and the engine would happily index BM25-only
chunks under what the user thought was a dense-enabled corpus.

## What changes

- **Split the two `dim=0` cases by `model_name`.** `NoopEmbedder` is
  the sentinel `"none"` model and keeps the silent skip — intentional
  BM25-only mode is unchanged. Any other embedder reporting `dim=0`
  now aborts with an explicit error message that flows back through
  the same return shape `IndexingStats` already exposes for the
  embedding-failure path.
- **`return` happens before the delete/upsert transaction**, so
  storage stays consistent (chunks + chunks_vec both untouched for
  this file).
- **Drive-by fix:** the single-file `index_file` return path was
  dropping `result["errors"]` on the floor. The embedding-failure
  path already populated `errors=[...]` but
  `IndexingStats(..., new_chunk_ids=...)` omitted `errors=` so the
  message never made it past the engine. Plumb it through too so the
  new abort message (and any pre-existing embedding-failure message)
  propagates to the caller — without this the new abort would log
  but its message would still be invisible to callers reading
  `IndexingStats.errors`.

## Test plan

- [x] `test_dim_zero_with_real_model_aborts` — abort when an embedder
      with `dimension=0` but `model_name="bge-m3"` (or any non-noop
      model) is asked to index. Asserts zero chunks landed in
      storage and that the error message names both the model and
      the BM25-only failure mode so users can grep for it.
- [x] `test_dim_zero_with_noop_provider_indexes_silently` — keeps
      the explicit BM25-only opt-in path working. NoopEmbedder
      (`model_name == "none"`) indexes normally with empty errors.
- [x] Mutation check: dropping the new gate (force the abort
      condition to `False`) fails the abort test, confirming the
      guard isn't being short-circuited elsewhere.
- [x] `pytest tests/test_indexing_engine.py -q` — 148 PASS
      (including the wider suite that already exercises NoopEmbedder
      paths under `bm25_only_components`).

## Out of scope

- Per-chunk dense/BM25 badge UI — deferred until coverage telemetry
  from #898 lands and we see how the rollup signal reads in
  practice.
- Recovery flow (e.g., `mm reindex --force` after an embedder fix)
  — orthogonal; the abort already leaves storage consistent so
  re-running indexing after fixing the embedder picks up the same
  set as a fresh run.

### DIFF
--- a/packages/memtomem/src/memtomem/indexing/engine.py
+++ b/packages/memtomem/src/memtomem/indexing/engine.py
@@ -568,6 +568,7 @@ class IndexEngine:
             skipped_chunks=result["skipped"],
             deleted_chunks=result["deleted"],
             duration_ms=duration,
+            errors=tuple(result.get("errors", ())),
             new_chunk_ids=tuple(result.get("new_chunk_ids", ())),
         )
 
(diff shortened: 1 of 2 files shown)


memtomem (approved):
LGTM

memtomem (approved):
Approving — Codex SHIP and CI is all green.

PR #882, opened 2026-05-09
feat(scope): ADR-0011 memory scope axis (PR-B + PR-C + PR-D)

pandas-studio:
## Summary

ADR-0011 §9 의 첫 세 phase (PR-B / PR-C / PR-D) 를 한 PR 으로 묶어 land 한다. ADR 자체는 별도 (`fe27394` on `feat/adr-0011-canonical-scope-hierarchy`) 에서 review 중이며 이 구현은 그 RFC 의 verbatim spec 을 따른다. PR-E (agents/skills/commands) 와 PR-F (Web UI + docs) 는 후속 PR.

- **PR-B** (`ec84a15`) — schema (`chunks.scope`, `chunks.project_root` columns) + `IndexingConfig.project_memory_dirs` + `_resolve_scope` / `_apply_scope` + `enforce_write_guard` Gate A. 사용자 facing CLI/MCP 표면 변화 없음.
- **PR-C** (`ce02e61`) — read surface: `ScopeFilter`, `scope_context_sql` always-on fragment, project-aware default merge per ADR-0011 §6, MCP read tools, CLI `--scope` (comma-list).
- **PR-D** — write surface. 두 단계로 분할:
  - `54e3934` — Gate A on `enforce_write_guard`, Gate B (`mem_add` / `mem_batch_add` explicit `scope` + `confirm_project_shared`), `mem_batch_add` per-entry `enforce_write_guard` (inline-scan bypass 제거), `mem_edit` inferred-scope.
  - `e3931f3` — `_mem_add_core` / `mem_batch_add` scope-aware **write target** (MCP `mem_add(scope='project_shared')` 가 실제 project tier 디렉터리에 land — CLI/MCP divergence 해소), `_validate_path` 의 base 에 `project_memory_dirs` 추가, `mem_delete` Gate B + bulk source-file scope-set probe (`list_scopes_by_source`), `mm mem add` CLI `--scope` / `--confirm-project-shared` / `--yes` 플래그, 신규 shared module `memtomem.memory_scope`, storage protocol methods (`list_scopes_by_source`, `update_chunks_scope_for_source`).
  - `59221c1` — `mem_consolidate_apply` cross-scope rejection (chunk_ids 가 truth source, mixed/project_shared 모두 user-visible 메시지로 skip).
[...]

PR #905, opened 2026-05-11
feat(rescan): mm mem rescan + mm context rescan privacy-only audit (#885)

### LINKED ISSUE #885: [ADR-0011 follow-up] Validate demand for mm mem rescan / mm context rescan
**Tracked from:** ADR-0011 §"Open questions for the implementation issues"
— fourth bullet, `mm mem rescan` / `mm context rescan` demand validation.

> The use cases for `mm mem rescan` and `mm context rescan` are spec'd
> to two scenarios in §5 (STM pattern sync; git-mv ingest bypass). If
> neither has a concrete user reporting before PR-D / PR-E, drop the
> subcommands from v1 and re-add when a real use case shows up.

**Status (post-#882):** Per the §"Open questions" default rule above,
`mm mem rescan` was **dropped from v1** in PR-D — neither STM-pattern
sync nor git-mv ingest bypass was reported as a concrete user need
before #882 shipped. `mm context rescan` was similarly scoped out of
PR-E by the same rule (recorded in the §9 PR-E phasing row of
[`docs/adr/0011-canonical-artifact-scope-hierarchy.md`](../blob/main/docs/adr/0011-canonical-artifact-scope-hierarchy.md)).

This issue exists as a **monitoring tracker**, not an active task.

### Re-add criteria

Re-open implementation when **either**:

1. A concrete user / team reports: "I edited `<project>/.memtomem/memories/foo.md`
   directly with my editor (or `git mv`'d a file into a tracked
   memory dir) and `mem_search` doesn't reflect the change. I expected
   memtomem to re-scan." Watcher misses by design (the trust boundary
   in `privacy.py:7-26` means ingest must run, not just an inotify
   touch).
2. STM ships a new secret-class pattern (the asymmetric sync rule in
   `privacy.py:7-26`) and a memtomem deployment with already-indexed
   project_shared content needs to retroactively re-scan against the
   new pattern set without re-indexing every chunk.

If either case is reported, design a minimal `--scope=<scope>` rescan
that runs `enforce_write_guard` again across the affected scope without
deleting/re-creating chunks.

### Out of scope until re-opened

- Auto-installing pre-commit hooks (explicit non-goal in §5).
- Hooking rescan into the watcher (§5 rationale: bypassable ingest
  scanning contradicts the trust boundary).

### Closing rule

If 6 months pass with zero reports against this issue, close it as
"deferred indefinitely" rather than letting it linger.

### References

- ADR-0011 §"Open questions" item 4
- ADR-0011 §5 trust boundary (`privacy.py:7-26`)
- PR #882 (PR-D shipped without `mm mem rescan`)

pandas-studio:
## Summary

Closes #885 (re-add criteria satisfied by explicit maintainer request).
Lands `mm mem rescan` and `mm context rescan` as **privacy-only audit**
subcommands — the scenario-2 shape from ADR-0011 §"Open questions"
item 4 (STM secret-pattern sync against already-indexed
`project_shared` content). Scenario 1 (`git mv` ingest bypass / full
reindex) stays out of scope and gets a separate `--reindex` follow-up
issue if reported.

### Policy override

Per ADR-0011 §"Open questions" item 4, both subcommands were dropped
from v1 in #882 (PR-D) and #889 (PR-E) under the "no concrete user
report → defer" rule. This PR overrides that default on explicit
maintainer request. The "6 months / zero reports → close as deferred
indefinitely" closing rule on #885 no longer applies; the issue body
is updated separately to record the override.

### What ships

`mm mem rescan --scope <scope> [--source <path>] [--json] [--quiet]`
- Streams stored chunks via the new
  `StorageBackend.iter_chunks_for_audit` enumerator, re-runs
  `enforce_write_guard(force_unsafe=False, record_outcome=False)`
  per chunk, and reports violations.
- `--scope` is the **guard decision scope**, required (no default).
  v1 only reaches `decision="pass"` or `"blocked"` — the
  `blocked_project_shared` branch at `privacy.py:493` is unreachable
  because `force_unsafe` is always `False`.
- `--source` resolves cwd-relative with strict normalization: file →
  exact source match, directory → component-aware descendant prefix
  via `Path.relative_to` (no fuzzy / substring). Nonexistent path
  exits 2.
- Chunk identity is preserved end-to-end: no delete, no re-embed, no
  validity-window mutation. Counters / bypass audit lines stay
  untouched (Test 9 pins the zero-drift invariant).

`mm context rescan --scope <scope> [--json] [--quiet]`
- v1 walks **only** the project-root generated-file set returned by
  `detect_agent_files` (`CLAUDE.md`, `.cursorrules`, `GEMINI.md`,
  `AGENTS.md`, `.github/copilot-instructions.md`).
- Agents / skills / commands runtime fanout (the scope-tiered
  `_runtime_targets.py:71` table) is **out of scope** and tracked
  separately.
- Reuses `scan_artifact_tree` with `on_blocked="skip_warn"` and the
  newly-added `record_outcome=False` parameter (default `True`,
  existing sync callers unchanged).

### Exit codes

- `0` — no violations
- `1` — at least one violation
- `2` — invocation / config error

### JSON schema

```json
{
  "scope": "project_shared",
  "scanned": 47,
  "violations": [
    {
      "chunk_id": "abc12345",
      "source": "/abs/docs/architecture.md",
      "scope": "project_shared",
      "decision": "blocked",
      "hits": [
        {"pattern_index": 6, "span_start": 240, "span_end": 272}
      ]
    }
  ]
}
```

`decision` is the raw `WriteGuardResult.decision` value (no scope
baked into the string). `scope` is a separate field. `hits` preserves
`pattern_index` + `span_start` + `span_end` from `RedactionHit`. Never
echoes matched bytes (per the `RedactionHit` docstring contract at
`privacy.py:260-269`). `mm context rescan` substitutes `path` for
`chunk_id`/`source` because the unit is a filesystem artifact, not a
stored chunk.

### Out of scope (named follow-ups)

- `mm mem rescan --reindex` for the `git mv` ingest-bypass shape.
- Agents / skills / commands runtime fanout rescan.
- Quarantine / soft-delete action on violating chunks.
- Migrating `mm add` / `mm recall` under `mm mem`.
- Pre-commit hook autoinstall + watcher-coupled rescan (both ADR-0011
  §5 non-goals — bypassable ingest scanning contradicts the trust
  boundary at `privacy.py:7-26`).

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_mem_rescan.py` —
      13 cases pass (clean, secret, pin-and-invert clean, JSON schema,
      `blocked` ≠ `blocked_project_shared`, scope filter, counter
      drift, `--source` exact / prefix / nonexistent).
- [x] `uv run pytest packages/memtomem/tests/test_context_rescan.py` —
      8 cases pass (clean, secret, pin-and-invert clean, JSON schema,
      `skip_warn` surfaces all violations, counter drift).
- [x] Mutation validation per
      `feedback_pin_test_mutation_validation`: forcing
      `decision="pass"` regardless of guard output fails 11 of the new
      tests; restoring the guard restores green.
- [x] Regression sweep `pytest -k "privacy or context or storage or
      cli or scope or rescan"` — 1661 pass.
- [x] `uv run ruff check` + `ruff format --check` clean on every
      modified file.
- [x] `uv run mypy` clean on the new source files (advisory).

### DIFF
--- a/packages/memtomem/src/memtomem/cli/__init__.py
+++ b/packages/memtomem/src/memtomem/cli/__init__.py
@@ -48,6 +48,7 @@ def _register() -> None:
     from memtomem.cli.gc_cmd import gc
     from memtomem.cli.indexing import index
     from memtomem.cli.ingest_cmd import ingest
+    from memtomem.cli.mem_cmd import mem
     from memtomem.cli.memory import add, recall
     from memtomem.cli.purge_cmd import purge
     from memtomem.cli.reset_cmd import reset
@@ -68,6 +69,7 @@ def _register() -> None:
     cli.add_command(search)
     cli.add_command(add)
     cli.add_command(recall)
+    cli.add_command(mem)
     cli.add_command(index)
     cli.add_command(ingest)
     cli.add_command(config)
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -65,7 +65,7 @@ from memtomem.context.generator import (
     extract_sections_from_agent_file,
 )
 from memtomem.context.parser import CONTEXT_FILENAME, parse_context, sections_to_markdown
-from memtomem.context.privacy_scan import PrivacyScanError
+from memtomem.context.privacy_scan import PrivacyScanError, scan_artifact_tree
 from memtomem.context.settings import (
     diff_settings,
     generate_all_settings,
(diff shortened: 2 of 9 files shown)


pandas-studio:
Codex P2 addressed in 6a56476.

- `iter_chunks_for_audit` `--source` directory prefix now uses `substr(source_file, 1, ?) = ?` instead of `LIKE '<prefix>/%' ESCAPE '\\'` so the match is case-sensitive (binary string compare), independent of any session PRAGMA, and avoids LIKE/GLOB metachar escape concerns entirely.
- New storage-level test file `tests/test_storage_iter_chunks_for_audit.py` pins the regression directly. The case-sensitivity tests use non-existent absolute paths so `Path.resolve(strict=False)` is a no-op — otherwise macOS's default case-insensitive APFS would fold `docs` / `DOCS` on resolve and hide the SQL-layer bug.
- Mutation validation: re-introducing the LIKE predicate fails the uppercase-sibling test; restoring substr-equality restores green.
- Regression sweep `pytest -k "privacy or context or storage or cli or scope or rescan or audit"` — 1673 pass.

pandas-studio:
Codex P2-b (Windows separator) addressed in 154045a.

- Prefix anchor now uses `os.sep` and `rstrip("/\\")` so the SQL `substr`-equality match works on Windows (`\\`-separator stored paths) and tolerates POSIX-style input on either platform.
- New `TestSourcePrefixSeparator` class simulates a Windows runtime by monkey-patching both `os.sep` AND `sqlite_backend.norm_path` — required because `Path.resolve()` on the host (POSIX in CI) keeps emitting `/` regardless of how `os.sep` is patched (per `feedback_pin_test_mutation_validation`).
- Mutation: reverting to `rstrip("/") + "/"` fails both new tests; restoring `os.sep` restores green.
- Regression sweep `pytest -k "privacy or context or storage or cli or scope or rescan or audit"` — 1675 pass (1673 + 2 new Windows-shape).

pandas-studio:
Windows CI failure addressed in 97ca552.

`TestSourcePrefixCaseSensitivity` fixtures used POSIX-style strings (`/__rescan_fixture__/docs/...`). On Windows, `norm_path` → `Path.resolve()` rewrote the filter prefix into `<drive>:\__rescan_fixture__\docs` form, and the production anchor uses `os.sep` (`\\`), so the stored POSIX strings never matched the Windows-form anchor → all three tests reported zero rows.

Fix: a new `_fx(*parts)` helper builds platform-native fixture paths with `os.sep` for both stored `source_file` and the monkey-patched `norm_path` return value. Same shape `TestSourcePrefixSeparator` already used for the explicit separator regression — case-sensitivity now goes through the same patch so macOS APFS case-insensitivity is still bypassed without re-introducing the cross-platform mismatch.

Local re-run: 30 pass. Windows CI re-running now.

memtomem (approved):
Reviewed Codex output + CI matrix (ubuntu/macOS/Windows all green after 97ca552 portability fix). Privacy-only audit contract is consistent with ADR-0011 §5 — chunk identity preserved, record_outcome=False threaded end-to-end, project_shared content stays in git history with no force-bypass valve. Approving for admin merge.

PR #907, opened 2026-05-11
Polish language toggle display

pandas-studio:
## Summary
- Show the current UI language as EN/KO instead of the next target language
- Style the language toggle as a smaller header chip
- Bump the stylesheet cache version

## Tests
- uv run pytest packages/memtomem/tests/test_i18n.py


memtomem (approved):
LGTM

Issue #900, opened 2026-05-11
[feature]: Extract shared Context Gateway artifact sync service


pandas-studio:
## Problem

Context Gateway artifact sync logic is converging on the same lifecycle across skills, agents, and commands:

1. Enumerate canonical artifacts.
2. Resolve runtime fan-out targets.
3. Apply drop/strict rules where relevant.
4. Apply per-vendor overrides where relevant.
5. Run sync-side Gate A privacy checks.
6. Write or promote final runtime outputs.
7. Return generated/skipped/dropped summaries for CLI, MCP, and Web surfaces.

Today that flow is still implemented separately across artifact modules. The duplication makes it harder to keep privacy behavior, skip codes, strict-drop semantics, and cross-surface error translation consistent.

## Proposed solution

Extract a shared artifact sync service after the scope model is settled.

Sketch:

- Define per-artifact adapters for parse/render/override/target resolution.
- Define a common sync plan dataclass that separates preflight from mutation.
- Keep artifact-specific output shapes (`dropped` for agents/commands, directory promotion for skills), but share the privacy and summary mechanics.
- Preserve existing public APIs (`generate_all_skills`, `generate_all_agents`, `generate_all_commands`) as thin wrappers.

## Acceptance criteria

- No behavior change in CLI/MCP/Web responses.
- Existing scope, privacy, runtime-table, and route tests remain green.
- New service-level tests cover shared failure modes: privacy block, unreadable source, `NO_PROJECT_FANOUT_FOR_RUNTIME`, unknown runtime, strict drop, and override bytes.

## Alternatives considered

- Leave the modules separate and rely on targeted tests. This is lower risk short-term but increases drift risk as scope-aware Web work lands.
- Extract only privacy scan helpers. That already exists; the remaining duplication is in planning and mutation ordering.

## Additional context

This should follow the Context Gateway scope-model RFC/design issue, since service interfaces should use the final project-root vs target-tier vocabulary.

pandas-studio:
# Behavior inventory: current artifact sync surface

> Status: read-only inventory of the existing `generate_all_skills` /
> `generate_all_agents` / `generate_all_commands` flows, intended as input
> for the future service extraction. **No service API names, dataclass
> shapes, or module layouts are proposed here.** Anchors reference
> `b4957c5` (`origin/main` at posting time).
>
> Pre-requisite: the scope-vocabulary RFC tracked in #901
[...]

PR #908, opened 2026-05-11
test(context): pin strict-drop partial-write boundary for #900 refactor

memtomem:
## Summary

- Add `test_strict_drop_preserves_earlier_writes` to both `TestStrictMode` classes (`test_context_agents.py`, `test_context_commands.py`).
- Lock in the pre-existing partial-write boundary in `generate_all_agents` / `generate_all_commands` Phase 2: when `on_drop="error"` hits a dropping canonical, earlier canonicals in the pending list have already been written to disk by `atomic_write_text`.
- Behavior-preserving — no production code changes. Two test files only.

## Why now

Issue #900 plans to extract the artifact sync flow into a shared service. The [behavior-inventory comment](https://github.com/memtomem/memtomem/issues/900#issuecomment-4419280062-anchor) on #900 calls out this `StrictDropError` mid-Phase-2 raise as a real, intended atomicity boundary distinct from the all-or-nothing `project_shared` privacy-block contract. Without a pinning test, a refactor that adds full-batch atomicity to the strict-drop path would silently change observable behavior. These tests catch that.

The new tests use the existing fixtures and `SAMPLE_*` constants in each file; canonicals are processed in sorted-name order (`list_canonical_agents` / `list_canonical_commands`), so `alpha-minimal` is guaranteed to be written before the raise on `beta-full`.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_context_agents.py::TestStrictMode packages/memtomem/tests/test_context_commands.py::TestStrictMode -v` — 6/6 pass.
- [x] `uv run pytest packages/memtomem/tests/test_context_agents.py packages/memtomem/tests/test_context_commands.py -q` — 119/119 pass (no regression in surrounding suites).
- [x] `uv run ruff check` + `ruff format --check` on the two test files — clean.

Refs #900

### DIFF
--- a/packages/memtomem/tests/test_context_agents.py
+++ b/packages/memtomem/tests/test_context_agents.py
@@ -277,6 +277,34 @@ class TestStrictMode:
         )
         assert len(result.generated) == 2
 
+    def test_strict_drop_preserves_earlier_writes(self, tmp_path):
+        # Pre-existing partial-write boundary documented on issue #900:
+        # Phase 2 raises StrictDropError for the first dropping canonical,
+        # but earlier canonicals in pending order have already been written.
+        # Canonicals iterate in sorted-name order (list_canonical_agents),
+        # so "alpha-minimal" is processed before "beta-full".
+        _make_canonical_agent(
+            tmp_path,
+            "alpha-minimal",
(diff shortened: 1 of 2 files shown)


tsdata (approved):
LGTM

Issue #901, opened 2026-05-11
[feature]: RFC: clarify Context Gateway project scope vs target tier model


pandas-studio:
## Problem

Context Gateway now has two different concepts that both read as "scope":

- `scope_id` in the Web multi-project UI means "which project root is selected".
- `scope` / `TargetScope` in the canonical artifact model means `user`, `project_shared`, or `project_local`.

The current split is workable while Web writes remain mostly cwd/project_shared-centric, but it is becoming easy to wire the wrong scope into routes, overview, Sync All, and future multi-scope write flows.

## Proposed solution

Write a small RFC or ADR that formalizes the Context Gateway Web request model before adding more tier-aware UI:

- Use explicit names such as `project_scope_id` for project-root selection and `target_scope` for canonical tier selection.
- Define which routes are project-root scoped only and which are tier scoped.
- Define overview semantics for `user`, `project_shared`, and `project_local`, including whether `project_local` should show canonical-only draft items despite having no runtime fan-out.
- Define Sync All behavior when multiple project roots or tiers are visible.
- Define how Settings hooks target scope relates to artifact target scope, since the overview currently passes scope only to settings.

## Acceptance criteria

- A short RFC/ADR describes the request/response vocabulary and route ownership boundaries.
- Follow-up implementation issues can point to this document instead of re-litigating scope semantics per route.
- Existing `scope_id` clients remain compatible or have an explicit migration path.

## Alternatives considered

- Keep `scope_id` for project roots and add a second unqualified `scope` query parameter. This is compact but likely to produce accidental misuse.
- Keep Web Context Gateway project_shared-only. This avoids immediate work but leaves CLI/core tier features disconnected from the Web UI.

## Additional context

The core model is already tier-aware via `canonical_artifact_dir()` and `runtime_fanout_root()`. The Web project selector is separate and currently maps `scope_id` to a project root through `resolve_scope_root()`.

pandas-studio:
# RFC draft: Context Gateway scope vocabulary and route ownership

> Status: draft for maintainer review. Posted as a comment so the original
> problem statement above remains the canonical issue body. Anchors below
> reference `b4957c5` (`origin/main` at posting time).

## TL;DR

The Web layer today overloads the word "scope" for two unrelated concepts:

- a **project-root selector** (`?scope_id=`, resolved through
  `discover_project_scopes`)
- a **canonical tier** (`TargetScope = Literal["user", "project_shared",
[...]

PR #909, opened 2026-05-11
fix(context): typed PARSE_ERROR skip for unreadable skill canonical/override

memtomem:
## Summary

- Closes the only remaining gap in the #900 behavior-inventory failure matrix: skills now emit a typed `PARSE_ERROR` skip when canonical or per-vendor override bytes are unreadable, matching `agents.py` / `commands.py` (`agents.py:688/748`, `commands.py:477/532`).
- Applied symmetrically to both code paths in `generate_all_skills`:
  - **project_shared batch path** — `_stage_skill` OSError and override `read_bytes()` OSError are caught; the offending pair is skipped and (on override failure after a successful stage) the orphaned staging tree is rmtreed. Privacy block remains the only failure that escalates to a raise.
  - **per-pair path (user / project_local)** — same try/except wrappers; existing `promoted = False` + `finally` cleanup handles the staging tree on override failure.

## Response shape change

This **is** a behavior change at the API boundary, intentionally:

- **Before**: unreadable canonical or override bytes raise `OSError` / `ClickException` out of `generate_all_skills`. CLI surfaces a traceback; Web routes return a non-200.
- **After**: the same condition lands in `SkillSyncResult.skipped` as `(skill_name, "unreadable: {exc}" | "override unreadable: {exc}", "parse_error")`.

This brings `SkillSyncResult` into structural parity with `AgentSyncResult` / `CommandSyncResult`, which is the same call-site contract used by the CLI, MCP, and Web layers. Documented in the #900 inventory's failure-mode matrix as a deliberate symmetry fix.

## Why now / #900 dependency

PR #908 pinned the strict-drop partial-write boundary; this PR closes the second of the two `#900`-inventory cleanup items that don't depend on the #901 §4 scope-vocabulary decisions. Both PRs land before the #900 service extraction so the eventual shared service inherits a uniform failure-mode contract across artifacts.

## Test plan

- [x] New `TestUnreadableSources` class in `test_context_skills.py` covers:
  - project_shared canonical OSError → healthy skill still generates; broken one yields typed skip; no orphaned `.staging-*.tmp` tree
  - project_shared override OSError → same shape with `"override unreadable"` reason prefix
  - user-scope canonical OSError (per-pair path) → typed skip; HOME redirected via existing `tests.helpers.set_home`
- [x] `uv run pytest packages/memtomem/tests/test_context_skills.py` — 34/34 pass
- [x] `uv run pytest -m "not ollama"` across `test_context_{skills,skills_staging,override,override_per_scope,sync_scope,atomic,generate_warnings,install,install_all,status,rescan,runtime_table_none_skips}.py` — 160/160 pass
- [x] `uv run ruff check` + `ruff format --check` on the two changed files — clean

Refs #900

### DIFF
--- a/packages/memtomem/src/memtomem/context/skills.py
+++ b/packages/memtomem/src/memtomem/context/skills.py
@@ -355,7 +355,17 @@ def generate_all_skills(
                 for lock_path in sorted({_lock_path_for(dst) for _, _, _, dst in work}, key=str):
                     stack.enter_context(_file_lock(lock_path))
                 for target, _gen, skill_dir, dst in work:
-                    staging = _stage_skill(skill_dir, dst)
+                    # Unreadable canonical: typed PARSE_ERROR skip rather than
+                    # an exception bubbling up — symmetric with agents.py /
+                    # commands.py read_bytes failure handling. Privacy block
+                    # is the only failure that still aborts the batch.
+                    try:
+                        staging = _stage_skill(skill_dir, dst)
+                    except OSError as exc:
+                        skipped.append(
+                            (skill_dir.name, f"unreadable: {exc}", skip_codes.PARSE_ERROR)
+                        )
+                        continue
                     staged.append((target, staging, dst))
 
                     vendor = GENERATOR_VENDOR.get(target)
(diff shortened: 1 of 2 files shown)


memtomem (commented):
LGTM

Issue #819, opened 2026-05-06
[bug]: Windows CI fails on test_server_sigterm.py + TestServerAliveRefuses after #818


memtomem:
### Summary

Windows CI on PR #818 (https://github.com/memtomem/memtomem/actions/runs/25419209194/job/74557245591) failed with 6 tests after the fcntl→portalocker swap landed. Two distinct root causes:

**Root cause A — Windows ``LockFileEx(LOCK_EX)`` blocks reads from other handles.**

POSIX ``flock``/POSIX ``fcntl`` is advisory and only governs other ``flock`` calls; reads from any handle proceed unaffected. Windows ``LockFileEx`` with ``LOCKFILE_EXCLUSIVE_LOCK`` blocks all I/O on the locked byte range from other handles, including reads. So ``pid_file.read_text()`` (which opens a new handle) raises ``PermissionError`` (``ERROR_LOCK_VIOLATION``) while another handle holds the lock.

Failing tests (both at ``pid_file.read_text()`` calls):
- ``test_server_sigterm.py::test_contended_server_start_preserves_pid_file_content`` (line 551) — the test holds ``LOCK_EX`` via ``holder`` and tries to verify content via a separate handle.
- ``test_server_sigterm.py::test_server_main_acquires_portalocker_pid_lock`` (line 636) — ``main()`` holds ``LOCK_EX`` via ``_lock_fp`` and the test tries to verify content via a separate handle.

Both pass on POSIX because flock semantics let reads proceed.

**Root cause B — ``TestServerAliveRefuses`` class-level skip removal was too coarse.**

#818 dropped the class-level Windows skip on ``TestServerAliveRefuses`` because the ``_hold_pid_lock`` helper now uses ``portalocker``. The lock-acquisition mechanics do work on Windows, but four tests in the class assert POSIX-specific user-facing text (``lsof``, the recorded pid in the message body) or POSIX-specific ``--force`` behavior (unlink-while-open).

Failing tests:
- ``test_refuses_when_server_alive_at_legacy_path`` — asserts ``str(os.getpid()) in result.output``; Windows output omits the pid in favor of "Find the holder via Sysinternals \`handle.exe\` or Resource Monitor."
- ``test_refuses_when_server_alive_at_runtime_path`` — same shape.
- ``test_refuses_with_unknown_pid_branch_when_pid_file_empty`` — asserts ``'lsof' in result.output``; Windows uses Sysinternals/Task Manager wording.
[...]

PR #910, opened 2026-05-11
test(server): pin Windows pid-content regression via stat().st_size

memtomem:
## Summary

Addresses a P2 review comment on the post-#819 server test:

> [P2] Keep Windows pid-content regression pinned — When this test runs on Windows, this guard skips the only assertion that ``main()`` wrote ``os.getpid()`` into ``server.pid``. That behavior is still required there because ``probe_pid_file`` reads the pid text for uninstall/status diagnostics; a Windows-only regression that creates and locks an empty pid file would now pass this cross-platform regression test.

The fix: add a ``pid_file.stat().st_size > 0`` assertion *before* the POSIX-only ``read_text().strip() == str(os.getpid())`` check. ``stat`` queries directory metadata and is unaffected by ``LockFileEx``, so it runs on every platform and pins the "non-empty" half of the contract that the platform skip removed.

## Why ``stat`` instead of removing the platform skip

``LockFileEx`` legitimately blocks content reads from a separate handle on Windows (#819) — that's why ``probe_pid_file`` itself catches ``OSError`` at the read site and falls back to ``pid=None`` during the live-server case. We can't read content from the test, but we *can* check size.

## What the new assertion catches

- ``main()`` calls ``_lock_fp.seek(0); _lock_fp.truncate(); _lock_fp.write(str(os.getpid())); _lock_fp.flush()`` (``server/__init__.py:407-410``).
- A regression that drops the ``write`` or otherwise leaves the file empty would still pass the existing cross-platform check, but ``probe_pid_file`` / ``mm uninstall`` / ``mm status`` would surface ``pid=None`` to users.
- ``stat().st_size > 0`` catches that regression on every platform.

## Test plan

- [x] ``uv run pytest packages/memtomem/tests/test_server_sigterm.py::test_server_main_acquires_portalocker_pid_lock`` — PASS
- [x] ``uv run pytest packages/memtomem/tests/test_server_sigterm.py`` — 11 passed, 1 skipped (existing platform-conditional skip)
- [x] ``ruff check`` + ``format --check`` clean

Refs the #818 / #819 / #820 portalocker cluster.

### DIFF
--- a/packages/memtomem/tests/test_server_sigterm.py
+++ b/packages/memtomem/tests/test_server_sigterm.py
@@ -642,13 +642,26 @@ def test_server_main_acquires_portalocker_pid_lock(
         server_mod.main()
 
         assert pid_file.exists(), "main() must create the pid file"
+        # Cross-platform: pid file must be non-empty. ``LockFileEx`` blocks
+        # *content* reads from other handles on Windows (#819), but file
+        # metadata via ``stat`` is unaffected — so a regression where
+        # ``main()`` creates and locks an empty pid file would still trip
+        # this assertion on Windows. ``probe_pid_file`` (and uninstall /
+        # status diagnostics) call ``read_text().strip()`` on the file when
+        # it isn't currently locked, so an empty pid file would surface as
+        # ``int("")`` → ``pid=None`` in ``ServerState`` — degraded UX even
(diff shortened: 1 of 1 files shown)


pandas-studio (approved):
LGTM

Issue #911, opened 2026-05-11
[ADR-0011 follow-up] Cross-DB `mm context memory-migrate` mode (deferred)


memtomem:
**Tracked from:** ADR-0011 §"Open questions" item 5 and #886's third deferral
bullet (cross-DB migration).

Cross-DB rename / migration is deferred pending a concrete use case.
ADR-0011 §4 argues for one DB with per-row scope tags, so cross-DB
movement would be a **storage-migration feature**, not an extension of v1
rename semantics. It is intentionally scoped *out* of #886, which
landed the low/medium-risk follow-ups (lineage regression test + glob
input + computed neighborhood display).

### Why this needs its own RFC

Single-DB chunk-id-stable rename works because ``chunks.id`` never
changes — the entire FK neighborhood (``chunk_links``, ``access_log``,
``chunk_entities``, ``chunk_relations``) survives untouched. Cross-DB
breaks that premise. A real implementation would need to cover:

- **FTS5 rebuild** on the receiver — ``chunks_fts`` rowids are local
  to the source DB; copied rows need a fresh FTS index pass.
- **Dense-vector backend skew** — receiver may run a different
  embedder (model name, dimension, normalization). Either re-embed
  on import or refuse mixed-embedder transfers.
- **Dedup hash reconciliation** — a chunk already present on the
  receiver (same content_hash, different ID) needs a merge policy.
- **Schema-version migration** — receiver may be on a different
  ``_memtomem_meta`` schema version; pin or upgrade in a single step.
- **``_memtomem_meta`` transfer** — the AI-summary cache rows are
  keyed by source-file path; same migration as in v1 but cross-DB.
- **Failure-mode policy across two DB locks** — what's the rollback
  contract when the receiver write succeeds but the source delete
  fails, or vice versa?

If a real product need surfaces, the two likely shapes have very
different semantics and should be **separate proposals**:

1. **Team onboarding** — export user-tier notes from teammate A's DB
   into project_shared tier of the team's shared DB. One-way export.
2. **Project archive split** — bulk-move an aging project's chunks
   from the main DB into a separate archive DB. Bulk move with
   referential integrity and FTS/vec rebuild on the receiver.

### Out of scope (intentional)

- ``v1`` single-DB chunk-id-stable rename — already shipped in #882
  PR-D and pinned by the regression test added in #886.
- Glob input — landed in #886.
- Computed ``chunk_links lineage`` display — landed in #886.

### References

[...]

Issue #887, opened 2026-05-09
[ADR-0011 follow-up] MCP parity for mm context init / sync / migrate (scope axis)


pandas-studio:
**Tracked from:** ADR-0011 §"Open questions for the implementation issues"
— sixth bullet, MCP exposure for `mm context init` / `sync` / `promote` / `demote`.

> v1 ships these as CLI / authoring tools only. MCP parity is a separate ask.

**Status (post-#882):** PR-E (canonical scope axis for agents / skills
/ commands) is still pending; once PR-E ships, the new subcommands
`mm context init --scope=...`, `mm context sync --scope=...`,
`mm context promote`, and `mm context demote` will be CLI-only by
design. MCP exposure of the same operations is **not** in PR-E or PR-F.

This issue exists as a **monitoring tracker** for the request, gated
behind PR-E.

### Why CLI-only in v1

ADR-0011 §"Open questions" item 6 records the v1 stance: these are
authoring tools — they stamp out canonical artifacts on disk in
`<project>/.memtomem/agents/<name>.md` etc. — so the MCP surface
(typically a tool inside an LLM session) is not the natural
invocation path. CLI invocation also keeps the audit trail in a
predictable shell history rather than a per-session MCP log.

### Re-add criteria

Re-open with a design pass when **either**:

1. A user / team reports they want to drive `mm context init` /
   `sync` from inside a Claude / Gemini / Codex session (e.g.,
   "agent generates a new skill and asks the user to sync it").
2. The MCP tool registry adopts a pattern where authoring tools sit
   alongside read tools (currently the registry separates them per
   `feedback_namespace_ownership.md` style boundary).

### Acceptance criteria when implemented

- New MCP tools follow the registry pattern (`@register` from
  `server/tool_registry.py` per CLAUDE.md MCP invariants).
- Tool names respect the 64-char limit per
  `feedback_proxied_tool_name_64char.md`.
- Each new tool has a confirmation gate analogous to the CLI
  `--confirm-project-shared` for write surfaces under
  `project_shared`.
- Hint prose copy mirrors the CLI hint per
  `feedback_hint_prose_parity_siblings.md`.

### Out of scope

- Read-side MCP coverage (`mm context list` already has read MCP
  parity inside #882's scope).
- Web UI parity (PR-F territory, separate from MCP).

### References

- ADR-0011 §"Open questions" item 6
- ADR-0011 §9 PR-E / PR-F roadmap
- CLAUDE.md MCP tool registry invariants

[...]

PR #926, opened 2026-05-11
feat(server): add mem_context_migrate MCP wrapper (#887)

memtomem:
## Summary

Closes the second half of #887 — the first half (`mem_context_diff` scope=) lands in #920. Wraps the existing CLI verb `mm context memory-migrate` so MCP callers can move markdown memory files between ADR-0011 memory scope tiers.

## Surface

```python
async def mem_context_migrate(
    source: str,                            # single file path or glob
    from_scope: str,                        # user | project_shared | project_local
    to_scope: str,                          # same vocabulary, must differ
    apply_: bool = False,                   # dry-run unless True
    confirm_project_shared: bool = False,   # required when to_scope='project_shared'
    ctx: CtxType = None,
) -> str
```

## Design notes

- **Dual-scope signature** — migrate is the one verb where source and target tiers differ. Both are validated up-front against the `{user, project_shared, project_local}` vocabulary; bogus values short-circuit before paying the `cli_components` bootstrap cost.
- **`--yes` does not survive the port** — it only suppresses the `click.confirm` prompt that the CLI falls through to when `--confirm-project-shared` is missing. MCP has no TTY, so there is nothing to suppress. Per gotcha #5 from `feedback_cli_mcp_parity_gotchas.md`, each CLI confirm gets its own MCP flag — here that's `confirm_project_shared`.
- **Gate A privacy translation** — `_memory_migrate_run` raises `click.exceptions.Exit(1)` with the rejection text on stderr for project_shared secret hits. The wrapper captures both streams via `redirect_stdout` / `redirect_stderr` and translates `Exit` into the `privacy block:` shape that `mem_context_init` / `mem_context_generate` already use, so MCP callers can branch on the same prefix across all context tools.
- **Cross-DB migration is out of scope** — deferred per ADR-0012 / #911.

## Tests

`packages/memtomem/tests/test_server_tools_context_migrate.py` pins the MCP-shape contract only:

1. Unknown `from_scope` returns `error: Unknown from_scope=...` without invoking the helper
2. `from == to` returns `error: --from and --to must differ.`
3. `to_scope='project_shared'` without `confirm_project_shared=True` returns `needs confirmation:` — helper never called
4. project_shared with confirm + secret content → `privacy block:` (Gate A translation)
5. Dry-run returns the plan via captured `click.echo`; no DB UPDATE
6. Apply against `project_local` calls the helper end-to-end; file moves, DB UPDATE awaited once
7. No-glob-match returns `error: No .md files matched ...`

The underlying chunk-id-stable rename / lockfile / compensation semantics are already pinned by `test_context_memory_migrate.py` — this PR does not re-cover them.

## Test plan

- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/test_server_tools_context_migrate.py` — 7 passed
- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/ -k "context_memory_migrate or test_server_tools_context"` — 38 passed (no regression in adjacent migrate / wrapper paths)
- [x] `uv run ruff check` + `uv run ruff format --check` on modified files — clean

## Follow-up

After this merges with #920, issue #887 closes.

### DIFF
--- a/docs/guides/mcp-clients.md
+++ b/docs/guides/mcp-clients.md
@@ -343,7 +343,7 @@ uses, so the output is identical. Useful as a sanity check between
 | **Data** | `mem_export`, `mem_import` |
 | **Config** | `mem_stats`, `mem_status`, `mem_config`\*, `mem_embedding_reset`\*, `mem_reset`\* |
 | **Evaluation** | `mem_eval` |
-| **Context** | `mem_context_detect`, `mem_context_init`, `mem_context_generate`, `mem_context_diff`, `mem_context_sync` (context tools accept `include="skills,agents,commands"` for canonical artifact workflows; `init`, `generate`, and `sync` accept `scope="project_shared\|user\|project_local"` — the canonical **tier** per ADR-0016 §2; `generate`/`sync` also accept `strict=True` to fail on sub-agent or command field drops) |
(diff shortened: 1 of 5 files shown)


pandas-studio (approved):
LGTM

pandas-studio (approved):
LGTM

PR #940, opened 2026-05-11
feat(context): polish tier filters

memtomem:
## Summary
- add explicit Web tier filters for Context Gateway overview and list surfaces
- default Web and CLI read surfaces to project_shared while keeping project_local opt-in
- mark project_local as draft/no-runtime-fan-out and disable sync affordances for that tier
- add CLI --scope support for context diff/status

Closes #936

## Tests
- uv run pytest packages/memtomem/tests/test_web_routes_context.py packages/memtomem/tests/test_web_routes_context_projects.py packages/memtomem/tests/test_context_status.py packages/memtomem/tests/web/test_context_gateway_lists.py
- uv run ruff check packages/memtomem/src/memtomem/web/routes/context_gateway.py packages/memtomem/src/memtomem/web/routes/context_projects.py packages/memtomem/src/memtomem/cli/context_cmd.py packages/memtomem/tests/test_web_routes_context.py packages/memtomem/tests/test_context_status.py

Issue #943, opened 2026-05-12
[ADR-0016 follow-up] Web context artifact write affordances for non-project_shared tiers


memtomem:
## Context

ADR-0011 / ADR-0016 shipped the three-tier canonical context model and the Web read/filter polish has landed (#924, #936, #940). The Web Context Gateway now exposes tier badges and `?target_scope=` list/detail routing, but artifact mutators still intentionally reject non-`project_shared` tiers.

Current route guards in `web/routes/context_{skills,agents,commands}.py` reject create/update/delete/sync/import when `target_scope != "project_shared"`. That keeps v1 safe, but the UI now lets users inspect `user` and `project_local` canonicals, so the write story needs an explicit follow-up rather than staying implicit.

## Scope

- Decide the Web write policy for `target_scope=user` and `target_scope=project_local` across Skills / Subagents / Commands.
- If writes remain blocked, surface that as intentional UI copy/disabled controls instead of a generic route error.
- If writes are enabled, add tier-aware create/update/delete/import/sync flows with the same privacy and no-fan-out semantics as CLI/MCP.
- Keep `project_local` no-runtime-fan-out intact: editing local drafts is okay, but sync/import affordances should not imply runtime materialization.
- Keep `project_shared` confirmation/privacy posture aligned with ADR-0011 Gate A/B.

## Acceptance criteria

- Web controls do not offer actions that are guaranteed to 400 solely because the selected tier is `user` or `project_local`.
- Detail/edit/delete/import/sync behavior is explicitly pinned for all three tiers.
- Tests cover at least one artifact type for `user`, `project_shared`, and `project_local`, plus shared helpers/guards for the other types.
- Public docs or in-app copy explain any intentionally read-only tier behavior.

## References

- #868
- #924 / #936 / #940 Web tier badge and filter polish
- ADR-0011 §3, §5
- ADR-0016 §5, §7

🤖 Generated with Codex

PR #955, opened 2026-05-12
fix(web): tier-aware overview header copy (#952)

### LINKED ISSUE #952: ctx-gateway: overview 'Project: <root>' header is misleading on user tier (PR #947 follow-up)
## Symptom

The Context Gateway overview dashboard renders a header
``Project: <project_root>`` (added in PR #947, commit ``29bb77e``) at
the top of the panel. The header is rendered unconditionally regardless
of the active tier filter — so when the user switches the tier filter
to ``user``, the dashboard counts switch over to ``~/.memtomem/...``
(user-scope canonicals, global per-host) but the header still reads:

> **Project:** ``/home/user/Work/some-repo``

That label is correct for ``project_shared`` and ``project_local``
tiers (where the artifacts live under the cwd's ``.memtomem/``), but
misleads on ``user`` tier — user-scope artifacts are global to the
host, not scoped to the cwd. A user reading "Project: …" while looking
at user-tier counts may incorrectly assume the listed artifacts are
project-scoped, or that switching projects would change them.

## Where

- ``packages/memtomem/src/memtomem/web/static/context-gateway.js``
  ``_renderCtxOverview`` header block (~L424–433): unconditionally
  emits ``<span class="ctx-overview-root-label">{Project}</span>
  <code>{projectRoot}</code>``. Reads ``data.project_root`` without
  branching on ``data.target_scope``.
- ``packages/memtomem/src/memtomem/web/routes/context_gateway.py``
  ``context_overview`` (~L283-287): the response always emits
  ``project_root`` (the server cwd) and ``target_scope`` (the active
  tier). The data shape already supports a tier-aware client; only the
  client render is missing.

The dashboard counts themselves are correct — ``diff_*`` /
``list_canonical_*`` are called with ``scope=target_scope`` so the
numbers reflect the right tier. This is purely a header-copy gap.

## What changes

Pick one of:

1. **Tier-aware label.** Render ``Project: <root>`` only on
   ``project_shared`` / ``project_local``; render
   ``User canonical: ~/.memtomem/`` (or equivalent) on ``user``. Two
   new i18n keys (``settings.ctx.user_canonical_label`` +
   ``settings.ctx.user_canonical_path``).
2. **Conditional hide.** Suppress the entire header on ``user`` tier
   — the project-scope identification is the whole point of the
   header, so it adds no signal on user tier. Cheaper, but loses the
   header's "you are looking at this canonical root" cue.
3. **Tier badge.** Keep ``Project: <root>`` and append a small tier
   chip (``[user]`` / ``[project_shared]`` / ``[project_local]``) so
   the header reads ``Project: <root> [user]``. Lowest UX disruption,
   but the chip is doing real work — the user has to know the
   semantics of the tier names.

Recommend (1) for the cleanest signal — the label is what the user
reads first, and it should match the tier they're looking at.

## Test plan

- ``test_context_gateway_overview.py``: mount with
  ``target_scope=user`` payload and assert the header label is the
  user-canonical wording (positive) AND not ``Project:`` (negative).
- ``test_i18n.py``: structural — both new locale keys present in
  ``en.json`` + ``ko.json`` with matching placeholders.
- Pre-existing tier-swap specs (``test_context_gateway_lists.py``,
  the ``ctx-tier-filter`` click handlers) already exercise the swap;
  the overview is tested separately and needs its own coverage.

## Background

Caught while triaging a PR-#950 review comment that misattributed an
empty-state-copy concern to that PR — the underlying observation
(tier swap doesn't update tier-specific UI affordances) is legitimate
but lives on the PR-#947 surface, not the deep-link carrier. PR #945
already wired ``data-write-blocked="<tier>"`` for write affordances;
this issue extends the same tier-awareness principle to the read-side
header copy.

## Status

**Not blocking.** Read-only header copy; the actual count data is
correct.

memtomem:
## Summary

- PR #947 added an unconditional `Project: <project_root>` header to the Context Gateway overview panel. Correct on project tiers; misleading on `user` tier where canonicals live under `~/.memtomem/` (host-global), not the cwd.
- Branch `_renderCtxOverview` on `data.target_scope`: user → `User canonical: ~/.memtomem/`, otherwise unchanged.
- Two new i18n keys (`settings.ctx.user_canonical_label` + `settings.ctx.user_canonical_path`). KO keeps `canonical` untransliterated to match existing locale style. Path string is shared across locales (filesystem path).
- Add `data-target-scope="<scope>"` to `.ctx-overview-root` so selectors can pin tier state without parsing text — mirrors PR #945's `data-write-blocked` shape.
- Server-side wire shape unchanged; the response already exposes `target_scope` (#940). Counts were already tier-correct — this is a read-side copy fix.

Closes #952.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/web/test_context_gateway_overview.py` — 17 passed (new `test_overview_header_user_tier_shows_user_canonical_label` + 16 regression pins on project-tier header + tile rendering).
- [x] `uv run pytest packages/memtomem/tests/test_i18n.py` — 44 passed (structural parity tests pick up the new keys automatically; no `{param}` placeholders so `test_placeholder_parity` is a no-op for them).
- [x] `uv run pytest packages/memtomem/tests/web/` — 70 passed (full browser suite, no cross-test fallout).
- [x] `uv run ruff check packages/memtomem/src packages/memtomem/tests` + `ruff format --check` — clean.

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/context-gateway.js
+++ b/packages/memtomem/src/memtomem/web/static/context-gateway.js
@@ -406,6 +406,19 @@ function _renderCtxOverview(data) {
   // path on langchange and would otherwise blow up.
   const runtimes = Array.isArray(data.detected_runtimes) ? data.detected_runtimes : [];
   const projectRoot = typeof data.project_root === 'string' ? data.project_root : '';
+  // #952: ``Project: <root>`` is misleading on ``user`` tier — user-scope
+  // canonicals live under ``~/.memtomem/`` (host-global), not the cwd.
+  // Branch the header label + path on ``target_scope`` so the heading
+  // matches the tier the counts are computed against. Defensive default
+  // ``project_shared`` matches the route's query-param default.
+  const targetScope = typeof data.target_scope === 'string' ? data.target_scope : 'project_shared';
+  const isUserTier = targetScope === 'user';
+  const rootLabel = isUserTier
+    ? t('settings.ctx.user_canonical_label')
+    : t('settings.ctx.project_root_label');
+  const rootPath = isUserTier
+    ? t('settings.ctx.user_canonical_path')
+    : projectRoot;
   const undetectedTitle = escapeHtml(t('settings.ctx.runtime_undetected_tooltip'));
   const chips = runtimes.map(rt => {
     const available = !!rt.available;
@@ -443,9 +456,9 @@ function _renderCtxOverview(data) {
   // this same render path use the same inline-``t()`` convention for the
   // same ordering reason.
   let html = `<div class="ctx-overview-header">
-      <div class="ctx-overview-root">
-        <span class="ctx-overview-root-label">${escapeHtml(t('settings.ctx.project_root_label'))}</span>
-        <code class="ctx-overview-root-path">${escapeHtml(projectRoot)}</code>
+      <div class="ctx-overview-root" data-target-scope="${escapeHtml(targetScope)}">
+        <span class="ctx-overview-root-label">${escapeHtml(rootLabel)}</span>
+        <code class="ctx-overview-root-path">${escapeHtml(rootPath)}</code>
       </div>
       <div class="ctx-overview-runtimes">
         <span class="ctx-overview-runtimes-label">${escapeHtml(t('settings.ctx.runtimes_label'))}</span>
--- a/packages/memtomem/src/memtomem/web/static/locales/en.json
+++ b/packages/memtomem/src/memtomem/web/static/locales/en.json
@@ -285,6 +285,8 @@
 
   "settings.ctx.overview_title": "Context Gateway",
   "settings.ctx.project_root_label": "Project",
+  "settings.ctx.user_canonical_label": "User canonical",
+  "settings.ctx.user_canonical_path": "~/.memtomem/",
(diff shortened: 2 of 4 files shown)


memtomem:
Declining the P3 — the citation doesn't hold up and the underlying concern lives elsewhere. Filed as #956 to track separately.

**Verification of the cited review:**

1. **Wrong line.** `context-gateway.js:176-178` is a comment paragraph documenting PR #943's tier-aware *write-gate* (`data-write-blocked`). No empty-state copy lives there.
2. **Quoted text does not exist.** "No … registered for this project." is not present anywhere in `context-gateway.js` or `locales/en.json`. The actual empty-state copy is `settings.ctx.empty_hint` (en.json:352) — "Place {type} under {canonical}/<name>/ then click Sync, or click Import to pull existing {type} from {scan_dirs} within this project."
3. **Not in this PR's diff.** Rendered from `_ctxRenderItemsHtml` (~L1122), which is the per-type list view — not the overview path PR #955 touches.
4. **Pre-existing.** The "within this project" wording predates PR #947 (the overview-header surface this PR fixes), so it's an orthogonal copy gap.

Tracked as #956 for the per-type-list empty hint tier-awareness.

PR #961, opened 2026-05-12
sec(web): fold codex review — thread CSRF in 5 missed SPA mutators (PR #958 follow-up)

memtomem:
Follow-up to PR #958 (which merged the CSRF/Origin/Host enforcement flip). Codex review of the merged commit flagged a class of regression the merged PR missed.

## Summary

PR #958 flipped `csrf_enforce` to `True` by default but four SPA mutators were still issuing unsafe `fetch(...)` without `X-Memtomem-CSRF` — those flows would 403 in production. Adding an AST-style guard test caught a fifth site neither I nor Codex spotted in the manual sweep.

## The 5 missed callsites

| File | Line | Route | Flow |
|---|---|---|---|
| `context-gateway.js` | 2286 | `POST /api/context/{type}/import` | bulk import |
| `context-gateway.js` | 2355 | `POST /api/context/{type}` | create artifact |
| `context-gateway.js` | 2394 | `POST /api/context/known-projects` | add project (path-picker callback) |
| `context-gateway.js` | 2159 | `POST /api/context/{type}/{name}/import` | runtime-only single-item import *(found by new test, not by Codex)* |
| `settings-hooks-watchdog.js` | 131 | `POST /api/context/settings/resolve` | hook conflict resolve |

All five now follow the established pattern: `const csrf = await ensureCsrfToken(); const headers = csrf ? {...,'X-Memtomem-CSRF':csrf} : {...}; fetch(URL, { method, headers, body })`.

## Regression guard

`tests/test_web_invariants_registry.py` gains `test_spa_unsafe_fetch_inline_headers_thread_csrf_token`. It scans `web/static/**/*.js` for unsafe-method `fetch(...)` calls and rejects any that pass `headers: {...}` as an inline literal without `X-Memtomem-CSRF`. Bare-identifier `headers` references (the threaded-variable convention) are trusted — this scoping keeps the check resistant to false positives from the long-handler patterns elsewhere in `context-gateway.js`.

## Autouse-fixture blind-spot pin

`tests/conftest.py`'s autouse `MEMTOMEM_WEB__CSRF_ENFORCE=0` fixture (added in PR #958) defaults route-unit tests to observe-only for ergonomics. Codex correctly noted this hides exactly the regression class fixed here. `test_production_create_app_enforces_csrf_without_token` in `test_web_csrf_middleware.py` explicitly clears the env, builds via `create_app()`, and asserts the middleware returns 403 for an unsafe `/api/...` request — pinning the production wiring end-to-end.

## Also

- Cache-bust bumps for the two edited SPA files (`context-gateway.js` v32→v33, `settings-hooks-watchdog.js` v5→v6).
- SECURITY.md wording fix: SPA fetches `/api/session` lazily on first unsafe-method request, not "on load" (Codex Minor).

## Test plan

- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/test_web_csrf_middleware.py packages/memtomem/tests/test_web_invariants_registry.py` → 30/30 ✅
- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/ -k 'web'` → 759/759 ✅
- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/` → 5037 passed, 10 skipped ✅
- [x] `uv run ruff check packages/memtomem/src packages/memtomem/tests` → All checks passed ✅
- [x] `uv run ruff format --check packages/memtomem/src packages/memtomem/tests` → 463 files already formatted ✅

References: #787, #958.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -31,8 +31,9 @@ every unsafe-method request to `/api/**` (POST / PUT / PATCH / DELETE) is
 gated by a single middleware that checks three things at once:
 
 1. **CSRF token** — `X-Memtomem-CSRF` must match the per-process token in
-   `app.state.csrf_token`. The SPA fetches it via `GET /api/session` on
-   load; the token rotates on every restart and is never persisted.
+   `app.state.csrf_token`. The SPA fetches it via `GET /api/session`
+   lazily on the first unsafe-method request (cached for the page
+   lifetime); the token rotates on every restart and is never persisted.
 2. **Origin / Referer** — when present, must resolve to a loopback host
    or an operator-supplied `--trusted-origin` entry. Defends drive-by
    tabs whose Origin reveals the attacker domain.
--- a/packages/memtomem/src/memtomem/web/static/context-gateway.js
+++ b/packages/memtomem/src/memtomem/web/static/context-gateway.js
@@ -1673,7 +1673,7 @@ function _ctxRenderConflictBanner(detailEl, userBuffer, freshContent) {
   banner.hidden = false;
 }
 
-async function _ctxHandleConflict(type, name, userBuffer, staleMtimeNs, detailEl, headers) {
+async function _ctxHandleConflict(type, name, userBuffer, staleMtimeNs, detailEl) {
   // ``staleMtimeNs`` is the mtime_ns the user's first Save was already
   // racing against — i.e. what they thought disk was. We thread it
   // through to the force PUT body so the server-side WARNING log
@@ -1702,6 +1702,10 @@ async function _ctxHandleConflict(type, name, userBuffer, staleMtimeNs, detailEl
   }
   if (choice === 'force') {
     try {
+      const csrf = await ensureCsrfToken();
+      const headers = csrf
+        ? { 'Content-Type': 'application/json', 'X-Memtomem-CSRF': csrf }
+        : { 'Content-Type': 'application/json' };
       const r2 = await fetch(
         _ctxWithTargetScope(`/api/context/${type}/${encodeURIComponent(name)}`),
         {
@@ -1905,7 +1909,7 @@ async function loadCtxDetail(type, name, opts = {}) {
           },
         );
         if (r.status === 409) {
-          await _ctxHandleConflict(type, name, content, mtime_ns, detailEl, headers);
+          await _ctxHandleConflict(type, name, content, mtime_ns, detailEl);
           return;
         }
         if (!r.ok) {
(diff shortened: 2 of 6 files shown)


memtomem:
## Handoff to next session — codex round 4 fold

Round 4 review (full transcript in `.dev-trio/log/dev-trio/latest-codex.log` at the time of HEAD `7c0aa60`) returned **NEEDS-FIX** with 3 Major parser-correctness gaps. All are hypothetical (zero occurrences in the codebase today) — strict-mode parser fold approved by maintainer.

### What to fix in `tests/test_web_invariants_registry.py`

1. **`_extract_headers_value` scope is too wide** (~line 495).
   Currently searches the whole 800-char fetch window for `headers:` — can be tricked by:
   - A commented-out `// headers: { 'X-Memtomem-CSRF': csrf }` line inside the options
   - A body payload like `body: JSON.stringify({ headers: { 'X-Memtomem-CSRF': csrf } })` where the nested `headers` is matched before any real request headers
   **Fix:** extract the second-arg text via `_scan_fetch_args`, strip comments + string contents, then scan only top-level properties of that object.

2. **Method detection misses computed keys** (~line 615).
   `{ ['method']: 'POST', body: '{}' }` produces no `method:` literal match → treated as default GET → unsafe write with no CSRF headers passes.
   **Fix:** detect any computed-key syntax `[...]:` at top level of the options object and fail with `method-not-inferable`, OR parse top-level properties and treat computed `['method']` as unclassifiable.

3. **Reassignment after binding** (~line 572).
   `let headers = csrf ? {...CSRF...} : {}; headers = {}; fetch(URL, { method: 'POST', headers })` — the prior CSRF-bearing binding still rescues even though `headers` was reassigned to a non-CSRF value before the fetch.
   **Fix:** for `let`/`var`, search the text between the binding and the fetch for `<ident> =` reassignment; if present, reject. OR simpler: only accept `const` bindings as a Shape B match.

### Branch state

- Branch: `issue-787-csrf-pr2-followup` (HEAD `7c0aa60`)
- Worktree: `.claude/worktrees/issue-787-csrf-pr1/`
- Test suite green: `5037 passed, 10 skipped` at HEAD
- All 4 rounds 1–3 findings already folded; only these 3 parser blind spots remain

### Verification pattern for the new session

The injection-based verification at the end of round 3 is the right template — write 3 small test cases that inject each bug shape, confirm the test catches it with the expected verdict, then revert. Use a temp dir + `cp backup.js`, NOT git-stash, so the file system stays clean.

### Stop signal

If round 5 finds more parser edges, that's the signal to either (a) move to a real JS AST parser (Node subprocess) or (b) accept remaining gaps as documented limitations. Don't keep adding regex layers past round 5.

pandas-studio (approved):
LGTM

memtomem:
recheck

memtomem:

memtomem:

PR #966, opened 2026-05-12
fix(web): Hooks Sync UI scope-aware label + needs_confirmation surfacing (#962)

### LINKED ISSUE #962: Hooks Sync UI: reduce status duplication and handle host-write confirmation explicitly
## Context

During the private Context Gateway manual test pass, the Hooks Sync API behaved correctly, but the Settings -> Hooks UI felt awkward and has one trust-gate behavior that needs explicit verification/fix before the checklist is promoted publicly.

Private checklist reference: `memtomem-docs/memtomem/testing/context-gateway-manual-testing.md` (private repo).

## Observed

With `.memtomem/settings.json` synced to the default `hooks.target_scope=user`, the Hooks panel rendered as a vertical stack similar to:

```text
Lifecycle hooks defined in .memtomem/settings.json, applied to the active runtime.

Hooks
Sync Now
All hooks are in sync
User-scope target: /tmp/mtm-cg-manual/home/.claude/settings.json
In sync
PreToolUse:Bash
```

The API responses were correct:

- `GET /api/settings-sync` reported `out_of_sync` with one pending `PreToolUse:Bash` rule before sync.
- `POST /api/settings-sync` with `{}` returned `results[0].status = "needs_confirmation"` for user-scope host writes.
- `POST /api/settings-sync` with `{ "allow_host_writes": true }` returned `ok`.
- Conflict flow worked: `conflicts` -> `POST /api/context/settings/resolve` -> `in_sync`.

## Issues to fix / verify

1. **Duplicated in-sync copy**
   - `All hooks are in sync` badge plus an `In sync` section heading repeats the same state.
   - The synced list should be visually quieter, or the badge/section should be consolidated.

2. **Target label is user-specific**
   - The UI says `User-scope target:`. That is accurate only when `hooks.target_scope=user`.
   - If the active target scope is `project_shared` or `project_local`, the label should reflect the actual scope or use neutral copy like `Target:`.

3. **Host-write confirmation handling must be explicit**
   - The Sync Now button opens a confirm modal, but the POST payload must include `allow_host_writes:true` when the user confirms a user-scope write.
   - If it posts `{}` and receives `needs_confirmation`, the UI must surface that state instead of showing a generic success toast.

4. **Long target path layout**
   - Long target paths should wrap cleanly and not push the Sync Now button or overflow the settings panel, including narrow widths.

## Suggested acceptance criteria

- In-sync Hooks state has no redundant status copy.
- Target path label is scope-neutral or reflects the active `hooks.target_scope`.
- Confirmed user-scope Sync Now either sends `{ "allow_host_writes": true }` or clearly renders the server `needs_confirmation` result as a warning/error.
- Long target paths wrap within the panel on desktop and narrow viewport widths.
- Add/update browser tests around the Sync Now confirmed payload and the `needs_confirmation` UI branch.

## Related

- ADR-0010 settings hooks target scope
- `packages/memtomem/src/memtomem/web/static/settings-hooks-watchdog.js`
- `packages/memtomem/src/memtomem/web/routes/settings_sync.py`

memtomem:
## Summary

Closes #962. The Hooks Sync API itself was already correct (manual test pass: pending → sync → conflict → resolve all green), but the Settings → Hooks UI was lying or silent in three different ways:

- **Duplicate in-sync copy** — the badge says \"All hooks are in sync\" and a separate \"In sync\" section heading repeated the same state. Heading now drops in the synced-only state; still kept when mixed with pending/conflicts so the separator stays load-bearing.
- **\"User-scope target:\" label was hardcoded** and lied for `project_shared` / `project_local` scopes. `GET /api/settings-sync` now surfaces `target_scope`; the panel renders the matching `settings.hooks.target_label_<scope>` key (falls back to a neutral `Target:` for unknown scopes or older payloads).
- **Host-write confirmation handling was silent.** Sync Now POSTed an empty body, the server returned `needs_confirmation` for the `~/.claude/settings.json` write, and the UI showed a generic success toast on top of it. The confirm modal IS the host-write consent (mirrors the CLI's `_confirm_settings_host_writes`), so the POST now carries `{\"allow_host_writes\": true}`. If the server still returns `needs_confirmation`, the panel renders an explicit banner naming the refused target paths and skips the success toast.
- **Long target path overflow** — inline-styled flex header is now a `.hooks-sync-header` CSS class with `flex-wrap: wrap` + `overflow-wrap: anywhere` on the path `<code>` element.

## Notes

- `target_label_user` / `target_label_project_shared` / `target_label_project_local` are new i18n keys; the old `target_label` stays as a neutral fallback to avoid breaking any older payload path.
- The `needs_confirmation` branch is a defensive surfacing path — once the first POST carries `allow_host_writes: true`, retrying with the same flag would loop, so the branch treats it as a terminal warning the user resolves via the CLI rather than auto-retrying.

## Test plan

- [x] Backend: `target_scope` round-trips for `user` / `project_shared` / `project_local` (added 3 tests under `TestSettingsSync`)
- [x] Browser: POST body equals `{\"allow_host_writes\": true}` after confirm
- [x] Browser: `project_shared` scope renders `Project (shared) target:` label, not the user-scope literal
- [x] Browser: `needs_confirmation` response surfaces the banner with the refused target path, no success toast
- [x] Existing browser tests for the cancel + confirm flows continue to pass
- [x] `uv run ruff check && uv run ruff format --check` clean
- [ ] Manual: spin up the web UI under each of the three `hooks.target_scope` values and verify the label updates (covered by tests + visual sanity)


memtomem:

memtomem:

PR #967, opened 2026-05-12
feat(web): promote Context Gateway to a top-level main tab (#962)

memtomem:
## Summary

Sibling PR to #966 — also part of the #962 thread. Manual Context Gateway testing showed the Gateway is buried under `Settings → Agent Integrations`, which costs an extra click and lets users wander into unrelated settings while diagnosing sync issues. Promote it to a top-level tab next to Search and Sources.

## Approach — restructure, don't duplicate

- Move `ctx-overview / ctx-skills / ctx-commands / ctx-agents / hooks-sync` sections from `#tab-settings` into a new `#tab-context-gateway` panel
- Keep existing ids (`settings-ctx-*`) and CSS classes (`.settings-nav-btn` / `.settings-section`) so `switchSettingsSection`'s DOM selectors resolve panels unchanged regardless of which tab hosts them
- A new `GATEWAY_SECTIONS` set is the routing pivot:
  - `activateTab('context-gateway')` restores the user's last gateway section (or falls back to `ctx-overview`)
  - `switchSettingsSection('ctx-*' | 'hooks-sync')` auto-redirects into the Gateway tab via `activateTab` with a `sectionOverride` arg
  - `activateTab('settings')` recognizes that the last section may be a gateway one and falls back to `config` so users don't get bounced back to Gateway by the redirect
  - `context-gateway.js`'s `langchange` listener now considers both `#tab-context-gateway` and `#tab-settings` as valid hosts (legacy fallback so a partial revert can't silently kill re-render)

Hooks moves with the Gateway tab — both the issue body's framing and the overview tile already treat Hooks as part of the Gateway flow.

## Tests

- `test_web_mode.py`: `ctx-overview` keeps the `--landing` modifier AND now lives under `#tab-context-gateway`. Symmetric pin on the leaves. New positive pin on the main-tab button presence + position. New negative pin that the Settings sidebar no longer surfaces gateway sections.
- `test_context_gateway_main_tab.py` (new browser test, 3 cases):
  - Clicking the main tab button lands on `ctx-overview` by default
  - Legacy `switchSettingsSection('ctx-skills')` auto-redirects to the Gateway tab
  - The Settings tab runtime no longer holds gateway nav buttons
- All 58 existing browser tests touching the gateway/hooks panels continue to pass (fixture redirect handles the move transparently)

## Test plan

- [x] `uv run pytest packages/memtomem/tests/web/test_context_gateway_*.py packages/memtomem/tests/web/test_settings_hooks_*.py` — 58 passed
[...]

PR #968, opened 2026-05-12
feat(web): surface detail meta + Hooks per-rule detail panel (#962)

memtomem:
## Summary

Third sibling PR in the #962 thread (alongside #966 Hooks UI fixes and #967 Gateway main-tab promotion). Manual testing showed two information-density problems with the Context Gateway:

- **Skills / Agents / Commands detail panel** rendered only the raw file content; the backend already exposed `description`, parsed fields (`role`/`isolation`/`model`/`allowed_tools`), and `target_scope`, but the UI ignored them. Users had to parse YAML frontmatter visually to confirm what artifact they were looking at.
- **Hooks panel** had no per-rule detail at all — synced/pending rules were unclickable rows showing only the `event:matcher` label, while skills/agents/commands had a full canonical pane.

## Approach

**Backend (additive, no behavior change):**
- `context_skills.py` / `context_agents.py` / `context_commands.py` — detail endpoints now echo `target_scope` and `layout` for parity with the list view
- Skills also surface `fields.description` parsed from SKILL.md frontmatter, with a first-non-blank-body-line fallback so the meta header never renders empty

**Frontend meta header:**
- New `_ctxRenderDetailMetaHeader` in `context-gateway.js` mounts above the Canonical|Diff tab strip
- Description / scope / layout / mtime rows for all three types
- Agents: chip row with role / isolation / kind / temperature
- Commands: chip row with argument_hint / allowed_tools / model
- Empty fields are dropped so an unset `temperature` won't render as the literal `undefined`

**Hooks per-rule detail:**
- Synced and pending rule rows are now `role="button"` with keyboard (Enter/Space) activation
- Click reveals a shared `#hooks-rule-detail` panel with event / matcher / type / command / timeout rows + the raw rule JSON
- Conflict cards intentionally untouched — their existing diff view IS the effective detail
- A new `_hooksRuleRegistry` indexes the full rule objects by `event:matcher` so the click handler renders without an extra round-trip

## Locale

21 new keys under `settings.ctx.detail.*` and `settings.hooks.detail.*` + their KO translations.

## Test plan

- [x] `test_context_gateway_detail_meta.py` (new, 4 cases): skill meta header + agent chip row + command chip row + negative pin that empty parsed fields don't render as `undefined` chips
- [x] `test_settings_hooks_detail.py` (new, 3 cases): clicking a synced rule opens the detail panel with the labelled fields and raw JSON; pending rules use the same panel (symmetry pin); the panel correctly swaps content on a second click instead of appending
- [x] All 80 existing browser tests + 247 backend regressions green locally
- [x] `uv run ruff check && uv run ruff format --check` clean
- [ ] Manual: open each artifact type's detail; click hook rules with and without `matcher` set; verify all chips/rows render correctly

## Notes for reviewers

- This PR can land in any order vs sibling PRs #966 and #967 — the only touch overlap is in `settings-hooks-watchdog.js` where #966's Sync Now POST body fix and this PR's clickable-row restructure both touch the synced/pending loop. Rebase resolves cleanly: take #966's POST body change AND this PR's clickable-row markup
- Skills aux-file tree polish (icons / sizes / tree layout) is intentionally NOT included — that's deferred per the original plan as a follow-up polish PR
- Conflict card behavior is fully unchanged in this PR


memtomem:

memtomem:

memtomem:

PR #971, opened 2026-05-12
Fix pending chunk highlight retry

### LINKED ISSUE #680: Preserve pendingActivateChunkId across Load All so large-source highlight retries
## Context

PR #676 added Timeline → Source jump that highlights a specific chunk
card via `STATE.pendingActivateChunkId`. PR #678 bumped the first fetch
to `limit=500` to cover up to 500-chunk sources.

For 500+ chunk sources the highlight silently no-ops — the toast
introduced in #678 (`toast.chunk_target_missing`) tells the user the
chunk wasn't located, but no recovery path exists. The "Load All"
button (`app.js:3092`) also issues `limit=500`, so clicking it is a
wasted click. The toast wording was therefore softened in #678
(commit 6d57ccc) to drop the "try Load All" guidance.

## What we want

When the chunk-target consumer in `browseSource` (`app.js:3232–3262`)
fails to find the card, **keep** `STATE.pendingActivateChunkId` set
(don't clear it). Modify `browseSource(path, limit)` so that when it's
called again with the same `path` (e.g. user clicks Load All, or a
future "Load more" pagination), the consumer re-tries the highlight on
the larger fetch. Clear only after a successful highlight or after an
explicit user pivot (manual `.source-item` click — already wired in
#678 fix #2).

The current "consume once" contract is intentional for the namespace-
chip / cold-load follow-through paths (PR #674) — make sure the
preservation is scoped to the chunk-id case only and doesn't regress
the path-level activation.

## Files

- `packages/memtomem/src/memtomem/web/static/app.js`
  - `browseSource()` pending-chunk consumer (lines 3232–3262)
  - Manual `.source-item` click handler already clears the chunkId
    (line 3061) — keep that as the explicit-pivot reset

## Out of scope

- Server-side pagination of `/api/chunks`. Today's `limit` ceiling is
  1000; a 1000+ chunk source is the next tail and would need a real
  pagination story.

## Acceptance criteria

- [ ] On a 500+ chunk source, click "Open in source" from Timeline →
      first fetch (limit=500) misses the target chunk → toast appears.
- [ ] Click "Load All" → second fetch (limit=500 today, or whatever the
      bumped value is in a follow-up) → if the chunk is now in the
      result set, the card scrolls into view + flashes.
- [ ] Manual `.source-item` click on a different source still clears
      the pending chunkId (no cross-source highlight leak).
- [ ] No regression on path-only `_navigateToSource` callers (Home
      recent, command palette).

## Difficulty

Good first issue: ≤ 2 files changed, no API break, behavioral test
already scaffolded by #676/#678.

memtomem:
## Summary
- preserve pending Timeline chunk highlight IDs across same-source misses so Load All can retry
- track the owning source path to avoid cross-source consumption
- add Vitest coverage for retry, path scoping, and manual source pivot clearing

## Tests
- npm test -- pending-activate-chunk-retry.test.mjs
- npm test

Fixes #680

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -39,9 +39,10 @@ const STATE = {
   // rendering the chunk list. Set by ``_navigateToSource`` when the
   // caller knows the specific chunk (e.g. Timeline → Source jump);
   // empty string means "source-level only, no chunk highlight".
-  // Cleared on consume in ``browseSource`` so a later same-source
-  // browse (e.g. Load All) doesn't re-flash an old target.
+  // Preserved across a same-source miss so a larger follow-up fetch
+  // (e.g. Load All / future pagination) can retry the highlight.
   pendingActivateChunkId: '',
+  pendingActivateChunkSourcePath: '',
   // Active Sources vendor sub-tab (one of ``user`` / ``claude`` /
   // ``openai``). Lazily populated from localStorage on first render so
   // the value survives reloads. Keeping it on STATE means re-renders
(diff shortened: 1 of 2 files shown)


memtomem:

memtomem:

pandas-studio (approved):
LGTM

PR #983, opened 2026-05-13
fix(web): clear stale model readiness errors

### LINKED ISSUE #982: Web model-readiness banner can stay failed after model loads
## Context

The Web UI can keep showing "Model failed to load - check Settings" even after
the lazy fastembed model subsequently loads successfully.

Observed locally:

- GET /api/system/model-readiness reported embedder.state=ready but reranker.state=error.
- The reranker cache file existed and a fresh process could load
  jinaai/jina-reranker-v2-base-multilingual successfully.
- Search responses were already using source=reranked, so the stale error state
  was only affecting readiness/banner reporting.

## Root Cause

_component_for() checks _load_error before _model. A lazy loader can retain a
stale _load_error from an earlier failed attempt even after a later request
successfully constructs _model.

## Scope

- Treat _model is not None as the authoritative ready state.
- Clear the returned readiness error in that case.
- Add regression tests for embedder and reranker stale-error cases.

## Acceptance Criteria

- model-readiness returns ready when the holder has _model even if _load_error
  still contains an older failure.
- The Web banner no longer stays on the failure message after the model is
  actually loaded.
- Focused readiness tests pass.

memtomem:
## Summary

- Treat an already loaded lazy model as ready even if a stale load error remains.
- Clear the returned readiness error in that ready state.
- Add embedder and reranker regression coverage for stale-error readiness.

Fixes #982

## Tests

- uv run ruff check packages/memtomem/src/memtomem/web/routes/system.py packages/memtomem/tests/test_web_model_readiness.py
- uv run pytest packages/memtomem/tests/test_web_model_readiness.py

### DIFF
--- a/packages/memtomem/src/memtomem/web/routes/system.py
+++ b/packages/memtomem/src/memtomem/web/routes/system.py
@@ -61,7 +61,6 @@ from memtomem.web.schemas.config import (
     EmbeddingResetResponse,
     EmbeddingStatusResponse,
     ModelComponent,
-    ModelComponentState,
     ModelReadinessResponse,
     PrivacyPatternEntry,
     PrivacyPatternsResponse,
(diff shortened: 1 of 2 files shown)


memtomem:

pandas-studio (approved):
LGTM

PR #981, opened 2026-05-13
feat(web): improve Sources navigation and filtering

memtomem:
## Summary

- Rework Sources vendor/category navigation with fixed sidebar controls, category menus, folder trees, and scroll-to-expand paging.
- Add flat file rendering for filter/chunks/size/recent sort modes, plus visible match highlighting.
- Add body/heading/summary-aware Sources filtering via a new content-matches API.

## Tests

- npm test
- uv run pytest packages/memtomem/tests/test_storage_extended.py packages/memtomem/tests/test_web_routes.py -q
- uv run ruff check packages/memtomem/src/memtomem/web/routes/sources.py packages/memtomem/src/memtomem/storage/sqlite_backend.py packages/memtomem/tests/test_web_routes.py packages/memtomem/tests/test_storage_extended.py

PR #992, opened 2026-05-13
fix(web): improve Search filter removal UX

### LINKED ISSUE #980: Improve Search tab filter add/remove UX
## Context

Review of the Web UI Search tab filter flow found several UX inconsistencies around adding and removing filters. The current UI makes it hard to tell which filters are active, especially when filters produce zero results or when advanced filters are collapsed.

## Findings

- Active filter chips are rendered after the empty-result early return, so chips disappear exactly when users most need to remove filters.
- Filter application is inconsistent: tag/namespace mostly update the badge until a search is run, while type/score/source/date rerender immediately.
- Active chips and the filter-count badge cover namespace/tag/type/score only; source and date filters can affect results without being visible in the active-filter summary.
- Source-only search is supported by the API but blocked by the UI guard in `doSearch()`.
- Load More only preserves query/tag, so namespace/source/context filters can be dropped when expanding result count.
- Advanced source filtering is hidden in a multi-select and is not reflected when the panel is collapsed.

## Acceptance criteria

- Active filter chips remain visible even when the current filters yield zero results.
- Source and date filters are included in active filter chips and the filter-count badge.
- Removing any active filter updates the results predictably and exposes a clear way to remove all filters.
- Source-only searches work from the Search tab when source filters are selected.
- Load More preserves the current server-side filter parameters.
- Basic UI behavior is covered by focused JS tests where practical.

## Notes

This issue tracks the Search tab add/remove filter UX separately from the broader Sources tab changes currently in progress.

memtomem:
## Summary

- Keep active Search filter chips visible when filters produce zero results.
- Include source/date filters in active chips and the filter-count badge.
- Split server-backed filter clearing from client-only filtering so removing filters does not flash stale intermediate states.
- Allow source-only Search requests and preserve namespace/source/context filters on Load More.
- Keep namespace in the main filter row, but render namespace dropdown options with full names inside stable human groups.

Closes #980.

## Tests

- `npm test -- search-filter-ux.test.mjs`
- `npm test`

## Worktree

Implemented in separate worktree `/private/tmp/memtomem-search-filter-ux` from `origin/main`, independent of #981.

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -49,15 +49,6 @@ const STATE = {
   // (filter typing, sort change) read the active vendor without
   // hitting localStorage on the hot path.
   sourcesActiveVendor: null,
-  // Per Sources vendor/category visible source-row or folder budget. Mirrors
-  // Search's paging shape, but the data is already client-side so scrolling
-  // just expands the rendered slice by 10 units.
-  sourcesCategoryLimits: {},
-  sourcesExpandedDirs: {},
-  sourcesActiveCategoryByVendor: {},
-  sourcesBodyFilterQuery: '',
-  sourcesBodyFilterPaths: null,
-  sourcesBodyFilterPending: false,
   dedupScanActive: false,
   dedupAbortCtrl: null,
   lastTagsData: [],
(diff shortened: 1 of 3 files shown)


memtomem:

memtomem:

PR #1000, opened 2026-05-13
Hide dev-only Home session stats in prod

memtomem:
## Summary

- Hide Home Sessions and Working Memory stat cards behind the existing dev-tier UI filter.
- Switch the Home stats grid to 4 columns in prod and 6 columns in dev, with responsive breakpoints preserved.
- Add grep coverage so the prod/dev Home stat behavior stays pinned.

## Test

- uv run pytest packages/memtomem/tests/test_web_mode.py -q


memtomem:

PR #1010, opened 2026-05-13
Clarify Home quick action flow

memtomem:
## Summary
- Rename Home Quick Actions so labels describe opening/preparing screens instead of implying immediate execution
- Add localized title/ARIA text plus context toasts for navigation-based actions
- Cover the labels, tooltip bindings, and quick-action toast keys in i18n tests

## Tests
- uv run pytest packages/memtomem/tests/test_i18n.py

Closes #989

Issue #989, opened 2026-05-13
[Web/Home] Clarify Quick Actions labels and execution flow


memtomem:
## Problem

Some Home `Quick Actions` labels imply immediate execution, but the current handlers mostly navigate to another tab or prepare a form state.

The clearest example is `Re-index All`: clicking it activates the Index tab and checks `index-force`, but does not start re-indexing. This is safer than immediate execution, but the label over-promises what happened.

## Relevant Code

- `packages/memtomem/src/memtomem/web/static/index.html`
  - Home Quick Actions buttons and localized labels.
- `packages/memtomem/src/memtomem/web/static/app.js`
  - Quick Action handlers for Search, Index, Re-index, Export, Dedup, Tags.

## UX Issues

- `Re-index All` reads like a destructive/expensive operation will start immediately.
- Export and Dedup silently route into More/Settings sections, which can feel like a context switch rather than an action.
- Emoji-prefixed labels vary by platform and are less consistent with the rest of the work-focused UI.

## Suggested Fix

- Rename actions to reflect their actual behavior, e.g. `Open Index`, `Prepare Full Re-index`, `Open Export`, `Open Dedup Scan`.
- Or make them true actions with confirmation/progress where appropriate.
- Add short accessible descriptions/tooltips for actions that navigate to another tab.
- Consider replacing emoji prefixes with consistent icon styling used elsewhere in the UI.

## Acceptance Criteria

- Each Quick Action label accurately describes what happens on click.
- Potentially expensive operations do not appear to run without confirmation.
- Navigation-based actions provide clear context after switching tabs/sections.
- Labels remain localized in English and Korean.

PR #1013, opened 2026-05-13
chore(web): polish Home Quick Actions helper + regression pin

memtomem:
## Summary

Follow-up polish for #1010 (already merged) addressing the three review notes that landed in the PR-review pass:

- Rename `_quickActionToast` → `showQuickActionToast` in `app.js`. The `_FOO` prefix in this file is reserved for module-private constants/regex (`_UNSAFE_METHODS`, `_HELP_VISIBLE_KEY`, `_UPLOAD_REDACTION_BLOCKED_RE`); function helpers in the Home view use plain names (`renderHomeStorage`, `switchSettingsSection`).
- Tighten `test_home_quick_actions_describe_navigation_flow`: add two negative-pin asserts that the pre-#989 brittle `.settings-nav-btn[data-section="export"|"dedup"]?.click()` DOM-poke does not return. Without these, a regression could re-pass the existing positive `switchSettingsSection(...)` asserts while silently reintroducing the original bug shape.
- Comment the leading-icon regex so a future reader knows it is the "no emoji prefix" guard from #989.

No behavior change.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_i18n.py` — 47 passed
- [x] `uv run ruff check packages/memtomem/src packages/memtomem/tests`
- [x] `uv run ruff format --check packages/memtomem/tests/test_i18n.py`

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -2133,7 +2133,7 @@ qs('home-search-input').addEventListener('keydown', e => {
 });
 
 // F. Quick Actions
-function _quickActionToast(key, focusTarget) {
+function showQuickActionToast(key, focusTarget) {
   showToast(t(key), 'info');
   if (focusTarget) {
     requestAnimationFrame(() => qs(focusTarget)?.focus());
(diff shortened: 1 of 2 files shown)


memtomem:

pandas-studio (approved):
LGTM

PR #1014, opened 2026-05-13
fix(web): async hot-reload reranker lazy load

### LINKED ISSUE #1012: Async-ify _sync_reranker so disk hot-reload doesn't block the event loop on lazy model load
## Context

Surfaced during review of PR #1009 (merged as `2ae47e9`).

`_sync_reranker` in `packages/memtomem/src/memtomem/web/hot_reload.py` runs synchronously inside `reload_if_stale`, which is invoked from every request handler that touches config. When an admin edits `~/.memtomem/config.json` on disk to flip `rerank.enabled=true` against a local provider (e.g. fastembed), the lazy-load check calls `reranker._get_model()` directly:

```python
if new_reranker is not None:
    load_model = getattr(new_reranker, "_get_model", None)
    if callable(load_model):
        try:
            load_model()
        except Exception:
            logger.exception(...)
            _close_async_component(new_reranker)
            return
```

That `load_model()` call is synchronous — it can pull tens of MB and run ONNX init, blocking the event loop for several seconds on the *next* request after the disk edit. All other in-flight clients freeze.

The companion PATCH path (`routes/system.py::_validate_reranker_ready`) correctly offloads via `await _asyncio.to_thread(load_model)`. The disk path doesn't have that affordance because `_sync_reranker` is synchronous all the way up through `apply_runtime_config_changes`.

## Not a regression

Pre-PR-#1009, disk-edit + enabling rerank silently did nothing until restart. So this issue describes a *worse-than-PATCH-but-better-than-pre-PR* state, not a regression.

## Options

1. **Document + accept.** Add a docstring note on `_sync_reranker` flagging the block; admins editing the YAML accept the cost. Simplest.
2. **Async-ify the path.** Make `_sync_reranker` async, propagate through `apply_runtime_config_changes` and `reload_if_stale`. Touches several call sites. Cleanest.
3. **Skip the lazy-load check on the disk path.** Revert disk-side semantics to "fail at first search" (pre-PR-#1009 behavior for the broken-load case). Asymmetric with PATCH.

Option 2 is correct but the broadest. Option 1 is the cheapest stopgap.

## Repro

1. Run `mm web` with `fastembed` not installed (or a local model file missing).
2. Edit `~/.memtomem/config.json`: set `rerank.enabled=true`, `rerank.provider=fastembed`.
3. Trigger any handler that calls `reload_if_stale` (e.g. `GET /api/config`).
4. Observe: the request hangs for the full lazy-load duration (or fails after that and keeps the previous reranker, per the new safety net).

## Related

- PR #1009 — `2ae47e9` (the merge that introduced the safety net)
- `packages/memtomem/src/memtomem/web/hot_reload.py::_sync_reranker`
- `packages/memtomem/src/memtomem/web/routes/system.py::_validate_reranker_ready` (correct async pattern)

memtomem:
## Summary

- make disk hot-reload runtime fanout async so reranker lazy model loads can run off the event loop
- await hot-reload from system routes and safely close replaced/rejected rerankers
- add regression coverage proving disk-edit lazy load runs on a worker thread

Fixes #1012

## Tests

- uv run ruff check packages/memtomem/src/memtomem/web/hot_reload.py packages/memtomem/src/memtomem/web/routes/system.py packages/memtomem/tests/test_web_hot_reload.py
- uv run pytest packages/memtomem/tests/test_web_hot_reload.py
- uv run pytest packages/memtomem/tests/test_web_routes.py

### DIFF
--- a/packages/memtomem/src/memtomem/web/hot_reload.py
+++ b/packages/memtomem/src/memtomem/web/hot_reload.py
@@ -353,13 +353,13 @@ async def _sync_reranker(
                 await asyncio.to_thread(load_model)
             except Exception:
                 logger.exception("Hot-reloaded reranker failed to load; keeping previous instance")
-                await _close_component_safely(new_reranker)
+                await _close_reranker_safely(new_reranker)
                 return
 
     if app is not None and _rerank_snapshot(getattr(app.state, "config", None)) != new_snapshot:
         logger.info("Skipping stale hot-reloaded reranker install; config changed while loading")
         if new_reranker is not None:
-            await _close_component_safely(new_reranker)
+            await _close_reranker_safely(new_reranker)
         return
 
     old_reranker = getattr(search_pipeline, "_reranker", None)
@@ -367,7 +367,7 @@ async def _sync_reranker(
     search_pipeline._rerank_config = new_cfg.rerank if new_cfg.rerank.enabled else None
 
     if old_reranker is not None and old_reranker is not new_reranker:
-        await _close_component_safely(old_reranker)
+        await _close_reranker_safely(old_reranker)
 
 
 def _rerank_snapshot(cfg: Any) -> tuple[object, ...] | None:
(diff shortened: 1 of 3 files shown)


memtomem:

pandas-studio (approved):

memtomem:

memtomem:

PR #1045, opened 2026-05-15
fix(dedup): default mem_dedup_merge to dry_run=True for safety parity

tsdata:
## Summary

- `mem_decay_expire` and `mem_cleanup_orphans` both default to `dry_run=True` so a typo in a maintenance script can't silently destroy data on first call. `mem_dedup_merge` was the odd one out — calling it with the wrong `keep_id`/`delete_ids` hard-deleted immediately.
- Web `POST /api/dedup/merge` is unaffected: the SPA shows a side-by-side scan preview before the confirm modal, so the safety gate lives in the UI there. The gap was the no-preview MCP/automation surface.
- Threads `dry_run` through `DedupScanner.merge` with a `False` default so the existing web call site stays write-through. The MCP wrapper (`mem_dedup_merge`) flips its default to `True`.

Found while running a risk-first QA pass against the destructive-action surface (web UI + MCP).

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_dedup.py` — 11 pass (4 new `TestMergeDryRun` cases + 7 existing scan tests)
- [x] `uv run pytest packages/memtomem/tests/test_usability_fixes.py::TestDedupMerge` — 2 new registry-pin tests pass alongside the existing `TestCleanupOrphans` parity assertion
- [x] `uv run pytest packages/memtomem/tests/test_web_routes_extended.py -k 'dedup or decay'` — 5 pass (web route signature unchanged)
- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/` — 5174 pass / 10 skip / 1 unrelated chromium playwright flake (`test_resolve_aborted_envelope_keeps_card_and_emits_error_toast`, passes in isolation)
- [x] `uv run ruff check packages/memtomem/src && uv run ruff format --check packages/memtomem/src`
- [x] `uv run mypy packages/memtomem/src/memtomem/search/dedup.py packages/memtomem/src/memtomem/server/tools/dedup_decay.py`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1046, opened 2026-05-15
fix(web): tighten decay confirm count, privacy hint, defensive escapes

tsdata:
## Summary

Bundles the P2/P3 follow-ups from the risk-first QA audit + regression pins for the destructive-action and trust-boundary findings the audit caught (sibling of #1045).

**User-visible**

- **F-D4-1**: Decay → Expire confirm modal now shows the would-be-deleted count (`{count}` placeholder + EN/KO locale updates) so a single click can't drop more chunks than the user expected.
- **F-Privacy-2**: Compose panel (Index → New memory) gains a fine-print line clarifying the privacy guard's scope: tokens / API keys / private keys, not PII like email or phone. The dynamic in-textarea pattern warning still fires; this hint sets context up front. EN/KO key added.

**Defensive HTML escapes**

- **F-XSS-1**: `chunk_type` badge inside `browseSource()`'s chunk-card template now routes through `escapeHtml`. Server controls the enum today, but the raw `${...}` interpolation was the only unescaped user-data slot the audit grep flagged in the card layout.
- **F-XSS-2**: `escapeHtml` gains the single-quote → `&#39;` replacement so single-quoted attribute contexts stay safe by default. `escapeAttr` collapses to a thin alias since the prior `.replace(/"/g, ...)` is now a no-op after `escapeHtml` emits `&quot;`.

**Regression pins** (new `tests/test_qa_audit_pins.py`, 10 cases)

- **D2** — admin namespace endpoints stay 404 in prod mode (mounted only on `_DEV_ONLY_ROUTERS` in `web/app.py`).
- **B2-c** — `POST /api/reset` rejects non-loopback ASGI clients even with a forged `X-Forwarded-For: 127.0.0.1` header.
- **S1 + S3** — `mem_search` keeps the parameter set the SPA actually sends; `mem_stats` stays zero-arg so the SPA's poll loop doesn't drift.
- **B5** — chunk_type badge, dedup-row chunk previews, search-result filename, and `escapeHtml`'s single-quote escape are all string-pinned so a future `\${...}` interpolation that drops the wrapper fails the pin instead of shipping an XSS surface.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_qa_audit_pins.py` — 10 new pins pass
- [x] `uv run pytest packages/memtomem/tests/ -k 'decay or compose or i18n or xss or escape or privacy'` — 184 pass (covers `confirm.expire_msg` placeholder parity, redacted JS pattern wiring, etc.)
- [x] `cd packages/memtomem/tests-js && npm test` (vitest + jsdom) — 48 pass
- [x] `uv run ruff check packages/memtomem/src && uv run ruff format --check packages/memtomem/src` — 250 files clean
- [x] `uv run pytest -m "not ollama" packages/memtomem/tests/` — 5179 pass; 13 local environmental flakes (12 in `test_web_cli.py` because the dev `mm web` server is currently running on this machine, and 1 test-ordering flake that passes in isolation), all unrelated to the changes here

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -4818,7 +4818,7 @@ async function browseSource(path, limit = 100) {
         );
         card.innerHTML = `
           <div class="chunk-card-meta">
-            <span class="badge badge-gray">${c.chunk_type.replace('_',' ')}</span>
+            <span class="badge badge-gray">${escapeHtml(c.chunk_type.replace('_', ' '))}</span>
             <span class="chunk-card-lines">lines ${c.start_line}–${c.end_line}</span>
             ${_tierBadgeHtml(c.target_scope)}
             ${c.heading_hierarchy.length ? `<span class="hierarchy-trail">${escapeHtml(c.heading_hierarchy.join(' › '))}</span>` : ''}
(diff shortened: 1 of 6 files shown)


pandas-studio (approved):
LGTM

PR #1051, opened 2026-05-16
test(qa-pins): add F-C4-1 embedding-reset / BM25-fallback regression pin

memtomem:
## Summary

- Adds `TestEmbeddingResetBm25Fallback` (2 cases) to `test_qa_audit_pins.py`, pinning the two invariants behind the kind-moth QA audit's F-C4-1 [Positive] finding: embedding reset concurrent with search degrades to BM25-only, never an HTTP 5xx.
- The probe itself ran 2026-05-15 (timing table in `memtomem-web-kind-moth-findings.md` §C4); this PR is just the regression guard so a future structural change to reset or to the dense leg cannot land without the reviewer re-running the smoke.
- No source changes. Total diff: +99 lines in one test file.

## What's pinned

1. **`search/pipeline.py` dense leg has a defensive `except Exception`** that resets `dense_results = []` and surfaces `dense_error = str(exc)`. Refactor that drops the catch ⇒ embedding-reset / dim-mismatch becomes a user-visible 503.
2. **`SqliteBackend.reset_embedding_meta` is one synchronous transaction** — DROP + CREATE + single `db.commit()`, no inter-statement `await`. Refactor that splits the transaction (background worker, lock, async checkpoint) ⇒ concurrent search could observe `chunks_vec` missing.

Either pin trips ⇒ the F-C4-1 smoke (BM25-only graceful fallback) must be re-run before merge.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_qa_audit_pins.py -v` — 12/12 pass (10 prior + 2 new)
- [x] `uv run ruff check packages/memtomem/src packages/memtomem/tests/test_qa_audit_pins.py`
- [x] `uv run ruff format --check packages/memtomem/tests/test_qa_audit_pins.py`
- [ ] CI green on push

### DIFF
--- a/packages/memtomem/tests/test_qa_audit_pins.py
+++ b/packages/memtomem/tests/test_qa_audit_pins.py
@@ -12,6 +12,9 @@ from the manual QA pass. Each ID maps to a finding in the audit doc:
 * **B5** — chunk-card / dedup-row renderers route every user-controlled
   field through ``escapeHtml`` or ``DOMPurify.sanitize`` before
   ``innerHTML`` so a malicious chunk body can't execute JS in the SPA.
+* **C4 / F-C4-1** — embedding reset concurrent with search degrades to
+  BM25-only (never 503 / sqlite error) because reset is one atomic
+  transaction *and* the dense leg has a defensive exception catch.
 
 The point is to catch a future PR that forgets the guard, not to
 re-derive it from scratch — the underlying enforcement code lives in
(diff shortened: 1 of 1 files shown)


memtomem:
Addressed the [P2] review comment in b7f2ff7.

The new test AST-walks `search/pipeline.py`, collects every `Try` whose body calls `dense_search`, and asserts at least one matches the full F-C4-1 shape — **unconditional `Exception` / `BaseException` handler, `dense_results = []` reset, and `dense_error = str(exc)` surface**. Narrowing the main-path handler from `except Exception as exc:` to e.g. `except RuntimeError as exc:` now drops the candidate out of the match set and fails with a per-Try diagnostic naming the exact regressed line.

Verified by temporarily narrowing line 747 to `except RuntimeError`:

```
AssertionError: search/pipeline.py main-path dense leg no longer matches the F-C4-1 fallback shape ...
  - Try at line 403: catches `Exception`; `dense_results = []`=False; `dense_error = str(exc)`=False
  - Try at line 738: catches `RuntimeError`; `dense_results = []`=True; `dense_error = str(exc)`=True
```

The rescue-path Try at line 403 (`_dense_leg`) is correctly rejected for not surfacing `dense_error`, and the main-path Try at line 738 is correctly rejected for the narrowed handler — exactly the scoping the review asked for.

12/12 pins pass, lint clean.

memtomem:
Addressed the round-2 review in 762ea3b.

The fixture is now `inspect.getsource(SearchPipeline.search)` (dedented), so the AST walk is structurally constrained to `search()`'s body. Helpers, the rescue-path `_dense_leg`, or any future dense_search call site outside `search()` are invisible to the test by construction — there is no codepath where a coincidental shape match on a sibling Try can mask a regression in the main leg.

Tightened the count assertion too: now requires **exactly one** Try wrapping a `dense_search` call in `search()`. Zero means the F-C4-1 structure was removed; more than one means `search()` grew a second dense path and the pin needs an explicit decision before relying on it.

Narrowing verification (line 747 → `except RuntimeError`):

```
AssertionError: SearchPipeline.search dense leg narrowed its handler from `except Exception` to `except RuntimeError` — embedding-reset or dim-mismatch errors not covered by this type will now surface as HTTP 5xx instead of the BM25-only fallback documented in the F-C4-1 runtime probe
```

12/12 pins pass, lint clean.

pandas-studio (approved):
L:GTM

PR #1052, opened 2026-05-16
test(qa-pins): close L5 — btn-danger contrast + setLang race guards

memtomem:
## Summary

Closes the last open row group from the kind-moth QA audit (L5
cross-cutting). Plan: `/Users/steve/.claude/plans/l5-hashed-lantern.md`
(local). Per-row decisions:

- **E-L5-a** (i18n key rendering) — already covered by
  `tests/test_i18n.py:50-68`; no new pin.
- **E-L5-b** (mobile 414×844 button visibility) — accepted; bbox check
  piggybacks on the c Playwright test.
- **E-L5-c / F-L5-1** (dark `.btn-danger` contrast + hover AA) —
  two-layer guard:
  - `tests/test_qa_audit_pins.py::TestBtnDangerContrastGuard` substring-greps
    `style.css` for the dark-only background override and hover
    `brightness(0.92)` rule (catches deliberate removal, <50 ms,
    browser-free).
  - New `tests/web/test_l5_btn_danger_contrast.py` measures effective
    contrast across all 4 `theme × hover` states using CDP
    `CSS.forcePseudoState` (catches the cascade- and filter-composition
    regressions a substring check cannot see). Reuses `mm_web_url`;
    disables CSS transitions so reads taken right after the force are
    not interpolated.
- **E-L5-d** (`I18N.setLang` race) — `TestSetLangAwaitOrder` slices the
  `setLang` body out of `i18n.js` and asserts
  `await _load(lang) < applyDOM() < dispatchEvent('langchange')`.
  `tests-js/i18n-apply-dom.test.mjs` (vitest/jsdom) stays as advisory
  backup; pytest is the required gate.

Deferred a11y items (`focus order, screen-reader announce, dynamic
modal stacking`) from the findings doc are spun out as a separate
`A11Y-1/2/3` audit pass — not relabelled into L5.

### Implementation note

`locator.hover()` does not reliably trigger CSS `:hover` in headless
Chromium — computed style stays at the no-hover values. Switched to
CDP `CSS.forcePseudoState`, which required:

1. Caching `nodeId` across calls because `DOM.getDocument` invalidates
   prior IDs.
2. Reading via `CSS.getComputedStyleForNode` (the JS-side
   `getComputedStyle` does not reflect the forced pseudo-state).
3. Injecting `* { transition: none !important }` so a read taken
   immediately after the force isn't an interpolated mid-transition
   value.

### Regression simulations (intentional corruption + revert)

- Drop `style.css:925` dark bg override → contrast test fails on
  `dark/*`.
- Drop `style.css:930` hover rule → contrast test fails on `*/True`
  (opacity falls back to global 0.85).
- Swap `await` order in `i18n.js:65` → `TestSetLangAwaitOrder` fails.

## Test plan

- [ ] `uv run pytest packages/memtomem/tests/test_qa_audit_pins.py packages/memtomem/tests/web/test_l5_btn_danger_contrast.py -v` — all 20 pass (4 new static + 6 new Playwright + 10 existing).
- [ ] `uv run ruff check packages/memtomem/tests/test_qa_audit_pins.py packages/memtomem/tests/web/test_l5_btn_danger_contrast.py` — clean.
- [ ] `uv run pytest -m "not ollama"` — full suite green outside pre-existing `test_web_cli.py` failures (reproduce on `main` without this branch).
- [ ] Manually corrupt `style.css:925` or `:930` or swap `await` in `i18n.js:65` and re-run the targeted pins to confirm each guard fires.


pandas-studio (approved):
LGTM

PR #1058, opened 2026-05-17
fix(config): coerce all_index_roots() to Path so /api/reindex stops 500ing

memtomem:
## Summary
- `Reindex all` (POST `/api/reindex`) was failing with `500 Internal server error`. Root cause: `config.indexing.memory_dirs` is declared `list[Path]` but `load_config_overrides` writes the JSON-loaded values via `setattr` without Pydantic validation, so entries stayed as raw `str`. `reindex_all` iterates `all_index_roots()` and calls `.expanduser()` directly → `AttributeError: 'str' object has no attribute 'expanduser'`. Single-dir `/api/index` survived because `trigger_index` defensively wraps `Path(req.path)`.
- Fix at the helper that already calls itself the "single source of truth" for index roots: coerce each entry to `Path` in `all_index_roots()`. Every other consumer either re-wraps `Path(d)` (no-op after the fix) or accepts `str | Path` already, so behaviour is unchanged elsewhere.
- Regression pin asserts the helper hands back `Path` instances even when `memory_dirs` is mutated via the same `setattr` path `load_config_overrides` takes.

## Test plan
- [x] `uv run pytest packages/memtomem/tests/test_all_index_roots.py` — 8 passed (incl. new `test_helper_coerces_str_entries_to_path`)
- [x] `uv run ruff check` + `ruff format --check` on changed files — clean
- [x] In-process repro against the user's live `~/.memtomem/config.json` (15 index roots, all raw-string entries): every root reindexes without exception
- [ ] Restart `mm web` and click **Reindex all** in the Sources tab — confirm toast `reindex_complete` instead of `reindex_failed`

### DIFF
--- a/packages/memtomem/src/memtomem/config.py
+++ b/packages/memtomem/src/memtomem/config.py
@@ -310,8 +310,14 @@ class IndexingConfig(BaseSettings):
         view, auto-discover migration); functional consumers that need
         to act on every indexable file go through this helper instead so
         a future scope addition does not fork the consumers.
+
+        Entries are coerced to ``Path`` to honour the declared return
+        type even when raw ``str`` values slipped past Pydantic via the
+        non-validating ``setattr`` path in ``load_config_overrides`` —
+        otherwise ``reindex_all`` blew up calling ``.expanduser()`` on
+        the unwrapped JSON strings.
         """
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #1059, opened 2026-05-17
fix(web): modal-aware global shortcut gate (#1053 PR #3)

memtomem:
## Summary

- Adds a modal-stack-aware gate to both global keydown listeners so bare
  shortcuts (`Cmd+K`, `?`, `/`, `h`, `j`, `k`) cannot pop a second
  overlay on top of one already open — the A11Y-3.1 bug the PR #1 audit
  flagged. `?` and `Cmd+K` are preserved as toggle-closes when their own
  modal owns the top of the stack, so the bare-key UX that opened the
  overlay can also dismiss it.
- Introduces `window.openModal(el, opts)` as the single open path that
  funnels through `openModalA11y`, plus `window.isAnyModalOpen()` /
  `window.isTopModal(el)` as the gate's single source of truth.
- Migrates `path-picker.js` and `context-gateway._ctxResolveConflict` to
  the new wrapper so every overlay actually joins `_ACTIVE_MODALS` (the
  `TestA11yModalToggleAntipattern` pin from the audit).

## Test plan

- [x] `uv run pytest packages/memtomem/tests/web/test_a11y_focus.py::TestA11yShortcutGate -v`
      — all 8 gate scenarios pass (`Cmd+K`/`?` blocked while
      `confirm-modal` open; `/`/`h`/`j`/`k` blocked; `Cmd+K`/`?`
      still toggle-close their own modal when it owns the top)
- [x] `uv run pytest packages/memtomem/tests/test_qa_audit_pins.py::TestA11yModalToggleAntipattern -v`
      — both `path-picker` and `_ctxResolveConflict` migration pins pass
      now that the xfail marker is gone
- [x] `uv run ruff check packages/memtomem/{src,tests}` +
      `ruff format --check` — clean
- [x] Full `pytest packages/memtomem/tests/web/test_a11y_focus.py
      packages/memtomem/tests/test_qa_audit_pins.py` — 78 passed, 2
      xfailed (PR4 + PR5 markers, untouched). 4 pre-existing async-loop
      failures in `TestNamespaceDeleteProdGuard` /
      `TestResetLocalhostGuard` reproduce on baseline when these files
      are collected alongside the Playwright tests — unrelated to this
      change.

Closes the PR #3 row in #1053.

### DIFF
--- a/packages/memtomem/src/memtomem/config.py
+++ b/packages/memtomem/src/memtomem/config.py
@@ -310,8 +310,14 @@ class IndexingConfig(BaseSettings):
         view, auto-discover migration); functional consumers that need
         to act on every indexable file go through this helper instead so
         a future scope addition does not fork the consumers.
+
+        Entries are coerced to ``Path`` to honour the declared return
+        type even when raw ``str`` values slipped past Pydantic via the
+        non-validating ``setattr`` path in ``load_config_overrides`` —
+        otherwise ``reindex_all`` blew up calling ``.expanduser()`` on
+        the unwrapped JSON strings.
         """
-        return list(self.memory_dirs) + list(self.project_memory_dirs)
+        return [Path(d) for d in (*self.memory_dirs, *self.project_memory_dirs)]
 
 
 class DecayConfig(BaseSettings):
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -606,6 +606,28 @@ function closeModal(modal) {
 window.openModalA11y = openModalA11y;
 window.registerModalCloser = registerModalCloser;
 window.closeModal = closeModal;
+
+// openModal(el, opts) — show + a11y in one call. The single open path that
+// keeps _ACTIVE_MODALS accurate; every modal-overlay opener should funnel
+// through here so window.isAnyModalOpen() reflects every overlay on screen
+// and the global shortcut gate (A11Y-3.1) can trust it. Returns the release
+// closure the caller must invoke on close (idempotent, so double-call is
+// safe — e.g. Esc + close-btn race).
+window.openModal = function (el, opts = {}) {
+  show(el);
+  return openModalA11y(el, opts);
+};
+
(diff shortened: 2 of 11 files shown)


pandas-studio (commented):
• The modal shortcut gate hides the command palette as intended, but it does not
  swallow the browser's Ctrl/Cmd+K default when another modal is active, allowing
  focus to escape the modal.

  Review comment:

  - [P2] Prevent default for blocked Ctrl/Cmd+K in modals — /Users/steve/Work/
    agent-harness/memtomem/packages/memtomem/src/memtomem/web/static/settings-
    namespaces.js:766-771
    When any non-command-palette modal is open, Ctrl/Cmd+K enters this modal-open
    guard but does not satisfy isTopModal(qs('cmd-palette')), so the listener
    returns without preventDefault(). In browsers where Ctrl/Cmd+K focuses the
    address/search bar, pressing it from settings/confirm/path-picker can move
    focus out of the trapped modal instead of being swallowed; prevent the default
    for the blocked Ctrl/Cmd+K path before returning.

memtomem:
Good catch on the leaky default — fixed in df76e90.

Approach: keep the existing toggle-close branch but always `preventDefault()` on Cmd/Ctrl+K when a modal is up, regardless of which modal owns the top. Cmd/Ctrl+K is a modifier shortcut (never a printable character) so unconditional preventDefault is safe — no typing-into-modal-input concern. The toggle-close path still fires only when the cmd-palette is on top, so the bare-key UX that opened the palette can still dismiss it.

Added `test_ctrl_k_default_prevented_while_modal_active` as a regression pin: probes `e.defaultPrevented` from a bubble-phase listener (fires after the capture-phase gate has run) while `confirm-modal` is open.

Did not extend the same pattern to bare `?` / `/` / `h` / `j` / `k`: those are printable characters and unconditional preventDefault would break typing them into a modal's text input. `?` still gets preventDefault only on the toggle-close path (which already had it). Happy to revisit if you see a leak there too.

pandas-studio (approved):
• No actionable correctness issues were found in the reviewed changes. The Cmd/
  Ctrl+K modal gate now prevents the browser default while preserving the command
  palette close behavior when it is the top modal.

PR #1060, opened 2026-05-17
fix(web): aria-live=polite on #results-list (#1053 PR #4)

memtomem:
## Summary

- Adds \`aria-live=\"polite\"\` to \`#results-list\` so SR users hear when \`renderResults()\` rewrites the list after a search. The A11Y-2.1 audit row explicitly endorsed placing the attribute on either the list or an adjacent summary node; this repo only has \`#results-list\`, so it lands there.
- Drops the \`_A11Y_XFAIL_PR4\` marker from \`TestA11yResultsLiveRegion\` — the pin already encodes the contract.

## Test plan

- [x] \`uv run pytest packages/memtomem/tests/test_qa_audit_pins.py::TestA11yResultsLiveRegion -v\` — PASSED
- [x] \`uv run ruff check\` + \`format --check\` — clean

Closes the PR #4 row in #1053.


pandas-studio (approved):
LGTM

PR #1061, opened 2026-05-17
fix(web): skip-to-main link + #main landmark (#1053 PR #5)

memtomem:
## Summary

- Adds a visually-hidden `.skip-link` as the first focusable child of `<body>` so the very first Tab from page load lands a "Skip to main content" affordance — keyboard users no longer have to step through ~6 header chrome controls + the tab nav to reach the active tab content.
- Wraps the seven tab panels in `<main id="main">` as the anchor target and primary content landmark. The `<nav>` tab strip stays outside `<main>` so it remains its own landmark; the skip jumps past header + nav both.
- Skip-link copy lives under `nav.skip_to_main` in en + ko.
- Drops the `_A11Y_XFAIL_PR5` marker on `TestSkipLinkPresent`.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_qa_audit_pins.py::TestSkipLinkPresent -v` — PASSED
- [x] `uv run pytest packages/memtomem/tests/web/test_a11y_focus.py::TestA11yBackgroundInert packages/memtomem/tests/web/test_a11y_focus.py::TestA11yStackedModalInertSurvives -v` — 9/9 PASSED (verifying `_recomputeBackgroundInert` still inerts `<main>` as a body child)
- [x] `uv run ruff check` + `format --check` — clean

Closes the PR #5 row in #1053.

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/index.html
+++ b/packages/memtomem/src/memtomem/web/static/index.html
@@ -6,7 +6,7 @@
   <title>memtomem</title>
   <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
   <link rel="apple-touch-icon" href="/favicon.svg" />
-  <link rel="stylesheet" href="/style.css?v=87" />
+  <link rel="stylesheet" href="/style.css?v=88" />
   <link rel="stylesheet" href="/vendor/prism-tomorrow.min.css?v=1" />
 </head>
 <body>
@@ -58,7 +58,7 @@
     <button id="tabbtn-settings" class="tab-btn" role="tab" aria-controls="tab-settings" aria-selected="false" tabindex="-1" data-ui-tier="prod" data-tab="settings" data-i18n="nav.more">More</button>
   </nav>
 
-  <main id="main">
+  <main id="main" tabindex="-1">
   <!-- ===== HOME TAB ===== -->
   <div id="tab-home" class="tab-panel" role="tabpanel" aria-labelledby="tabbtn-home" hidden>
     <div class="home-layout">
--- a/packages/memtomem/src/memtomem/web/static/index.html
+++ b/packages/memtomem/src/memtomem/web/static/index.html
@@ -6,7 +6,7 @@
   <title>memtomem</title>
   <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
   <link rel="apple-touch-icon" href="/favicon.svg" />
-  <link rel="stylesheet" href="/style.css?v=86" />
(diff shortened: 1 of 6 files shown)


memtomem:
Good catch — fixed in 966e695.

Added the four declarations to \`#main\` so it stays transparent to body's column flex layout:

\`\`\`css
#main {
  flex: 1;
  min-height: 0;
  display: flex;
  flex-direction: column;
}
\`\`\`

\`flex: 1\` + \`min-height: 0\` is the same combo \`.tab-panel\` already uses; pairing them on \`#main\` re-establishes the sizing chain from body down to the active panel's internal scroll container.

Also pinned the fix with \`test_main_landmark_preserves_tab_flex_layout\` so a future CSS cleanup can't silently re-introduce the scroll-clipping regression. \`style.css?v=87\` bump on the cache-buster.

memtomem:
P1 + P2 addressed in cfc0efd.

**P1 — \`tabindex="-1"\` on \`<main>\` + runtime pin.** Spot-on: the DOM contract pin was happy but the runtime behavior was broken in exactly the way you described. Added \`tabindex="-1"\` so hash-link activation moves focus to the landmark, paired with \`#main:focus { outline: none }\` + \`#main:focus-visible { outline: 2px solid var(--accent); outline-offset: 2px }\` matching your sketch. New Playwright pin \`test_activating_skip_link_moves_focus_into_main\` drives Tab → Enter and asserts \`document.activeElement\` is inside \`#main\` — exactly the property you flagged as untested.

**P2 — regex-tightened the layout pin.** Replaced the substring search with \`re.search(r"^\s*#main\s*\{([^}]*)\}", css, re.M)\` for the rule lookup and \`\b<prop>\s*:\s*<val>\s*;\` for each declaration. Formatter putting the brace on a new line, comma-combined selectors, and \`flex: 1.5\` all now fail the pin the way they should.

**P3 — banners outside \`<main>\`.** Filed mentally as a follow-up. Agree the \`embedding-mismatch\` banner is the awkward case (it carries actionable buttons that the skip-link bypasses); the other two are pure \`role="status"\` with no interactive content. Likely the right shape is reorganizing the alert-class banner to live inside \`<header>\` or marking its action region so it joins the post-skip flow — both bigger than this PR.

**Nits.** Cache buster v=85→v=88 (absorbed v=86 from PR #4, v=87 from the layout fix, v=88 now). Noted on \`clip-path: inset(50%)\` for a future visually-hidden consolidation pass.

\`style.css?v=88\` bumped.

pandas-studio (approved):
LGTM

Issue #1062, opened 2026-05-18
a11y(web): manual VoiceOver smoke pass (complement to #1053 axe coverage)


memtomem:
## Context

#1053 (A11Y-1/2/3) landed 5 fix PRs (#1055 #1057 #1059 #1060 #1061) plus regression pins (#1054), all driven by **vendored axe-core via CDP**. The issue itself flagged that axe auto-detects only ~**30%** of WCAG issues — the remaining ~70% (announcement quality, traversal order *as heard*, label clarity, modal context cues) needs a human with a screen reader.

This issue tracks that **manual VoiceOver smoke pass** so the gap is visible instead of implicit.

## Scope

WCAG 2.1 AA on `packages/memtomem/src/memtomem/web/static/` — same surface as #1053. Run on macOS VoiceOver (Safari + Chrome). No code changes expected; if defects are found, file follow-up issues and link back here.

## Smoke checklist

For each item, capture: ✅ pass / ⚠️ minor / ❌ defect + brief note.

### Focus & traversal (A11Y-1)
- [ ] Skip-to-main link is announced as first focusable, activates and moves focus to `#main`
- [ ] Tab order through home page matches visual order
- [ ] All 8 modals: opening moves focus into modal, Tab cycles **within** modal, ESC closes and returns focus to opener
  - `confirm-modal`, `path-picker-modal`, `cmd-palette`, `expand-modal`, `source-preview-modal`, `settings-modal`, `shortcuts-modal`, `ctx-conflict-modal`

### Announcements (A11Y-2)
- [ ] Search results update is announced (aria-live polite on `#results-list` from #1060) — verify it doesn't over-announce on every keystroke
- [ ] Icon-only buttons announce a meaningful name (settings, lang/theme toggle, group/view toggle, similar/source close)
- [ ] Form inputs announce a label (home-search, tag-filter, score-threshold, d-editor, d-tag-input, memory-add)
- [ ] Language toggle: `lang` attribute change is reflected in VO pronunciation

### Modal stacking & context (A11Y-3)
- [ ] With `confirm-modal` open, Cmd+K does NOT open command palette (gate from #1059)
- [ ] Modal opening announces role=dialog + accessible name
- [ ] Background content is not reachable via VO rotor while modal is open

## Out of scope

- WCAG AAA, cognitive load, motion-sensitivity (2.3.3) — separate follow-ups if wanted
- NVDA/JAWS — VoiceOver only for this pass; cross-SR can be a later issue

[...]

PR #1065, opened 2026-05-18
fix(web): icon-only close button names + view toggle aria-label sync

memtomem:
## Summary

- Pin accessible names on four icon-only modal close buttons (`expand`, `source-preview`, `settings`, `shortcuts`) that were rendering only the `×` glyph — screen readers had nothing to announce beyond the character itself. Adds `data-i18n-aria-label` + fallback `aria-label`, with en/ko strings.
- Sync the search view-toggle's runtime `aria-label` with its `title`/`textContent` on click. Previously only the visible text and tooltip flipped between card/list; the SR-facing label was frozen at the startup value.
- Two regression pins under `TestIssue1062IconButtonNames` in `tests/test_web_a11y.py` — source-level grep on `index.html` for the four close buttons, and a runtime-block check on `app.js` for the view-toggle listener.

Found while preparing the manual VoiceOver smoke pass tracked in #1062. The smoke pass itself still needs a human VO run; this PR is the easy-to-spot subset that didn't need a screen reader to verify.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/web/test_a11y_focus.py packages/memtomem/tests/test_web_a11y.py -q` → 48 passed
- [x] `uv run ruff check packages/memtomem/src packages/memtomem/tests/test_web_a11y.py` → passed
- [x] `uv run ruff format --check packages/memtomem/src packages/memtomem/tests/test_web_a11y.py` → passed
- [x] `jq empty` on both locale JSONs → valid
- [ ] Manual VO pass — deferred to #1062 (this PR doesn't close that issue)

## Refs

- #1053 (parent a11y audit, CLOSED)
- #1062 (manual VoiceOver smoke — sibling, still open)


memtomem:
Addressed the review in `483ccab`:

1. **Langchange clobber** — dropped `data-i18n-title` / `data-i18n-aria-label` from `#view-toggle`, extracted `_syncViewToggle()`, and registered it as a `langchange` listener so the per-state label survives a locale switch. Updated the static `title` / `aria-label` boot fallbacks to "Switch to list view" so the no-JS state is also accurate for the default card view.
2. **Key naming** — `modal.close_btn_aria` → `modal.close_expand_aria` for symmetry with the three sibling close keys (settings / shortcuts / source_preview). No other call sites used the old key.
3. **Test fragility** — bound `test_view_toggle_updates_runtime_aria_label` by intrinsic anchors (`qs('view-toggle').addEventListener` … `renderResults(STATE.lastResults);`) instead of the `// ---` comment delimiter.
4. **New regression pin** — `test_view_toggle_label_survives_langchange` asserts the HTML has neither i18n attr on `#view-toggle` and that `_syncViewToggle` is wired to `langchange`. This is the guard for (1).

49 passed (was 48 — new test landed green). Ruff check + format and `jq empty` on both locales also clean.

pandas-studio (approved):
LGTM

PR #1066, opened 2026-05-18
fix(web): help-toggle accessible name (#1062 F1)

memtomem:
## Summary

- #1062's manual VoiceOver smoke pass surfaced `help-toggle` (`index.html:27`) as the only header icon-only button without `aria-label` — it relied on `title` alone, which VoiceOver does not reliably announce in form-controls rotor mode.
- Adds `aria-label` + `data-i18n-aria-label` (reusing the existing `header.help_title` key, matching the pattern already on `settings-btn` / `lang-toggle` / `theme-toggle` two lines above).
- Pins it in `TestIssue1062IconButtonNames` (`tests/test_web_a11y.py`) so the gap can't silently re-open.

Same class of defect as #1065 just landed for the view-toggle / close buttons; closes one of the items called out on #1062.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_web_a11y.py::TestIssue1062IconButtonNames -v` — all 4 pass (3 existing + new)
- [x] `uv run ruff check` + `uv run ruff format --check` on the test file — clean
- [ ] Manual VoiceOver smoke (optional, for whoever picks up the remaining #1062 checklist): Cmd+F5, Tab to header `?` button, confirm VO announces "Toggle help hints (h), toggle button, pressed/not pressed" instead of just "button".

## Out of scope

The other items on #1062 (modal announcement quality, live-region timing, rotor traversal) still need a human VO pass — this PR only closes the one programmatically-discoverable gap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1068, opened 2026-05-18
fix(web): aria-label on residual icon-only buttons (#1062 sweep)

memtomem:
## Summary

Static sweep follow-up to #1062 after #1066 landed. Found 10 icon-only `<button>` instances across 6 ids/classes that still relied on `title` alone or had no accessible name at all — same defect class as #1065 / #1066.

| Selector | Before | After |
|---|---|---|
| `#save-star-btn` ☆ | `data-i18n-title` only | + `data-i18n-aria-label="search.save_title"` (reuse) |
| `#delete-query-btn` ✕ | `title="Delete selected"` hardcoded | + new `search.delete_query_aria` |
| `#d-prev-btn` ◀ / `#d-next-btn` ▶ | `title=...` hardcoded | + `data-i18n-aria-label="search.prev_title"` / `next_title` (reuse) |
| `#history-close-btn` ✕ | **nothing** — no title, no label | + new `search.history_close_aria` |
| `.tab-help-bar-dismiss` ✕ (×5) | **nothing** | + new `common.dismiss_help` |

Three new locale keys added with both en and ko translations; existing keys reused where possible.

## Pinning

Extends `TestIssue1062IconButtonNames`:
- Parametrized test over the five fixed id'd buttons
- Class-sweep over every `.tab-help-bar-dismiss` so a new tab can't ship a help bar without a name

## Out of scope

Static sweep only — covers what's grep-discoverable. The non-static items on #1062 (announcement quality, rotor traversal as heard, live-region timing, modal context cues) still need a human VoiceOver session.

Text-input accessible names are also out of scope here. The 6 inputs originally flagged on #1062 were handled in earlier PRs; broader placeholder-only inputs (sources-filter, index-path, dedup/decay numerics, etc.) are a separate judgment call.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_web_a11y.py -v` — 23 pass (10 in `TestIssue1062IconButtonNames`)
- [x] `uv run pytest packages/memtomem/tests/test_i18n.py -v` — 47 pass (locale parity intact)
- [x] `uv run ruff check` + `uv run ruff format --check` on changed files — clean
- [x] Both locale JSONs parse
- [ ] Manual VoiceOver smoke on the 6 button types (left for whoever picks up the remaining #1062 checklist)

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/index.html
+++ b/packages/memtomem/src/memtomem/web/static/index.html
@@ -160,7 +160,7 @@
   <div id="tab-search" class="tab-panel active" role="tabpanel" aria-labelledby="tabbtn-search">
     <div class="tab-help-bar" id="help-search" role="status" hidden>
       <span class="tab-help-bar-text" data-i18n="search.help">Semantic search across indexed memories. Filters: tag, namespace, type, score. Pipeline: BM25 (keyword) + Dense (semantic) → RRF fusion.</span>
-      <button class="tab-help-bar-dismiss" data-help-tab="search">✕</button>
+      <button class="tab-help-bar-dismiss" data-help-tab="search" data-i18n-title="common.dismiss_help" data-i18n-aria-label="common.dismiss_help" title="Dismiss help" aria-label="Dismiss help">✕</button>
     </div>
     <div class="search-bar">
       <button id="filter-toggle" class="btn-ghost btn-icon has-tip" data-i18n-title="search.filter_toggle_title" title="Show/hide tag, type, sort and other filters" aria-controls="search-filters" aria-expanded="false"><span data-i18n="search.filter_toggle">▾ Filters</span><span id="filter-count-badge" class="filter-count-badge" hidden></span></button>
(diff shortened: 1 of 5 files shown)


memtomem:
Addressed in f386be6.

The static `aria-label="Save current search"` on `#save-star-btn` was indeed wrong in the starred state — the button toggles to a *remove* action when ★. Same root cause as the original `#view-toggle` regression that landed earlier in this issue, so I applied the same pattern:

- HTML drops `data-i18n-title` / `data-i18n-aria-label` (the literal `title` + `aria-label` stay as the initial-state default, since empty query == unsaved == "Save current search").
- New `_syncSaveStar()` helper in `settings-namespaces.js` owns glyph + `.starred` class + `title` + `aria-label`, reading the current query and saved list.
- Click handler now delegates; input listener and a new `langchange` listener also call it. The "Save" vs "Remove from saved searches" announcement now flips with the glyph.
- Added `search.unsave_title` to en + ko.
- Test pin `test_save_star_label_survives_langchange` (mirrors `test_view_toggle_label_survives_langchange`) enforces: HTML has no `data-i18n-*` hooks, `_syncSaveStar` exists, click handler delegates to it, langchange listener is wired.

`test_web_a11y.py` (23 pass) + `test_i18n.py` (47 pass) green locally; CI will exercise the JS path via the Playwright smoke matrix.

pandas-studio (approved):
LGTM

PR #1069, opened 2026-05-18
feat(server): Add remote transport options to MCP Server

powerzist:
Added transport options to `memtomem-server` so that the MCP Server can be run remotely, in addition to the existing default `stdio` mode.

Previously, the MCP Server command ran with `stdio` by default. Since running a `stdio`-based MCP server directly in a terminal environment is not useful, `memtomem-server` now displays a guidance message and exits instead of continuing execution in that case.

Added `sse` and `http` (`streamable-http` alias) as supported `transport` options for remote MCP Server execution. Users can start a remote server by providing the `transport`, `host`, `port`, and `url` arguments.

Also updated the `-h` / `--help` text for `memtomem-server` to reflect these changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+- **`memtomem-server` direct TTY launches now exit with guidance.** Running the
+  stdio MCP server directly in a terminal prints MCP-client setup and
+  network-transport examples instead of waiting on stdin.
+
 ## [0.2.1] — 2026-05-13
 
 Patch release focused on Web UI stability and release polish after
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+- **`memtomem-server` now supports network MCP transports.** Adds
+  `--transport sse|http` plus `--host`, `--port`, `--url`, `--allowed-host`,
+  `--allowed-origin`, and `--disable-dns-rebinding-protection` for deployments
+  that need SSE or streamable HTTP instead of stdio. Treat sse/http transports
+  as trusted-network only; place an authenticated reverse proxy in front before
+  exposing publicly.
 - **`memtomem-server` direct TTY launches now exit with guidance.** Running the
   stdio MCP server directly in a terminal prints MCP-client setup and
   network-transport examples instead of waiting on stdin.
--- a/packages/memtomem/src/memtomem/server/__init__.py
+++ b/packages/memtomem/src/memtomem/server/__init__.py
@@ -322,13 +322,133 @@ def _try_hold_legacy_flock(legacy_pid: Path) -> object | None:
     return legacy_fp
 
 
-def main() -> None:
+def _is_direct_stdio_terminal() -> bool:
+    """Return True when stdio mode was launched directly in a terminal."""
+    import sys
+
+    return sys.stdin.isatty() and sys.stdout.isatty()
+
+
+def _print_direct_stdio_help() -> None:
+    """Explain why bare stdio server launches exit immediately."""
+    print(
+        "\n".join(
+            [
+                "memtomem-server is an MCP stdio server.",
+                "",
+                "This command is normally launched by an MCP client over stdin/stdout.",
+                "Do not run it directly in a terminal.",
+                "",
+                "Configure your MCP client with:",
+                "  command: uvx",
+                '  args: ["--from", "memtomem", "memtomem-server"]',
+                "",
+                "Example:",
+                "  claude mcp add memtomem -s user -- uvx --from memtomem memtomem-server",
+                "",
+                "For a manually started network server, use:",
+                "  memtomem-server --transport sse --host 127.0.0.1 --port 8000",
+                "  memtomem-server --transport http --host 127.0.0.1 --port 8000",
+                "",
+                "No MCP client is connected; exiting.",
+            ]
+        )
+    )
+
+
+def _parse_server_args(argv: list[str] | None = None):
+    """Parse ``memtomem-server`` transport options."""
+    import argparse
+
+    parser = argparse.ArgumentParser(
+        prog="memtomem-server",
+        description="Run the memtomem MCP server.",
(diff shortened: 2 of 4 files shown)


memtomem:
## Review — remote transport options

CI is green on all 9 required jobs. Branch is behind `main`; rebase before merge.

### Correctness — mostly clean
- **Exit code 2 for the TTY guard** is conventionally "usage error"; the user invoked the binary intentionally, so exit `0` would feel friendlier for what is really a help message. Minor.
- **Banner goes to stdout.** The TTY-guard fires before `mcp.run()`, so it can't corrupt the stdio handshake today — but routing both `_print_direct_stdio_help` and `_print_network_server_info` to `sys.stderr` would make that guarantee structural rather than ordering-dependent, and matches Unix banner convention.
- **`mcp.settings` is a module-level singleton** mutated by `_configure_network_transport`. Inside `main()` that's fine. Tests don't snapshot/restore it, so a future test ordering change could see leaked `streamable_http_path`/`host`/`port` from a prior test. A small fixture that snapshots `mcp.settings` would harden this.

### Security — sane defaults, one help-text gap
DNS-rebinding protection is on by default, and the allowed-hosts/origins derivation (URL hostname + bind host + `localhost`/`127.0.0.1`/`[::1]`, with `0.0.0.0`/`::` correctly skipped) matches MCP guidance. One ask:

- `--disable-dns-rebinding-protection`'s `help=` says "Advanced" but doesn't flag the risk. Suggest: `"Advanced/dangerous: disables DNS rebinding protection — only safe behind an authenticated reverse proxy."`

### Test coverage — good, four gaps worth filling
Covered: TTY guard exits, stdio-pipe runs silently, `http` alias → `streamable-http`, SSE mount/endpoint split, trailing-slash URL normalization, missing-endpoint-path rejection. Missing:

- `_normalize_endpoint_url` rejecting non-http(s) or schemeless URLs (`file://...`, `127.0.0.1:8000/mcp`).
- `--disable-dns-rebinding-protection` actually setting `enable_dns_rebinding_protection=False` and skipping allowed-host derivation.
- IPv6 host bracket-wrapping in `_host_patterns` (`fe80::1` → `[fe80::1]`, `[fe80::1]:*`).
- Auto-default URL path when `--url` is omitted for `--transport sse`/`--transport http` (the `_default_network_url` branch).

Minor nit in `tests/test_server_cli.py:64`: `assert "mm status" not in out` is a curious negative assertion with no obvious motivation — a one-line comment on what regression it pins against would help future readers.

### Behavior change worth a release note
`memtomem-server` invoked directly in a terminal used to start (and hang on stdin); it now exits with code 2 and prints guidance. Anyone with a `memtomem-server &` wrapper in a TTY will see this as a regression. Worth an explicit line in the release notes.

### Style nit
`server/__init__.py` is now 670+ lines with eight transport helpers added. Splitting them into `server/_transport.py` would keep this file focused on the lock dance. Not blocking — just for future reference.

### Verdict
Address the help-text security warning and consider stderr for the banners; the test gaps and file-split are nice-to-haves. Looks good once rebased.

powerzist:
Addressed the review feedback except for the intentional stdout choice and the optional transport-helper split.

Changes made:
- Added a comment in `server.main()` explaining why the direct-stdio and network startup banners are intentionally safe on stdout.
- Updated `--disable-dns-rebinding-protection` help text to call out that it is dangerous and only safe behind an authenticated reverse proxy.
- Added an `Unreleased` changelog entry for direct TTY `memtomem-server` launches now exiting with guidance.
- Added a test fixture to snapshot/restore `mcp.settings` so transport settings do not leak across tests.
- Added URL validation tests for non-http and schemeless `--url` values.
- Added coverage for `--disable-dns-rebinding-protection` setting `enable_dns_rebinding_protection=False` and skipping allowed host/origin derivation.
- Added default URL coverage:
  - `--transport sse` + omitted `--url` defaults to `/sse`
  - `--transport http` + omitted `--url` defaults to `/mcp`
  - `--host 0.0.0.0` + omitted `--url` derives the public origin from `127.0.0.1`
- Added a short comment explaining why the test asserts `"mm status"` is absent from the direct-launch guidance.

On stdout vs stderr:
I kept the informational banners on stdout intentionally. The direct stdio help only fires in the TTY guard before `mcp.run()` and exits immediately, so it cannot enter the JSON-RPC stdio stream. The network startup banner only runs for non-stdio transports, where stdout is not the MCP transport. I added a code comment to make that ordering/safety contract explicit.

Not changed:
- I did not split the transport helpers out of `server/__init__.py` in this PR since that was noted as non-blocking and would broaden the scope. Happy to do that as a follow-up cleanup.
- I left the direct TTY guard exit code as `2` for now, since the current behavior and test treat this as a usage guard. I can change it to `0` if you feel strongly.

Validation:
- `uv run pytest packages/memtomem/tests/test_server_cli.py`

pandas-studio:
Nice work — code is clean and tests are thorough. Two small things I'd like to land before merge:

1. **CHANGELOG**: the current entry only covers the stdio TTY guard. Please add a line for the main feature (new `--transport sse|http` options + host/port/url/allowed-host/allowed-origin/disable-dns-rebinding-protection flags).
2. **Security note in help/CHANGELOG**: the network transports ship without built-in authentication. The help text currently mentions `0.0.0.0 for direct network access` without a warning. Please add a one-liner along the lines of *"Treat sse/http transports as trusted-network only — place an authenticated reverse proxy in front before exposing publicly."* in both the `--help` epilog and the CHANGELOG entry.

Two follow-up items that are fine to defer:
- streamable-http multi-segment URL path (e.g. `--url https://example.com/api/mcp`) isn't covered by tests; worth a single case to confirm SDK routing matches the SSE mount-path behavior.
- `mcp.settings.*` is mutated as a process-global; tests restore it via fixture, but worth a comment if a single-process stdio+http coexistence is ever expected.

powerzist:
Addressed the latest feedback.

- Added a `CHANGELOG.md` entry for the main network transport feature: `--transport sse|http`, `--host`, `--port`, `--url`, `--allowed-host`, `--allowed-origin`, and `--disable-dns-rebinding-protection`.
- Added the trusted-network/authenticated reverse proxy warning to the changelog entry.
- Added the same security guidance to the `memtomem-server --help` epilog.
- Tweaked the `--host` help text so `0.0.0.0` is described as trusted direct network access, not generic public exposure.

Validation:
`uv run pytest packages/memtomem/tests/test_server_cli.py`

PR #1056, opened 2026-05-17
Retry OpenAI and Anthropic 429 responses

pntech20:
## Summary
- retry OpenAI and Anthropic chat generation when providers return HTTP 429
- honor Retry-After delays through the existing retry helper
- cover successful retry recovery for both providers

Fixes #1037

## Tests
- uv run pytest packages/memtomem/tests/test_llm_providers.py
- uv run --python 3.12 --with ruff ruff check packages/memtomem/src/memtomem/llm/openai.py packages/memtomem/src/memtomem/llm/anthropic.py packages/memtomem/tests/test_llm_providers.py

PR #1084, opened 2026-05-19
refactor(retry): share RateLimitError + promote parse_retry_after

### LINKED ISSUE #1083: Follow-up nits from #1056: dedupe _RateLimitError, promote parse_retry_after, exhaustion tests
## Context

Follow-up nits from #1056 review (retry OpenAI/Anthropic 429s). The PR itself is non-blocking on these — merging as-is, tracking cleanup here.

### 1. Deduplicate `_RateLimitError` and promote `_parse_retry_after` to public API

Currently `_RateLimitError` is defined verbatim in both `packages/memtomem/src/memtomem/llm/openai.py` and `.../llm/anthropic.py`. Each file also reaches into `memtomem.embedding.retry` for the private `_parse_retry_after`.

`retry.py` is the natural home for both — its `with_retry` decorator already inspects `exc.retry_after` via `getattr` (see `retry.py:64`), so the exception class lives in the same conceptual layer as the contract.

Proposed:
- Move `RateLimitError` (drop the leading underscore) into `memtomem/embedding/retry.py`.
- Rename `_parse_retry_after` → `parse_retry_after` (or keep both with a deprecation alias).
- Update `openai.py` / `anthropic.py` to import from the public surface.

Bonus: any future provider (Ollama, etc.) gets the same retry behavior by raising the shared exception.

### 2. Add retry-exhaustion test for both providers

The `except _RateLimitError` block in `OpenAILLM.generate()` and `AnthropicLLM.generate()` (the wrapping into `LLMError` with the "after retries" message) is currently uncovered. Add a test per provider with `side_effect=[rate_limited, rate_limited, rate_limited]` asserting `LLMError` with `match="after retries"`.

### 3. Optional: `Retry-After` parsing edge cases

Add a small parametrized test against `parse_retry_after` for:
- HTTP-date format (RFC 7231)
- Missing / empty header (falls back to exponential backoff)
- Unparseable string

These would live in a `tests/test_retry.py` (or wherever the retry helper tests already are), not the provider tests.

## Out of scope

The provider error-message string ("Please wait before retrying or upgrade your plan.") is currently duplicated across providers. Fine until a third provider lands — defer.

pandas-studio:
## Summary

Follow-up to #1056 (#1083). Three providers — `llm/openai.py`, `llm/anthropic.py`, `embedding/openai.py` — each defined an identical local `_RateLimitError(Exception)` class and reached into `memtomem/embedding/retry.py` for the private `_parse_retry_after` helper.

- Move `RateLimitError` into `retry.py` alongside the `with_retry` contract it satisfies (decorator already inspects `exc.retry_after` via `getattr`).
- Promote `_parse_retry_after` → `parse_retry_after`. The symbol was already imported across modules under its private name, so the underscore was misleading.
- Update all three providers and the two test modules that referenced the old names.
- Add retry-exhaustion tests for both LLM providers (the `LLMError` "after retries" wrapping in `OpenAILLM.generate()` / `AnthropicLLM.generate()` was uncovered).
- Add a `RateLimitError` smoke test that pins the `with_retry` interaction: `retry_after=7.0` wins over `base_delay=1.0` exponential backoff.

No behavior change for callers — same 3-attempt retry budget, same `Retry-After` honoring, same `LLMError` / `EmbeddingError` wrapping.

Closes #1083

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_retry.py packages/memtomem/tests/test_llm_providers.py packages/memtomem/tests/test_embedding_providers.py -m "not ollama"` → 104 passed
- [x] `uv run ruff check` + `uv run ruff format --check` on all touched files → clean
- [x] `grep -rn "_RateLimitError\|_parse_retry_after" packages/memtomem/src/` → no hits

### DIFF
--- a/packages/memtomem/src/memtomem/embedding/openai.py
+++ b/packages/memtomem/src/memtomem/embedding/openai.py
@@ -9,18 +9,10 @@ from typing import Sequence
 import httpx
 
 from memtomem.config import EmbeddingConfig
-from memtomem.embedding.retry import _parse_retry_after, with_retry
+from memtomem.embedding.retry import RateLimitError, parse_retry_after, with_retry
 from memtomem.errors import EmbeddingError
 
 logger = logging.getLogger(__name__)
-
-
-class _RateLimitError(Exception):
-    """Raised on HTTP 429 to trigger retry via with_retry decorator."""
-
-    def __init__(self, retry_after: float | None = None) -> None:
-        self.retry_after = retry_after
-        super().__init__(f"Rate limited (retry_after={retry_after})")
 
 
 class OpenAIEmbedder:
@@ -57,7 +49,7 @@ class OpenAIEmbedder:
     @with_retry(
         max_attempts=4,
         base_delay=1.0,
-        retryable_exceptions=(httpx.ConnectError, httpx.TimeoutException, _RateLimitError),
+        retryable_exceptions=(httpx.ConnectError, httpx.TimeoutException, RateLimitError),
     )
     async def _embed_batch_with_retry(self, batch: list[str]) -> list[list[float]]:
         """Send a single batch to OpenAI with retry on transient errors and 429."""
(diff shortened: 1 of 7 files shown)


memtomem (approved):
LGTM — refactor cleanly consolidates the 3-way duplicate _RateLimitError into retry.py and promotes _parse_retry_after to public. Exhaustion tests close the previously uncovered LLMError 'after retries' wrapping. Merging.

PR #1085, opened 2026-05-19
fix(dedup): align scan timeout to 120s + cover 408 path

pandas-studio:
## Summary

The dedup scan was timing out at 30 s on larger workspaces. The server raised a 408 with a helpful `Try reducing max_scan (current: N).` detail, but the client's `AbortController` fired first (also 30 s) and surfaced a generic `Scan timed out (30 s)` toast instead of the server hint. Symptom: on a 1000+ chunk workspace a scan that would have completed in ~60 s aborted with an unhelpful message.

- **Server** (`web/routes/dedup.py`): bump `_DEDUP_SCAN_TIMEOUT` 30 → 120 s.
- **Client** (`web/static/settings-maintenance.js`): bump fetch abort timer to 120 000 ms and lift it into a named constant. When `res.ok` is false, parse the JSON detail and surface it — the server's `Try reducing max_scan (current: N)` message is more actionable than the bare status text.
- **Test** (`tests/test_web_routes_extended.py`): add `test_dedup_scan_timeout_returns_408` that monkeypatches the timeout to 0.001 s and asserts both the 408 status and the exact detail string. The 408 path was uncovered.

No DB-touching changes; dedup is a read-only candidate scan.

## Test plan

- [x] `uv run pytest packages/memtomem/tests/test_web_routes_extended.py::TestDedup -m "not ollama"` → 3 passed (incl. new test)
- [x] `uv run ruff check` + `uv run ruff format --check` on touched files → clean

### DIFF
--- a/packages/memtomem/src/memtomem/web/routes/dedup.py
+++ b/packages/memtomem/src/memtomem/web/routes/dedup.py
@@ -18,7 +18,7 @@ from memtomem.web.schemas.core import chunk_to_out
 
 router = APIRouter(prefix="/dedup", tags=["dedup"])
 
-_DEDUP_SCAN_TIMEOUT = 30  # seconds
+_DEDUP_SCAN_TIMEOUT = 120  # seconds
 
 
 @router.get("/candidates", response_model=DedupScanResponse)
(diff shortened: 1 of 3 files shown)


memtomem (approved):
LGTM — server/client timeout alignment + new test pins the 408 path that was uncovered. Merging.

PR #1086, opened 2026-05-19
fix(web): allow returning to Server CWD via project dropdown (#1071)

### LINKED ISSUE #1071: ui(ctx-gateway): cannot return to Server CWD via project dropdown (#962 follow-up)
Surfaced in the 2026-05-19 Context Gateway implementation/UI-UX review (local report `docs/reports/context-gateway-implementation-ui-ux-review-2026-05-19.md`, gitignored). Follow-up to #962.

## Problem

`_ctxWireProjectControls()` ignores any changed select value that is falsy. Server CWD is intentionally represented by an empty `scope_id` in the legacy route-default path, so after a user selects an added project, choosing Server CWD in the dropdown becomes a no-op.

## Code references

- `packages/memtomem/src/memtomem/web/static/context-gateway.js:140-148`
- `packages/memtomem/src/memtomem/web/static/context-gateway.js:224-238`
- `packages/memtomem/src/memtomem/web/static/context-gateway.js:241-260`

## Impact

- The only obvious recovery path is to delete the active project, clear local storage, or reload into a state where `_ctxNormalizeActiveScope()` falls back.
- Project switching feels broken exactly after the Add Project success path makes the new project active.

## Recommendation

- Allow `next === ""` as a valid selection and only early-return when `next === _ctxActiveScopeId`.
- Consider rendering the dropdown with an explicit sentinel value for Server CWD, then mapping it back to empty scope internally; this avoids future truthiness mistakes.
- Add a browser or JS test that selects an added project, then selects Server CWD and asserts the next list/overview fetch omits `scope_id`.

pandas-studio:
## Summary

- Drop the `!next` truthiness guard in `_ctxWireProjectControls` so the empty `scope_id` that represents Server CWD is a valid selection again. Only `next === _ctxActiveScopeId` short-circuits the change handler.
- Add a one-line code comment on the contract (empty `scope_id` ⇒ Server CWD) so the next reviewer doesn't reintroduce the same truthiness mistake.
- Pin both directions with a vitest regression test: positive (switching from an added project to Server CWD persists `''` and triggers `loadCtxOverview`) and negative (re-picking the active scope still no-ops).

Closes #1071.

## Why

`_ctxFetchProjects`'s legacy fallback (`context-gateway.js:207`) and `_ctxNormalizeActiveScope` both treat the Server CWD scope as having `scope_id = ''`, so `_ctxScopeParam` can leave the param off and preserve the single-project URL shape. The change handler's `if (!next || ...)` short-circuited on that same `''`, making it impossible to return to Server CWD from the dropdown once the user had switched to an added project — clearing localStorage or reloading was the only recovery.

## Test plan

- [x] `npx vitest run` in `packages/memtomem/tests-js` (11 files, 50 tests pass).
- [x] Mutation check: revert the source change, rerun new test — positive case fails (asserts `localStorage` flips to `''`); negative case stays green because it pins the unaffected no-op path.
- [x] `uv run ruff check packages/memtomem/src` + `ruff format --check` clean.

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/context-gateway.js
+++ b/packages/memtomem/src/memtomem/web/static/context-gateway.js
@@ -243,8 +243,10 @@ function _ctxWireProjectControls() {
     if (select.dataset.scopeWired === 'true') return;
     select.dataset.scopeWired = 'true';
     select.addEventListener('change', () => {
+      // Server CWD is intentionally represented by an empty scope_id, so
+      // treat '' as a valid selection and only short-circuit a no-op re-pick.
       const next = select.value || '';
-      if (!next || next === _ctxActiveScopeId) return;
+      if (next === _ctxActiveScopeId) return;
       _ctxActiveScopeId = next;
       _ctxNormalizeActiveScope(_ctxProjectsCache);
       _ctxBumpActiveScopeDetailSeq();
--- a/packages/memtomem/tests-js/ctx-project-switch-server-cwd.test.mjs
+++ b/packages/memtomem/tests-js/ctx-project-switch-server-cwd.test.mjs
@@ -1,0 +1,130 @@
+/* Regression guard for #1071 — return to Server CWD via project dropdown.
+ *
+ * Server CWD is intentionally represented by an empty ``scope_id`` (the
+ * legacy route-default URL shape omits ``scope_id`` for the server-CWD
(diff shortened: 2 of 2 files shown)


memtomem (approved):
LGTM — one-line truthiness fix on the Server CWD scope_id contract + symmetric pin/invert regression test. Mutation validation in PR body confirms the positive case fails without the source change.

PR #1109, opened 2026-05-25
feat(context): fan out hooks to Codex + Gemini runtimes (ADR-0018)

memtomem:
## What

Fan out canonical `.memtomem/settings.json` hooks to **Codex** and **Gemini**, not just Claude. Hooks were the only context-gateway artifact locked to a single runtime; they now fan out to all three from one canonical source. (ADR-0018, layering on ADR-0010.)

Two commits (kept separate for cleaner review/revert/release-notes):
1. `feat(context): fan out hooks to Codex + Gemini runtimes (ADR-0018)`
2. `docs(context): correct Codex command fan-out wording` — drive-by: commands fan out to Claude/Gemini only (no `CodexCommandsGenerator` exists; the docstring + guide claimed otherwise).

## How

- **Codex** (`~/.codex/hooks.json`): same event names + native `Bash`/`Edit`/`Write` matchers → reuses the Claude additive merge; drops `Notification`/`SessionEnd` (unsupported) with a warning.
- **Gemini** (`~/.gemini/settings.json`): event remap (`PreToolUse`→`BeforeTool`…) + tool-name matcher remap (`Bash`→`run_shell_command`, `Edit`/`MultiEdit`→`replace`, `Write`→`write_file`, `Read`→`read_file`) + handler `name` synthesis; unmappable events/matchers dropped with a warning.
  - `UserPromptSubmit`→`BeforeAgent` and `Stop`→`AfterAgent` are **best-effort** lifecycle mappings (approximate timing — Gemini documents no exact analog; recorded as a policy choice in code + ADR-0018, **not** a verified 1:1).
- `SettingsGenerator.target_file()` widened to `Path | None` (Codex/Gemini have no `project_local` hooks target); `generate_all_settings` / `diff_settings` / `host_write_targets` / `detector.detect_settings_files` treat `None` as skipped.
- Web Settings hooks sync no longer masks a per-runtime `error`/`aborted` result as success.

Mappings verified against official docs: Claude (code.claude.com/docs/en/hooks), Codex (developers.openai.com/codex/hooks), Gemini (gemini-cli `docs/hooks/writing-hooks.md` + `docs/tools/file-system.md`).

## Scope / follow-ups (intentional)
- Rich Web surfacing of per-runtime conversion status (runtime-readiness doctor + per-artifact compatibility badges) is **PR2** in the roadmap; this PR keeps the Web change minimal.
- A `--strict` mode (fail instead of warn on lossy conversion) is future work.

## Test plan
- `uv run pytest packages/memtomem/tests/test_context_settings.py packages/memtomem/tests/test_context_settings_multiruntime.py -q` → **88 passed**
- `uv run pytest packages/memtomem/tests/web/test_settings_hooks_sync_confirm.py -q` → **6 passed** (incl. new per-runtime error → no-silent-success pin)
- `uv run ruff check packages/memtomem/src` + `ruff format --check` → clean
- `uv run mypy packages/memtomem/src/memtomem/context/settings.py` → clean

## Review
Codex (3-agent reviewer) round-3 verdict: **SHIP**. Round-1 (silent-failure masking + matcher normalization) and round-2 (Gemini `replace` / lifecycle mapping) were NEEDS-FIX — all resolved and re-verified.

### DIFF
--- a/docs/adr/0018-multi-runtime-hook-fanout.md
+++ b/docs/adr/0018-multi-runtime-hook-fanout.md
@@ -1,0 +1,101 @@
+# ADR-0018: Multi-runtime hook fan-out (Codex + Gemini)
+
+**Status:** Accepted
+**Date:** 2026-05-25
+**Context:** ADR-0010 established the `hooks.target_scope` tiers but fan-out
+was Claude-only (`~/.claude/settings.json`). Codex CLI and Gemini CLI have
+since shipped their own hook systems, so the single canonical
+`.memtomem/settings.json` hooks record can now fan out to all three runtimes —
+the highest-automation-value artifact in the gateway, previously the only one
+locked to one runtime. This ADR records that decision and the conversion
+rules. It **layers onto** ADR-0010 rather than amending it, per the repo
+convention (ADR-0010 §"Considered & rejected" rejects in-place amendment of an
+Accepted ADR; ADR-0007 / ADR-0008 layer onto ADR-0001 the same way).
+
+## Decision
+
+1. **Three generators.** `SETTINGS_GENERATORS` gains `CodexSettingsGenerator`
+   and `GeminiSettingsGenerator` next to `ClaudeSettingsGenerator`. The
+   canonical hooks record (keyed by Claude event name) fans out to:
+
+   | Runtime | user scope | project_shared | project_local |
+   |---|---|---|---|
+   | Claude | `~/.claude/settings.json` | `<proj>/.claude/settings.json` | `<proj>/.claude/settings.local.json` |
+   | Codex  | `~/.codex/hooks.json` | `<proj>/.codex/hooks.json` | — (none) |
+   | Gemini | `~/.gemini/settings.json` | `<proj>/.gemini/settings.json` | — (none) |
+
+2. **Codex is near-identical.** Codex shares Claude's event names and accepts
+   `Bash` / `Edit` / `Write` matchers natively, so it reuses the same
+   `_merge_hooks_record` additive merge as Claude. Events Codex lacks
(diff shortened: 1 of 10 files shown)


memtomem:
Thanks for the catch — the underlying point about memtomem updating its own generated rules is valid and now tracked in #1110. But I don't think it's a regression for this PR, and two details in the comment need correcting:

1. **Gemini hook fan-out is new in this PR.** `GeminiSettingsGenerator` isn't on `main`, so no released version has ever written `memtomem-…` Gemini hook names. There's no upgraded-install population carrying an old name format — the feature ships atomically with the current name + ms-timeout. The only install that could hold the old `-{idx}` names is one built from an intermediate commit of *this branch*, not a release.

2. **The new name is a content hash, not a positional ordinal.** `_ensure_gemini_handler_names` emits `memtomem-{event}-{slug}-{digest}`, where `digest = sha1(f"{original_matcher}\x00{idx}\x00{command}")[:8]`. There is no cross-rule `{rule_ordinal}` — I deliberately avoided a positional index so the name stays reorder-stable and re-syncs are idempotent (pinned by `test_synthesized_name_is_deterministic_across_syncs`).

The real kernel — `_merge_hooks_record` keys conflicts by `(event, matcher)`, so a same-matcher rule whose content changed is treated as a *user* conflict and skipped rather than updated — is a **pre-existing property of the additive merge that applies to Claude and Codex too** ("user rules always win"). It isn't introduced or worsened here, and a first-time sync is unaffected. Making memtomem-owned rules self-updating across output changes is a broader, cross-runtime change with its own clobber-risk tradeoff, so I've split it out to #1110 rather than fold it into this PR.

PR #1111, opened 2026-05-25
feat(context): mark memtomem-owned hook rules so re-sync can update them (ADR-0019)

### LINKED ISSUE #1110: context-gateway: re-sync can't update memtomem's own previously-emitted hook rules (additive merge treats them as user conflicts)
### Summary

`_merge_hooks_record` in `packages/memtomem/src/memtomem/context/settings.py` does an additive merge of the canonical hooks record into each runtime's settings file, keying conflict detection on `(event, matcher)`. When a same-`(event, matcher)` rule already exists with **different content**, the existing rule is kept and a warning is emitted — the documented "user rules always win" semantics.

That's correct for genuinely user-authored rules, but it also means **memtomem cannot update a rule it previously wrote itself**. If memtomem's generated output for a given `(event, matcher)` changes between two releases a user syncs across (a handler `name`, `timeout`, or command-format change), the next `mm context sync --include=settings` sees the old memtomem-written rule as a same-matcher *user* conflict, refuses to replace it, and the install stays warning/out-of-sync until the user manually deletes the stale rule.

### Scope

Pre-existing; applies to **all three** settings runtimes (Claude, Codex, Gemini), not Gemini-specific.

Surfaced during review of #1109 (introduces Gemini fan-out; changed the synthesized handler-`name` format mid-PR). **#1109 itself is not affected for released users** — Gemini fan-out is new there and ships atomically (no prior release emitted the old name format). This issue is the forward-looking upgrade path the *next* time any generated output changes.

### Possible approaches

- Mark memtomem-authored rules (handler `name` prefix `memtomem-`, or a `_source: memtomem` marker) and let the merge **replace** an existing memtomem-owned rule on the same `(event, matcher)` instead of treating it as a user conflict.
- Risk to weigh: a user who hand-edits a `memtomem-`named rule (keeping the name) would be clobbered — needs an explicit "`memtomem-` is a reserved namespace" contract + tests across all three runtimes.

### Repro sketch

1. Sync settings so memtomem writes a hook rule for some `(event, matcher)`.
2. Change memtomem's generated output for that rule (name/timeout/command).
3. Re-sync → conflict warning, stale rule retained, no update.

memtomem:
## Summary

`_merge_hooks_record` keyed conflict detection on `(event, matcher)` with no way to tell a **user-authored** rule from one **memtomem wrote in a prior release**. So when memtomem's generated output for an `(event, matcher)` changed between two releases (a `command`, `timeout`, or handler-`name` change), the next `mm context sync --include=settings` treated memtomem's own stale rule as a user conflict, refused to update it, and the install stayed warning/out-of-sync until the user deleted the rule by hand. Pre-existing; applied to all three runtimes (Claude, Codex, Gemini).

Closes #1110. Records the decision in **ADR-0019** (layers onto ADR-0018).

## Approach — official-field ownership marker

A custom marker key (`_memtomem`) was **rejected**: the runtime hook schemas don't document tolerance for unknown keys, and a strict validator could reject the whole settings file. The marker reuses only **documented** handler fields:

| Runtime | Marker field | Reserved value |
|---|---|---|
| Claude / Codex | command-handler `statusMessage` | prefix `memtomem · ` |
| Gemini | handler `name` | prefix `memtomem-` |

- **Stamping** is idempotent and preserves any author-supplied `statusMessage` after the prefix; Gemini's `memtomem-` `name` is always (re)stamped so memtomem owns it.
- **Merge** replaces a same-`(event, matcher)` memtomem-owned rule in place (FIFO per matcher), prunes owned rules canonical no longer emits, and keeps the **user-wins** warning for unmarked rules — sharpened for legacy command-matches, but **never silently clobbered**.
- **Functional comparison** ignores the marker-carrier fields, so a byte-identical user rule is not flagged as a conflict the moment memtomem stamps its copy.
- All three Claude **web paths** (GET diff, `resolve`, POST sync) stamp/classify through one shared helper; the GET diff mirrors the merge's owned-slot resolution exactly (add / update / conflict / prune / surplus).

## Tests

New coverage in `test_context_settings.py` (Claude), `test_context_settings_multiruntime.py` (Codex/Gemini), and `test_web_routes_extended.py` (web): re-sync update, user-wins + legacy warning, idempotency, preserve-author-text, Gemini name override + no `statusMessage` leak, stale-event / surplus-duplicate / empty-canonical pruning, and GET↔merge consistency. `ruff` + `mypy` clean.

## Reviewed

Multi-round Codex review (dev-trio); all blocker/major findings on the merge engine and primary write paths addressed. One **deferred** follow-up (filed separately): `resolve_conflict` matches conflicts by `(event, matcher)` and can't disambiguate *duplicate* same-matcher rows — a **pre-existing** limitation (other mutation endpoints already use `rule_index`/`rule_hash`), out of scope for #1110.

### DIFF
--- a/docs/adr/0019-hook-rule-ownership-marker.md
+++ b/docs/adr/0019-hook-rule-ownership-marker.md
@@ -61,15 +61,16 @@ an Accepted ADR).
    rule, so memtomem does **not** silently overwrite it (a genuine user rule
    that merely shares a command must not be clobbered). Instead, when an
    unmarked colliding rule shares a `command` with the contribution, the warning
-   is sharpened to "looks like a memtomem-managed rule from a previous version —
-   remove it and re-run sync." After one marked sync, the rule self-updates.
+   explains that it may be a memtomem-managed rule from a previous version and,
+   if so, should be removed before re-running sync. After one marked sync, the
+   rule self-updates.
 
-5. **All Claude write paths agree.** The CLI merge, the web `GET
-   /settings-sync` diff (`_compare_hooks` stamps the canonical before comparing
-   and classifies a differing memtomem-owned target rule as *pending*, not a
-   conflict), and the web `POST /settings-sync/resolve` "use memtomem's version"
-   action all stamp through the one shared helper, so no path writes an unmarked
-   rule.
+5. **All Claude write paths agree.** The CLI merge, settings-migrate target
+   writes, the web `GET /settings-sync` diff (`_compare_hooks` stamps the
+   canonical before comparing and classifies a differing memtomem-owned target
+   rule as *pending*, not a conflict), and the web
+   `POST /settings-sync/resolve` "use memtomem's version" action all stamp
+   through the one shared helper, so no path writes an unmarked rule.
 
 ## Consequences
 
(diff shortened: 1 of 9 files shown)


pandas-studio (approved):
LGTM

PR #1113, opened 2026-05-25
docs(hooks): fix Claude hook timeout example to seconds (was 5000/10000 ms)

memtomem:
## Summary

The Claude Code integration guide (`docs/guides/integrations/claude-code.md`) showed hook timeouts of `5000` / `10000` with surrounding language implying **5 s / 10 s** ("ensures hooks don't block the session"; "searches complete within 100–500ms"). But Claude Code — and Codex — interpret a hook `timeout` as **seconds** (default 60). So a user copying `"timeout": 5000` gets a ~83-minute cap, which defeats the non-blocking intent.

## Changes

- Example hook config: `5000` → `5`, `10000` → `10` (seconds).
- FAQ: `timeout: 5000` → `timeout: 5`, with a "(seconds)" clarification.
- Added a callout: hook `timeout` is in seconds; don't use millisecond values; `mm context sync` converts the canonical (seconds) value to milliseconds for Gemini automatically.

## Context

Surfaced while reviewing the Gemini timeout conversion during #1110 (PR #1111). The seconds→milliseconds conversion in `_remap_for_gemini` is **correct** for a Claude-shaped canonical (Claude=seconds → Gemini=ms); the real footgun was this docs example steering users toward millisecond values that Claude then over-interprets. Docs-only; affects Claude/Codex users directly, independent of the Gemini fan-out.

### DIFF
--- a/docs/guides/integrations/claude-code.md
+++ b/docs/guides/integrations/claude-code.md
@@ -182,7 +182,7 @@ Add the following to `~/.claude/settings.json`:
       "hooks": [{
         "type": "command",
         "command": "mm session start --idempotent --auto-end-stale 24h --agent-id claude-code 2>>/tmp/mm-hook.log || true",
-        "timeout": 5000
+        "timeout": 5
       }]
     }],
     "UserPromptSubmit": [{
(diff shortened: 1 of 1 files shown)


pandas-studio (approved):
LGTM

PR #1116, opened 2026-05-25
fix(web): dedup project-fetch toast + preserve active scope on transient failure (#1101, #1102)

### LINKED ISSUE #1101: ui(ctx-gateway): project-fetch failure toast re-fires per panel (no dedup) (#1099 follow-up)
Follow-up to PR #1099. The new "Project list failed to load" toast fires
once per `_ctxFetchProjects()` invocation, but the function is called from
at least three independent panel-load paths, so a single persistent
outage stacks N near-identical toasts as the user navigates.

## Problem

`_ctxFetchProjects()` has three entrypoints in the current tree:

- `context-gateway.js:933` — `loadCtxOverview`
- `context-gateway.js:1847` — settings projects panel
- `settings-hooks-watchdog.js:373` — hooks sync flow

Each independently triggers the new `warn` toast on every call. There is
no cooldown, no last-fired memo, no de-dup on `warn.detail`. A user who
opens the context overview, then settings, then hooks sync against a
persistent 5xx will see three identical toasts in quick succession.

## Code reference

- `packages/memtomem/src/memtomem/web/static/context-gateway.js:240-242`

## Impact

- Noisy UX during a real outage (one user-visible symptom → many toasts).
- Not severe enough to block ship, but it inverts the "non-blocking" intent
  in the PR #1099 commit message.

## Recommendation

Two viable shapes — decide in the issue before opening a PR:

1. **Module-level last-fired memo** — cache `(warn.kind, warn.detail)` and
   skip the toast if it matches the last fire within e.g. 30s. Cheapest.
2. **Per-detail dedup with explicit clear** — store the last fired key,
   clear on first successful fetch. Avoids stale memos across long
   sessions.

Either way, extend `tests-js/ctx-project-fetch-failure.test.mjs` with a
"two consecutive 503s → one toast" cell that mutation-validates by
removing the memo (cell flips red).

Scope: 1 file (`context-gateway.js`) + 1 test file. No locale, no API
change.

### LINKED ISSUE #1102: ui(ctx-gateway): transient project-fetch failure permanently clears active scope (#1099 follow-up)
Pre-existing behavior surfaced by PR #1099's new toast. A single transient
5xx response from `/api/context/projects` discards the user's persisted
active-project selection — when the endpoint recovers, the selection is
gone, with no warning that state was dropped.

## Problem

On the synthetic-fallback path, `_ctxFetchProjects` still calls
`_ctxNormalizeActiveScope(_ctxProjectsCache)` against the one-element
Server-CWD scope. The lookup for the user's previous `_ctxActiveScopeId`
(e.g. `proj-abc`) misses, normalization rewrites the in-memory id to
`''`, and `_CTX_ACTIVE_SCOPE_KEY` in localStorage is overwritten at
`context-gateway.js:191`.

When the endpoint recovers on the next page load, `proj-abc` is back in
the real `scopes` list, but the persisted selection no longer points to
it — the user is silently demoted to Server-CWD.

The PR #1099 commit message argues this normalization is intentional
("a transient 5xx doesn't leave `_ctxActiveScopeId` dangling at a
now-absent project"). The in-memory part is fine; **persisting** the
synthetic-collapse to localStorage is the surprising bit, because the
project is not actually absent — only this fetch failed.

## Code reference

- `packages/memtomem/src/memtomem/web/static/context-gateway.js:191`
  (localStorage write inside `_ctxNormalizeActiveScope`)
- `packages/memtomem/src/memtomem/web/static/context-gateway.js:239`
  (call site that runs against synthetic data after fallback)

## Impact

- User-visible: project switcher silently resets to Server-CWD after any
  transient 5xx, even though the toast says "falling back" (implying
  temporary).
- Pre-existing: yes — the localStorage write predates #1099. But until
  #1099 the failure was silent, so users had no signal that anything had
  happened. The toast now points at a state mutation users will reasonably
  expect to be reversible on the next successful fetch.

## Recommendation

When `_ctxFetchProjects` falls through to the synthetic fallback, either:

1. **Skip the localStorage write** — keep `_ctxActiveScopeId` reset to
   `''` in memory (so downstream requests work) but do **not** persist;
   the next successful fetch restores from localStorage. Smallest change.
2. **Suppress normalization entirely on synthetic** — pass a flag so
   `_ctxNormalizeActiveScope` knows the scope list is synthetic and
   should not commit to localStorage.

Add a JS test cell that asserts `localStorage.getItem(_CTX_ACTIVE_SCOPE_KEY)`
is unchanged after a 503-then-recovery sequence. Mutation-validated by
removing the suppression: cell flips red.

Scope: 1-2 files (`context-gateway.js`, possibly a small refactor of the
`_ctxNormalizeActiveScope` signature) + 1 test file. No locale, no API
change.

memtomem:
## Summary

Two PR #1099 follow-ups in `_ctxFetchProjects` (`context-gateway.js`), both keyed off the existing `warn` signal so toast behavior and active-scope behavior stay aligned.

### #1101 — failure toast re-fires per panel (no dedup)
The "Project list failed to load" toast fired once per `_ctxFetchProjects()` call, but the function runs from three independent panel-load paths (overview, settings projects, hooks sync), so a single persistent outage stacked N near-identical toasts as the user navigated.

A module-level `_ctxProjectsFetchWarnKey` memo now de-dups on the `kind:status:detail` failure shape and **clears on a clean (or silent-404) fetch**, so a *later, distinct* outage still notifies.

### #1102 — transient failure permanently clears active scope
A transient 5xx ran `_ctxNormalizeActiveScope` against the synthetic one-element Server-CWD list and **persisted** that collapse to `localStorage`, silently demoting the user's active project; recovery couldn't restore it.

Normalization is now gated on `!warn` — the same signal that drives the toast:

| Outcome | Toast (`warn`) | Normalize? | Active scope |
| --- | :---: | :---: | --- |
| 200 real scopes | no | yes | authoritative |
| 404 / network throw | no (silent) | yes | clear now-stale `proj-*` (older-deploy contract) |
| 5xx / non-404 4xx | **yes** | **no** | **preserved** — restored on recovery |
| 200 + malformed JSON | **yes** | **no** | **preserved** |

`_ctxScopeParam` already omits `scope_id` when the active id is absent from the cache, so requests during the degraded window safely fall back to Server-CWD.

### Draft-scope consistency (surfaced by the #1102 fix)
Preserving the active id while only the synthetic Server-CWD scope is cached means the **request** scope (`_ctxScopeParam` omits the uncached id → Server-CWD) and the **conflict-draft** key (`_ctxStashKey`, which used the raw active id) could diverge — a draft saved/restored during the outage would be keyed under `proj-abc` and cross-contaminate the real project after recovery. Both now route through a single `_ctxEffectiveScopeId` helper (which also backs the `_ctxRestoreDraft` legacy-fallback guard), so the draft namespace always matches the scope the request actually used.

## Test plan

Extended `tests-js/ctx-project-fetch-failure.test.mjs` (JSDOM + Vitest), all mutation-validated:

- **#1101** — two consecutive 503s → **one** toast (mutation: drop the memo → 2); re-notifies after recovery + a fresh outage.
- **#1102 symmetric pair** — 503 preserves the persisted scope and restores it on recovery (mutation: `normalize-always` → red); 404 clears a now-stale id (mutation: `skip-always` → red).
- **Draft scope** — during a 503 the preserved-but-uncached id keys drafts under `__default__`, not `proj-abc` (mutation: raw-id key → red), and `_ctxScopeParam()` is empty.

- [x] `npx vitest run` — 63/63 pass (was 58; +5 new cells)
- [x] Each new cell mutation-validated (flips red under the inverted production logic)
- [x] No Python touched (ruff/mypy unaffected); locale/API unchanged

`context-gateway.js?v=44` → `?v=45` so the silent build isn't cached.

## Review

Two Codex review rounds folded in: (1) the initial `!synthetic` skip regressed the silent-404/older-deploy path (stale active id leaking into `_ctxRestoreDraft`) → re-gated on `!warn`; (2) preserving the id surfaced the request-vs-draft scope divergence above → unified via `_ctxEffectiveScopeId`. An untracked `node_modules` symlink flagged in round 1 was removed (replaced with a real `npm ci`'d, gitignored dir).

Closes #1101
Closes #1102

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/context-gateway.js
+++ b/packages/memtomem/src/memtomem/web/static/context-gateway.js
@@ -126,6 +126,16 @@ const _CTX_ACTIVE_SCOPE_KEY = 'memtomem_ctx_active_scope_id';
 let _ctxActiveScopeId = '';
 let _ctxProjectsCache = [];
 
+// De-dup memo for the `/api/context/projects` failure toast (#1101).
+// ``_ctxFetchProjects`` runs from three independent panel-load paths
+// (overview, settings projects, hooks sync), so a single persistent outage
+// would otherwise stack one near-identical toast per path as the user
+// navigates. We remember the last fired ``kind:status:detail`` key and skip
+// re-toasting an identical failure; a successful (or silent-404) fetch clears
+// the memo so a *later*, distinct outage notifies again instead of being
+// swallowed by a stale key.
+let _ctxProjectsFetchWarnKey = null;
+
 try {
   _ctxActiveScopeId = localStorage.getItem(_CTX_ACTIVE_SCOPE_KEY) || '';
 } catch {
@@ -137,14 +147,29 @@ function _ctxTargetScopeParam() {
   return `target_scope=${encodeURIComponent(_ctxTargetScope)}`;
 }
 
-function _ctxScopeParam(scopeId = _ctxActiveScopeId) {
+// The scope a request *effectively* targets, as a bare id ('' === Server-CWD).
+// An active id that isn't an available, non-Server-CWD scope in the cache —
+// Server-CWD itself, a now-``missing`` scope, or a selection preserved across
+// a transient projects-fetch outage where only the synthetic Server-CWD scope
+// is cached (#1102) — collapses to Server-CWD. This is the single source of
+// truth for "what scope are we really on"; ``_ctxScopeParam`` (request URL) and
+// ``_ctxStashKey`` (conflict-draft key) both route through it so they can never
+// disagree — otherwise a draft saved while the request silently fell back to
+// Server-CWD would be keyed under the preserved project id and leak across
+// scopes after recovery.
+function _ctxEffectiveScopeId(scopeId = _ctxActiveScopeId) {
   if (!scopeId) return '';
   const activeScope = (_ctxProjectsCache || []).find(scope =>
     scope && scope.scope_id === scopeId && !scope.missing);
+  if (!activeScope || _ctxScopeIsServerCwd(activeScope)) return '';
+  return scopeId;
+}
+
+function _ctxScopeParam(scopeId = _ctxActiveScopeId) {
   // Server CWD is the route default. Leaving scope_id off preserves the
   // legacy single-project URL shape while still sending ids for added projects.
-  if (!activeScope || _ctxScopeIsServerCwd(activeScope)) return '';
-  return `scope_id=${encodeURIComponent(scopeId)}`;
+  const eff = _ctxEffectiveScopeId(scopeId);
+  return eff ? `scope_id=${encodeURIComponent(eff)}` : '';
 }
 
 function _ctxWithTargetScope(url, opts = {}) {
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #1119, opened 2026-05-29
fix(context): refuse to overwrite agent files when context.md has zero sections

memtomem:
## Problem
`mm context sync` (and its MCP twin `mem_context_sync`) wrote `gen.generate(sections)` to every detected agent file even when `parse_context()` returned `{}` — which happens for any `context.md` that has content but no `## Heading` delimiters (a stub, an all-prose file, or one the user is mid-edit on). `ClaudeGenerator.generate({})` yields only header boilerplate and `CursorGenerator` yields `""`, so `sync` silently overwrote the user's existing `CLAUDE.md` / `GEMINI.md` / `.cursorrules` with near-empty output — **data loss with no `.bak`**.

The sibling `generate` verb has always guarded this (`context_cmd.py:932`); `sync` and `mem_context_sync` did not.

## Fix
Mirror the empty-sections guard on both surfaces: when `sections` is empty, emit a "refusing to overwrite" notice and skip the write loop. The happy path (sections present) is unchanged.

## Tests
`test_context_sync_empty_guard.py` pins both the refusal (existing file preserved) and the happy path (file still regenerated) for CLI + MCP. Full context suite green; ruff clean.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -1086,24 +1086,36 @@ def sync_cmd(
 
     if ctx_path.exists():
         sections = parse_context(ctx_path)
-        files = detect_agent_files(root)
-
-        if files:
-            agents_to_sync = {f.agent for f in files}
-
-            for agent_name in sorted(agents_to_sync):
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #1120, opened 2026-05-29
fix(context): stop overrides/ leaking into skill fan-out and clear false diff drift

memtomem:
Two coupled per-vendor-override bugs in the canonical→runtime fan-out.

## M3 — `overrides/` leak (skills)
`generate_all_skills` staged the whole canonical skill tree via `copy_tree_atomic`, which copied the canonical-side `overrides/` SOURCE directory into every runtime tree (`.claude/skills/<n>/overrides/gemini.md` ended up in Claude's tree). That leaks every other vendor's override bytes cross-vendor, and lets one vendor's override secret block the whole fan-out at scan time.

**Fix:** `copy_tree_atomic` gains a keyword-only `skip_top_level` set applied **only at the root** of the call (not propagated into recursion); `_stage_skill` passes `{"overrides"}` on the two fan-out paths so the source is **never written** to staging — no crash window, no silent leak if a post-copy delete failed. A legitimate nested `scripts/overrides/` is preserved. Import (runtime→canonical) and migrate (canonical→canonical) keep the default and retain the source.

## M2 — permanent false "out of sync"
`diff_{agents,commands,skills}` compared the **plain** canonical render against the override-**applied** runtime file, so any artifact carrying a per-vendor override reported "out of sync" forever — in `mm context diff`/`status`, the web status routes, and the MCP diff tool — unresolvable by re-sync.

**Fix:** agents/commands diff compare the override **byte-exact** against the runtime (sync writes raw bytes via `atomic_write_bytes`; decoding as UTF-8 + strip crashed on non-UTF-8 overrides and masked whitespace-only drift). `diff_skills` compares against the effective fan-out tree via `_skill_effective_equal`, which mirrors `copy_tree_atomic` exactly: `COPY_SKIP_NAMES` + symlinks excluded at every depth, top-level `overrides/` excluded canonical-side only, `SKILL.md` replaced by the override bytes. An unreadable resolved override reports drift (matching sync's typed skip) instead of masking it.

## Tests
overrides/ does not leak (+ nested `scripts/overrides/` survives); diff in-sync after override fan-out for all three kinds; byte-exact non-UTF-8 override (agents+commands); unreadable override → drift; `__pycache__` no longer causes false drift. Full context + wiki/install/atomic suites green; ruff clean.

### DIFF
--- a/packages/memtomem/src/memtomem/context/_atomic.py
+++ b/packages/memtomem/src/memtomem/context/_atomic.py
@@ -155,7 +155,13 @@ def atomic_write_text(
     atomic_write_bytes(path, text.encode(encoding), mode=mode)
 
 
-def copy_tree_atomic(src: Path, dst: Path, *, mode: int = 0o644) -> int:
+def copy_tree_atomic(
+    src: Path,
+    dst: Path,
+    *,
+    mode: int = 0o644,
+    skip_top_level: frozenset[str] | None = None,
+) -> int:
     """Recursively mirror *src* → *dst*, each file via :func:`atomic_write_bytes`.
 
     Returns the number of files written. ``mode`` (default ``0o644``) is the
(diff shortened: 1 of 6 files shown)


pandas-studio (approved):
LGTM

PR #1121, opened 2026-05-29
fix(context): resolve web project root by walking up, matching CLI/MCP

memtomem:
## Problem
The web lifespan pinned `app.state.project_root = Path.cwd()` with no ancestor walk, while `cli/context_cmd.py` and `server/tools/context.py` both walk up to 10 levels for a `.git`/`pyproject.toml` marker. Launching `mm web` from a project subdirectory therefore wrote canonical artifacts to `<subdir>/.memtomem` (via `resolve_scope_root` → `canonical_artifact_dir`) while the same user's `mm context sync` wrote to `<repo>/.memtomem` — the two surfaces silently targeted **different canonical trees** for one project.

## Fix
Extract root detection into a single shared `find_project_root()` in `context/scope_resolver.py` (its natural home, beside `project_root_from_artifact_path`) and route all three surfaces through it: the web lifespan now walks, and the CLI/MCP `_find_project_root` become thin wrappers — also removing the byte-for-byte duplicated helper so root detection is defined **once**. Prior CLI/MCP semantics preserved (walk 10, `.git` OR `pyproject.toml`, fall back to original cwd, no resolve).

The memory write path (`system.py`) already prefers `_resolve_project_context_from_dirs` and only falls back to `app.state.project_root`, so the walked root only **improves** that fallback; no consumer regresses.

## Tests
`find_project_root` unit cases (cwd marker, subdir→repo-root walk, pyproject ancestor, no-marker fallback, explicit start), a CLI/MCP wrapper-delegation parity pin, and a source-level lifespan pin (`inspect.getsource`) that fails if `app.py` reverts to `Path.cwd()`. 30 scope-resolver tests + full context suite green; ruff clean.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -80,7 +80,7 @@ from memtomem.context.settings_migrate import (
     format_plan_summary,
     plan_migration,
 )
-from memtomem.context.scope_resolver import canonical_artifact_dir
+from memtomem.context.scope_resolver import canonical_artifact_dir, find_project_root
 from memtomem.context.skills import (
     diff_skills,
     extract_skills_to_canonical,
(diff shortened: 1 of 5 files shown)


pandas-studio (approved):
LGTM

Issue #1100, opened 2026-05-22
ui(ctx-gateway): shape-invalid 200 still silently collapses to Server CWD (#1080 / #1099 follow-up)


memtomem:
Follow-up to #1080 / PR #1099. The fix split the catch into HTTP / parse /
404 / network branches, but a 200 response with a *structurally* unexpected
body still escapes both warning branches and reproduces the original
"unreadable store masquerading as 'no registered projects'" symptom.

## Problem

`_ctxFetchProjects()` only treats a response as failing when:

- `!res.ok` (HTTP error), or
- `res.json()` throws (SyntaxError).

If the body parses but is not the expected `{scopes: Array}` shape — `null`,
`{}`, `{"error": "db locked"}`, a string, an array — the function silently
falls through to either a TypeError (for literal `null`) or an empty
`_ctxProjectsCache = []`, with **no toast**. This is the same UX state #1080
was filed to eliminate, just reached via a different code path than the one
PR #1099 covered.

## Code reference

- `packages/memtomem/src/memtomem/web/static/context-gateway.js:214-219`
  (the inner `try { data = await res.json() } catch (parseErr)` only catches
  parse failures, not shape failures)
- `packages/memtomem/src/memtomem/web/static/context-gateway.js:238`
  (`data.scopes || []` — crashes on `data === null`, silently empties on
  `data = {}`)

## Impact

- A backend bug that returns `200 OK` with a malformed body looks identical
  to "user has no registered projects" — the exact failure mode #1080
  describes.
- For `data === null` (server returns the JSON literal `null`),
  `data.scopes` throws and the caller's catch surfaces a generic "Failed to
  load overview", with the toast line never reached.

## Recommendation

After `data = await res.json()` succeeds, validate the shape and route
shape failures through the same `warn = { kind: 'shape', detail }` path
PR #1099 added for parse failures. Roughly:

```js
data = await res.json();
if (!data || !Array.isArray(data.scopes)) {
  warn = { kind: 'shape', detail: `unexpected response shape: ${typeof data}` };
  throw new Error(warn.detail);
}
```

Extend `tests-js/ctx-project-fetch-failure.test.mjs` with two cells —
`200 + null body` and `200 + {} body` — mirroring the symmetric-pin
pattern (`feedback_pin_invert_symmetric_assertion.md`) already used there.
Mutation-validated by deleting the new shape branch: both new cells flip
red.

Scope: 1 file (`context-gateway.js`) + 1 test file. No API change, no
[...]

PR #1122, opened 2026-05-29
fix(web): localize context-gateway error/empty/diff text, polish a11y + mobile

memtomem:
UI/UX gaps in the context gateway surface that the rest of the panel's heavy localization had skipped.

## i18n
The list/detail/diff loaders, the Create form labels/placeholder, the tier-filter buttons, and the import receipt rendered hardcoded English — so a Korean user hit "Failed to load", "Diff failed", "No project scopes", raw `project_shared`/`project_local` tier tokens, and "Name"/"Content"/"Skipped" mid-panel. Routed through `t()` with **16 matching `settings.ctx.*` keys in both locales** (placeholder parity preserved); `t()` output escaped before every `innerHTML` interpolation.

## a11y + mobile
The three injected banners (runtime-only remediation, deep-link filter, write-blocked) now carry `role="status"` so screen readers are told when a tier flip blocks writes or a filter narrows the list. The conflict-modal "Your edits / Current on disk" preview row now stacks (`flex-direction: column`) at `<=520px` instead of splitting into two unreadable ~225px panes during the destructive Reload-vs-Force choice.

## Correctness
The settings overview tile (nav button **and** remediation pointer) no longer deposits an inert `?filter=` deep-link when navigating to a section that can't consume it (`hooks-sync`), which left a stale/misleading shareable URL. `loadCtxOverview` gains the #1100 null/shape guard, and the create-form error branch guards `r.json()` with `.catch()` like every other handler. Cache-bust → `v=47`.

## Tests
i18n key/placeholder guard (47), Playwright context-gateway suite (63), Vitest ctx suite (22) — all green.


pandas-studio (approved):
LGTM

Issue #1123, opened 2026-05-29
Context Gateway: deferred Minor/Nit cleanup backlog (2026-05-29 full-stack audit)


memtomem:
Backlog of confirmed **Minor/Nit** findings from a full-stack context-gateway audit (12 dimensions, adversarially verified — 47 confirmed / 7 refuted). The 4 **Major** fixes shipped as #1119 (M1 sync data-loss), #1120 (M2+M3 override leak/diff drift), #1121 (M4 web project-root walk); several Minor/Nit were folded into #1122 (UI/UX). This issue tracks the **remaining deferred items**, grouped into focused-PR buckets. Each bucket should branch off `main` **after #1119–1122 land** (they touch the same files).

Legend: 🐛 bug · 🔁 parity · 📝 doc · 🎨 ux · 🕳 gap. Line numbers are point-in-time — re-verify before editing (this repo's reviews have a documented stale-line failure mode).

### Already shipped — do NOT re-file
- #1119 M1 · #1120 M2+M3 · #1121 M4 (incl. `_find_project_root` dedup)
- #1122 folded: list/detail/diff + Create-form i18n, tier-filter friendly labels, banner `role=status`, settings-tile stale `?filter=`, mobile conflict-modal CSS, create-error `.catch()`, `loadCtxOverview` null-guard
- #1145 Bucket 3 (Concurrency & atomicity, B3-1…B3-5) — also fixed a pre-existing Windows test-path failure split out as #1146
- #1132 Bucket 1 (B1-1/B1-2) · #1133 Bucket 2 (all) · #1135 Bucket 4 (all) · #1137 Bucket 5 (B5-3/4/5/6) · #1138 + #1140 Bucket 6 (all) · #1134 Bucket 7 (B7-4/B7-5)

**Deferred design items — split to #1147:** B1-3 (preamble-before-first-heading), B1-4 (Rules/Style round-trip), B5-1, B5-2, B7-1, B7-2, B7-3. Each is a design / behavior-contract decision (not a mechanical fix), intentionally deferred in its bucket PR; now tracked in #1147 so this issue's cleanup buckets are complete. (Verified against `main` 2026-05-31; per-item rationale inline below.)

---

### Bucket 1 — Parser / generator hardening — ◑ (#1132): B1-1/B1-2 fixed · B1-3/B1-4 → #1147
[...]

PR #1132, opened 2026-05-30
fix(context): harden section parser against round-trip data loss (#1123 B1)

memtomem:
## What

Bucket 1 of the #1123 deferred context-gateway audit backlog: harden the
section parser against three silent round-trip data-loss bugs.

`parse_context` and `extract_sections_from_agent_file` each duplicated a naive
`^## ` heading loop that lost content in three ways:

- **B1-1** — a `##` line *inside a fenced code block* was parsed as a section
  delimiter, truncating the real section and minting a spurious one. (Both
  ` ``` ` and `~~~` fences.)
- **B1-2** — a repeated `## Name`, or two headings aliasing to one canonical
  key (e.g. `Rules` + `Coding Rules` → `Rules`), overwrote the earlier copy.
- **B1-3a** — a whitespace-only `##` heading opened a section under an
  empty-string key, stranding its body there.

## How

Extract a single fence-aware `iter_markdown_sections()` helper in `parser.py`
and route **both** round-trip halves through it, so they can no longer drift.
Merge (concatenate) on duplicate keys instead of overwriting.

## Deferred (tracked in #1123, not in this PR)

These two B1 items are **design questions**, not clean fixes, so they are
deliberately out of scope here:

- **B1-3b preamble capture** — capturing text before the first heading under a
  generic key would make `parse_context` swallow the `# Project Context` H1,
  which `sections_to_markdown` re-prepends → a duplication regression; it would
  also misattribute the heading-less Project body that cursor/copilot emit. It
  changes the dict→markdown contract.
- **B1-4 Rules+Style reverse-merge** — already has a shipped warning
  (`b340bc7`); every fix option trades off the compact-output design intent for
  cursor/codex/copilot.

## Test

Adds `TestParserHardening` (6 cases) covering all three fixed bugs plus a
happy-path round-trip regression guard.

- `uv run pytest -k context -m "not ollama"` → **1183 passed**
- `ruff check` ✓ · `ruff format --check` ✓ · `mypy` clean

Refs #1123 (Bucket 1)

### DIFF
--- a/packages/memtomem/src/memtomem/context/generator.py
+++ b/packages/memtomem/src/memtomem/context/generator.py
@@ -2,10 +2,11 @@
 
 from __future__ import annotations
 
-import re
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Protocol
+
+from memtomem.context.parser import iter_markdown_sections
 
 
 class AgentGenerator(Protocol):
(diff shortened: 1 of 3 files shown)


memtomem:
Follow-up fix in `ee122e6` for the nested-fence finding.

The initial B1-1 fix toggled a single `in_code` flag on **any** fence line, so a mismatched inner fence closed the block early — a `~~~` inside a ` ```markdown ` example flipped code-mode off and the in-code `##` was parsed as a real heading again.

Repro (now fixed) — before: `Architecture`, `not a section`, `Rules`; after: `Architecture`, `Rules`.

The parser now tracks the **opening** fence's marker char + length and only closes on a matching fence (same marker, length ≥ opener, no trailing info string), per CommonMark; leading indentation up to 3 spaces is allowed. Added regression tests for the nested-fence case and for a too-short inner fence that must not close a longer block. Context suite + ruff + mypy green.

pandas-studio (approved):
LGTM

PR #1134, opened 2026-05-30
fix(context): surface init settings status + narrow dead tier Literal (#1123 B7)

memtomem:
## What

Bucket 7 of the #1123 deferred context-gateway audit backlog: two small
correctness/clarity fixes.

| # | Finding | Fix |
|---|---|---|
| B7-4 | `mm init`'s settings step looped over `generate_all_settings` results handling only `ok`/`skipped`, **silently dropping** `needs_confirmation` (host-write guard refused) and any other status | Add an explicit `needs_confirmation` branch pointing at `mm context sync --include=settings`, plus a defensive `else` so no status is ever swallowed. |
| B7-5 | `ProjectScope.tier` was typed `Literal["user", "project"]` but `discover_project_scopes` only ever constructs `"project"` — the `"user"` arm was never produced | Narrow the type to `Literal["project"]`. (The `"user"` tier in `settings_doctor` is a separate dataclass and is unaffected.) |

## Deferred (tracked in #1123, not in this PR)

These three B7 items each have a legitimate as-designed defense, so they are out of scope here:

- **B7-1** — `~/.claude/projects` decoder treats every `-` as a path separator. The encoding (`/`→`-`) is fundamentally lossy; the docstring already owns this and the caller filters via `is_dir()`. A real fix is a design change.
- **B7-2** — the "already exists with different config" merge warning. This is the intended "user rules always win" contract; rewording risks muddying it.
- **B7-3** — `runtime_artifact_names` returns raw on-disk stems without `validate_name`. That's a deliberate listing of whatever is physically on the runtime fan-out root; validating could crash a diff on a pre-existing oddly-named file.

## Test

`test_init_settings_status.py` — drives `_step_settings` with the interactive bits stubbed and asserts every status is surfaced.

- `pytest test_init_settings_status.py` → **3 passed**
- init / projects / settings_doctor suites → green
- `ruff check` ✓ · `ruff format --check` ✓ · `mypy` clean

Refs #1123 (Bucket 7)

### DIFF
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -964,6 +964,18 @@ def _step_settings(state: dict) -> None:
                 click.secho(f"  Merged → {r.target}", fg="green")
             elif r.status == "skipped":
                 click.secho(f"  skipped {name}: {r.reason}", fg="yellow")
+            elif r.status == "needs_confirmation":
+                # The host-write guard refused this target. Surface it instead
+                # of silently dropping the status (#1123 B7-4) — the user-scope
+                # fan-out can legitimately hit a host file that needs opt-in.
+                click.secho(
+                    f"  {name}: needs confirmation — re-run "
+                    "`mm context sync --include=settings` to apply",
+                    fg="yellow",
+                )
+            else:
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #1135, opened 2026-05-30
fix(context): migrate lockfile cleanup, user-tier status, settings-migrate apply re-classify (#1123 B4)

memtomem:
## What

Bucket **B4 (migrate/status)** of the Context-Gateway deferred backlog (#1123) — three correctness fixes:

### B4-1 — dangling lockfile entry → status "missing"
`migrate_scope` moved a canonical artifact out of `project_shared` but never touched `lock.json`, so a later `mm context status` iterated the stale entry, found the canonical gone from the project_shared dest, and reported the (now-moved) artifact as **missing**.

- New `Lockfile.remove_entry(asset_type, name)` — atomic (sidecar-lock + `atomic_write_bytes`), mirrors `upsert_entry`; preserves sibling entries and unknown top-level/per-entry fields per ADR-0008; returns `False` and leaves the file untouched when there's nothing to remove.
- `migrate_scope` calls it **best-effort** after the move commits, only when `src_scope == "project_shared"` (the sole lockfile-tracked tier). The canonical move already committed inside the lock, so a bookkeeping failure must never undo or fail the migration — it logs and continues.

### B4-2 — `status --scope user` returned empty
The `--scope user` flag was wired but nothing enumerated the user tier, so it always filtered an empty row set.

- New `scan_user_artifacts()` walks `~/.memtomem/{agents,commands,skills}/` (shared draft-walk helper with `project_local`), surfaced by the CLI **only for `--scope user`**.
- **Design decision:** the user-tier walk is kept OUT of `classify_status`. Folding the global `~/.memtomem` walk into it would couple every status read — and every test exercising one — to the caller's real home directory (a local↔CI flake source). `classify_status` stays project-rooted; the CLI surfaces the global tier on demand.
- Render now keys its "draft" columns/counts on `state == "local-draft"` (behavior-preserving for project_shared & project_local). User drafts get dash columns but **no `(draft, no fan-out)`** tag — user artifacts, unlike project_local, *do* fan out to runtime dirs.

### B4-3 — settings-migrate applied stale plan-time classification (TOCTOU)
`apply_migration` trusted the `already_at_target` / `conflict_at_target` flags frozen at **plan** time. If the target tier drifted between plan and apply (another writer, a manual edit, a second run), it could append a canonical rule the target already grew — a same-matcher duplicate Claude Code fires twice — or write past a freshly introduced conflict.

- `apply_migration` now re-reads and **re-classifies** the target against its live state, deriving the write-set and source-clean-set from the current target, not the snapshot. Drift discovered at apply time is refused per-move (no append, source entry left intact) and recorded in `MigrateResult.warnings`.
- The CLI prints those warnings and turns them into **exit 1**, matching the existing plan-time conflict contract.

## Tests
- `Lockfile.remove_entry`: delete + return True; absent → False with **no rewrite** (byte-identical); sibling + unknown-field round-trip.
- Migrate `project_shared → {project_local, user}` drops the entry; status shows no `missing` row.
- User-tier scan: dir + flat layout, skills-flat-not-recognised, `classify_status` stays project-rooted (no user rows), CLI `--scope user` lists artifacts / shows the empty message.
- Apply-time drift: drift-to-conflict refuses + warns + leaves source intact; drift-to-identical skips the redundant write but cleans source; CLI surfaces the warning and exits 1.

## Gates
- `ruff check` ✅ · `ruff format --check` ✅ · `mypy` ✅ (changed files clean) · full suite **5455 passed, 10 skipped**.

## Deferred
- **MCP parity** for status/settings-migrate is out of scope here (consistent with the bucket's audit-vs-runtime split); the user-tier status surface and the apply-time drift guard are CLI/engine-side. Tracked under #1123.

Closes part of #1123 (B4).

### DIFF
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -59,7 +59,7 @@ from memtomem.context.migrate import (
 )
 from memtomem.context.projects import KnownProjectsStore
 from memtomem.context.lockfile import LockfileVersionError
-from memtomem.context.status import classify_status, load_with_recovery
+from memtomem.context.status import classify_status, load_with_recovery, scan_user_artifacts
 from memtomem.context.generator import (
     GENERATORS,
     extract_sections_from_agent_file,
@@ -1562,7 +1562,13 @@ def status_cmd(scope_flag: TargetScope) -> None:
 
     wiki = WikiStore.at_default()
     wiki_head, rows = classify_status(root, wiki=wiki)
-    rows = [row for row in rows if row.tier == scope_flag]
+    if scope_flag == "user":
+        # User-tier artifacts live in the global ~/.memtomem store, outside
+        # the project root classify_status walks; enumerate them on demand
+        # only when this scope is requested (#1123 B4-2).
+        rows = sorted(scan_user_artifacts(), key=lambda r: (r.asset_type, r.name))
+    else:
+        rows = [row for row in rows if row.tier == scope_flag]
 
     if lockfile_error is not None:
         click.secho(f"  ✗ lock.json: {lockfile_error}", fg="red", err=True)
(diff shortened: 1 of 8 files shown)


pandas-studio (approved):
LGTM

PR #1137, opened 2026-05-30
fix(context): close MCP/web parity gaps from the B5 backlog (#1123 B5-3/4/5/6)

memtomem:
Part of #1123 (bucket **B5 — MCP parity**). Four small, independent CLI→MCP / cross-route parity gaps from the 12-dimension context-gateway audit. The two design-laden items in the bucket are **deferred** (see below).

## Fixes

- **B5-3 — cross-tier duplicate-hook warnings on the MCP settings path.** The CLI surfaces these inside the real generate / diff / sync workflow via `_print_duplicate_tier_warnings` (ADR-0010 §4), but the MCP settings branches dropped them — so an MCP caller never learned a memtomem-managed hook was duplicated in a non-active tier. New `_settings_dup_tier_warnings` helper, threaded into `mem_context_generate` / `mem_context_diff` / `mem_context_sync` settings branches (resolved lazily inside the `settings` branch, so artifact-only callers don't pay the detector cost).

- **B5-4 — `on_drop=warn` was unreachable via MCP.** `generate` / `sync` only exposed the boolean `strict` and never passed `on_drop` to `generate_all_agents` / `generate_all_commands`, so the helper always saw the `ignore` default (or `error` via `strict`). Added an `on_drop` arg (`ignore` / `warn` / `error`) with up-front validation mirroring the CLI's `click.Choice(ON_DROP_LEVELS)`; `strict` kept as the documented legacy alias. Precedence (per `_sync_atomic`): an explicit non-default `on_drop` wins; `strict` only bites when `on_drop` is still `ignore`.

- **B5-5 — rename `mem_context_migrate` `apply_` → `apply`.** The trailing underscore was a CLI-internal Click convention that leaked into the MCP tool schema. (CLI-internal `apply_` in `migrate_cmd` / `_memory_migrate_run` is unchanged.)

- **B5-6 — `create_skill` returned HTTP 400 on conflict.** A bare `ValueError("already exists")` mapped to 400 via the app's `value_error_handler`, while the sibling `create_agent` / `create_command` routes raise `HTTPException(409)` — and the Web UI conflict-resolution flow keys off 409. Aligned to 409.

## Deferred (not in this PR — need a maintainer decision)

- **B5-1** — artifact scope-tier + flat→dir `mm context migrate` modes have no MCP counterpart (the existing `mem_context_migrate` only mirrors `memory-migrate`).
- **B5-2** — `mem_context_migrate` is named as if it mirrors `mm context migrate` but only covers the memory subset.

These two are entangled: closing B5-1 means **building a new MCP surface** (wrap the unified `migrate` verb's `asset_type` dispatch) and B5-2 is the naming decision that falls out of it. That's a build-vs-document call of the same shape as B6, so it's left in #1123 for a decision rather than shipped speculatively.

## Testing

- New `test_server_tools_context_on_drop.py` (on_drop reachability + strict/on_drop precedence both directions) and `test_server_tools_context_dup_warnings.py` (exact `format_warning` string surfaced on generate/diff/sync + negative pin).
- Updated `test_server_tools_context_migrate.py` (`apply` rename) and `test_web_routes_context.py` (skill conflict → 409).
- Docs: `docs/guides/mcp-clients.md` Context row updated for `on_drop` + the `apply` rename.
- `ruff check` + `ruff format --check` clean; `mypy` clean (249 files); full suite **5470 passed, 10 skipped** (`-m "not ollama"`).


pandas-studio (approved):
LGTM

PR #1138, opened 2026-05-30
docs(adr): correct non-existent `mm context list` refs + flip ADR-0011 to Accepted (#1123 B6)

memtomem:
Part of #1123 (bucket **B6 — docs/ADR**). Docs-only ADR accuracy fixes. The user decided to **amend the docs** (not build a tier-spanning `mm context list` verb — that's feature work beyond this Minor/Nit backlog).

## B6-1 / B6-2 — the non-existent `mm context list` verb (6 refs)

The ADRs cite `mm context list`, which **does not exist**. Real verbs: `detect / diff / generate / init / install / memory-migrate / migrate / rescan / settings-doctor / settings-migrate / status / sync / update`.

- 5 refs → **`mm context status`** — the real per-tier listing verb, which already emits the exact `(draft, no fan-out)` annotation the ADRs attribute to `list` (`context/status.py`).
- The ADR-0016 §3 *"CLI spans by default"* claim was **false for artifacts**: only `mm mem search` (memory) and the Claude Code runtime span by default. Canonical artifact reads have **no all-tier surface** (`status` is per-tier; the Web overview is single-tier-by-default per ADR-0015 §4a/§4f). Reworded so it no longer implies a single all-tier view; cross-tier collision visibility is "inspect each tier."

## B6-3 — ADR-0011 status

Flip `**Status:** Proposed` → `Accepted`. ADR-0011 is the oldest gateway ADR (2026-05-09), fully implemented via PR-E (#889/#890/#893); all peers (ADR-0009/0010/0015/0016) are already Accepted.

## B5-1 / B5-2 (deferred, documented here per maintainer decision)

The existing ADR-0011 *"MCP exposure for `mm context init`/`sync`/`migrate`"* open-question item is updated to current reality: #887 landed MCP parity for `init`/`sync`/`generate`/`diff`, and memory migration is `mem_context_migrate` — but the **artifact** scope-tier + flat→dir modes of `mm context migrate <kind> <name>` remain **CLI-only by design** (no agent/skill/command MCP migrate path). Recorded as a deliberate gap rather than built.

## Scoped out → follow-up

Different surface (not ADR prose), so kept out of this focused PR:
- **B6-4** — plugin `SKILL.md` frontmatter (`context/`, `setup/`) lists non-core `mem_related` / `mem_config`.
- **B6-5** — web client-side 409 conflict-flow has no test.

[...]

Issue #763, opened 2026-05-04
RFC: 409 mtime_ns conflict — diff/merge/force-save UX


memtomem:
## Motivation

The Context Gateway editor uses optimistic concurrency: `GET` returns
the canonical content + `mtime_ns`; `PUT` echoes back `mtime_ns` and
the server returns 409 if the file changed underneath the editor.

Today the front-end's only response to 409 is:

```js
// packages/memtomem/src/memtomem/web/static/context-gateway.js:569-571
if (r.status === 409) {
  showToast(t('settings.ctx.mtime_conflict'), 'warning');
  loadCtxDetail(type, name);
}
```

A toast + silent reload — the user's edits are simply replaced by
whatever's on disk. If the conflict is real (another editor / runtime
migration / git pull mid-session), the user has no way to:

- see what changed,
- merge their edit on top, or
- override the on-disk version with their text intact.

This is fine when conflicts are rare. With multiple sessions per
project and the imminent Settings prod expansion (RFC #761),
conflicts will become more common, and silently
discarding the user's text is the wrong default.

## Current state

- Backend `PUT /context/{type}/{name}` returns 409 with body
  `{"detail": "..."}` and the current `mtime_ns` in the
  detail. (Verify — currently the body is plain text "File was
  modified by another process. Reload and retry.")
- Front-end loses the user's edits on 409. There is no diff display,
  no force-save, no merge.
- The Diff tab can already render a 3-way-ish view (canonical vs.
  runtime), but it's not wired to the conflict path.

## Proposed change

Introduce a **conflict resolution dialog** triggered on 409:

```
┌─ Conflict: someone else modified this file ─────────────┐
│                                                          │
│  Your edits           │   Current on-disk                │
│  ─────────────────    │   ─────────────────              │
│  <user buffer>        │   <fresh GET content>            │
│                                                          │
│  [ Reload (discard my edits) ]                           │
│  [ Force save (overwrite their changes) ]                │
│  [ Open diff in editor ]                                 │
└──────────────────────────────────────────────────────────┘
```

Key UX rules:

- **No silent reload** — the toast-and-reload behavior is gone. Edits
  are not destroyed without explicit user choice.
- **"Force save" requires backend support**: a `force: true` field
  on the PUT body that bypasses the mtime_ns check. Server logs the
  override at WARN with both mtime values.
- **"Open diff in editor"** stashes the user's buffer, switches to
  the Diff tab, and renders user-buffer-vs-on-disk. The user copies
[...]

PR #1140, opened 2026-05-30
fix(plugin,web): core-mode skill tools + 409 conflict-modal test (#1123 B6-4/B6-5)

memtomem:
Part of #1123 (bucket **B6**, residual non-ADR items split out of #1138). Two unrelated-but-small fixes on different surfaces.

## B6-4 — plugin skills declared non-core tools

The plugin runs the MCP server in the default `core` mode (it sets no `MEMTOMEM_TOOL_MODE`), and every other plugin skill uses core-only tools. Two skills broke that convention by declaring tools the `core` mode removes:

- **`context/SKILL.md`** used `mem_related` (in the `relations` pack — standard/full only). Step 2 is rewired to **`mem_do` with `action="related"`** — `mem_do` is a core-9 tool and the documented gateway to non-core actions, so the skill now works in the default mode.
- **`setup/SKILL.md`** listed `mem_config` (full-only — an `_ASTERISK_TOOLS` member per `test_docs_guards.py`) that the skill body **never invokes**, and omitted **`mem_search`** which step 3 *does* use. Dropped the unused non-core tool, added the used core one.

## B6-5 — the 409 conflict-resolution flow had no test

`_ctxResolveConflict` (`context-gateway.js`, #763) opens the three-choice dialog when `PUT /context/{type}/{name}` returns 409 (on-disk file changed under unsaved edits). New `tests-js/ctx-conflict-resolution.test.mjs` drives it through the existing `window.openCtxConflictModal()` dev entry point and pins:

- all four outcomes — Force-save → `'force'`, Reload → `'reload'`, Open-diff → `'diff'`, Escape / backdrop → `null`;
- the a11y safety contract: the modal focuses the non-destructive **Reload** button, never **Force**, so a reflexive Enter can't silently overwrite the other writer's edits.

## Verification

- New test: 6 pass. Full `vitest run`: **72 passed (13 files)**.
- `test_docs_guards.py`: 12 passed (unaffected — it validates hooks.json + the Config-tools doc, not SKILL.md `allowed-tools`).
- No Python source touched; docs-guard + plugin tests green.

### DIFF
--- a/packages/memtomem-claude-plugin/skills/context/SKILL.md
+++ b/packages/memtomem-claude-plugin/skills/context/SKILL.md
@@ -2,7 +2,7 @@
 name: context
 description: Inject relevant memories as structured context for the current task. Use when starting complex work that may benefit from past decisions or notes.
 argument-hint: [topic or question]
-allowed-tools: mcp__memtomem__mem_search, mcp__memtomem__mem_related
+allowed-tools: mcp__memtomem__mem_search, mcp__memtomem__mem_do
 ---
 
 Search for relevant memories about: $ARGUMENTS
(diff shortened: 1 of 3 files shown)


tsdata (approved):
LGTM.

PR #1145, opened 2026-05-30
fix(context): close settings TOCTOU, fix skill promote rollback leak, collapse override double-write, document gateway lock layers (#1123 B3)

memtomem:
Closes the **B3 "Concurrency & atomicity"** bucket of the context-gateway audit (#1123). Branched off `main`; disjoint from the already-merged B1/B2/B5/B6/B7 PRs.

The five B3 findings resolve to **three real per-file fixes** plus **two clarifications** (a code comment + a docstring). A gateway-wide cross-process lock was considered and **deliberately rejected** — analysis below.

## Fixes

**B3-3 · settings read-merge-write TOCTOU** (`context/settings.py`)
`generate_all_settings` read the target, merged, re-stat'd the mtime, then wrote — leaving an unguarded window between the mtime recheck and the atomic rename where a separate-process `mm context sync` / MCP writer could land and have its hook rule silently dropped. Now holds a per-target `_file_lock(_lock_path_for(target_path))` across read→merge→recheck→write (the same sidecar `portalocker` primitive skills already uses). `_write_json`→`atomic_write_text` is lock-free, so calling it under the held lock does not self-deadlock. The `st_mtime_ns` recheck is **kept** as a second layer that still catches a non-gateway direct disk edit. One lock per iteration (distinct target per generator) → no cross-generator ordering cycle.

**B3-5 · override double-write** (`context/_sync_atomic.py`)
Phase 2 wrote the rendered `content` then immediately overwrote the same file with the per-vendor `override_bytes` — wasted I/O plus a crash window where the runtime file briefly held the non-override bytes. Now computes the final payload once (`override_bytes` if present, else `content.encode("utf-8")`) and writes it with a single atomic `os.replace`. **Byte- and mode-identical** to the old no-override path (`atomic_write_text` defaulted to utf-8 + `0o600`). `render()` still runs for its dropped-fields side effects.

**B3-4 · skill promote rollback leak** (`context/skills.py`)
`_promote_staging`'s rollback did a bare `os.replace(old, dst)`; if that rename itself failed it masked the original promote error and left `old` — the only surviving copy of the original tree — orphaned with no trace. Now wraps the rollback: on failure it logs a breadcrumb naming `old`/`dst` so an operator can recover, then re-raises the original error with the rollback failure chained (`raise promote_exc from rollback_exc`).

## Clarifications (no behavior change)

**B3-1 · agents/commands take no cross-process lock** — documented as a code comment, intentionally **not** changed. After B3-5 each `out_path` is one idempotent atomic write; project_shared all-or-nothing is enforced in Phase 1 and user/project_local partial writes are an intentional contract (#908), so there is no cross-file invariant a lock would protect. A lock would be deadlock-free but pointless — the comment records this so a future reader doesn't "fix" it.

**B3-2 · `_gateway_lock` docstring** — rewrote the over-claiming cross-process note into an accurate **two-layer model**: layer 1 is the in-process `asyncio.Lock` (serializes same-process async handlers only); layer 2 is per-engine and asymmetric (skills hold a file lock across their swap; settings now hold a per-target file lock via B3-3; agents/commands rely on single atomic `os.replace`).

### Why not a gateway-wide cross-process lock
A single `.memtomem/.gateway.lock` was rejected because: (a) there is **no all-or-nothing snapshot invariant** across settings+agents+commands+skills to protect — partial cross-file fan-out is pinned as intentional (#908); (b) the web handlers run the engine **synchronously inside** `async with _gateway_lock`, so a blocking `portalocker` lock there would stall the event loop; (c) it would **nest** with the per-target `_file_lock`s into a non-reentrant `flock`/`LockFileEx` self-deadlock (empirically reproduced during design).

## Tests
- `test_context_settings.py` — lock-wraps-write ordering pin + 8-thread no-torn-file contention test
- `test_context_skills_staging.py` — rollback-failure breadcrumb + exception-chaining pin
- `test_sync_atomic_engine.py` — override single-write + no-override byte-equivalence pins
- `test_web_routes_context_locks.py` — two-layer docstring pin

Each new test fails on the pre-fix code (verified as genuine regression pins, not vacuous).

## Gates
- `ruff check` ✅ · `ruff format --check` ✅ · `mypy` clean ✅
- full suite (CI filter): **5462 passed, 10 skipped**

### DIFF
--- a/packages/memtomem/src/memtomem/context/_atomic.py
+++ b/packages/memtomem/src/memtomem/context/_atomic.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 import logging
 import os
 import tempfile
+import time
 from contextlib import contextmanager
 from datetime import datetime, timezone
 from pathlib import Path
@@ -74,7 +75,7 @@ update until the user manually deletes the ``.bak``.
 
 
 @contextmanager
-def _file_lock(lock_path: Path) -> Iterator[None]:
+def _file_lock(lock_path: Path, *, timeout: float | None = None) -> Iterator[None]:
     """Cross-process exclusive lock on a sidecar lockfile.
 
     Locking the data file directly does **not** survive ``os.replace`` —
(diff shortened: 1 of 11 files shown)


tsdata (approved):
LGTM

PR #1148, opened 2026-05-31
fix(init): don't gate HTTP embedding providers on unused client packages

memtomem:
## Problem

`mm init -y --provider ollama` (or `--provider openai`) **aborts in a base install**, reporting the `ollama`/`openai` PyPI client as a required missing extra — even for `--provider openai --api-key …`. This breaks valid scripted setup flows.

But the Ollama/OpenAI **embedding providers use `httpx` directly** (a core dependency) and never import the `ollama`/`openai` Python packages:
- `embedding/ollama.py` / `embedding/openai.py` → `import httpx` only.
- `embedding/factory.py` imports those httpx-based embedders directly.
- Zero `ollama`/`openai` imports anywhere under `packages/memtomem/src`.

So the gate is a **false failure path** rejecting configs that would work fine.

## Why removing it is safe

The package was never the right signal. `mm init -y` already does **not** prove the Ollama server is running or that an OpenAI key is set (the *real* preconditions), so removing the client-SDK gate does **not** reintroduce the runtime failure #402 guarded against. The real preconditions remain checked: Ollama via the binary/server probe, OpenAI via the API key.

The gating was applied consistently across the non-interactive path, the interactive wizard, **and** a source-scan test — so a partial (non-interactive-only) fix would leave the interactive wizard printing a now-false *"-y will refuse"* hint. This PR fixes all of them.

## Change

- `_collect_missing_extras`: drop the ollama/openai PyPI-client probes + appends.
- `_REQUIRED_EXTRAS_FOR_NON_INTERACTIVE`: `{onnx, ollama, openai, korean}` → `{onnx, korean}`.
- interactive `_step_embedding`: drop the `"<x> Python client not installed"` warning + `_y_refuse_hint` for ollama/openai (Ollama still checks the binary/server; OpenAI still collects an API key).
- tests: invert the four ollama/openai abort / missing-extra tests to assert the config **is** written; update the source-scan refuse-hint pin (onnx/korean keep it, ollama/openai must not); correct the test-class docstring premise.

`onnx` (fastembed) and `korean` (kiwipiepy) gating is **unchanged** — those packages ARE imported at runtime.

## Behavior change

`mm init -y --provider ollama|openai` in a base install now **writes the config** (exit 0) instead of aborting; embedding runs once the Ollama server is up / with the OpenAI key. The interactive wizard no longer claims `-y` will refuse for these providers.

## Not in scope

The `[ollama]`/`[openai]` `pyproject` extras (which install the unused PyPI clients) are left as-is — removing them is a separate, user-facing API decision (`memtomem[all]` references them).

## Tests

Full suite green: `uv run pytest -m "not ollama"` → 5499 passed, 10 skipped, 46 deselected. Lint + format + mypy clean.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/init_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/init_cmd.py
@@ -478,37 +478,21 @@ def _step_embedding(state: dict) -> None:
 
     elif choice == 3:
         provider = "ollama"
-        # Two orthogonal preconditions for Ollama: the runtime binary
-        # (``_ollama_available()`` = ``shutil.which("ollama")``) and the
-        # ``ollama`` Python client (``_have_module("ollama")``). The
-        # ``-y`` extras gate (#402) checks the Python client, not the
-        # binary — so the refuse-hint must fire whenever the Python
-        # client is missing, whether or not the binary is on PATH.
-        have_ollama_binary = _ollama_available()
-        have_ollama_module = _have_module("ollama")
-
-        if not have_ollama_binary:
+        # The only Ollama precondition that matters is the runtime SERVER (the
+        # ``ollama`` binary, ``_ollama_available()`` / ``_ollama_running()``):
+        # the embedder talks to it over HTTP via ``httpx`` (a core dependency).
+        # The ``ollama`` PyPI client is NOT imported by the embedder, so it is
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

Issue #1147, opened 2026-05-31
Context Gateway: deferred design-decision items from the #1123 audit (B1-3/B1-4/B5-1/B5-2/B7-1/B7-2/B7-3)


memtomem:
Seven items from the full-stack context-gateway audit (#1123) were **intentionally deferred** during the per-bucket cleanup PRs because each is a **design decision / behavior-contract change**, not a mechanical fix. Splitting them out so #1123's cleanup buckets (B1–B7) can close. Verified still-open against `main` 2026-05-31.

Legend: 🐛 bug · 🔁 parity · 📝 doc · 🎨 ux · 🕳 gap. Line numbers are point-in-time — re-verify before editing.

### Parser / generator (deferred from Bucket 1, #1132)
- [ ] 🐛 **B1-3 (preamble half)** `context/parser.py` — text *before the first `##` heading* is silently dropped. The whitespace-only-heading half was fixed in #1132; capturing leading preamble was deferred because a naive capture swallows the `# Project Context` H1 and breaks the dict→markdown round-trip contract. Needs a decision on how preamble maps into the section model.
- [ ] 🐛 **B1-4** `context/generator.py` — Rules+Style merge is irreversible on reverse-import (Style folded into Rules) for cursor/codex/copilot targets. #857 added a CLI warning (`_warn_rules_style_merge`) but the round-trip is still lossy. A real fix needs to preserve Rules vs Style separately in the generated `.cursorrules`/`AGENTS.md`/`copilot-instructions.md`, or an explicit accepted-lossy decision.

### MCP parity surface (deferred from Bucket 5, #1137)
- [ ] 🔁 **B5-1** `server/tools/context.py` — artifact scope-tier + flat→dir migration are CLI-only; there is no MCP tool. Deferred pending a maintainer decision on whether to build a new MCP migration surface.
- [ ] 🎨 **B5-2** `server/tools/context.py` — the `mem_context_migrate` MCP tool name implies it mirrors CLI `mm context migrate` but actually only covers `memory-migrate` (discoverability hazard). Entangled with B5-1 — resolve together (rename and/or add the missing surface).

### Behavior-contract / misc (deferred from Bucket 7, #1134)
[...]

PR #1149, opened 2026-05-31
feat(context): rename mem_context_migrate → mem_context_memory_migrate + deprecated alias (#1147 B5-2)

memtomem:
## What

Renames the MCP tool `mem_context_migrate` → `mem_context_memory_migrate` and
keeps the old name as a thin **deprecated alias**. Closes the **B5-2** item of
#1147.

## Why

`mem_context_migrate` only ever covered **memory**-tier migration (it mirrors
the CLI `mm context memory-migrate`), but its bare name implied parity with the
full CLI `mm context migrate`, which also performs artifact flat→dir and
scope-tier moves. That discoverability hazard (called out in #1147 B5-2) blocks
adding the artifact-migration MCP surface (**B5-1**) under a coherent name —
so this lands first.

## How

- Rename the impl to `mem_context_memory_migrate` (`mem_do` action
  `context_memory_migrate`).
- Keep `mem_context_migrate` as a deprecated alias tool that forwards every
  argument unchanged. The alias **repeats the explicit parameter signature**
  (not `**kwargs`) because the MCP schema is built by inspecting the function
  signature — a `**kwargs` alias would publish an empty parameter schema.
- The old `mem_do` action name keeps working via
  `_ALIASES["context_migrate"] = "context_memory_migrate"` (validated by the
  existing `test_aliases_resolve_to_real_actions` guard).

## Tests

- Existing wrapper parity pins (`test_server_tools_context_migrate.py`) now
  target the canonical name.
- New `test_server_tools_context_migrate_alias.py` pins the deprecation
  contract: the alias delegates, its docstring opens with `DEPRECATED`, the
  renamed action is registered while the old action name is not, and
  `mem_do(action="context_migrate")` still routes.
- Docs (`mcp-clients.md`) + `CHANGELOG.md` updated; stale wrapper reference in
  `cli/context_cmd.py` corrected.

`uv run pytest -m "not ollama"` (context + server-tool + registry suites,
1280 tests) green; `ruff check` / `ruff format --check` clean; `mypy` clean on
changed files.

Part of the #1147 B5 track: **PR-1 (B5-2, this) → PR-2 (B5-1: add
`mem_context_artifact_migrate`)**.


pandas-studio (approved):
LGTM

PR #1150, opened 2026-05-31
fix(context): capture runtime-file preamble on reverse-import (#1147 B1-3)

memtomem:
## What

Stops `mm context init` / `mem_context_init` from silently dropping the
**preamble** (text before the first `##` heading) when reverse-importing a
runtime agent file. Closes the **B1-3** item of #1147.

## Why

For the **cursor** and **copilot** generators the leading text *is* the project
description — they emit the `Project` body directly with no H1. So a user's
hand-authored `.cursorrules` / `.github/copilot-instructions.md` prose was lost
on import.

The round-trippable fix lives on the **reverse-import** side, not the canonical
parser: capturing preamble in `parse_context` would break the dict→markdown
round-trip contract (the `# Project Context` wrapper) and silently mutate
hand-edited `context.md`. (Canonical-parser preamble capture is consciously
deferred — noted on #1147.)

## How

- Add `parser.split_preamble(text) → (preamble, rest)`, sharing the fence /
  heading state machine with `iter_markdown_sections` so a `##` inside a code
  fence (B1-1) or a whitespace-only `##` (B1-3 #1132) is not a false boundary.
- Make `extract_sections_from_agent_file(content, source=...)` source-aware:
  for a known source it strips that generator's boilerplate — the
  `# CLAUDE.md` / `# GEMINI.md` / `# AGENTS.md` H1s and intro sentences plus the
  `# Project Context` wrapper — via a data-driven table kept in lock-step with
  the generators, then maps the remaining prose onto canonical `Project`,
  **prepended** to any existing `## Project` body to preserve document order.
  The CLI and MCP callers pass `source=best.agent` / `f.agent`.
- `source=None` keeps the **exact prior behavior** (preamble dropped) — the
  canonical parser contract is unchanged and existing callers are unaffected.

## Tests

`split_preamble` fence/whitespace awareness; per-source
generate→import→generate **idempotence** with no boilerplate leakage (claude /
gemini / codex / cursor / copilot); the cursor user-prose capture (the bug);
the collision **prepend** rule; wrapper-H1 strip; and the `source=None`
back-compat drop.

`uv run pytest -m "not ollama"` (context suites, 1235 tests) green;
`ruff check` / `ruff format --check` clean; `mypy` clean on changed modules.

Independent of the other #1147 items.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -811,12 +811,12 @@ def init_cmd(
             best = max(files, key=lambda f: f.size)
             click.echo(f"Extracting from {best.agent}: {best.path.name} ({best.size} bytes)")
             content = _read_agent_file(best.path)
-            sections = extract_sections_from_agent_file(content)
+            sections = extract_sections_from_agent_file(content, source=best.agent)
             for f in files:
                 if f.path == best.path:
                     continue
                 other_content = _read_agent_file(f.path)
-                other_sections = extract_sections_from_agent_file(other_content)
+                other_sections = extract_sections_from_agent_file(other_content, source=f.agent)
                 for key, val in other_sections.items():
                     if key not in sections and val.strip():
                         sections[key] = val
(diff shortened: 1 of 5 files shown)


pandas-studio (approved):
LGTM

PR #1151, opened 2026-05-31
fix(context): FS-guided reconstruction of kebab-case ~/.claude/projects names (#1147 B7-1)

memtomem:
## What

Replaces the blind `-`→`/` decoder for the experimental `~/.claude/projects/`
scan with an **FS-guided reconstruction** so kebab-case paths resolve. Closes
the **B7-1** item of #1147.

## Why

Claude Code's on-disk project-dir encoding maps `/`, `.` **and** literal `-` all
to `-`. The old decoder blind-replaced every `-` with `/`, so any path with a
literal-dash component (`agent-harness`) or a hidden dir (`.claude-worktrees`)
mis-decoded and never resolved — the project silently never surfaced. (Verified
against a live `~/.claude/projects/`: this repo's own
`-Users-…-Work-agent-harness-memtomem` is an instance; `…-prep--claude-worktrees-…`
shows the `/.` → `--` double-dash.)

## How

- Each encoded `-` becomes a **3-way** choice — path separator, `.`, or a
  literal dash — and only branches whose committed directory prefix actually
  exists survive, so the filesystem prunes the otherwise `3**k` partition space
  to the paths that really exist (memoized `scandir` prefix-prune + a frontier
  cap as a runaway backstop). `_decode_claude_project_dirname` now returns a
  candidate **list**.
- `known_projects.json` roots and the server **cwd** are passed as
  authoritative decode **anchors**: any whose encoding equals the dir name are
  returned directly — but **all** matches (the encoding is many-to-one, so
  distinct roots can collide), leaving the accept-one decision to the caller.
- The caller accepts exactly one candidate; **zero → skip + warn** (point at
  `known_projects.json`), **>1 → skip + warn ambiguous** (never guess). The
  `is_dir()` fail-closed guarantee and the **off-by-default** experimental gate
  are preserved.

## Tests

POSIX-only and hermetic (monkeypatched import-bound `_CLAUDE_PROJECTS_DIR` +
real fixture dirs under `tmp_path` so the absolute `is_dir()` probes hit
fixtures): kebab-case single-match resolves; a dotted hidden-dir component
(`.config-dir`) reconstructs; genuinely-ambiguous twin dirs (`a-b/c` vs `a/b-c`)
skip + warn; no-match skips + warns; a known-root **anchor** resolves an
otherwise-ambiguous name. Existing misdecoded-filter / dedup / experimental-off
behaviors preserved.

`uv run pytest -m "not ollama"` (project suites, 348 tests) green;
`ruff check` / `ruff format --check` clean; `mypy` clean. Feature stays
experimental / off by default — no config flip.

Independent of the other #1147 items.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/sync_doctor_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/sync_doctor_cmd.py
@@ -233,11 +233,11 @@ def check_claude_slug(cwd: Path, *, home: Path | None = None) -> CheckResult:
        previous one-shot ``str(cwd).replace("/", "-")`` only handled POSIX
        and produced an invalid slug on Windows (drive prefix + backslashes
        remained), causing false failures for valid Windows layouts.
-    2. **Slow path** — iterate real entries and decode-then-``samefile`` each
-       against ``cwd``. Catches cases where the encoded form isn't
-       predictable (e.g., the entry on disk uses an encoding variant we
-       didn't enumerate) but skips entries whose name has literal dashes
-       and so doesn't round-trip to a real directory.
+    2. **Slow path** — iterate real entries and FS-guided-decode-then-
+       ``samefile`` each against ``cwd`` via
+       :func:`memtomem.context.projects._decode_claude_project_dirname`, which
+       reconstructs ``/``, ``.`` and literal-``-`` segments (so kebab-case and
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #1156, opened 2026-05-31
fix(context): mirror Claude Code's real ~/.claude/projects slug rule (#1155)

### LINKED ISSUE #1155: context: _encode_claude_project_path is POSIX-only → dotted-slug test red on windows (#1151)
## Summary

`_encode_claude_project_path` in `context/projects.py` is **POSIX-only**, so on
Windows it produces the wrong `~/.claude/projects/<slug>` directory name. This
surfaced as a **Windows-only test failure** that landed on `main` via #1151
(B7-1): `test (windows-latest)` is not a required check, so the squash merged
while that job was red.

## The failing test

```
FAILED packages/memtomem/tests/test_sync_doctor_cli.py::TestCheckClaudeSlug::test_dotted_cwd_slug_matches
  AssertionError: assert 'fail' == 'pass'
```

ubuntu + macos pass; only `windows-latest` fails. (Run on the #1151 CI matrix.)

## Root cause

```python
def _encode_claude_project_path(root: Path) -> str:
    # Mirrors Claude Code's dirname scheme: every os.sep and '.' becomes '-'.
    return str(root).replace("/", "-").replace(".", "-")
```

It only replaces `/` and `.`. On Windows the path separator is `\` and absolute
paths carry a drive `C:`, so `str(root)` is e.g. `C:\Users\foo\.config-dir`,
which this encodes to `C:\Users\foo--config-dir` — `\` and `:` are left intact
instead of being normalized. The doctor's `check_claude_slug` then can't match
the on-disk dir, so the round-trip check returns `fail`.

The test's own helper `_claude_slug_for` already handles this
(`.replace("\\", "-").replace(":", "")`), and `test_dotted_cwd_slug_matches`
calls the **production** encoder directly via
`proj_mod._encode_claude_project_path(cwd.resolve())` — so the test correctly
exercises the real code path and correctly fails on Windows. The bug is in the
production encoder, not the test.

## Why not fixed inline

The correct Windows encoding depends on **how Claude Code actually names
`~/.claude/projects/` directories on Windows** (drive-letter handling, `\` vs
`/`, colon treatment). That upstream behavior should be confirmed before
changing the encoder — guessing risks encoding to a slug Claude Code never
produces, which would make the decoder/discovery silently miss real project
dirs on Windows. Filing instead of shipping a guessed fix.

## Severity

Low. The affected surface (`mm sync-doctor`'s `check_claude_slug` + the
`~/.claude/projects` discovery in `discover_project_scopes`) is a diagnostic /
opt-in discovery that fails closed (`is_dir()` gated). No data loss; Windows
users just don't get the kebab-path reconstruction. But `main` currently has a
**red windows test**, so this should be fixed (or the test skipped on Windows
with an xfail referencing this issue) before the next release.

## Proposed fix (needs the upstream check first)

Make `_encode_claude_project_path` separator- and drive-agnostic — mirror
`_claude_slug_for`'s `os.sep`/`os.altsep` + drive-colon handling — once the
Windows slug convention is confirmed. Add the corresponding decode/anchor
handling in `_decode_claude_project_dirname` if needed. Then `test (windows-latest)`
goes green.

Refs #1151, #1147.

memtomem:
## Summary

`_encode_claude_project_path` (the helper that mirrors how Claude Code names
`~/.claude/projects/<slug>` directories) was **POSIX-only**:

```python
return str(root).replace("/", "-").replace(".", "-")
```

On Windows the `\` separators and the drive `:` weren't normalized, so the
dotted-slug doctor test went **red on `test (windows-latest)`** (it landed on
`main` via #1151 because the windows job isn't a required check). It was also
wrong on **POSIX**: it left `_`, spaces and non-ASCII characters intact.

## Root cause — what Claude Code actually does

Claude Code replaces **every character outside ASCII `[A-Za-z0-9]`** with a
single `-` — equivalent to `re.sub(r"[^A-Za-z0-9]", "-", abspath)`. This is
documented (with a reproduced example + reverse-engineered impl) in
[anthropics/claude-code#19972](https://github.com/anthropics/claude-code/issues/19972),
and confirmed empirically: on a live macOS install the temp-dir hash
`…_b0w_…` is stored under `~/.claude/projects/` as `…-b0w-…`, i.e. `_`→`-`.
The drive colon is **not** special-cased, so `C:\dev\repo` → `C--dev-repo`;
non-ASCII (Korean/CJK/accented) chars each become `-`.

## Changes

- **Encoder** → `re.sub(r"[^A-Za-z0-9]", "-", str(root))`; param widened
  `Path` → `PurePath` (only `str(root)` is used) so a `PureWindowsPath` fixture
  can exercise the `C--` drive form on a POSIX CI host.
- **`check_claude_slug` fast path** → collapse the 4-entry candidate hand-list
  (3 entries kept `_`/`.` or used a wrong colon-dropped `C-`) to the single
  correct encoder output; drop the now-dead separator guard.
- **Decoder (`_decode_claude_project_dirname`)** → the encoder change makes it
  build realistic CC names containing `_`/non-ASCII that the hardcoded
  `(".", "-")` reconstruction couldn't reverse. Generalize the non-separator
  branch to read the **actual** next character from each existing child
  (FS-guided, bounded by child count), so a `-` reverses to whatever the real
  path holds (`.`, `_`, literal `-`, space, non-ASCII, …). Also try the
  **anchors before** the leading-`-` gate so a registered Windows `C--…` root
  resolves through the (platform-agnostic) anchor fast-path — it previously
  returned `[]` before anchors were considered. The slow-path FS walk stays
  POSIX-root-only (a `C--…` slug has no leading `-`).

## Tests

- `_claude_slug_for` now delegates to the production encoder (its old form kept
  `_`/`.` and dropped the drive colon → stale fixtures).
- New literal-assertion encoder tests (incl. `PureWindowsPath(r"C:\Users\foo")`
  → `C--Users-foo`, plus underscore/dot and Hangul cases) — expected values are
  literals, never recomputed via `re.sub`.
- New decoder tests: `_`/Hangul FS-walk round-trip, and a Windows-slug-via-anchor
  regression (cross-platform, runs on the windows job too).
- Mutation-checked: reverting the encoder to the old body fails the new tests.

`ruff check` + `ruff format` clean; `mypy` clean (249 files); full suite
**5565 passed / 10 skipped**.

## Scope / follow-up

Encoder fix necessarily came with the decoder generalization (the encoder
change made the existing FS-walk tests build realistic CC names the old decoder
couldn't reverse). The `_`/non-ASCII gap is now fully closed. The only remaining
limitation: **unregistered** Windows dirs in the experimental, off-by-default
`~/.claude/projects` scan still can't be FS-walk-reconstructed (the `C--…` slug
has no leading `-`); registered roots resolve via the anchor path. Tracked as a
follow-up.

Closes #1155. Refs #1151, #1147, anthropics/claude-code#19972.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/sync_doctor_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/sync_doctor_cmd.py
@@ -228,16 +228,18 @@ def check_claude_slug(cwd: Path, *, home: Path | None = None) -> CheckResult:
     Skipped silently when ``~/.claude/projects/`` is absent (user doesn't run
     Claude Code). Two-tier match:
 
-    1. **Fast path** — encode ``cwd`` (POSIX ``/`` → ``-``, Windows ``\\`` /
-       drive ``:`` → ``-``) and look up the resulting directory directly. The
-       previous one-shot ``str(cwd).replace("/", "-")`` only handled POSIX
-       and produced an invalid slug on Windows (drive prefix + backslashes
-       remained), causing false failures for valid Windows layouts.
+    1. **Fast path** — encode ``cwd`` to Claude Code's slug via
+       :func:`memtomem.context.projects._encode_claude_project_path` (every
+       character outside ASCII ``[A-Za-z0-9]`` → ``-``, so POSIX ``/``/``.``/``_``
+       and Windows ``\\`` / drive ``:`` all collapse — ``C:\\dev\\repo`` →
+       ``C--dev-repo``) and look up that directory directly. This single encoder
+       is the source of truth, so the slug matches on every platform.
     2. **Slow path** — iterate real entries and FS-guided-decode-then-
        ``samefile`` each against ``cwd`` via
        :func:`memtomem.context.projects._decode_claude_project_dirname`, which
-       reconstructs ``/``, ``.`` and literal-``-`` segments (so kebab-case and
-       dotted directories round-trip, unlike the old blind ``-`` → ``/``).
+       reverses each ``-`` to the real character on disk (``/``, ``.``, ``_``,
+       literal ``-``, space, non-ASCII, …) so kebab-case, dotted and
+       underscored directories round-trip, unlike the old blind ``-`` → ``/``.
     """
     projects = (home or Path.home()) / ".claude" / "projects"
     if not projects.is_dir():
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #1159, opened 2026-05-31
fix(context): reconstruct unregistered Windows `C--…` ~/.claude/projects dirs (#1157)

### LINKED ISSUE #1157: context: decoder FS-walk can't reconstruct unregistered Windows `C--…` ~/.claude/projects dirs
## Summary

After #1156 the `~/.claude/projects/` slug **encoder** is correct on every
platform, and `_decode_claude_project_dirname`'s **anchor fast-path** resolves
Windows `C--…` slugs (it re-encodes the anchors — cwd + `known_projects.json`
roots — and string-matches). The remaining gap is the **slow-path FS walk**:
it's POSIX-root-only, so it can't reconstruct an **unregistered** Windows dir
in the experimental `~/.claude/projects` scan.

## Where

`packages/memtomem/src/memtomem/context/projects.py` —
`_decode_claude_project_dirname`:

```python
# anchors tried first (platform-agnostic) … then:
if not name.startswith("-"):
    return []
body = name[1:]                     # leading "-" is the POSIX "/" root
...
frontier: list[tuple[Path, str]] = [(Path("/"), "")]
```

A Windows slug like `C--Users-foo` has no leading `-` (the drive letter `C`
leads, `:`/`\` both became `-`), so the FS walk bails with `[]`. There is no
"start at the drive root" branch.

## Impact

- **Low severity.** The `~/.claude/projects` scan is **experimental and off by
  default** (`experimental_claude_projects_scan`), and the path **fails closed**
  (`[]` → the dir is simply not surfaced; no wrong result, no crash).
- **Registered roots are unaffected** — cwd and `known_projects.json` entries
  resolve via the anchor fast-path on Windows. Only the discovery of
  *unregistered* `~/.claude/projects` dirs on Windows is missing.
- `mm sync-doctor`'s `check_claude_slug` is **unaffected**: it matches the cwd
  via the encoder fast-path (the slow-path decode is only a fallback).
- The FS-walk decoder tests are `@_WIN_SKIP` (POSIX-only), so this isn't covered
  on the windows job today.

## Possible fix (sketch — not prescriptive)

Detect a Windows drive-rooted slug (CC encodes `C:\` → `C--`, i.e. a leading
`[A-Za-z]--`) and seed the FS walk at the drive root (`PureWindowsPath` /
`C:\\`) consuming the drive prefix, instead of bailing on the missing leading
`-`. Then drop `@_WIN_SKIP` from the reconstruction tests (or add Windows
variants). Mind the lossiness: `C--…` could also be a POSIX path component, so
keep it FS-guided / fail-closed.

## Workaround

Register the project in `known_projects.json` (or run from it as cwd) — the
anchor fast-path resolves it regardless of platform.

Refs #1156, #1147.

memtomem:
## Summary

Closes #1157. The decoder's slow-path FS walk (`_decode_claude_project_dirname`)
reconstructs an absolute path from a `~/.claude/projects/<slug>` directory name.
It was POSIX-root-only: it seeded the frontier at `Path("/")` after stripping a
leading `-`, so a Windows drive-rooted slug like `C--Users-foo`
(`C:\Users\foo` — the drive `C` leads; `:` and `\` both became `-`, so **no**
leading `-`) bailed with `[]`. Registered roots already resolved via the
platform-agnostic anchor fast-path, but the experimental `~/.claude/projects`
scan could not discover an **unregistered** Windows dir.

## What changed

**`context/projects.py`**

- New `_decode_seed(name)` maps an absolute slug to its `(root, body)` walk seed
  for both encodings:
  - POSIX: leading `-` → `/` root (unchanged).
  - Windows: `<letter>--…` ← `<letter>:\…` → seed at the drive root, consume the
    drive prefix from the body.
- The drive branch is gated on `seed_root.is_absolute()`. On Windows
  `Path("C:\\")` is a real absolute `WindowsPath` (scandir / join / `is_dir` hit
  the actual drive); on POSIX it is a **relative** `PosixPath` (a one-component
  name), so the slug **fails closed by construction** instead of being walked
  relative to cwd. Claude Code never emits a drive-rooted slug for a POSIX path
  anyway.
- Hardened `children()`: a single unreadable `DirEntry` now skips only itself
  instead of the outer `except OSError` dropping the whole directory listing.
  The drive-root walk necessarily scans busy system dirs (`C:\`, `C:\Users` with
  its `Default User` / `All Users` junctions) where one entry's `is_dir()` can
  raise while the real project ancestors are fine.

## Tests

- `test_decode_seed_recognizes_posix_and_windows_roots` — string-only, runs on
  every host. POSIX proves a drive slug fails closed (`None`, and end-to-end
  `[]`); Windows asserts the drive seed.
- `test_decode_windows_drive_root_walk` (Windows-only) — exercises the real
  `C:\`-rooted walk by handing the slug straight to the decoder (no
  `~/.claude/projects/<long-name>` fixture, so no `MAX_PATH` risk).
  Canonicalizes the target with `resolve()` first so the slug spelling matches
  what `os.scandir` returns (GitHub's hosted Windows runners can expose the temp
  dir via an 8.3 short alias like `C:\Users\RUNNER~1\...`).
- The 9 deep-`tmp_path` reconstruction cases stay `@_WIN_SKIP`: their fixture
  creates a directory whose **name** is the full encoded absolute path, which
  risks Windows `MAX_PATH` (260) — a fixture limit, not a decoder limit. The
  skip rationales (and the two `/tmp` discovery skips) are corrected so they no
  longer claim the encoding itself is POSIX-only.

## Severity / scope

Low. The `~/.claude/projects` scan is experimental and off by default
(`experimental_claude_projects_scan`), and the path **fails closed** on both
platforms (`[]` → the dir is simply not surfaced; no wrong result, no crash).
Registered roots are unaffected (anchor fast-path). Refs #1156, #1147.

## Verification

- `uv run pytest packages/memtomem/tests/test_context_projects.py` — 36 passed,
  1 skipped (Windows-only); `test_sync_doctor_cli.py` — 39 passed.
- `ruff check` + `ruff format --check` clean; `mypy` clean; no SyntaxWarnings.
- Windows-specific paths can only be exercised on the `test (windows-latest)` CI
  job — that job is the real gate for this PR.

### DIFF
--- a/packages/memtomem/src/memtomem/context/projects.py
+++ b/packages/memtomem/src/memtomem/context/projects.py
@@ -263,6 +263,49 @@ def _encode_claude_project_path(root: PurePath) -> str:
     return re.sub(r"[^A-Za-z0-9]", "-", str(root))
 
 
+# A Windows drive-rooted slug: the drive letter survives (ASCII-alnum) and the
+# ``:`` + ``\`` of ``<letter>:\`` both collapse to ``-``, so ``C:\dev`` encodes to
+# ``C--dev``. Anchored at the start; a POSIX absolute slug always leads with ``-``
+# (the ``/`` root) so it never matches here. UNC slugs (``\\host\share`` →
+# ``--host-share``) DO lead with ``-`` and fall through to the POSIX branch — they
+# are out of scope for the drive-root walk and simply fail closed there.
+_WIN_DRIVE_SLUG_RE = re.compile(r"([A-Za-z])--")
+
+
+def _decode_seed(name: str) -> tuple[Path, str] | None:
+    """Return the ``(root, body)`` seed for the FS-walk, or ``None`` if *name* is
+    not a walkable absolute slug.
+
+    Two absolute-root encodings are recognized:
+
+    * **POSIX** — a leading ``-`` is the ``/`` root (``-home-foo`` ← ``/home/foo``).
+    * **Windows** — a drive-rooted slug ``<letter>--…`` ← ``<letter>:\\…`` (the
+      drive ``:`` and the ``\\`` separator both collapsed to ``-``). The walk is
+      seeded at the drive root and the drive prefix is consumed from the body.
+
+    ``Path(...)`` is the native flavor, so on a Windows host the drive seed is a
+    ``WindowsPath`` whose ``os.scandir`` / ``/`` join / ``is_dir`` all hit the real
+    drive. On a non-Windows host ``Path("C:\\")`` is a *relative* ``PosixPath`` (a
+    one-component name, ``is_absolute()`` is False), not a drive root — so the drive
+    branch is gated on ``is_absolute()`` and returns ``None`` there, failing closed
+    by construction rather than depending on the cwd not happening to contain a
+    literal ``C:\\`` child. This is harmless either way: Claude Code never emits a
+    drive-rooted slug for a POSIX path (and vice versa).
+    """
+    if name.startswith("-"):
+        return Path("/"), name[1:]  # the leading "-" is the root "/"
+    m = _WIN_DRIVE_SLUG_RE.match(name)
+    if m:
+        # "C--rest" ← "C:\rest": seed at the drive root, walk "rest". Only honor
+        # it when the seed is a genuine absolute root on THIS host — a real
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #1162, opened 2026-05-31
fix(web): bind gateway/config locks per event loop to de-flake tests

memtomem:
### Symptom
`test_web_routes_context_locks.py::test_settings_sync_and_skills_sync_share_lock`
failed on **ubuntu-latest** in v0.2.2's post-merge `main` CI with:

```
RuntimeError: <asyncio.locks.Lock object ...> is bound to a different event loop
```

The release commit (#1161) changed no source — only `pyproject`/`uv.lock`/
`CHANGELOG` — and the same test passed on macos, on the PR run, and on a rerun.
Order-dependent flake, not a regression.

### Root cause
`_gateway_lock` / `_config_lock` were module-level bare `asyncio.Lock()`s. A
bare `Lock` binds to an event loop the first time it takes its **contended**
slow path (the uncontended fast path never calls `_get_loop()`), then raises
"bound to a different event loop" when contended again from another loop.
Production runs one long-lived loop, but pytest gives each `async` test its own;
the flaky test contends the lock via `asyncio.gather`, so whichever test bound
it first could strand a later test on a fresh loop.

### Fix
Wrap both locks in `_LoopLocalLock`, a drop-in proxy keying the underlying
`asyncio.Lock` per running loop:

- **Semantics preserved** — within one loop all callers share one lock
  (serialization intact); distinct loops get distinct locks.
- **Call sites unchanged** — `async with _gateway_lock:` and `.locked()` still
  work; module-singleton `is` identity still holds (the proxy *is* the
  singleton). `.locked()` reports the running loop's lock, or — from the worker
  thread that runs the synchronous engine call under the lock — whether any
  loop's lock is held (one lock in production).
- **Bounded registry** — a `WeakKeyDictionary` can't help here (a stored lock
  strongly refs its bound loop, keeping the weak key alive), so the registry is
  a plain dict pruned of closed-loop entries on the new-loop path. *(Caught in
  review — initial draft used a leaking `WeakKeyDictionary`.)*

### Tests
Two deterministic pins (the original flake was order-dependent):
- `test_locks_survive_distinct_event_loops` — contends each lock across two
  `asyncio.run` loops; raises the exact CI `RuntimeError` on a bare
  `asyncio.Lock`, passes on the proxy (mutation-verified on Python 3.12).
- `test_loop_local_lock_registry_prunes_closed_loops` — asserts closed loops are
  pruned instead of leaking one entry per loop (mutation-verified: 5 without the
  sweep).

Local: 126 passed (`test_web_routes_context_locks` + `_mutators` +
`test_web_hot_reload`, py3.12), ruff + format clean, mypy clean.

### DIFF
--- a/packages/memtomem/src/memtomem/web/routes/_locks.py
+++ b/packages/memtomem/src/memtomem/web/routes/_locks.py
@@ -53,5 +53,79 @@ from __future__ import annotations
 
 import asyncio
 
-_config_lock = asyncio.Lock()
-_gateway_lock = asyncio.Lock()
+
+class _LoopLocalLock:
+    """An ``asyncio.Lock`` proxy that resolves to a per-event-loop lock.
+
+    A bare module-level ``asyncio.Lock`` binds to the first event loop that
+    acquires it and then raises ``RuntimeError: ... is bound to a different
+    event loop`` when reused from another loop. In production this never
+    happens — the web server runs a single long-lived loop — but pytest gives
+    each ``async`` test its own loop, so whichever test acquires the lock first
+    binds it and a later test on a fresh loop fails non-deterministically.
+
+    Keying the underlying lock by the running loop keeps the contract intact:
+    *within* one loop every caller shares the same lock (so two concurrent
+    gateway mutators still serialise), while distinct loops get distinct locks.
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #1164, opened 2026-06-01
feat(init): split Codex memories into per-subdir namespaces

memtomem:
## Summary

`mm init`'s provider preset routed everything under `~/.codex/memories/` into a
single flat `codex` namespace. That directory actually holds three distinct
memory classes:

| subtree | role |
| --- | --- |
| `rollout_summaries/` | per-session episodic recaps Codex auto-writes |
| `extensions/` | the ad-hoc note inbox awaiting consolidation |
| top-level (`MEMORY.md`, `memory_summary.md`, …) | the consolidated long-term memory |

Folding all three into one namespace defeats per-class search and time-decay
(e.g. you can't decay high-volume rollout recaps harder than the curated core).

This splits the codex preset into three **ordered, literal-namespace** rules,
mirroring the existing `claude:<slug>` convention:

```
~/.codex/memories/rollout_summaries/**  → codex:rollout_summaries
~/.codex/memories/extensions/**         → codex:extensions
~/.codex/memories/**                    → codex:global   (catch-all, last)
```

A single `{parent}`/`{ancestor:N}` template can't express this (codex's depth
varies), so the split is N ordered rules resolved first-match-wins. No new
placeholder is introduced, so the RFC #304 `_VALID_PRESET_PLACEHOLDERS` lock is
left intact — RFC #304's broader vendor/product hierarchy can still supersede
these later.

## Changes

- **Preset machinery now supports multiple rules per category.**
  `_PROVIDER_RULE_TEMPLATES` values become ordered tuples of `(path_glob,
  namespace)`; `_proposed_rule_for_category` → `_proposed_rules_for_category`
  (returns a list). All three build sites and the module-level assert updated.
- **Re-run migration.** The exact legacy flat `codex` rule is removed before the
  split is appended, so the subdir rules land *before* the catch-all
  (first-match-wins). Surfaced with a `↻` banner line. A user's hand-edited
  codex rule (custom namespace) is matched exactly on path_glob **and**
  namespace, so it is never touched.
- **Shadow guard.** If a surviving recursive catch-all (e.g. a user-kept
  `~/.codex/memories/** → custom`) would make a proposed subdir rule
  unreachable, the subdir rule is dropped rather than written as dead config,
  and reported with a `⚠` banner line.
- **Docs / changelog.** `docs/guides/configuration.md` rule example updated to
  the split (also demonstrates first-match-wins ordering); `CHANGELOG.md`
  `[Unreleased]` entry added.

## Migrating existing data

The preset only governs *new* config generation. An already-indexed instance
moves to the new namespaces with a forced re-index (no `--namespace`, so policy
rules apply):

```bash
mm index ~/.codex/memories --force
```

Force re-index re-namespaces chunks in place (chunk identity and per-chunk
personalization preserved, ADR-0005); deleted-source orphans, if any, clear via
the normal delete-by-source path.

## Out of scope

The separate `mm ingest codex-memory` importer (which writes the
`codex-memory:` prefix) is intentionally **not** touched — unifying its prefix
would be a breaking, documented CLI change and belongs in its own PR.

## Review

Design and implementation were both reviewed by Codex (reviewer agent). The
diff-level pass returned no blockers; its one Major — a user's custom catch-all
silently shadowing freshly-appended subdir rules — is fixed by the shadow guard
above (`test_banner_reports_shadowed_subdir_rule`,
`test_write_preserves_user_rule_on_same_glob_diff_namespace`).

## Tests

New: codex 3-rule ordering, flat→split re-run migration, shadow-drop, and the
`↻`/`⚠` banner lines. Updated: the banner skip/all-skipped and idempotency
tests to non-superseded presets. Full suite: 5585 passed, 11 skipped; ruff +
mypy clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+- **`mm init` splits Codex memories into per-subdir namespaces.** The provider
+  preset for `~/.codex/memories/` now generates three ordered namespace rules
+  instead of one flat `codex` rule: `codex:rollout_summaries` (per-session
+  episodic recaps), `codex:extensions` (the ad-hoc note inbox), and
+  `codex:global` (the consolidated top-level memory) as the catch-all. Folding
+  those three distinct memory classes into one namespace defeated per-class
+  search and time-decay. Rules use literal namespaces only, so the RFC #304
+  placeholder lock is unaffected. Re-running `mm init` after upgrading migrates
+  an existing flat `codex` rule in place: the stale catch-all is removed and the
+  split takes its position (reported with a `↻` line), so the new subdir rules
+  resolve correctly under first-match-wins instead of being shadowed. A
+  hand-edited codex rule (custom namespace) is never touched. Existing indexed
+  data is migrated by re-running `mm index ~/.codex/memories --force`.
+
 ## [0.2.2] — 2026-05-31
 
 Patch release on top of 0.2.1: adds network MCP transports, a `mm web`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #1169, opened 2026-06-01
fix(indexing): enforce provider index-file conventions on every index path

memtomem:
## What

Provider index-file conventions — the meta/TOC files an agent keeps inside a memory dir (Claude Code's `MEMORY.md`/`README.md`, Codex's `README.md`) — are now centralized in one table and enforced on **every** index path, not just `mm ingest`.

## Why

The exclude set lived only in `cli/ingest_cmd.py`, so the general `memory_dirs` indexing path (`engine._discover_files` / `_index_file` / the file watcher) and `mm index <dir>` indexed `MEMORY.md` as searchable content. Since `MEMORY.md` is a pointer-only table of contents, it then surfaced as a high-score duplicate on every query. Verified live: this project's auto-memory dir had `MEMORY.md` indexed with 10 chunks.

## How

- **`config._PROVIDER_INDEX_CONVENTIONS`** — one frozen `ProviderIndexConvention` per `ProviderCategory` (`index_file` + `exclude_filenames`), plus `index_excluded_filenames()` / `provider_index_file()` helpers and a keys-vs-categories lock assert mirroring the existing `_PROVIDER_CATEGORY_PATTERNS` lock.
- **`engine._path_is_excluded`** (the single funnel for `_discover_files`, `_index_file`, `index_file`, and the watcher) now also excludes a file whose name matches the convention of its **owning** `memory_dir` root. Ownership uses the existing `resolve_owning_memory_dir` (most-specific, longest-prefix root), so a nested configured root overrides its parent's convention.
- **`cli/ingest_cmd.py`** derives `_EXCLUDE_FILENAMES` / `_CODEX_EXCLUDE_FILENAMES` from the central helper (behavior identical).
- **`mm purge --matching-excluded`** routes through `_path_is_excluded`, so it skips the same files and can reclaim `MEMORY.md` chunks indexed before this fix.

Provider-scoped: a `MEMORY.md` in a plain user folder is still indexed as real content.

## Tests

- `test_indexing_engine.py::TestPathIsExcludedProviderConventions` — claude-memory index/meta excluded, user-dir `MEMORY.md` kept, codex `README.md` excluded, and a nested-root regression (deepest root wins).
- `test_purge_cli.py` — provider-convention match + provider-scoping; existing calls updated for the new `memory_dirs` arg.
- `test_config_overrides.py` — convention lock + helper behavior.
- Full suite: **5617 passed**, ruff + mypy clean.

## Review

Design and diff reviewed by Codex. Its one Major — nested configured roots inheriting a parent provider's exclude set — was fixed by resolving the owning root with longest-prefix semantics, with a regression test.

First PR of a small series adding `mm memory doctor` (memory-index hygiene); this one is the conventions refactor + the exclude-enforcement bug fix it depends on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+- **Provider index-file conventions enforced on every index path (fixes
+  `MEMORY.md` pollution).** The exclude set for agent index/TOC files — Claude
+  Code's `MEMORY.md`/`README.md` and Codex's `README.md` — was previously
+  honored only by `mm ingest`, so the general `memory_dirs` walk, the file
+  watcher, and `mm index <dir>` indexed `MEMORY.md` as searchable content: a
+  pointer-only table of contents that then surfaced as a high-score duplicate
+  on every query. The conventions now live in one table
+  (`config._PROVIDER_INDEX_CONVENTIONS`) consulted by the shared
+  `_path_is_excluded` funnel, so ingest, the engine walk, the watcher, and
+  `mm purge --matching-excluded` all skip the same files. `mm purge
+  --matching-excluded --apply` consequently reclaims `MEMORY.md` chunks indexed
+  before this fix. The convention is provider-scoped — a `MEMORY.md` in a plain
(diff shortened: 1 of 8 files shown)


pandas-studio (approved):
LGTM

PR #1170, opened 2026-06-01
feat(cli): add `mm memory doctor` — read-only memory hygiene report

memtomem:
## What

Adds `mm memory doctor` — a **read-only, report-only** hygiene command that surfaces the 3-way drift between what's on disk, what an agent index/TOC file (e.g. Claude Code's `MEMORY.md`) points at, and what's actually indexed in the searchable DB.

PR 2 of the `mm memory doctor` plan; builds on the provider index-file conventions enforced in #1169.

## Why

A registered `memory_dir` can be **barely indexed** — the fs watcher only reacts to live events, so files that landed while the server was down stay invisible until a forced re-walk — and the index file can drift from the files on disk. When that happens, `mem_search` *silently* can't find the un-indexed files. There was also no single-file delete CLI, so deleting a memory file leaves its chunks orphaned in the DB. This command makes all of that visible.

## Checks (per configured `memory_dir`)

- **db_coverage** — files on disk the engine would index but with zero DB chunks (`indexed N/M`). Headline signal.
- **stale_source** — DB chunks whose `source_file` was deleted from disk.
- **convention_violation** — an index/meta file (`MEMORY.md`/`README.md`) indexed as content despite the provider convention.
- **broken_link** — index links classified `missing_target` / `outside_root` (`url` / `anchor` are recognised and *not* reported).
- **index_orphan** — files on disk the index file doesn't list ("not in the TOC" ≠ "not indexed").
- **budget** — index file over its byte / line / per-line-char hot-cache budget (per-line measured in characters, not bytes).
- **cold_candidate** — indexed files never accessed since indexing (informational).

Human output by default; `--json` for CI. Exits `1` only on **definite** inconsistencies (`stale_source`, `convention_violation`, `broken_link`); coverage/orphan/budget/cold are advisory and don't fail the exit code (a partially-indexed dir is a legitimate steady state) — mirrors `mm sync-doctor`, with a JSON + exit code like `mm context settings-doctor`.

## Read-only contract

- Config loads via `Mem2MemConfig` + `load_config_d(quiet=True)` + `load_config_overrides(migrate=False)` — never triggers the legacy auto-discover config rewrite.
- The DB is read through a bare `sqlite3` URI `mode=ro` connection, **never** `SqliteBackend.initialize()` (which would create the file/parent dir, run schema migration, and checkpoint the WAL on close). `mode=ro` still surfaces committed rows in a live writer's WAL; a missing / too-old / corrupt DB degrades to disk/index-only checks instead of being created.
- Discovery reuses the engine's own `discover_indexable_files` with **most-specific-owner** attribution (symmetric with the DB row bucketing), so coverage/orphan counts match the real indexer even with nested configured roots.

## Tests

`tests/test_memory_doctor.py` (25 tests): parser (`..`/absolute/url/anchor/whitespace/non-ASCII), link classification, budget (byte vs char), full drift detection against a **real** `SqliteBackend` + on-disk `claude-memory` dir, golden human + `--json` output, exit codes, and read-only regression pins — missing DB not created, parent dir not `mkdir`-ed, old-schema not migrated, corrupt DB doesn't crash, nested roots don't false-flag coverage.

## Validation

`ruff check` + `ruff format --check` clean; `mypy` clean. Run against the real auto-memory dir it correctly reports the known drift (incl. a deleted file's lingering chunk → `stale_source`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,23 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+- **`mm memory doctor` — read-only hygiene report for memory stores.** A
+  registered `memory_dir` can be barely indexed (the fs watcher only reacts to
+  live events, so files that landed while the server was down stay invisible
+  until a forced re-walk) and its index/TOC file can drift from what's on disk
+  — `mem_search` then silently can't find the un-indexed files. The new
+  `mm memory doctor` makes the 3-way drift between disk, the agent index file
+  (e.g. `MEMORY.md`), and the searchable DB visible: per configured dir it
+  reports DB coverage gaps, DB chunks whose source file was deleted, index/meta
+  files indexed as content, broken index links (classified
+  `missing_target`/`outside_root`, leaving `url`/`anchor` alone), files absent
+  from the TOC, hot-cache budget overruns, and never-accessed "cold" files.
+  Human output by default, `--json` for CI; exits `1` only on definite
+  inconsistencies (deleted-source chunks, convention violations, broken links).
+  Report-only — it never writes disk, DB, or config (loads config with
+  `migrate=False` so it can't trigger the legacy auto-discover rewrite). The
+  curation `--fix` path lands later behind its own ADR.
+
 - **Provider index-file conventions enforced on every index path (fixes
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #1171, opened 2026-06-01
docs(reference): document `mm memory doctor` + pin the contract with parity tests

memtomem:
## What

`mm memory doctor` (the read-only memory-hygiene report from #1170) shipped without user-facing docs. This adds them and pins the documented contract with tests so the docs can't drift from the implementation.

- **`docs/guides/reference.md` §5 "Memory hygiene — `mm memory doctor`"** — what the command does (the 3-way drift between disk, the index/TOC file, and the searchable DB), usage, the full **check → severity → remediation** table (all 11 checks), example human + `--json` output, exit-code semantics, and the read-only-by-design note.
- **`tests/test_memory_doctor.py::TestDocsParity`** — docs-as-tests guards.

## Docs-as-tests binding

The parity test **parses the check/severity table out of `reference.md`** and asserts it equals the `{check: severity}` pairs **extracted from the command's own `Finding(...)` call sites via AST** — so adding, renaming, or re-classifying a check fails the test until the published table is edited to match (and vice versa). Mutation-validated: flipping a severity in the source makes `test_docs_table_matches_source` fail with a clear diff. The AST extractor fails loudly on any non-literal `check`/`severity` rather than silently skipping it. Budget caps, the `--json` status rule, and the `--help` text are pinned too.

## Verified before writing

| Documented claim | Verified against |
|---|---|
| Check names + severities (11) | AST of `Finding(...)` call sites in `memory_doctor_cmd.py` (12 sites; `index_missing` ×2) |
| Exit code `1` on `stale_source`/`convention_violation`/`broken_link` | `memory_doctor` exit logic + `--help` |
| `--json` shape `{status, dirs[], summary}`; `status="issues"` on error/warn | live `mm memory doctor --json` |
| Budget caps 24,400 B / 200 lines / 200 chars | `_INDEX_MAX_*` constants |
| `db_coverage` → `mm index <dir> --force` | `mm index --help` |
| `convention_violation` → `mm purge --matching-excluded --apply` | `mm purge --help` / `purge_cmd.py` |
| `stale_source` → `mem_do(action="cleanup_orphans")` | `reference.md` §5 / `dedup_decay.py` |
| `unowned_chunks` → `mm gc orphan-projects` only for deleted project roots | `mm gc orphan-projects --help` / `gc_cmd.py` |
| Example human output format | live `mm memory doctor` |

## Tests

[...]

PR #1172, opened 2026-06-01
docs(adr): ADR-0020 — index-file write contract for `mm memory doctor --fix`

memtomem:
## What

ADR-0020 records the **write contract** for a future `mm memory doctor --fix`, before any code lands. PR4 of the memory-doctor series (PR1 #1169, PR2 #1170, PR3 #1171); the `--fix` implementation is a later PR.

`mm memory doctor` shipped read-only by design and deferred curation "to its own ADR". The decision matters because **memtomem does not write index files today** — `MEMORY.md` is agent-curated by the Claude Code memory hook, so `--fix` would be memtomem's *first* write to that artifact.

## The contract

- **Subtractive-only, provably-dead-only.** `--fix` may only *delete* `broken_link`/`missing_target` pointer lines (target inside the root but gone from disk). No add/reorder/reformat/budget-trim — so it can't fight the agent's curation.
- **Byte-exact preservation.** Surviving lines kept verbatim incl. CRLF/LF + trailing-newline, via line-splice on a **newline-preserving** read of the original text (`read_bytes().decode` / `open(newline="")`, *not* `read_text()` which normalizes CRLF→LF) — not reconstruction from the parser's terminator-stripped fields. Round-trip test required.
- **Dry-run default + `--apply`** (mirrors `mm purge`); the #1170/#1171 read-only contract is unchanged.
- **Atomic + concurrency-aware.** `atomic_write_text` (mkstemp + `os.replace`) preserving the existing file mode; sidecar lock for memtomem-vs-memtomem; fresh re-read + re-validate (still-present *and* still-dead) for memtomem-vs-agent — which **bounds but does not eliminate** the irreducible sub-`os.replace` race (the agent doesn't honor the lock; no CAS at the replace point). The ADR documents and accepts that residual race honestly rather than claiming "never clobbers", and requires `--apply` to report removed lines so loss is auditable.
- **Deferred (Tier 3):** budget trim, `index_orphan` add, `outside_root` — TRACKER.md row, event-driven.

Layers onto ADR-0011/0008/0005.

## Review

Design-gated by Codex over **four rounds** — it caught five distinct real design issues before SHIP: concurrency overstatement, byte-round-trip infeasibility from stripped parser fields, file-mode downgrade (`0o600` default), recoverability overstatement, and CRLF normalization at the read step. Final verdict: SHIP. All cited APIs/ADRs verified against source.

### DIFF
--- a/docs/adr/0020-memory-index-write-contract.md
+++ b/docs/adr/0020-memory-index-write-contract.md
@@ -1,0 +1,199 @@
+# ADR-0020: Index-file write contract for `mm memory doctor --fix`
+
+**Status:** Accepted
+**Date:** 2026-06-01
+**Context:** `mm memory doctor` (shipped in #1170, documented in #1171,
+`docs/guides/reference.md` §5) is **read-only by design**: it reports the
+3-way drift between disk, the agent index/TOC file (`MEMORY.md` for a
+`claude-memory` dir), and the searchable DB, but writes nothing. Its module
+docstring and the reference guide both defer the curation `--fix` path to
+"its own ADR" — this is that ADR.
+
+The decision matters because memtomem **does not write index files today** —
+they are agent-curated artifacts. The Claude Code memory hook owns `MEMORY.md`:
+it appends pointer lines, trims the budget, and orders entries. A `--fix` that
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

Issue #688, opened 2026-05-02
RFC: Tag management — rename / delete / merge in the Tags tab


memtomem:
## Background

`packages/memtomem/src/memtomem/web/routes/tags.py` only has two
routes today: `GET /tags` (list with counts) and `POST /tags/auto`
(extract tags). There is no way to rename, delete, or merge tags
through the API or the UI — every fix-up requires either editing
chunks one by one in the Search-tab detail panel or re-running
`auto_tag` with `overwrite=true`, both of which are error-prone for
even small taxonomy clean-ups.

A reviewer flagged this in a recent UX walk-through (Tags tab → "no
manage UI"). The Tier-1/Tier-2 polish PRs (#682–#687) intentionally
deferred this because it's a backend-shaped change with retrieval-
correctness implications, not a paint job. Filing as an RFC so we can
align on shape before any code lands.

## Goals

1. Rename one tag globally: `old → new`, applied to every chunk
   that carries `old`. Idempotent if `new` already coexists.
2. Delete one tag globally: drop it from every chunk that carries
   it. Chunks that end up tag-less stay indexed; we don't re-tag.
3. Merge N tags → 1: same as rename but for a set, with the
   resulting chunk-tag list deduplicated.

Out of scope for this RFC: tag taxonomies / hierarchies, synonyms,
per-namespace tag scopes, undo history.

### Backend

```
PUT    /api/tags/{name}           # rename body { new_name: str }
DELETE /api/tags/{name}           # delete tag from all chunks
POST   /api/tags/merge            # body { sources: [str], target: str }
```

All three return:

```jsonc
{
  "tag": "<resolved name>",
  "affected_chunks": <int>,
  "dry_run": <bool>,
}
```

`dry_run=true` (URL param) → no writes, just return the count and
sample chunk ids (cap 10) so the UI can show a confirmation modal.

Implementation notes:
- Iterate matching chunks via the existing
  `storage.list_chunks_by_tag(tag)` (add if missing) and rebuild
  `ChunkMetadata.tags` per chunk.
- Single transaction per request (atomic across the
  `upsert_chunks` batch) so partial failure can't strand a tag in
  a half-renamed state.
- Embeddings stay valid: tag changes touch `metadata.tags` only,
  not `content`, so BM25 / dense indexes don't need a rebuild.
  (Worth pinning with a test that asserts `embedding`,
  `content_hash`, `created_at` survive a rename.)
- Reject system / reserved prefixes (`validity:`, `system:`,
[...]

PR #1175, opened 2026-06-01
fix(tags): make dry-run tag sample global to match count/apply (#688)

memtomem:
## Why

The Tags **rename / delete / merge** ops (#688 PR1, shipped in v0.1.36) are
**global** — they mutate every chunk carrying the tag regardless of scope
tier — and `count_chunks_by_tag` counts globally to match. The dry-run
**sample** that feeds the confirm-modal preview, however, routed through
`recall_chunks(tag_filter=...)`, which always appends the ADR-0011 §6
scope-context fragment. With no project context pinned, that narrowed the
sample to `scope='user'`.

Result: the three paths disagreed on which rows the op touches.

| path | query | scope |
| --- | --- | --- |
| apply (`rename/delete/merge_tags`) | `chunks` whole table | **global** |
| count (`count_chunks_by_*`) | `chunks` whole table | **global** |
| sample (`list_chunks_by_tag`) | via `recall_chunks` | **`scope='user'` only** ❌ |

So a tag living only on project-tier chunks previewed as **"N affected, 0
samples"** while the apply still wiped all N rows — a destructive op with a
misleadingly empty confirmation preview. The mix case (some user, some
project) under-sampled the same way.

## Fix

Query the `chunks` table directly with the **same `EXISTS(json_each ...)`
membership** as `count_chunks_by_tag`, so sample / count / apply all agree
on one global row set. `ORDER BY created_at DESC, id` keeps the sample
deterministic under second-precision timestamp ties (the prior
`recall_chunks` path used a scope-priority tie-break we explicitly don't
want here).

The only caller of `list_chunks_by_tag` is the tag-management dry-run path
(`services/tag_management.py`), so decoupling it from `recall_chunks` has
no search-facing impact.

## Test

`test_list_chunks_by_tag_samples_globally_across_scopes` seeds
user + project_local + project_shared chunks under one tag and pins sample
membership to the global count. Mutation-validated: reverting only the
`sqlite_backend.py` change makes it fail (`assert 1 == 3` — old sample
returned the user-scope row only); the fix makes it pass.

```
uv run pytest packages/memtomem/tests/test_storage.py \
  packages/memtomem/tests/test_services_tag_management.py \
  packages/memtomem/tests/test_mcp_tag_management_parity.py \
  'packages/memtomem/tests/test_web_routes_extended.py::TestTagManagementRoutes'
# 74 passed; ruff check + format clean; mypy clean
```

## Scope

Pre-fix for #688 — surfaced while re-auditing the RFC state before picking
up the remaining UI (PR2) and CLI (PR3) work, both of which make this
dry-run preview user-facing. Not a close of #688.

### DIFF
--- a/packages/memtomem/src/memtomem/storage/sqlite_backend.py
+++ b/packages/memtomem/src/memtomem/storage/sqlite_backend.py
@@ -1426,12 +1426,25 @@ class SqliteBackend(
         return int(row[0]) if row else 0
 
     async def list_chunks_by_tag(self, tag: str, limit: int = 10) -> list[Chunk]:
-        # Thin wrapper over ``recall_chunks(tag_filter=...)`` — the post-fusion
-        # tag-filter path (#750) already implements the json_each EXISTS match
-        # and ORDER BY created_at DESC LIMIT shape this helper needs. Keeping
-        # the SQL in one place means the dry-run sample path inherits any
-        # future correctness fixes for free.
-        return await self.recall_chunks(tag_filter=tag, limit=limit)
+        # Dry-run sample for the global tag-management ops (rename / delete /
+        # merge). Those ops mutate EVERY chunk carrying the tag regardless of
+        # scope tier, and ``count_chunks_by_tag`` counts globally to match, so
+        # the sample must draw from the same global row set. Routing through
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #1176, opened 2026-06-01
feat(cli): add `mm tags` — list / rename / delete / merge tags (#688 PR3)

memtomem:
## Why

PR3 of the #688 tag-management split. PR1 shipped the shared
`services.tag_management` service plus the Web `/api/tags` routes and MCP
`mem_tag_*` tools (v0.1.36); #1175 fixed the dry-run sample scope. The CLI
was the remaining gap — `mm` had no way to rename/delete/merge tags, so
scripts and the UI were asymmetric.

This adds a thin CLI surface that funnels through the **same** service, so
the gate (strip, same-name reject, in-process write lock, `updated_at`
curation bump, search-cache invalidation) lives in one place and all three
surfaces — Web, MCP, CLI — stay symmetric. The CLI only renders and gates
the write.

## Commands

```
mm tags list                            # tags + chunk counts (most frequent first)
mm tags rename OLD NEW [--apply] [--yes]
mm tags delete NAME    [--apply] [--yes]
mm tags merge SRC...   --into TARGET [--apply] [--yes]
```

UX mirrors the existing `mm gc orphan-projects`:

- **dry-run by default** — prints the affected count + a sample (cap 10),
  no write;
- `--apply` performs the write behind a confirmation prompt;
- `--apply --yes` is the non-interactive form;
- `--yes` without `--apply` is a usage error.

Mutations are **global** (every chunk carrying the tag, regardless of scope
tier), so the preview count/sample matches exactly what apply will touch
(this is what #1175 guaranteed). `search_pipeline` is threaded only into
the apply call — a dry-run never invalidates the search cache. A 0-row
preview short-circuits before the write even under `--apply --yes`.

## `mm tags list`

Included as a 4th read-only subcommand for parity with MCP `mem_tag_list`
and so the group has a way to discover tag names before renaming — beyond
the RFC's original three, but read-only and trivial.

## Tests

`test_tags_cmd.py` mocks `services.tag_management` and pins the CLI glue:
dry-run default, the `--apply`/`--yes` gate, confirm-abort, the 0-row
short-circuit, `ValueError` surfacing, and `search_pipeline` threaded only
on apply. Mirrors the mock-based CLI tests for `mm gc` (`test_orphan_gc.py`
`TestCli`); the service SQL surface is covered in
`test_services_tag_management.py` and `test_storage.py`.

```
uv run pytest packages/memtomem/tests/test_tags_cmd.py \
  packages/memtomem/tests/test_services_tag_management.py \
  packages/memtomem/tests/test_storage.py packages/memtomem/tests/test_orphan_gc.py
# 87 passed; ruff check + format clean; mypy clean; `mm tags --help` smoke OK
```

## Scope

Closes the CLI gap for #688 (PR3). Remaining: PR2 — the Tags-tab UI
(hover Rename/Merge/Delete + dry-run confirm modal + i18n). Not a close of
#688.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/__init__.py
+++ b/packages/memtomem/src/memtomem/cli/__init__.py
@@ -60,6 +60,7 @@ def _register() -> None:
     from memtomem.cli.shell import shell
     from memtomem.cli.status_cmd import status
     from memtomem.cli.sync_doctor_cmd import sync_doctor
+    from memtomem.cli.tags_cmd import tags
     from memtomem.cli.uninstall_cmd import uninstall
     from memtomem.cli.upgrade_cmd import upgrade
     from memtomem.cli.watchdog_cmd import watchdog
@@ -84,6 +85,7 @@ def _register() -> None:
     cli.add_command(activity)
     cli.add_command(status)
     cli.add_command(sync_doctor)
+    cli.add_command(tags)
     cli.add_command(watchdog)
     cli.add_command(schedule)
     cli.add_command(web)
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #1177, opened 2026-06-01
feat(web): Tags-tab rename / merge / delete actions (#688 PR2)

memtomem:
## Why

PR2 of the #688 tag-management split — the last piece. PR1 (#795) shipped
the shared service + Web `/api/tags/*` routes + MCP tools; #1175 fixed the
dry-run sample scope; #1176 added the `mm tags` CLI. The Tags tab itself
still had no way to rename / merge / delete a tag — every fix-up meant
editing chunks one at a time. This wires the UI to the existing routes.

## What

Each tag row gets a **hover-revealed action menu** — Rename / Merge /
Delete — and a confirmation modal that:

1. runs a **dry-run first** (`?dry_run=true`) → shows *"This will affect N
   chunks across all scopes"* + a sample (cap 10);
2. writes only on an explicit confirm.

`rename` and `merge` collect a value (new name / merge target) in an
**input phase**, then transition to the shared **preview phase**; `delete`
opens straight into preview. Mutations are global (every chunk carrying
the tag, regardless of scope tier), and the preview count/sample matches
exactly what apply touches — guaranteed by #1175.

### Correctness: apply can only write a previewed value

A monotonic request token (`_tagReqSeq`) tags every dry-run / apply.
Editing the value after a preview — or while a dry-run is still in flight —
bumps the token, so the stale response is dropped and the modal reverts to
the input phase. `apply()` writes `previewedValue` (the snapshot), never
the live input, with a defence-in-depth guard. So there is no window where
the count/sample shown describes a different value than the one written.

### Other behaviour

- Backend 400 (empty / same-name) is shown inline; the modal stays open.
- `affected_chunks === 0` disables the apply button.
- a11y reuses the existing `showConfirm` pattern (`openModalA11y` + focus
  trap + ESC/backdrop); for the destructive **delete**, initial focus is
  **Cancel**, not the danger button.
- Action buttons `stopPropagation` so they don't also trigger the row's
  search-by-tag click.
- All strings i18n-keyed (en + ko parity).

## Tests

`tests-js/tags-manage-actions.test.mjs` — 9 jsdom/vitest tests pinning the
wiring: dry-run-before-write, input → preview → apply phases, **edit after
preview requires a fresh dry-run and applies the NEW value**, **a stale
mid-flight dry-run is dropped**, 400-keeps-open, empty-rejected,
cancel-no-write.

```
cd packages/memtomem/tests-js && npm test
# 15 files, 83 tests passed (74 existing + 9 new)
# i18n parity (test_i18n.py): 47 passed; en.json + ko.json valid
```

Static assets bumped: `app.js?v=119`, `style.css?v=92`.

## Scope

Closes the UI gap for #688. With PR1 + #1175 + #1176 (CLI) + this, all
three surfaces — Web, MCP, CLI — can rename/delete/merge tags through one
shared service. #688 can close once this lands.


pandas-studio (approved):
LGTM

PR #1178, opened 2026-06-02
fix(retry): honor string retry_after attributes in with_retry (#1029)

### LINKED ISSUE #1029: retry: let with_retry honor string retry_after attributes
## Summary

`with_retry()` honors an exception's `retry_after` attribute only when it is an `int` or `float`:

```python
ra = getattr(exc, "retry_after", None)
if ra is not None and isinstance(ra, (int, float)):
    delay = max(delay, float(ra))
```

The same module already has `_parse_retry_after()` for numeric strings and RFC 7231 HTTP-date values, but `with_retry()` does not use it for string-valued attributes.

## Scope

1. Update `packages/memtomem/src/memtomem/embedding/retry.py` so `with_retry()` accepts `retry_after` attributes that are numbers or strings.
2. Reuse `_parse_retry_after()` for string values instead of duplicating parsing logic.
3. Add focused tests in `packages/memtomem/tests/test_retry.py`.

## Acceptance criteria

- Numeric `retry_after` behavior stays unchanged.
- `retry_after = "5"` delays for 5 seconds when larger than the exponential backoff delay.
- An HTTP-date `retry_after` string is parsed through `_parse_retry_after()`.
- Invalid string values are ignored and fall back to exponential backoff.
- `uv run pytest packages/memtomem/tests/test_retry.py` passes.

This is a small helper-level change. Avoid changing provider-specific OpenAI/Anthropic/Ollama code in the same PR.

memtomem:
Closes #1029.

## What

`with_retry()` honored an exception's `retry_after` attribute only when it was already an `int`/`float`. The module already ships `parse_retry_after()` (handles numeric strings **and** RFC 7231 HTTP-dates), but `with_retry()` never used it for string-valued attributes — so a provider that surfaces the `Retry-After` header verbatim as a string fell back to plain exponential backoff.

## Change

`packages/memtomem/src/memtomem/embedding/retry.py` — string `retry_after` now routes through `parse_retry_after()`:

```python
ra = getattr(exc, "retry_after", None)
ra_delay: float | None = None
if isinstance(ra, (int, float)):
    ra_delay = float(ra)
elif isinstance(ra, str):
    ra_delay = parse_retry_after(ra)
if ra_delay is not None:
    delay = max(delay, ra_delay)
```

- Numeric (int/float) behavior is unchanged.
- Unparseable strings return `None` → fall back to backoff; a bad header can't break the delay.
- The `max(delay, ...)` floor means a string smaller than the backoff can't shrink it either.

> Note: the issue referenced the helper as `_parse_retry_after`; the real public helper is `parse_retry_after` (no leading underscore). No duplicated parsing logic.

## Tests

`packages/memtomem/tests/test_retry.py` — new `TestWithRetryStringRetryAfter`:
- `retry_after="5"` → sleeps 5.0 (wins over 1.0 backoff)
- HTTP-date string → routed through `parse_retry_after`, ~10s
- `"not-a-number"` → falls back to 1.0 backoff
- `"0.2"` → loses to the 1.0 backoff floor

`uv run pytest packages/memtomem/tests/test_retry.py` → 22 passed. ruff check + format clean.

### DIFF
--- a/packages/memtomem/src/memtomem/embedding/retry.py
+++ b/packages/memtomem/src/memtomem/embedding/retry.py
@@ -72,10 +72,19 @@ def with_retry(
                     last_exc = exc
                     if attempt < max_attempts - 1:
                         delay = min(base_delay * (2**attempt), max_delay)
-                        # Honour Retry-After from rate-limit responses
+                        # Honour Retry-After from rate-limit responses. The
+                        # attribute may be numeric (seconds) or a string (a raw
+                        # Retry-After header: seconds or an RFC 7231 HTTP-date);
+                        # strings go through parse_retry_after, which returns
+                        # None for unparseable values so we fall back to backoff.
                         ra = getattr(exc, "retry_after", None)
-                        if ra is not None and isinstance(ra, (int, float)):
-                            delay = max(delay, float(ra))
+                        ra_delay: float | None = None
+                        if isinstance(ra, (int, float)):
+                            ra_delay = float(ra)
+                        elif isinstance(ra, str):
+                            ra_delay = parse_retry_after(ra)
+                        if ra_delay is not None:
+                            delay = max(delay, ra_delay)
                         logger.warning(
                             "Retry %d/%d for %s after %.1fs: %s",
                             attempt + 1,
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #1179, opened 2026-06-02
test(webhooks): pin _validate_webhook_url URL safety behavior (#1030)

### LINKED ISSUE #1030: test(webhooks): pin webhook URL validation behavior
## Summary

`server/webhooks.py` has URL safety checks in `_validate_webhook_url()`, but `packages/memtomem/tests/test_webhooks.py` currently covers lifecycle behavior and does not pin the validation cases directly.

The helper already rejects unsupported schemes and private/reserved IP literals. Adding direct tests would make future changes safer.

## Scope

Add focused tests in `packages/memtomem/tests/test_webhooks.py` for `_validate_webhook_url()`:

- Accepts `https://example.com/hook`.
- Rejects unsupported schemes such as `file://...`.
- Rejects loopback/private/reserved IP literals such as `http://127.0.0.1/hook`, `http://10.0.0.1/hook`, and `http://[::1]/hook`.
- `WebhookManager` disables itself when configured with a rejected URL.

## Acceptance criteria

- Tests assert on the helper's returned error strings at a stable, not overly-specific level.
- No network access is needed.
- Existing webhook behavior is unchanged.
- `uv run pytest packages/memtomem/tests/test_webhooks.py` passes.

This is test coverage only unless a test exposes a real bug. Do not broaden webhook validation to DNS pinning in this PR; that would be a separate security design change.

memtomem:
Closes #1030.

## What

`server/webhooks.py::_validate_webhook_url()` rejects unsupported schemes and private/reserved IP literals, and `WebhookManager` self-disables when constructed with a rejected URL. `test_webhooks.py` already covered lifecycle (fire / event-filtering) but never pinned the validation contract directly — so a regression in the safety checks would slip past CI.

## Tests added — `TestValidateWebhookUrl`

- `https://example.com/hook` and `http://example.com/hook` (public DNS host) → accepted
- `file:///etc/passwd` → rejected (asserts the **"scheme"** reason)
- `http://127.0.0.1/hook`, `http://10.0.0.1/hook`, `http://[::1]/hook` → rejected (asserts the **"private/reserved"** reason; covers loopback, private, and IPv6 loopback)
- `WebhookManager(enabled=True, url=<rejected>)` → effective config is disabled, and stays inert (`_client is None`) even when a configured event fires

Assertions are at the rejection-**reason** level (scheme vs IP), not the exact message, per the issue's "stable, not overly-specific" guidance — wording can change without churning tests.

## Notes

Test coverage only — no production behavior change, no network access, no DNS-pinning scope creep (that would be a separate security design change, per the issue).

`uv run pytest packages/memtomem/tests/test_webhooks.py` → 19 passed. ruff check + format clean.

### DIFF
--- a/packages/memtomem/tests/test_webhooks.py
+++ b/packages/memtomem/tests/test_webhooks.py
@@ -46,6 +46,65 @@ class TestWebhookManager:
         assert mgr._client is None
 
 
+class TestValidateWebhookUrl:
+    """Pin the URL safety checks in ``_validate_webhook_url`` (#1030).
+
+    Assertions stay at the rejection-reason level (scheme vs IP) rather than
+    pinning the exact message, so wording can change without churning tests.
+    """
+
+    def test_accepts_https_url(self):
+        from memtomem.server.webhooks import _validate_webhook_url
+
+        assert _validate_webhook_url("https://example.com/hook") is None
+
+    def test_accepts_http_dns_host(self):
+        from memtomem.server.webhooks import _validate_webhook_url
+
+        # A public DNS name over http is allowed — only IP literals are checked.
+        assert _validate_webhook_url("http://example.com/hook") is None
+
+    def test_rejects_unsupported_scheme(self):
+        from memtomem.server.webhooks import _validate_webhook_url
+
+        err = _validate_webhook_url("file:///etc/passwd")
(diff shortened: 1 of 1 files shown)


pandas-studio (approved):
LGTM

PR #1192, opened 2026-06-02
feat(context): read-only runtime/client registration registry (ADR-0021 PR1)

memtomem:
Implements the runtime-status surface of **ADR-0021 Context Portal** (#1191). Backend + MCP/HTTP wrappers only; no UI yet (Project Portal / chips land in later PRs).

## What

- **`context/runtime_registry.py`** — read-only registry reporting per-provider-client install + `memtomem`/`mms` registration. Multi-location probes: Claude (user `~/.claude.json` mcpServers / local `projects."<root>".mcpServers` / project `<root>/.mcp.json`), Antigravity (CLI + IDE + VS Code-side `mcp.json` with the `servers` key), Codex (`~/.codex/config.toml` TOML), Kimi (`~/.kimi/mcp.json` / `$KIMI_SHARE_DIR`).
- **`GET /api/context/runtimes`** — the client/provider axis.
- **`overview.detected_runtimes`** — additively enriched with `installed` / `memtomem_registered` (mapped `gemini`→antigravity via `RUNTIME_TO_CLIENT`); `name` + `available` preserved (**backward compatible**).
- **`mem_context_detect(include_runtimes=True)`** — reports registration via a *separate bool* so the shared sync/generate/init/diff `include` set is **not** widened (`mem_context_sync(include="runtimes")` still rejects).

## Trust boundary (ADR-0021 §B)

Read-only; never writes a client config. Returns only booleans, location kinds, and `$HOME`-collapsed paths plus a coarse `error_kind` — **never raw config bytes/values, never an exception message** (parsed config never escapes `_probe_location`). **Never imports `memtomem_stm`.** `docs/guides/mcp-clients.md` is the location source-of-truth, pinned by a conformance test.

## Tests

20 new (multi-location detection, Antigravity `servers` key, Codex TOML, Kimi share-dir, coarse error classification, **secret-non-egress pinned to `privacy.scan`**, **mcp-clients.md SoT-conformance**, route shape, overview enrichment, MCP detect, sync-include invariant). Existing context route + MCP suites green (253). `ruff` clean. Core module reviewed by codex (SHIP).

> Provider-client set is Claude / Antigravity / Codex / Kimi (Antigravity = gemini-family client; standalone Gemini CLI out of scope). Best reviewed after #1191.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1191, opened 2026-06-02
docs(adr): ADR-0021 Context Portal (gateway → multi-project state board)

memtomem:
## What

Adds **ADR-0021** recording the decision to elevate the Context Gateway dashboard into a **Context Portal** — a project × runtime × artifact state board. Doc-only (`docs/adr/0021-context-portal.md` + two `docs/adr/TRACKER.md` rows).

## Decision (v1 scope)

(A) multi-project **Project Portal**, (B) **runtime/provider status** chips + filter, (C) **Sync All** extension.

## How it relates to prior ADRs

- **Narrowly supersedes ADR-0009**: multi-project is **read/manage only**; artifact writes stay one `(project_root, target_scope)` per invocation (preserves **ADR-0016 §5**); registry-lifecycle mutations (unregister, label) carry no artifact-tier dimension and sit outside that rule (**ADR-0015** project-root-only).
- Runtime-registration detection is constrained to **read-only** client-config probing, `privacy.scan` redaction (no raw config egress), and **no `memtomem_stm` coupling**; `docs/guides/mcp-clients.md` is the location source-of-truth.
- **Sync All reuses the existing `ctx-sync-all-btn`** (no backend endpoint — avoids `_gateway_lock` reentrancy and the skills exclusion in `_sync_atomic`).

## Deferred (TRACKER rows / ADR open questions)

Docs Drift Doctor, atomic backend `/sync-all` (both TRACKER), plus inline conflict editor / destructive delete / MCP Sync-All parity (ADR-only v1 scope-outs).

## Notes

Provider-client surface is **Claude / Antigravity / Codex / Kimi** (Antigravity is the gemini-family client; standalone Gemini CLI is out of scope). Plan + ADR reviewed across multiple adversarial codex rounds (Blocker 0).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1193, opened 2026-06-02
feat(context): project health + label rename + lazy counts (ADR-0021 PR2)

memtomem:
## Context Portal v1 (ADR-0021) — PR2: project health + label rename + lazy counts

Backend slice of the Gateway → Portal epic. Builds on PR0 (ADR-0021, #1191) and
PR1 (runtime registry, #1192). Frontend Project Portal section is PR4.

### What changed

**`context/projects.py`**
- `ProjectScope.stale` + `ProjectHealth` / `annotate_project_health(scope)`.
  *stale* = root exists but has no `.memtomem/` store (registered/discovered yet
  never initialized → the Portal will offer "Initialize"); *missing* = root
  gone. The two are mutually exclusive.
- `discover_project_scopes` is now **label-aware**: a stored `known_projects`
  label wins precedence (**stored > "Server CWD" > basename**) — it was ignored
  before.
- `KnownProjectsStore.update_label_by_scope_id` — atomic + lock-guarded, updates
  every matching row (mirrors `remove_by_scope_id`).

**`web/routes/context_projects.py`**
- `GET /api/context/projects`: per-type counts are now **opt-in via
  `?include=counts`** (default `counts: null`) — each scope costs several
  artifact scans, so the N-project list stays cheap. Response also carries
  `stale`.
- New **`PATCH /api/context/known-projects/{scope_id}` `{label}`** — rename a
  registration (label only; blank clears to basename; 404 mirrors `DELETE`).
- `POST` now normalizes a blank label to `None` too, so the new precedence can't
  render an all-whitespace name.

**`web/static/context-gateway.js`** (no-regression coupling)
- The shared `_ctxFetchProjects` loader requests `?include=counts` so the
  scope-picker count badges keep working under the lazy default.
- The fetch-failure fallback scope gains `stale` + the full four-key `counts`
  dict to match the API shape.

### Compatibility
The only semantic change to the existing surface is `counts` going
`{…}` → `null` by default; the live UI is kept whole by the one-line loader
change above. Response shape is otherwise additive (`stale` added). No new
`memtomem_stm` import; no raw config/secret egress (labels + paths only).

### Tests / verification
- `tests/test_context_projects.py`: store label update (duplicate-row + clear),
  label-aware + stale/missing discovery, `annotate_project_health`.
- `tests/test_web_routes_context_projects.py`: `?include=counts` opt-in +
  default-omit + unknown-token, stale flip, PATCH round-trip/clear/404, POST
  blank-label normalization. Existing count asserts migrated to
  `?include=counts`.
- Over-narrow Playwright projects stub broadened to match the query string.
- ruff + mypy clean; context suite (1311) + vitest (21) + Playwright (33) green.
- Reviewed by codex (**SHIP**, 0 blocker/major) + a consumer-sweep workflow.

### DIFF
--- a/packages/memtomem/src/memtomem/context/projects.py
+++ b/packages/memtomem/src/memtomem/context/projects.py
@@ -32,9 +32,11 @@ logger = logging.getLogger(__name__)
 
 __all__ = [
     "ProjectScope",
+    "ProjectHealth",
     "KnownProjectsStore",
     "compute_scope_id",
     "discover_project_scopes",
+    "annotate_project_health",
 ]
 
 
@@ -91,7 +93,12 @@ class ProjectScope:
     # stops advertising a tier the producer can't return.
     tier: Literal["project"]
     sources: tuple[str, ...]
+    # ``missing`` — the root is gone (registered but no longer a directory);
+    # ``stale`` — the root exists but has no ``.memtomem/`` store (never
+    # initialized as a memtomem project). Mutually exclusive: a missing root is
+    # never also stale. See ``annotate_project_health``.
     missing: bool = False
+    stale: bool = False
     experimental: bool = False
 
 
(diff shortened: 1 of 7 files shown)


pandas-studio (approved):
LGTM

Issue #1074, opened 2026-05-19
ui(ctx-gateway): Sync All partial mutation leaves dashboard stale on later failure (#962 follow-up)


memtomem:
Surfaced in the 2026-05-19 Context Gateway implementation/UI-UX review (local report `docs/reports/context-gateway-implementation-ui-ux-review-2026-05-19.md`, gitignored). Follow-up to #962.

## Problem

`ctx-sync-all-btn` posts each visible artifact type sequentially, then settings. `loadCtxOverview()` is called only after all posts and settings result handling complete. If skills sync succeeds but agents or settings fails, the catch path shows an error toast and skips overview refresh.

## Code references

- `packages/memtomem/src/memtomem/web/static/context-gateway.js:1081-1170`

## Impact

- The filesystem may have changed while the dashboard still shows pre-sync counts.
- Users get a failure toast without seeing which earlier phases succeeded.
- Retrying can be confusing because the visible state is stale relative to disk.

## Recommendation

- Collect per-phase results and refresh overview in a `finally` or after any phase that started.
- Surface a compact partial-result toast or status block:
  - `Skills synced, Agents failed: <reason>`
  - `Hooks need confirmation`
- Add a browser test where the first POST succeeds and the second fails, then assert overview is refreshed or the UI shows a partial-sync summary.

PR #1197, opened 2026-06-03
feat(web): Context Portal — Sync All per-phase progress + summary + (project, tier) pin (ADR-0021 PR6)

memtomem:
Final slice of the **Context Portal v1** epic (ADR-0021). Extends the existing `ctx-sync-all-btn` — **no new backend route** (per ADR-0021 §"Sync orchestration": a single `/sync-all` endpoint would re-enter the non-reentrant `_gateway_lock` and can't model the skills/settings result shapes).

## What

**1. Pin both `(project, tier)` dimensions of a run.** Snapshot the *effective* scope id **and** the target tier at confirm-time and pass them fixed to every phase URL via new `_ctxWithTargetScope` opts (`targetScope`, `scopeResolved`). Previously only `_ctxActiveScopeId` was snapshotted while `_ctxWithTargetScope` re-read the mutable `_ctxTargetScope` global *and* re-resolved the scope against the live `_ctxProjectsCache` on each phase — so a mid-run **tier flip** *or* a **cache refresh** could send later phases to a different `(project, tier)`, violating ADR-0016 §5 ("one (project, tier) per invocation"). Tier/project controls are also disabled for the run's duration.

**2. Per-phase live progress + result summary.** New `#ctx-sync-status` region (declarative `_renderCtxSyncStatus`, not the SSE-only `makeChunkProgressRenderer`) shows each phase `pending → syncing → done | failed | not_run`, with a generated/dropped/skipped summary per artifact phase. Settings `needs_confirmation` renders as a distinct **attention** state so the row matches the "complete except Settings" toast. The region lives outside `#ctx-overview-content` (survives the post-sync reload), re-translates on `langchange`, and clears on a scope/tier switch.

All existing toast/severity routing and the #1074 overview-reload-in-`finally` are preserved unchanged.

## Tests

- Pin the `mcp-servers` phase the happy-path previously let the conftest catch-all swallow (5-phase order now asserted).
- New Playwright regressions — mid-run **tier-flip** and **cache-refresh** — both flip the bare global mid-run via a `window.fetch` wrapper and assert every phase URL stays pinned; both carry a non-vacuous guard proving the flip actually fired.
- New vitest for the progress/summary rendering (done-summary, mid-run-failure `not_run`, `needs_confirmation` → attention, clear-on-tier-change).

## Verification

- Full no-browser suite: **5610 passed**
- Browser (Playwright + Chromium): **172 passed**
- vitest: **95 passed**
- `ruff check` + `ruff format --check` + `mypy`: clean
- i18n en/ko key + placeholder parity: clean (10 new keys)
- Cache-bust: `context-gateway.js` v50→51, `style.css` v92→93

## Review

Reviewed by a read-only multi-lens adversarial pass before commit. It caught a **blocker the original plan missed** — only the tier was pinned; the scope still re-resolved per phase, so a mid-run cache refresh could collapse later phases to Server-CWD. Fixed with the `scopeResolved` bypass + the cache-refresh regression test above, plus three minors (the `attention` settings state, an unnecessary `escapeHtml` on a controlled class name, and the clear-on-tier-change test).


pandas-studio (approved):
LGTM

Issue #1080, opened 2026-05-19
ui(ctx-gateway): project endpoint failure silently collapsed to Server CWD only (#962 follow-up)


memtomem:
Surfaced in the 2026-05-19 Context Gateway implementation/UI-UX review (local report `docs/reports/context-gateway-implementation-ui-ux-review-2026-05-19.md`, gitignored). Follow-up to #962.

## Problem

`_ctxFetchProjects()` catches every fetch failure and creates a fallback single Server CWD scope. That was useful for older deployments and tests, but in current prod it can hide a broken `/api/context/projects` endpoint or unreadable project store as if all registered projects simply disappeared.

## Code reference

- `packages/memtomem/src/memtomem/web/static/context-gateway.js:195-221`

## Impact

- Users cannot tell the difference between "I have no added projects" and "the project management API failed".
- If the store becomes unreadable, the UI may look like it lost project registrations without offering a diagnostic.

## Recommendation

- Keep the fallback for compatibility, but surface a non-blocking warning toast/banner when the endpoint exists and returns an error.
- Add a JS test for project-fetch failure so the compatibility behavior and the new warning are both intentional.

PR #972, opened 2026-05-12
Fix tier-aware Hooks sync stale response handling

memtomem:
## Summary
- thread target_scope through Hooks sync GET/POST/resolve routes and Web calls
- render tier controls and scope-specific Hooks target labels
- guard loadHooksSync responses/errors by request sequence and requested tier so stale responses cannot paint the wrong tier
- update docs, styling, and regression pins for tier-aware Hooks sync

## Tests
- uv run pytest packages/memtomem/tests/test_context_settings.py packages/memtomem/tests/test_settings_doctor.py packages/memtomem/tests/test_web_routes_extended.py packages/memtomem/tests/test_web_mode.py packages/memtomem/tests/web/test_settings_hooks_detail.py

PR #1200, opened 2026-06-03
fix(web): gate projects-cache commit behind caller seq guard (#1194)

### LINKED ISSUE #1194: Context Gateway: _ctxFetchProjects clobbers shared _ctxProjectsCache before caller seq-guard (stale-fetch race)
## Problem

`_ctxFetchProjects` (`web/static/context-gateway.js`) unconditionally commits
`_ctxProjectsCache = data.scopes` **inside the helper, before any caller's
sequence guard runs**. Every caller (`loadCtxList`, `loadCtxOverview`,
`loadCtxProjects`) guards its own *render* against a stale response (seq /
scope), but none can prevent a **superseded fetch from clobbering the shared
cache**: if a newer load starts (higher seq) and an older in-flight
`_ctxFetchProjects` resolves *after* it, the older payload still lands in
`_ctxProjectsCache`. Subsequent reads of that cache (active-scope
normalization via `_ctxScopeParam` / `_ctxNormalizeActiveScope`, and any
section that renders from it) then operate on stale data until the next fetch.

Pre-existing since the multi-project work (#1080–#1102); no user report, low
probability (needs two overlapping fetches where the older resolves later).
Surfaced by the Codex review of the Context Portal PR (ADR-0021 PR4), which
**insulated the new Portal board** by rendering from a post-guard local
snapshot (`_ctxPortalScopes`) rather than the shared cache — so the Portal is
already correct; this issue is about the shared helper.

## Suggested fix

Split fetch from cache-commit: a fetch-only helper returns the payload, and
each caller commits `_ctxProjectsCache` **only after re-checking its own
seq/scope guard** (mirroring how each already gates its render). Preserve the
existing failure-handling contract in `_ctxFetchProjects` (#1080 shape
validation, #1100/#1101 toast de-dup, #1102 fallback) when refactoring.

## Scope
- `web/static/context-gateway.js`: `_ctxFetchProjects`, `loadCtxList`, `loadCtxOverview`
- `web/static/context-portal.js`: `loadCtxProjects` (already snapshots locally; align with the new contract)

memtomem:
## Problem
`_ctxFetchProjects` committed the shared `_ctxProjectsCache` and normalized +
persisted `_ctxActiveScopeId` **inside the helper, before any caller's sequence
guard ran**. A superseded in-flight projects fetch resolving *after* a newer one
therefore clobbered both the shared cache and the persisted active-scope
selection with stale data — even though every caller correctly skipped its own
stale *render*. Stale readers (`_ctxScopeParam` request URLs, the scope switcher,
conflict-draft keys) then operated on the wrong scope until the next successful
fetch. Pre-existing since the multi-project work (#1080–#1102); low-probability
(needs two overlapping fetches where the older resolves later).

## Fix
- Split `_ctxFetchProjects` into a pure **`_ctxFetchProjectsData`** (fetch only,
  returns `{data, warn}`, no global writes) and **`_ctxCommitProjects`** (cache +
  `_ctxNormalizeActiveScope` + the #1101 failure-toast memo).
- Every caller commits **only after its own seq/scope guard passes**:
  `loadCtxOverview`, `loadCtxList`, the portal's `loadCtxProjects`, and
  `loadHooksSync`.
  - **4th caller beyond the issue's listed three:** `settings-hooks-watchdog.js`'s
    `loadHooksSync` had the same pre-guard clobber; its commit is now gated too.
- `loadCtxOverview` additionally pins the tier captured at entry **and** the
  resolved effective scope, so the projects fetch and the overview fetch can't
  split across a mid-fetch tier flip (ADR-0021 §C; mirrors the portal's #972
  `requestedScope` guard).
- Legacy `_ctxFetchProjects` wrapper (= Data + Commit) kept for the
  #1080/#1101/#1102 direct-call tests.

## Tests
- New `tests-js/ctx-projects-stale-fetch-race.test.mjs`: a purity pin + a
  two-overlapping-fetch race (older resolves later), asserted via **public
  behaviour** (localStorage active scope + `_ctxProjectControls`) per the #1102
  no-poke convention. **Mutation-verified** — reverting any caller to
  commit-before-guard flips it red (active scope demotes to `''`).
- `test_i18n` overview structural guard updated to tolerate the tier-drift clause
  and require all **three** fetch-dependent guard paths (`>= 3`, not `>= 2`).
- Green: full vitest (98), browser-lane context/portal/hooks (56), full
  non-browser pytest (5611), `ruff check` + `format --check`.

## Review
Design reviewed by Codex before implementation (SHIP on the split; it flagged the
`loadCtxOverview` tier-pin, folded in as Major 1). Post-implementation adversarial
multi-agent review: 1 confirmed finding (guard-test threshold) fixed; 3 dismissed
on re-verification.

Closes #1194

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/context-gateway.js
+++ b/packages/memtomem/src/memtomem/web/static/context-gateway.js
@@ -235,7 +235,22 @@ function _ctxNormalizeActiveScope(scopes) {
   return active;
 }
 
-async function _ctxFetchProjects() {
+// Fetch ``/api/context/projects`` and classify the outcome, WITHOUT mutating
+// any module global. Returns ``{ data, warn }``:
+//   - ``data``  always a ``{scopes: [...]}`` object — the real scopes on
+//               success, or the synthetic Server-CWD fallback on any failure.
+//   - ``warn``  null on success / silent-404, else ``{kind, status?, detail}``
+//               describing a "loud" (toastable) failure shape.
+//
+// Pure by contract (#1194): a superseded, still-in-flight fetch that resolves
+// AFTER a newer one must not be able to clobber the shared ``_ctxProjectsCache``,
+// the normalized+persisted ``_ctxActiveScopeId``, or the failure-toast memo.
+// So this helper only READS — callers commit the result via
+// ``_ctxCommitProjects`` ONLY after re-checking their own sequence/scope guard.
+// ``opts.targetScope`` pins the tier when a caller must fetch projects and a
+// sibling resource (e.g. overview) under one tier, so a mid-flight tier flip
+// can't split the two requests across tiers (ADR-0021 §C).
+async function _ctxFetchProjectsData(opts = {}) {
   let data;
   // Four failure shapes need to stay distinguishable per #1080:
   //   - 404 / network throw   → older deployment or absent endpoint; the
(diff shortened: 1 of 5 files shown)


pandas-studio (approved):
LGTM

PR #1198, opened 2026-06-03
feat(context-gateway): project scope center and runtime coverage matrix

memtomem:
## Summary
Implements a project scope center and a runtime coverage matrix in the Context Gateway. 

### Changes
* **Backend**:
  * Extracted common runtime status probing into `memtomem/context/runtime_coverage.py`.
  * Integrated probing helper inside `/api/context/projects` DTO and deleted scan-heavy `Last Synced` fields.
* **Frontend**:
  * Implemented project scope matrix table in the Context Gateway Overview section.
  * Separated client installation status (`installed`) and memtomem registration status (`memtomem_registered`) in the runtime coverage badges.
  * Corrected sync disabled logic to properly evaluate the global `_ctxTargetScope === 'project_local'` instead of using the local scope tier.
  * Added `anyPhaseStarted` reload logic in `_ctxSyncProjectScope` to ensure dashboard indicators reload correctly after partial artifact sync success or settings sync errors/aborts.
  * Fixed click handler on Server CWD row sync button to correctly pass blank scope IDs.
* **Tests**:
  * Unit tests for the new `compute_runtime_coverage` helper.
  * Updated web routes schema verification for `runtime_coverage`.
  * Playwright E2E browser tests in `test_context_gateway_matrix_sync.py` verifying badge rendering, scope selection, sync button calls, and dashboard refresh flows.

PR #1201, opened 2026-06-03
fix(context-gateway): pin matrix sync scope/tier + gate runtime_coverage + i18n badges

pandas-studio:
## Summary
Post-merge review follow-up to #1198 (Project Scope Matrix). Addresses one correctness blocker, a default-path performance regression, an i18n gap, and a robustness/test-masking issue surfaced by review (Claude + adversarial Codex pass).

## Changes

**🔴 Blocker — matrix sync re-opened a (project, tier) race**
`_ctxSyncProjectScope` re-resolved scope **and** tier on every phase: it called `_ctxWithTargetScope(…, { scopeId })` inside the fan-out loop, and that helper re-reads the mutable `_ctxTargetScope` global and re-resolves the id against the live `_ctxProjectsCache` per call. A mid-run tier-filter flip — or a projects-cache refresh marking the scope missing — could send later phases to a different `(project, tier)`, the exact invariant the canonical Sync All pins against (ADR-0016 §5 / ADR-0021 §C). Now snapshots the effective scope id + tier once after confirm and passes `{ scopeResolved: true, targetScope }` to every phase.

**🟠 Major — `runtime_coverage` broke "cheap by default"**
`_scope_to_dict` computed `runtime_coverage` unconditionally, outside the `with_counts` gate, so the default `/api/context/projects` paid a `probe_all_runtimes` pass (per-client config-file reads) for *every* scope. Now opt-in via `?include=runtime_coverage`, symmetric with `counts`; the matrix's shared loader requests `?include=counts,runtime_coverage`.

**🟠 Major — badges weren't internationalized**
Badge labels, their tooltips, the disabled-sync title, and the inventory tooltip were hardcoded English while the rest of the matrix used `t()`. Routed through `t()` with 15 new keys in **both** `en.json` and `ko.json` (i18n parity guard enforces this).

**🟢 Minor / 🆕 robustness**
- Dropped the redundant ` (Reg)` suffix on the `Active` state (Active already implies registered).
- Empty runtime cell uses the base `badge` class (was an undefined `matrix-badge`).
- `counts === null` (the API default) now renders a muted `—` instead of a misleading all-zero inventory row.

## Tests
- Backend `runtime_coverage` opt-in test + updated default-omits assertion.
- Browser: new `counts: null` dash test and a **Korean-langchange** badge-localization test; updated the `Active` badge assertion.

## Validation (all CI gates green locally)
- `ruff check` + `ruff format --check`: pass
- `pytest` (route + i18n parity): 97 passed
- Playwright matrix browser tests: 8 passed
- vitest `js-unit`: 98 passed

## Deferred (noted, out of scope)
- Runtime column list is still hardcoded vs `KNOWN_RUNTIMES` (stable set; low value).
- Matrix Sync lacks the canonical "all-canonicals-empty" runtime-only guard (harmless no-op fan-out; needs per-scope canonical-empty state the matrix doesn't cheaply have).


memtomem (approved):
LGTM

PR #1203, opened 2026-06-03
feat(context): per-project sync enrollment + filtered auto-display

pandas-studio:
## What

Decouples project **discovery** from **sync participation** in the Context Gateway Projects portal. **Backend only** — the frontend enroll/toggle UI is a separate follow-up.

### Auto-display (configured projects)
- `discover_project_scopes` filters `~/.claude/projects` scan candidates to roots carrying a runtime marker (`auto_display_configured_projects`, **on by default**; `experimental_claude_projects_scan` stays the unfiltered escape hatch).
- Adds **`.codex`** to `_MARKER_DIRS` so codex-configured projects are recognized. The marker set now covers claude / antigravity (`.gemini`) / codex / kimi + canonical `.memtomem`. Verified against the Kimi Code CLI docs that per-project config is `.kimi/`.

### Sync enrollment
- New **`enabled`** flag on `known_projects` entries. A disabled project stays discoverable but is excluded from sync: the web Sync button (via new wire field), Sync All, and CLI `mm context update --all`.
- `/api/context/projects` gains **`enabled`** + derived **`sync_eligible`** (`server-cwd OR (enrolled AND enabled)`).
- `enabled` is read strictly from the `known_projects` entry during discovery (`known_by_key`), so a scan/cwd source can **never** silently re-enable a paused project.

### PATCH atomicity & fixes
- `PATCH /known-projects/{id}` updates label and/or `enabled` in a **single locked store write** (`update_entry_by_scope_id`); each field applies only when present (`label` via `model_fields_set`, `enabled: null` = unchanged) so neither clobbers the other.
- Rename now **preserves `enabled`**; legacy rows default to enabled (schema stays v1); the `no_runtime_marker` warning derives from `_MARKER_DIRS`.

## Why
Previously the only project sources were server-cwd + explicit Add-Project + an off-by-default experimental scan, and every discovered project was implicitly a sync target. Goal: configured projects auto-appear, while sync stays an explicit, reversible per-project decision.

## Review
Design was adversarially reviewed (NEEDS-FIX → all findings fixed): non-atomic combined PATCH, scan re-enabling a paused project, label/enabled clobber, stale marker warning.

## Deferred (separate follow-ups)
- Frontend enroll/toggle UI + Playwright + i18n.
- API-direct write guard (`resolve_writable_scope_root` → 409) across artifact + `settings_sync` routes.
- `~/.kimi-code` install-detection marker (ADR-0021 §1 drift).

## Test plan
- `uv run pytest -m "not ollama"` → **5842 passed**, ruff clean, mypy advisory unchanged.
- New tests: store `enabled` round-trip + legacy default + rename-preserves-enabled + atomic combined update; discovery filter + `sync_eligible` + paused-not-reenabled-by-scan regression; route PATCH enabled/label independence + null-unchanged + combined; CLI `update --all` skips disabled; suite-wide scan-isolation fixture.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -1383,7 +1383,10 @@ def _run_update_all(
         )
         return
 
-    project_roots = [e.root for e in project_entries if e.root.is_dir()]
+    # Sync enrollment: a disabled (paused) known project is excluded from the
+    # batch, matching the web Sync gate and ``sync_eligible``. ``enabled``
+    # defaults True for legacy entries, so this only skips explicitly-paused ones.
+    project_roots = [e.root for e in project_entries if e.root.is_dir() and e.enabled]
     if not project_roots:
         click.echo(
             "No registered projects exist on disk; nothing to update.",
--- a/packages/memtomem/src/memtomem/config.py
+++ b/packages/memtomem/src/memtomem/config.py
@@ -779,8 +779,18 @@ class ContextGatewayConfig(BaseSettings):
     known_projects_path: Path = Path("~/.memtomem/known_projects.json")
     # Opt-in reverse-decode of ``~/.claude/projects/<encoded>`` directory
     # names. Off by default — the encoding is fragile around dash-containing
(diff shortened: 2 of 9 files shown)


memtomem (approved):
LGTM

PR #1205, opened 2026-06-03
fix(context): don't badge auto-displayed configured projects as experimental

pandas-studio:
## What
`discover_project_scopes` computed `experimental = sources == {"claude-projects"}`. That was right when `~/.claude/projects` discovery was opt-in only, but **#1203** made the **filtered** scan (`auto_display_configured_projects`) **on by default** — so a marker-bearing scan row is now admitted by default yet was still flagged `experimental`. The existing Portal/Gateway UI renders that as an **"opt-in experimental `~/.claude/projects` scan"** warning, mislabeling normal auto-displayed configured projects.

## Fix
A scan-only row is experimental **only when it lacks a runtime marker** — i.e. it is present solely because the unfiltered `experimental_claude_projects_scan` gate is open:

```python
experimental = sources == {"claude-projects"} and not has_runtime_marker(resolved)
```

Marker-bearing scan rows (admitted by the default filtered auto-display path) are normal configured projects and carry no experimental badge. cwd / known-projects unions already clear the flag.

Follow-up to **#1203** (P3 review finding).

## Test plan
- `uv run pytest -m "not ollama"` → **5847 passed**; ruff clean.
- New tests: marker-bearing scan row (default config) → `experimental=False`; unmarked scan row (experimental gate on) → `experimental=True`. Existing assertions unchanged (unmarked `/tmp` scan row stays `True`; cwd-union stays `False`).

### DIFF
--- a/packages/memtomem/src/memtomem/context/projects.py
+++ b/packages/memtomem/src/memtomem/context/projects.py
@@ -732,10 +732,13 @@ def discover_project_scopes(
     scopes: list[ProjectScope] = []
     for key in order:
         resolved, sources, missing, stored_label = by_resolved[key]
-        # ``experimental`` is true iff the *only* source is the opt-in scan.
-        # cwd / known-projects unions clear the flag so the most-trusted
-        # source wins for display purposes.
-        experimental = sources == {"claude-projects"}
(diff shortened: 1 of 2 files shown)


memtomem (approved):
LGTM

PR #1199, opened 2026-06-03
feat: add CLI session command tracing and Langfuse v4 integration

memtomem:
Adds tracing to CLI session commands with local JSONL logging and Langfuse SDK v4 tracing support. See walkthrough.md in artifact logs for details.

PR #1209, opened 2026-06-03
fix(observability): redact secrets in session-trace metadata sinks

pandas-studio:
## What

Follow-up to the merged session-tracing feature (#1199): **redact secrets in trace *metadata*, not just the payload.**

`format_payload()` already redacts the trace payload, but metadata is exported through three separate paths that were never redacted — Langfuse propagated attributes (initial + final `propagate_attributes`), the Langfuse span output, and the local JSONL row. So a secret living in metadata leaked verbatim in `metadata` and `redacted` modes:

- the wrapped command line — `mm session wrap -- tool --api-key sk-...` lands in `metadata["command"]`, or
- an error string echoing it (`ctx["metadata"]["error"] = str(exc)` on command failure).

## How

- Add `_redact_metadata(metadata, mode)` and apply it at **all four** metadata egress points (3 Langfuse sinks + the JSONL row). It reuses the existing `_redact_value` primitive and redacts for every mode **except `full`** (the explicit verbatim opt-in for both payload and metadata). Redaction runs *before* `format_propagated_metadata()`'s 200-char truncation, so a secret can't be split/hidden by truncation.
- Harden `_redact_value` so non-string values don't slip through: an object / `Exception` / `bytes` value whose `__str__` contains a secret is stringified by the exporters anyway (`json.dumps(default=str)` for JSONL, `sanitize_metadata_value()`'s `str()` for Langfuse), so redact its string form too. Genuine JSON scalars (`int/float/bool/None`) pass through untouched → serialized output is unchanged for non-secret values.

## Tests

- **unit**: redaction across all three modes; the non-string-object leak; scalars are not stringified.
- **integration**: JSONL redacted vs `full`-verbatim; a **mocked-Langfuse** test asserting redaction at the initial/final `propagate_attributes` and `span.update(output=...)` sinks (the JSONL tests force Langfuse off, so they wouldn't catch a regression there).
- isolate `HOME` in the existing `mm session start` test so it no longer writes the developer's real `~/.memtomem/.current_session`.

Full suite: `5860 passed, 11 skipped` (`-m "not ollama"`); ruff + format clean. `mypy` (advisory) gains one error of the pre-existing `ctx`-untyped-union class on this file.

## Scope

Config validation, rollback (web/MCP/CLI), recursive *payload* redaction, and Langfuse telemetry failure-isolation are **already on main** (#1199) — this PR is the metadata-redaction follow-up only.

## Review trail

The non-string-value leak and the Langfuse-sink test gap were caught in review (Codex → NEEDS-FIX) and both addressed here; an independent read-only review confirmed the 4-sink coverage and mode logic.

### DIFF
--- a/packages/memtomem/src/memtomem/observability/session_tracing.py
+++ b/packages/memtomem/src/memtomem/observability/session_tracing.py
@@ -145,8 +145,39 @@ def _redact_value(val: Any) -> Any:
         val = re.sub(p2, r"\1\2***", val)
 
         return val
-    else:
+    elif val is None or isinstance(val, (bool, int, float)):
+        # JSON scalars cannot carry a secret string — leave them untouched so
+        # we don't stringify legitimate numbers/bools.
         return val
+    else:
+        # Any other object (Exception, bytes, custom type) is stringified by
+        # the exporters (json.dumps(default=str) for the JSONL row,
+        # sanitize_metadata_value's str() for Langfuse), so a value whose
+        # __str__ contains "--api-key sk-…" would otherwise leak verbatim.
+        # Redact its string form here too.
+        return _redact_value(str(val))
+
+
+def _redact_metadata(metadata: Dict[str, Any], mode: str) -> Dict[str, Any]:
+    """Redact metadata values unless ``full`` payload mode is selected.
+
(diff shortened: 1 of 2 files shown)


memtomem (approved):
LGTM

PR #1210, opened 2026-06-03
fix(web): enforce sync-eligibility on runtime-writing API routes (#1203 §1i)

pandas-studio:
## What & why

Closes the last deferred item of #1203 (**§1i**): a **defense-in-depth backend gate** so a direct API call can't write into a **paused** (enrolled→disabled) or **never-enrolled** (discovery-only scan row) project's runtime. Until now, pause was enforced only at the **Web UI** and `mm context update --all` — both bypassable by hitting the HTTP API directly.

## How

- New FastAPI dependency **`resolve_writable_scope_root`** (sibling of `resolve_scope_root`, sharing the extracted `_resolve_selected_scope` resolver — no behavior change to the read path). Raises **409** with a structured `detail` carrying a machine-readable `reason_code` (`sync_paused` / `sync_not_enrolled`) when the resolved scope has `sync_eligible=False`. Reuses the existing `ProjectScope.sync_eligible` derivation — no duplicated logic. `server-cwd` is always eligible (you can't pause the running dir; it coalesces with a paused entry).
- Applied to the **seven always-runtime-writing mutators**:
  - `POST /context/{agents,commands,skills,mcp-servers}/sync`
  - `POST /settings-sync` (apply), `POST /settings-sync/resolve`, `POST /settings-sync/rules/delete`
- **Cascade deletes** are conditionally gated via **`resolve_scope_root_cascade_gated`**: `DELETE /context/{agents,commands,skills}/{name}` removes only the canonical source when plain (**ungated**), but `?cascade=true` ALSO unlinks the generated **runtime** copies — a runtime write — so the cascade variant is gated. (MCP cascade is a documented v1 no-op, so it stays ungated.)
- **Not gated** (canonical-direction writes — pause must not block preparing a project's artifacts): `create/update` agent·command·skill·mcp, `import_*`, `rules/promote`, and plain (non-cascade) `delete`.

### Tier handling
The guard gates **both project-runtime tiers** — `project_shared` (`.claude/settings.json`) and `project_local` (`.claude/settings.local.json`). The `user` tier targets global `~/.claude` and stays **exempt**.

## Adversarial review — two blockers caught & fixed

1. **`project_local` bypass** (multi-agent review): gating only `project_shared` left the three **settings** routes open — unlike the artifact routes they have no `_reject_non_shared_write` backstop and legitimately accept `project_local`, so `?target_scope=project_local` slipped a runtime write into a paused project (live-reproduced). Fixed by gating every project-runtime tier (`target_scope != "user"`).
2. **Cascade-delete bypass** (Codex review): `DELETE …?cascade=true` unlinks runtime copies through a route still on the ungated resolver. Fixed with the cascade-aware resolver above.

Both are regression-tested.

## Tests

`packages/memtomem/tests/test_web_sync_write_guard.py`: paused / not-enrolled **409 across all gated routes**; `server-cwd` + coalesced cwd-paused **pass**; the `project_local` hole closed (409 + no-write); `user`-tier exempt; **cascade delete gated while plain delete + MCP-cascade stay ungated**; canonical routes ungated on a paused project; `not_enrolled` reason_code via a real `claude-projects` scan row.

- Full `pytest -m "not ollama"`: **5884 passed**. The one unrelated failure, `test_calibrate_portfolio::test_collect_tagged_chunks_counts_match_corpus`, is a live-`corpus_v2`-fixture env/ordering check byte-identical to `main` (fails the same way before this change) — not caused by this PR.
- `ruff check` + `ruff format --check`: clean. `mypy`: advisory; the 2 pre-existing `_scope_to_dict` notes are untouched.

## Out of scope
CLI (`mm context update --all` already skips disabled) and MCP-tool sync parity — this PR is the **web-API** backstop.

### DIFF
--- a/packages/memtomem/src/memtomem/web/routes/context_agents.py
+++ b/packages/memtomem/src/memtomem/web/routes/context_agents.py
@@ -28,7 +28,11 @@ from memtomem.context.agents import (
 )
 from memtomem.context.detector import AGENT_DIRS
 from memtomem.context.privacy_scan import PrivacyScanError
-from memtomem.web.routes.context_projects import resolve_scope_root
+from memtomem.web.routes.context_projects import (
+    resolve_scope_root,
+    resolve_scope_root_cascade_gated,
+    resolve_writable_scope_root,
+)
 from memtomem.web.routes._locks import _gateway_lock
 
 # Flat list of project-relative runtime scan paths reported on list / import
@@ -366,7 +370,7 @@ def _safe_rel(p: Path, project_root: Path) -> str:
 async def delete_agent(
     name: str,
     cascade: bool = Query(False),
-    project_root: Path = Depends(resolve_scope_root),
+    project_root: Path = Depends(resolve_scope_root_cascade_gated),
     target_scope: TargetScope = Query(
         "project_shared",
         description="Canonical-residency tier to delete from. Non-shared tiers rejected (ADR-0011).",
@@ -470,7 +474,7 @@ class SyncRequest(BaseModel):
 @router.post("/context/agents/sync")
 async def sync_agents(
     body: SyncRequest | None = None,
-    project_root: Path = Depends(resolve_scope_root),
+    project_root: Path = Depends(resolve_writable_scope_root),
     target_scope: TargetScope = Query(
         "project_shared",
         description="Canonical-residency tier to fan out. Non-shared tiers rejected (ADR-0011).",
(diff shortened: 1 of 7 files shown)


memtomem (approved):
LGTM

pandas-studio:
Force-pushed `6bf0034` after a Codex review round. Codex flagged a **NEEDS-FIX** the earlier multi-agent review missed: `DELETE /context/{agents,commands,skills}/{name}?cascade=true` unlinks the generated *runtime* copies but was still on the ungated `resolve_scope_root`. Fixed with a new `resolve_scope_root_cascade_gated` resolver (plain delete stays ungated; cascade is gated like a sync push; MCP cascade is a v1 no-op so it stays ungated). Added cascade regression tests. Full suite green (the lone `test_calibrate_portfolio` failure is a pre-existing live-corpus env check, identical on main).

PR #1211, opened 2026-06-03
fix(web): handle structured 409 (sync paused/not-enrolled) in the UI (#1210 follow-up)

pandas-studio:
Follow-up to #1210 (§1i write-guard, backend-only). #1210 made the runtime-writing routes return a **structured** 409 for sync-ineligible projects:

```json
{"detail": {"reason_code": "sync_paused" | "sync_not_enrolled", "message": "...", "project_scope_id": "..."}}
```

The cascade-delete and settings-sync error handlers passed `err.detail` (now an **object**) straight into `showToast` / `new Error`, both of which render `textContent` — so the user saw the literal **`[object Object]`** instead of an explanation.

## What this PR does

- **`_ctxErrDetail(detail, fallback)`** (context-gateway.js) — maps `reason_code` → i18n key, else `detail.message`, else passes a plain string through unchanged (back-compat with every non-gated 4xx), else the caller fallback. A **`_hooksErrDetail`** trampoline in the decoupled `settings-hooks-watchdog.js` delegates to it behind a `typeof` guard.
- **Wires the extractor into every in-scope write error site** that reaches a #1210-gated route: cascade-delete, the four `/sync` routes, settings-sync apply/resolve/rules-delete, and the **Sync-All fan-out** (via `_ctxErrorMessageFromResponse`, so it surfaces the same localized reason the per-row Sync button does).
- **Proactive gating** — disables the cascade-**Delete** button and the Hooks **Sync-Now** button when the active project is sync-ineligible (mirrors the matrix Sync gate, same `data-i18n-title` tooltip keys). The error handling above remains the safety net for the render→paused race.
- **i18n** — `settings.ctx.error_sync_paused` / `_not_enrolled` (en + ko).
- **Cache-bust** — bumps `index.html` `context-gateway.js?v=51→52` and `settings-hooks-watchdog.js?v=13→14` so already-loaded sessions actually receive the fix.

## Scope decisions

- The structured 409's **sole source** is the `resolve_writable_scope_root` / `resolve_scope_root_cascade_gated` dependencies (`context_projects.py`). Every changed error site maps to a route that `Depends()` on them.
- **Left as plain-string handling** (cannot emit the structured 409, extractor is back-compat anyway): the GET `/api/settings-sync` reads in the watchdog, and the `POST /api/context/known-projects` add/rename routes.
- The **PUT content-save** 409 is intercepted as a merge-conflict *before* its `!r.ok` branch — not a missed site.
- The **Hooks Sync-Now** gate honors the backend's tier exemption (`target_scope != "user"`): a user-tier sync targets global `~/.claude`, not the project runtime, so it stays enabled even when the active project is paused.

⚠️ **Reviewer note (§5a):** disabling the whole **Delete** button on an ineligible scope also blocks a *canonical-only* (`cascade=false`) delete, which the backend leaves ungated. This is intentional — it matches the "this project isn't syncable" intent and the user resumes/enrolls on the Projects board — but flag if you'd prefer the error-handling-only fallback.

## Tests

- `tests/web/test_context_gateway_409_errors.py` — cascade-delete 409: paused / not-enrolled / unknown-reason→message fallback / **KO** locale / proactive-gate (paused + not-enrolled).
- `tests/web/test_settings_hooks_sync_409.py` — settings-sync 409 paused/not-enrolled + proactive gate + **user-tier-stays-enabled** regression.
- `tests-js/ctx-sync-enrollment.test.mjs` — `_ctxErrDetail` unit cases incl. the string-passthrough back-compat and the `[object Object]` negative pin.

No new routes → web-invariants registry is N/A. Verified: `pytest -m "not ollama"` **5949 passed / 11 skipped**, vitest **115 passed**, ruff clean.

This PR was reviewed by an adversarial multi-agent pass (verdict: SHIP-WITH-NITS, 0 blockers); the cache-bust, user-tier exemption, and Sync-All fan-out items above came out of that review.

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/context-gateway.js
+++ b/packages/memtomem/src/memtomem/web/static/context-gateway.js
@@ -90,11 +90,18 @@ async function _ctxErrorMessageFromResponse(resp, fallback) {
   if (contentType.includes('application/json')) {
     const err = await resp.json().catch(() => ({}));
     const detail = err.detail;
-    if (typeof detail === 'string' && detail) return detail;
-    // Defensive fallback for structured payloads shaped like `{detail: "..."}`.
+    // Defensive branch for structured payloads shaped like `{detail: "..."}`
+    // (ProjectTierBlocked / redaction) — a different shape than #1210's 409, so
+    // handle it before delegating to the shared extractor.
     if (detail && typeof detail === 'object' && typeof detail.detail === 'string' && detail.detail) {
       return detail.detail;
     }
+    // String detail, or #1210's ``{reason_code, message}`` write-guard 409, both
+    // via the shared extractor — so the Sync All fan-out surfaces the same
+    // localized paused / not-enrolled reason the per-row / per-section Sync
+    // buttons do, instead of a generic English fallback.
+    const extracted = _ctxErrDetail(detail, null);
+    if (extracted) return extracted;
   } else {
     const text = await resp.text().catch(() => '');
     if (text.trim()) return text;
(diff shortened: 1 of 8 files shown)


memtomem (approved):
LGTM

PR #1212, opened 2026-06-03
feat(context): web routes for version snapshots + label pointers (ADR-0022 PR1)

pandas-studio:
## What

PR1 of surfacing **ADR-0022** (canonical artifact version snapshots + label pointers) in the Context Portal. **Backend only** — new web routes over the already-merged `context/versioning.py` store; no UI yet (that's PR3).

```
GET    /api/context/{type}/{name}/versions        list versions + label map
POST   /api/context/{type}/{name}/versions        freeze working canonical → vN
PUT    /api/context/{type}/{name}/labels/{label}   promote (== rollback) a pointer
DELETE /api/context/{type}/{name}/labels/{label}   drop a label pointer
```

## Why these boundaries (ADR-0022)

- **agents + commands only** — skills are tree-snapshot artifacts (invariant 7) → `404`
- **directory layout only** — flat artifacts have no per-artifact version home (invariant 3) → `409` "migrate first" on mutate; the read route returns a benign `migrate_required` flag so the UI can hint instead of erroring
- **project_shared writes only** this release, mirroring the other context write routes (`_reject_non_shared_write`) → `400`
- **`latest` reserved** + version-shaped label names rejected (delegated to `versioning.py`) → `400`
- **promote == rollback** — both just move a label pointer to a frozen version

## Trust boundary

The snapshot bytes are **not** privacy-scanned at create time: a labeled sync re-scans the frozen `versions/vN.md` at deploy time (Gate A), keeping the trust boundary at fan-out exactly as the working-file path has it — so these routes are redaction-exempt by the same logic as `create_agent`. `VersionError` (a `ValueError`) is caught and mapped to accurate status codes (404/409/400) with the **same filesystem-path redaction** the app-level handler applies, so no raw path egress.

The 3 mutators are classified in `test_web_invariants_registry` (CSRF-protected + redaction-exempt with justification) — the gate every new web mutator must pass.

## Tests

22 route tests: create/promote/rollback/delete/list across agents **and** commands; skills/unknown → 404; flat → 409; non-shared → 400; reserved + version-shaped label → 400; **path-redaction on a corrupt manifest**; **TimeoutError → 503**.

- Full `pytest -m "not ollama"` green (**5962 passed**, incl. browser lane)
- `ruff check` + `ruff format --check` clean; `mypy` clean

## Follow-ups (not this PR)

- **PR2** — MCP `mem_context_version` / `mem_context_promote` (headless parity)
[...]

PR #1213, opened 2026-06-03
feat(web): detail-panel version/label manager (ADR-0022 PR3)

pandas-studio:
## What

PR3 of surfacing **ADR-0022** in the Context Portal: a **"Versions" section** in the Context Gateway detail panel for **agents + commands**. Frontend over PR1's `context_versions` routes (#1212).

The section lists immutable version snapshots (newest first), shows the label pointers (`production`/`staging`/…) that land on each, and offers:
- **Freeze current** → `POST /api/context/{type}/{name}/versions`
- per-row **Promote** (a label `<select>` + button) → `PUT /api/context/{type}/{name}/labels/{label}` — *promote and rollback are one act: both move a pointer*
- per-label **remove (×)** → `DELETE /api/context/{type}/{name}/labels/{label}`

![placement: list card → detail panel]

## Why these boundaries (ADR-0022)

- **agents + commands only** — skills are tree-snapshot artifacts (invariant 7). The hidden placeholder mounts only for versionable types, so skills/flat artifacts never flash an empty box.
- **directory layout only** — a flat artifact's versions GET returns `migrate_required`, so the section shows a *"run `mm context migrate` first"* hint instead of controls (invariant 3).
- **tier-gated** — freeze/promote/remove join `_CTX_WRITE_BUTTON_SELECTOR`, so non-`project_shared` tiers dim the buttons and the capture-phase intercept blocks the POST; `_ctxRefreshWriteBlockedState()` re-runs after the async paint.
- **race-safe** — the async paint shares `loadCtxDetail`'s `_ctxDetailSeq` guard; the mount's `seq` is *captured* (not re-read) so a post-mutation reload bails if the user has since navigated away (#972 pattern).

## i18n / CSRF

All user strings are `t()`-keyed; `en.json`/`ko.json` keysets stay identical with matching placeholders; button text + the migrate/empty hints carry `data-i18n` so a langchange re-translates them. Korean avoids josa-on-Latin on `{tag}`/`{name}`/`{label}`. CSRF threads via the inline-binding header shape (`test_web_invariants_registry` shape B).

## Tests

- **vitest** `ctx-detail-versions` (7): render + label chip, freeze POST, promote PUT, remove DELETE, flat→migrate-hint, skills→no-mount, langchange re-translate.
- **Playwright** `test_context_gateway_versions` (4, `pytestmark=browser`): section renders, freeze/promote/remove hit the right routes.
- Full `pytest -m "not ollama"` green (**5943 passed**); full **vitest 117**; ruff clean. `context-gateway.js?v` 51→52, `style.css?v` 93→94.

A read-only adversarial review (multi-agent) returned **SHIP-WITH-NITS** (0 blockers/majors); all 5 minor nits were folded in (seq-capture hardening, early seq guard, `data-i18n` on hints, langchange test).

## Follow-up (not this PR)

- **PR4** — per-list-card label chip + `GET /context/{type}?include=versions` enrichment that feeds it.
- **PR2** — MCP `mem_context_version` / `mem_context_promote` (headless parity).

Stacked on the backend routes in #1212 (no file overlap; both branch off `main`).


memtomem (approved):
LGTM

PR #1214, opened 2026-06-04
feat(web): list-card label chips + ?include=versions enrichment (ADR-0022 PR4)

pandas-studio:
## ADR-0022 PR4 — list-card label chips + `GET /context/{type}?include=versions`

Stacked on the merged ADR-0022 work (PR1 version store + routes `#1212`, PR3 detail-panel version manager `#1213`). PR3's commit explicitly deferred this slice to keep itself focused.

Brings the edit/deploy split to the **list** view: each agents/commands card shows read-only `production → v2` label chips, so a user sees which frozen version a label points at without opening the detail panel.

### Backend — opt-in enrichment
- New `include_has()` + `version_summary()` in `context_versions.py`, imported by **both** `list_agents` / `list_commands` (one implementation, no copy-paste divergence; no circular import — `context_versions` imports the `context.*` package, never the web-route modules).
- Each canonical item gains a nested `versions: {labels, count, versionable, migrate_required}`. **Opt-in**: without `?include=versions` the list path is byte-identical with zero extra I/O.
- Flat-layout artifact → `versionable:false, migrate_required:true` (inv 3). A **corrupt `versions.json` is isolated per-artifact** (`error:true`, empty labels) — one bad sidecar can't 500 the whole list, mirroring how the sync engine isolates a per-item skip.

### Frontend — read-only chips
- `renderLabelChips()` renders escaped accent pills (visually derived from PR3's detail-panel chip). Label names are **charset-unrestricted server-side**, so every interpolation is `escapeHtml`'d (load-bearing, not defensive).
- The card is `role=button`, so there is **no interactive element inside a chip** (avoids the nested-interactive a11y trap); label pointers are echoed into the card `aria-label` for SR parity (the visible `→` glyph is `aria-hidden`).
- `include=versions` requested only for agents/commands (skills out of v1, inv 7). `langchange` re-renders the list, keeping inline `t()` tooltip/aria locale-correct.
- 2 new en/ko i18n keys (josa-safe on Latin `{label}`/`{tag}`). Cache-bust `style.css 94→95`, `context-gateway.js 53→54`.

Scope: agents + commands only; chips are **read-only** — freeze/promote/remove stay in the PR3 detail panel.

### Tests
- **Backend +17**: no-include unchanged (agents **and** commands), dir/flat, labels+count, corrupt-manifest isolation w/ healthy sibling, commands, `include_has` parametrized, skills-ignores-`include`.
- **vitest +6** (`ctx-list-label-chips`): render, no-labels, aria parity, locale switch, enrichment-off, malicious-label **XSS-escape** pin.
- **Playwright +1**: chips render + `?include=versions` requested + aria parity.

Full `pytest -m "not ollama"`, full vitest (128), `ruff check`/`format --check`, and `mypy` all green. Reviewed by a 5-dimension adversarial agent pass; 4 nits found and fixed (XSS-escape pin, CSS chip-size comment, skills-ignores-include test, commands default-path test).

### DIFF
--- a/packages/memtomem/src/memtomem/web/routes/context_agents.py
+++ b/packages/memtomem/src/memtomem/web/routes/context_agents.py
@@ -33,6 +33,7 @@ from memtomem.web.routes.context_projects import (
     resolve_scope_root_cascade_gated,
     resolve_writable_scope_root,
 )
+from memtomem.web.routes.context_versions import include_has, version_summary
 from memtomem.web.routes._locks import _gateway_lock
 
 # Flat list of project-relative runtime scan paths reported on list / import
@@ -120,8 +121,17 @@ async def list_agents(
             "when explicitly requested."
         ),
     ),
+    include: str | None = Query(
+        None,
+        description=(
+            "Comma-separated optional enrichments. ``versions`` adds a per-item "
+            "``versions`` summary (label pointers + count) to feed the list-card "
+            "chips (ADR-0022 PR4); omitted by default so the list stays I/O-free."
+        ),
+    ),
 ) -> dict:
     """List canonical agents. Accepts project selector aliases like list_skills."""
+    want_versions = include_has(include, "versions")
     canonicals = list_canonical_agents(project_root, scope=target_scope)
     diffs = diff_agents(project_root, scope=target_scope)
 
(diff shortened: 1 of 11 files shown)


memtomem (approved):
LGTM

PR #1215, opened 2026-06-04
feat(context): MCP version/promote tools + label-aware sync (ADR-0022 PR2)

pandas-studio:
## ADR-0022 PR2 — MCP version/promote tools + label-aware sync

Headless-agent parity for the **edit/deploy split**. Agents driving memtomem over MCP (no shell, no web UI) now get the same freeze → promote → deploy flow the CLI `mm context version` group + `--label` and the web `context_versions` routes (PR1 #1212 / PR3 #1213) already expose. Everything reuses the pure-filesystem `context/versioning.py` store and the existing `generate_all_{agents,commands}(label=)` engine path — **no logic re-implemented**. `agents` + `commands` only (skills deferred, ADR-0022 inv 7).

### New `@register("context")` tools
core-9 default mode unchanged; reachable via `mem_do` in core mode.

- **`mem_context_version(artifact_type, name, action="list"|"create", note, scope, confirm_project_shared)`** — list versions + label pointers (read-only default) or freeze the working canonical into an immutable `versions/vN.md`. **Gate A on create** mirrors the CLI `version_create_cmd`: read once + snapshot the **same** bytes (`source_bytes=`) so a concurrent edit can't slip unscanned bytes into git-tracked `versions/vN.md`; `project_shared` hard-refuses a privacy hit *before anything lands*, `user`/`project_local` permissive.
- **`mem_context_promote(artifact_type, name, label, version, delete, scope, confirm_project_shared)`** — move a label pointer (promote == rollback) or `delete=True` to drop it.

### Deploy half (folded in)
`mem_context_sync` / `mem_context_generate` gain a **`label`** param threaded to `generate_all_{agents,commands}(label=)`:
- `""` / `"latest"` → the working file (byte-for-byte today's behavior — no caller changes);
- a label or bare version tag → fans out the **frozen** bytes for agents/commands;
- ineligible included kinds run label-less with a note (inv 10); an unknown/dangling label or flat-layout artifact isolates as a per-artifact **skip**, never a whole-run error;
- `mem_context_diff` stays label-unaware (inv 8).

Full headless flow: **`mem_context_version`** (freeze) → **`mem_context_promote`** (label) → **`mem_context_sync(label=)`** (deploy).

### Boundaries
Flat layout — `list` hints, `create`/`promote` refuse (inv 3); reserved `latest` + version-shaped label names rejected by the shared store; explicit `project_shared` writes need `confirm_project_shared` (git-tracked tree, MCP can't prompt — mirrors `mem_context_init`); the implicit default scope stays frictionless.

### Review
An adversarial multi-lens review caught + fixed a **blocker**: the originally-added `mem_do` alias `"version"` shadowed the existing `mem_version` action (memtomem-stm protocol negotiation), because `mem_do` resolves aliases before real actions. Dropped that alias (kept the collision-free `"promote"`), added a guard test that no alias key may shadow a registered action, plus a regression pin that `mem_do("version")` still returns the protocol JSON.

### Tests
- `test_server_tools_context_version.py` — 32 real-FS pins: list/create/promote/delete, flat-layout, privacy block + user-tier permissive, confirm gate, `mem_do` routing, and label-aware deploy (frozen-not-working / latest / unknown-label skip / ineligible-kind note / generate-by-version-tag).
- `test_meta_tool.py` — alias-collision guard + protocol-negotiation pin.
- Full `pytest -m "not ollama"` green (6003 passed). The 12 `test_web_cli` failures are a pre-existing stale `web.pid` in the local env, reproduced on the base branch — not introduced here.

### DIFF
--- a/packages/memtomem/src/memtomem/server/__init__.py
+++ b/packages/memtomem/src/memtomem/server/__init__.py
@@ -132,6 +132,8 @@ from memtomem.server.tools.context import (
     mem_context_memory_migrate,
     mem_context_migrate,  # deprecated alias for mem_context_memory_migrate (#1147 B5-2)
     mem_context_artifact_migrate,
+    mem_context_version,  # ADR-0022 PR2 — version snapshots (list/create)
+    mem_context_promote,  # ADR-0022 PR2 — label pointers (promote/delete)
 )
 from memtomem.server.tools.ingest import (
     mem_ingest,  # no @mcp.tool; import triggers @register("ingest") for mem_do routing
--- a/packages/memtomem/src/memtomem/server/tools/context.py
+++ b/packages/memtomem/src/memtomem/server/tools/context.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, cast
 import click
 
 from memtomem.config import TargetScope
+from memtomem.context import versioning
 from memtomem.context.scope_resolver import find_project_root
 from memtomem.server import mcp
 from memtomem.server.context import CtxType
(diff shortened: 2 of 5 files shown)


memtomem (approved):
LGTM

PR #1218, opened 2026-06-04
feat(web): enable versioning for flat web-created artifacts (ADR-0022 rank 6)

pandas-studio:
## What

Web-created agents/commands are written **flat** (`.memtomem/agents/<name>.md`) with no per-artifact directory, so they can't carry a `versions/` store and were **permanently locked out of ADR-0022 versioning**. The version section's `migrate_required` hint pointed at `mm context migrate` — but that command **provably skips these files**: a flat file with no lockfile entry classifies as `skip_manual` ("resolve manually") because it sits outside the wiki install/upgrade lifecycle. The named escape hatch refused the very files it named. (Context Gateway UX audit, **rank 6 [P1]**.)

This adds an explicit **"Enable versioning"** action that adopts a flat canonical into directory layout, after which the existing version routes work normally.

## Changes

- **`context/migrate.py`** — `adopt_flat_to_dir()`: a single byte-identical `os.replace` of `<type>/<name>.md` → `<type>/<name>/<manifest>`, **bypassing the lockfile-lifecycle gate** (install provenance is irrelevant to giving a file a directory home), with a path-escape guard and a flat+dir collision refusal.
- **`web/routes/context_versions.py`** — `POST /context/{type}/{name}/versions/enable`: `project_shared`-only (canonical-write tier policy), gateway-lock-held, **idempotent** on dir layout (`migrated: false`), path-redacted `404/409/400/503` mapping. Classified in the web-invariants registry (CSRF + redaction-exempt).
- **`context-gateway.js`** — the `migrate_required` branch now renders an **"Enable versioning"** button (write-gated, CSRF-threaded, seq-guarded); on success it reloads the whole detail panel so the layout / file-count chips refresh too. Cache-bust `?v=55 → 56`.
- **locales (en/ko)** — 3 new keys + revised the misleading `migrate_required` copy (it pointed at the dead-end command).

## Design notes

- **ADR-0022 invariant 3 preserved**: `version create` still **refuses** a flat artifact — "enable" is a separate, deliberate action, *not* auto-adopt.
- **No privacy re-scan** in adopt: the rename is byte-identical and stays in-scope; a labeled sync still re-scans the frozen `versions/vN.md` at deploy time (Gate A).
- Write-path 409 + docstrings now surface **both** affordances (enable for UI-created, `mm context migrate` for CLI-installed).
- **Scope**: web-only this PR. CLI/MCP `version create` share the same flat dead-end; the reusable `adopt_flat_to_dir` core makes that a small follow-up.

## Tests

`adopt_flat_to_dir` unit tests (incl. the no-lockfile `skip_manual` dead-end it unblocks) · the enable web route (flat→dir, idempotent, commands, 404/400) · a browser test for the button render + POST. **5839 backend + 6 version-browser + 128 vitest pass**; ruff + mypy clean. (The 12 `test_web_cli.py` failures locally are the known live-server pidfile flake, not a regression.)

Reviewed by a multi-agent adversarial pass; the one confirmed finding (stale `mm context migrate` guidance on the write-path 409) is fixed in this PR.


memtomem (approved):
LGTM

PR #1219, opened 2026-06-04
feat(web): rank-10 Sync/Import confirm previews — dry-run + count/destination threading

pandas-studio:
## What & why

Rank 10 of the Context Gateway UX audit: Sync / Sync-All / Import confirmed with **generic, count-less prose** and only revealed *what* and *where* they touched in a post-hoc toast — users committed blind, and Import never named its destination tier/project. This threads the known context into the confirm dialog **before** the user commits.

## Frontend — confirm threading (`context-gateway.js`, `?v` 55→56)

| Action | Before | After |
|---|---|---|
| **Section Sync** | "Fan out skills to all runtimes?" | "Fan out **2** skills to all runtimes?" (count from the section dataset — no fetch) |
| **Sync-All / portal Sync** | "Fan out all canonical artifacts to all runtimes?" | "Fan out all canonical artifacts in **{project}** to all runtimes?" |
| **Import** | "Import skills from runtimes into .memtomem/?" | "Import skills from runtimes into **{project} · Project (shared)**? **2 to import, 1 already exist.** Enable Overwrite to replace the existing ones." |

Import fetches a **dry-run preview** (`POST .../import?dry_run=1`) before the modal so it can name the destination + would-import/skip counts. It's **best-effort** — any failure (offline, lock contention, malformed body) falls back to a destination-only confirm, never blocking Import. When some artifacts already exist, the count is tied to the existing Overwrite checkbox so "already exist" doesn't read as "will be skipped".

## Backend — `?dry_run=1` preview

`extract_{skills,commands,agents}_to_canonical` gain a keyword-only `dry_run` flag that runs the **full** scan + name validation + Gate A privacy walk + cross-runtime dedup + canonical-exists check, then **skips only the final write** (`copy_skill` / `atomic_write_*`). Because the skip decision evaluates `dst.exists()` before any write in both modes, the preview is **accurate** without mutating disk.

The three **bulk** import POST routes expose `dry_run` as a query param and echo a `dry_run` flag in the response. No new route → the CSRF/redaction invariants registry is unaffected (verified). Single-item import routes are untouched.

## Tests

- Backend: `dry_run` extract tests (skills/commands/agents — both command write-branches) assert the would-import set is reported with **nothing written to disk**, and a real run afterwards still writes; route-level test pins `?dry_run=1` → `dry_run:true` + disk untouched.
- Frontend: new browser suite (`test_context_gateway_rank10_confirm.py`) pins all four confirm sites — section-sync count, import destination + dry-run preview + overwrite hint, best-effort fallback, and per-project portal-sync project name.
- i18n en/ko key + placeholder parity kept; new `confirm_import_preview` / `confirm_import_overwrite_hint` keys.

## Verification

- Full `pytest -m "not ollama"`: **6019 passed**, 0 failures (excluding `test_web_cli.py`, which fails only because a live local `mm web :8080` server holds the pidfile — documented env flake, unrelated to this change).
- ruff check + format clean; mypy clean on changed source.
- Reviewed via a multi-agent adversarial pass (backend dry-run: all-clear; i18n: all-clear); the two confirmed frontend findings (overwrite-count clarity, malformed-preview hardening) are addressed in the diff.


memtomem (approved):
LGTM

pandas-studio:
**Codex review (dev-trio): NEEDS-FIX → 0 blockers, 2 major — both addressed** (force-pushed in `a725e4f`).

- **① Skip-reason conflation** — the import preview counted *every* `skipped` row as "already exist", but the backend `skipped` list also carries `already_imported` (cross-runtime dedup), `invalid_name`, parse errors, and privacy blocks. The preview now counts **only `reason_code === 'canonical_exists'`** toward "already exist" and the Overwrite hint; the other skips still surface in the post-import result panel. New browser test pins it (`test_import_confirm_counts_only_canonical_exists_as_already_exist`).
- **② Scope/tier drift** — the preview POST, the destination label, and the final import POST all read live `_ctxActiveScopeId`/`_ctxTargetScope`; the gateway UI isn't inert until the modal opens, so a mid-flight active-project change could make them disagree. Effective `scope_id`+tier are now **snapshotted once at click time** and reused for all three — the same one-(project, tier)-per-invocation pinning Sync All enforces (ADR-0016 §5).

Codex independently verified the backend dry-run guards (all write sites incl. both command branches), confirmed the preview is accurate vs a real run, and checked i18n placeholder parity — all clear. This matches the earlier multi-agent adversarial pass (backend + i18n all-clear).

PR #1003, opened 2026-05-13
Fix Home pinned chunk navigation

memtomem:
## Summary
- make Home pinned chunk rows focusable and activatable by mouse or keyboard
- resolve pins via GET /api/chunks/{id}, then navigate to the source with the chunk selected
- mark missing chunks as stale and let users remove stale pins without navigating

Fixes #991

## Tests
- uv run pytest packages/memtomem/tests/web/test_home_pinned_chunks.py
- uv run pytest packages/memtomem/tests/test_i18n.py
- uv run ruff check packages/memtomem/tests/web/test_home_pinned_chunks.py
- git diff --check

PR #1220, opened 2026-06-04
fix(web): a11y — settings-nav→nav list, un-nest project-remove ×, label Hooks controls (rank 14/15/22)

pandas-studio:
## What

Context Gateway UX-audit follow-up — three **P2 frontend-only** a11y/correctness ranks. No backend or route changes.

### rank 14 — sidebar ARIA contract → navigation list
The Gateway + Settings/More sidebars (`.settings-nav`, shared) declared `role=tablist` with `role=tab` buttons but **no `aria-controls`, no `tabpanel` roles, and no arrow-key nav** — a broken tab contract that announced *"tab, N of M"* with no associated panel. The sidebars also interleave **interactive collapsible group-header buttons**, which a strict tablist must not contain, so "completing" the tab pattern wasn't achievable cleanly.

**Fix (chosen direction):** demote to a navigation list — `<nav aria-label>`, drop `role=tab`/`aria-selected` from the 17 nav buttons, mark the active item with `aria-current="page"` (managed by `switchSettingsSection`). The four **real** tablists (top tab-nav, sources vendor, index-mode, detail sub-tabs) are untouched.

### rank 15 — un-nest the destructive project-remove ×
The × was interpolated **inside** the `<summary>` activation control (nested-interactive antipattern, #1003) with only a mouse-path `stopPropagation` guard. It's now a **sibling of `<details>`**, wrapped in `.ctx-scope-group-wrap` and pinned over the summary's right edge by CSS. It can't be a plain `<details>` child — native `<details>` hides non-`<summary>` children when collapsed (× would vanish) — so the wrapper keeps it always visible and a standalone keyboard-focusable button.

| collapsed | open |
|---|---|
| × stays visible (native `<details>` would otherwise hide it) | × pinned top-right, no badge overlap |

### rank 22 — label the Hooks control row
Hooks inherited the artifact-section project switcher + tier toggle but rendered them bare above a single "Sync Now" button, reading as accidental chrome. Both controls are functional (they pick which canonical `settings.json` is read/synced), so they're kept but wrapped in a labeled `.hooks-controls` row with a caption naming that effect.

## Tests
- New source-pin contracts in `test_web_a11y.py` (8) — lock each rank's contract so a future edit can't silently restore the broken patterns.
- `vitest` **128/128**, browser suite **165/165** (gateway · portal · ui_smoke · hooks · a11y · i18n), i18n parity (en/ko, 3 new keys), `ruff` clean.
- rank 15 × geometry visually verified against the real `style.css` (collapsed / open / non-removable Server-CWD; no badge overlap).
- Cache-bust bumped for the 4 edited static assets (style 97 / app 124 / hooks 15 / gateway 56).

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -1462,14 +1462,19 @@ function switchSettingsSection(sectionName) {
   try { localStorage.setItem(LAST_SECTION_KEY, sectionName); } catch {}
   document.querySelectorAll('.settings-nav-btn').forEach(b => {
     b.classList.remove('active');
-    b.setAttribute('aria-selected', 'false');
+    // ``.settings-nav`` is a navigation list (role-less <nav>), not a tablist
+    // (rank 14): the group headers are interactive collapse toggles, which a
+    // strict ARIA tablist may not contain. Mark the active entry with
+    // ``aria-current="page"`` — the standard nav-selection cue — instead of
+    // ``aria-selected``, which only belongs on a role=tab.
+    b.removeAttribute('aria-current');
   });
   document.querySelectorAll('.settings-section').forEach(s => s.classList.remove('active'));
   const btn = document.querySelector(`.settings-nav-btn[data-section="${sectionName}"]`);
   const section = document.getElementById(`settings-${sectionName}`);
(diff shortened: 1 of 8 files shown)


memtomem (approved):
LGTM

PR #1221, opened 2026-06-04
refactor(web): hoist active-project + tier controls to one gateway header bar (rank 11)

pandas-studio:
## Context Gateway UX audit — rank 11

The active-project `<select>` and the canonical-tier filter were re-emitted into **6 of the 7** gateway sections (Overview, Skills, Commands, Agents, MCP, Hooks) — the same picker painted six times for one piece of module-global state (`_ctxActiveScopeId` / `_ctxTargetScope`). This hoists them into **one persistent host `#ctx-control-bar`** above the section panels.

### How it works
- **`_ctxRenderControlBar()` always self-sources the type** from the active gateway section (`.settings-section.active`), never a caller-supplied value. The bar is a single shared, *visible* host and the loaders that repaint it are async, so a stale `loadCtxOverview`/`loadCtxList` that resolves after the user navigated away paints the bar for the section they're **actually** on (or hides it on the Projects portal) instead of hijacking it back and mis-routing the next tier/project change.
- `switchSettingsSection` repaints it synchronously on every gateway switch; each loader + the `langchange` listener repaint it too (the latter covers `hooks-sync`, which has no per-section re-issue branch).
- A **Sync All lock** now lives in a module flag (`_ctxSyncControlsLocked`) re-applied after every repaint, so navigating / a loader / a locale flip can't silently re-enable the shared controls mid-run.
- The deep-link banner keeps the tier-aware write-blocked banner (#943) at the very top, mirroring the runtime-only remediation banner's anchoring.

### Supersedes rank 22 (#1220)
#1220 kept the Hooks controls in-section, wrapped in a labeled `.hooks-controls` row + caption. Hoisting them out leaves nothing there to wrap, so this removes the now-dead `_hooksControlsHtml` helper, the `.hooks-controls` / `.hooks-controls-caption` CSS, the `settings.hooks.controls_caption` i18n key (en/ko), and re-points the two rank-22 a11y source pins to assert the hoist. The tier still scopes which canonical `settings.json` is read/synced — that function is preserved, just driven from the shared bar. **rank 14** (nav landmark) and **rank 15** (project-remove × un-nest) from #1220, and **rank 6** (#1218) version manager, are untouched (rebased onto both).

### Review
Two adversarial review passes (multi-agent in-house workflow + Codex):
- **BLOCKER** (in-house): a stale async overview render could hijack the shared bar's section → fixed by making `_ctxRenderControlBar()` self-source the active section.
- **Major** (Codex): Sync All lock lost on repaint → fixed with the module-flag lock re-applied per repaint.
- **Minor** (Codex): deep-link banner could land above the write-blocked banner → fixed with the established anchor pattern.

### Tests (all green)
- vitest **136** (new `ctx-control-bar-hoist.test.mjs`, 8 cases: render-once, hide-on-projects, loader-routing, stale-render-no-hijack, hooks-sync langchange, sync-lock persistence)
- browser gateway/portal/hooks **129**; retargeted 8 Playwright tier-filter selectors `#ctx-skills-list` → `#ctx-control-bar`
- full non-browser pytest **5850 passed** (the only failures are 12 `test_web_cli.py` — a local flake from a live `mm web` server holding the pidfile, not this change)
- `test_web_mode` / `test_web_a11y` / `test_i18n` source pins updated; ruff clean; `?v` cache-busts bumped above main (style 98 / app 125 / hooks 16 / gateway 58)

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/app.js
+++ b/packages/memtomem/src/memtomem/web/static/app.js
@@ -1478,6 +1478,16 @@ function switchSettingsSection(sectionName) {
   }
   if (section) section.classList.add('active');
   ensureActiveGroupExpanded(sectionName);
+  // rank 11: repaint the persistent gateway control bar synchronously on every
+  // gateway section switch. Its visibility (hidden on the Projects portal) and
+  // selection must update the instant the section flips — the async loader below
+  // also repaints it post-fetch, but this synchronous call avoids a stale or
+  // wrong-section bar in the interim (and hides it immediately on ctx-projects,
+  // which has no loader that touches the bar). Guarded for non-gateway sections
+  // and for script load order (context-gateway.js defines _ctxRenderControlBar).
+  if (GATEWAY_SECTIONS.has(sectionName) && typeof _ctxRenderControlBar === 'function') {
+    _ctxRenderControlBar();
+  }
   // Section-specific loads (reuse existing functions)
   if (sectionName === 'config') loadConfig();
   if (sectionName === 'namespaces') loadNamespacesTab();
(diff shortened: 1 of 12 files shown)


memtomem (approved):
LGTM

PR #1223, opened 2026-06-04
refactor(web): unify artifact-section toolbars into one capability-mapped source (rank 21)

pandas-studio:
## What

The Skills / Commands / Agents / MCP Servers section toolbars were four hand-copied blocks of inline button markup in `index.html`. The divergence (MCP drops Import) was structural and drift-prone — a button added to one section wouldn't propagate, and MCP's missing Import read as an accidental gap rather than a declared capability. (Context Gateway UX audit, **rank 21 [P3/M]** — maintainability/consistency.)

## How

- **One source.** `_ctxToolbarHtml(type)` + `_CTX_TOOLBAR_CAPS` in `context-gateway.js` emit the buttons. `import: false` for `mcp-servers` makes the omission an **explicit capability flag** — there is no per-type `/import` route (servers come from the single `.mcp.json` source; `context_mcp_servers.py` ships no import endpoint).
- **Containers.** `index.html` keeps an empty `<div class="ctx-toolbar" data-type="…">` per section; `_ctxRenderToolbars()` fills them once at init.
- **Zero behavior change.** Buttons carry the exact classes / `data-type` / `data-i18n*` the static markup used, so the existing click bindings, the write-block sweep (`_CTX_WRITE_BUTTON_SELECTOR`), and i18n `applyDOM` are unchanged. Rendered before the bindings and before `DOMContentLoaded`'s first `applyDOM`, so they behave like the old static buttons (English fallback text, translated on the first pass). The `.ctx-toolbar` div inherits the existing `.ctx-header div` / `.ctx-header > div` flex CSS — no CSS change.
- **Hooks untouched** — a structurally distinct single-button header, out of scope for the artifact-toolbar unification.

## Scope decision

The audit's optional "one-line MCP note" is **not** added: rank 7's empty-state hint + the section desc (".mcp.json") already cover the user-facing side, so a separate always-visible note would duplicate it. The capability map's code comment documents the omission for maintainers. Trivial to add if preferred.

## Tests

- New `tests-js/context-toolbar.test.mjs` (3): MCP omits Import while skills/commands/agents include it; Sync stays rightmost/primary; the hyphen→underscore i18n-key bridge holds.
- Full vitest (131) + browser `test_context_gateway_write_blocked.py` (8) + `test_i18n.py` (63): green.
- Full `pytest -m "not ollama"`: **6060 passed** (the 12 `test_web_cli.py` failures are the local live-`mm web`-server pidfile contention, green in CI).
- Codex review: **SHIP** (0 findings).

Cache-bust: `context-gateway.js` v58 → v59.

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/context-gateway.js
+++ b/packages/memtomem/src/memtomem/web/static/context-gateway.js
@@ -640,9 +640,10 @@ function _ctxWireShowAllScopes(type, listEl) {
 //       and fires a toast — the per-button handler never sees the event,
 //       so no POST is ever issued.
 //
-// Per-section buttons (.ctx-create-btn / .ctx-import-btn / .ctx-sync-btn)
-// live in the static HTML so the refresh applies on every render that
-// touches the tier filter; per-item buttons (.ctx-detail-edit-btn /
+// Per-section toolbar buttons (.ctx-create-btn / .ctx-import-btn /
+// .ctx-sync-btn) are generated once at init by ``_ctxRenderToolbars`` (rank 21)
+// and then persist in the DOM, so the refresh still applies on every render
+// that touches the tier filter; per-item buttons (.ctx-detail-edit-btn /
 // .ctx-detail-delete-btn) are minted by ``loadCtxDetail`` so its callers
 // reapply the refresh after the detail innerHTML lands.
 //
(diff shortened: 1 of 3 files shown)


memtomem (approved):
LGTM

Issue #1229, opened 2026-06-10
Context Gateway full-stack review (2026-06-10): verified findings catalog


memtomem:
## Context Gateway full-stack review — verified findings catalog (2026-06-10)

Method: 12-lens multi-agent static review (backend, routes, frontend logic/state,
API contract, UX/IA, i18n, ADR conformance) + hands-on Playwright walkthrough.
Every finding below survived independent adversarial verification (2 verifiers:
refutation lens + manifestation-path lens; UX items: accuracy+impact judge).
The copy/terminology and diagnostics packages already landed as #1227 / #1228.

**23 verified findings remain open** (1 critical / 12 major / 10 minor).
High-priority items should be split into individual tracker issues as they are picked up.

#### [CRITICAL] Kimi managed-block replacement uses re.sub with an unescaped template — backslashes/newlines in hook commands corrupt config.toml on every re-sync
- `context/settings.py:698` (backend-settings)
- _replace_kimi_managed_block calls `pattern.sub(block, existing)` with the freshly rendered managed block as the replacement STRING, so Python's replacement-template escape processing is applied to the block text. Hook commands are TOML-encoded via json.dumps, so a command containing a backslash (any regex like grep '\bfoo', awk, sed scripts, Windows paths) produces `\\` sequences in the block, which re.sub halves to a single `\` — and a command containing a newline produces a literal `\n` that re.sub converts to a real newline inside the TOML basic string. The FIRST sync (no existing block; pl
- **Suggested fix:** Replace via a callable so no template processing occurs: `pattern.sub(lambda m: block, existing)` (or `re.escape`-free `block.replace('\\', '\\\\')` on the replacement). Add a round-trip test: render → replace → tomllib.loads for commands containing backslashes, quotes, and newlines.

#### [MAJOR] diff_agents/diff_commands key artifacts by file stem while sync targets the frontmatter name — permanent phantom drift after a successful sync
- `context/agents.py:971` (backend-artifacts)
[...]

PR #1232, opened 2026-06-10
fix(web): per-name diff routes resolve runtime targets at the requested tier (#1229)

memtomem:
## Summary

Fixes the #1229 verified finding (routes-api, **major**): *Per-name /diff endpoints ignore `target_scope` when resolving runtime targets — detail diff contradicts the list view on user/project_local tiers.*

The per-name `GET /api/context/{skills,commands,agents}/{name}/diff` routes resolved the **canonical** side scope-aware (`_canonical_skill_dir(..., scope=target_scope)` etc.) but probed every **runtime** target via `gen.target_dir(project_root, name)` / `gen.target_file(project_root, name)`, which default to `scope="project_shared"`:

- On the **user** tier the detail panel compared user-tier canonical content against the project's `.claude/...` files and reported phantom `missing target` / `out of sync` / `missing canonical` rows that contradict the scope-aware list diff.
- On **project_local** (a NO_FANOUT tier per ADR-0011 §3) it fabricated runtime comparisons that cannot exist.

## Change

Pass `scope=target_scope` through to the generator target resolution in the three per-name diff handlers. NO_FANOUT tiers then return `None` and are skipped by the existing guard — matching the engine diff (`diff_skills` / `diff_commands` / `diff_agents`) that the list endpoints use.

Same-shape audit: the remaining scope-less `gen.target_*` calls in these route files are the cascade-delete paths, which are gated by `_reject_non_shared_write`, so `project_shared` (the default) is the only tier that can reach them — left unchanged deliberately.

## Tests

- 6 new regression tests; all fail pre-fix with phantom `missing target` rows and pass post-fix:
  - `test_diff_user_tier_resolves_user_runtime` (skills / commands / agents): user-tier canonical + identical `~/.claude/...` runtime copy → claude row reports `in sync`.
  - `test_diff_project_local_has_no_runtime_rows` (skills / commands / agents): project_local canonical → `runtimes == []`.
- `test_web_routes_context.py`: 101 passed. `ruff check` + `ruff format --check`: clean.

Part of #1229.

### DIFF
--- a/packages/memtomem/src/memtomem/web/routes/context_agents.py
+++ b/packages/memtomem/src/memtomem/web/routes/context_agents.py
@@ -480,7 +480,9 @@ async def diff_agent(
 
     runtimes = []
     for gen_name, gen in AGENT_GENERATORS.items():
-        target = gen.target_file(project_root, name)
+        # Match the canonical side's tier — see context_skills.py diff_skill
+        # for the rationale (#1229). NO_FANOUT tiers return None → skipped.
+        target = gen.target_file(project_root, name, scope=target_scope)
         if target is None:
             continue
         if canonical_content is None and not target.is_file():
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #1233, opened 2026-06-10
fix(context): key diff canonicals by frontmatter name and survive undecodable files (#1229)

memtomem:
## Summary

Fixes two #1229 verified findings (backend-artifacts, both **major**) in the artifact diff engine, plus one same-family guard named in the finding's evidence.

**1. Stem-vs-frontmatter-name phantom drift.** The sync engine targets the parsed frontmatter `name:` (`_sync_atomic` Phase 1 → `name_of`), but `diff_agents` / `diff_commands` keyed their canonical index by file stem / dir name. A canonical whose stem differs from its frontmatter name (e.g. `.memtomem/agents/reviewer.md` containing `name: code-reviewer`) synced correctly, then reported two phantom rows per runtime — `('reviewer', 'missing target')` + `('code-reviewer', 'missing canonical')` — forever. Re-syncing never cleared them, and the bogus `missing canonical` row invited an import that would duplicate the artifact under a second name.

**2. One non-UTF-8 byte aborted the whole diff.** The same functions read canonical and runtime files with strict UTF-8; a single stray latin-1 byte raised an uncaught `UnicodeDecodeError` (only `AgentParseError` / `CommandParseError` were caught), turning the per-type list routes into 500s and `mm context diff` into a traceback — losing status for **all** artifacts — while sync decodes the same bytes with `errors="replace"` and succeeds.

## Change

- `diff_agents` / `diff_commands` now parse each canonical **once up front** (bytes → decode `errors="replace"` → `_parse_canonical_*_text`) and key the index by the parsed name — the same identity derivation sync uses. Unreadable/unparseable canonicals keep the stem / dir-name key so their `parse error` row still names the file (existing wire statuses unchanged).
- Runtime-side reads use the same lenient decode; an undecodable byte now degrades to a per-artifact `out of sync` (matching what sync actually wrote). `OSError` on a runtime read also reports `out of sync` — parity can't be asserted, so report drift, never mask it (the existing unreadable-override contract).
- Same-family guard (named in the finding's evidence): `diff_skills` no longer lets a `PermissionError` from `_skill_effective_equal`'s `read_bytes` abort the whole skills diff — the comparison is guarded per artifact with the same drift-not-crash contract.

Side benefit: each canonical is parsed once per diff call instead of once per generator.

## Tests

- 7 new regression tests; all fail pre-fix (phantom rows / `UnicodeDecodeError` / `PermissionError`) and pass post-fix:
  - `test_frontmatter_name_wins_over_stem_after_sync` (agents / commands): sync then diff → single name, all `in sync`.
  - `test_non_utf8_canonical_byte_does_not_abort_diff` (agents / commands): `0xe9` byte in canonical → diff agrees with what sync wrote.
  - `test_non_utf8_runtime_byte_reports_drift_not_crash` (agents / commands): corrupted runtime copy → `out of sync`, others untouched.
  - `test_unreadable_runtime_file_reports_drift_not_crash` (skills, POSIX non-root only): chmod-0 runtime manifest → `out of sync`, no crash.
- Full suite (`-m "not ollama"`): 6089 passed; the only failures are pre-existing `test_session_tracing` config-rollback cases that reproduce identically on a clean `origin/main` checkout (environment-dependent, unrelated to this change). `ruff` clean; `mypy` on the three changed files: clean.

Part of #1229.

### DIFF
--- a/packages/memtomem/src/memtomem/context/agents.py
+++ b/packages/memtomem/src/memtomem/context/agents.py
@@ -988,10 +988,26 @@ def diff_agents(
     pre-PR-E3 behavior.
     """
     results: list[tuple[str, str, str]] = []
-    canonical_index = {
-        path.parent.name if layout == "dir" else path.stem: (path, layout)
-        for path, layout in list_canonical_agents(project_root, scope=scope)
-    }
+    # Key canonicals by the parsed frontmatter ``name:`` — the identity sync
+    # targets (``_sync_atomic`` Phase 1: bytes → decode errors="replace" →
+    # parse → ``name_of``). Keying by file stem reported permanent phantom
+    # drift after a successful sync whenever stem and frontmatter name
+    # disagree: ('<stem>', 'missing target') + ('<name>', 'missing
+    # canonical') on every runtime, forever (#1229). The lenient decode also
+    # keeps a stray non-UTF-8 byte from aborting the whole diff with an
+    # uncaught UnicodeDecodeError while sync handles the same file fine.
+    # Unreadable / unparseable canonicals keep the stem / dir-name key
+    # (``None`` value) so their "parse error" row still names the file.
+    canonical_index: dict[str, SubAgent | None] = {}
+    for path, layout in list_canonical_agents(project_root, scope=scope):
+        fallback_name = path.parent.name if layout == "dir" else path.stem
+        try:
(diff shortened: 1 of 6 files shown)


pandas-studio (approved):
LGTM

PR #1245, opened 2026-06-11
ux(context): give parse-error and invalid-name rows a diagnostic surface (#1229)

memtomem:
U7 from the Context Gateway review catalog (#1229) — the last open finding.

## "Parse error" told you something broke, never what or where

The diff layer emitted the bare status string with no message — in `diff_agents`/`diff_commands` the exception was swallowed **without even logging** — and the UI faithfully rendered a red badge with no file path, no cause, no fix guidance on tile, list card, or diff pane. Verification also found a fresh contradiction: the per-name diff endpoints never re-parsed the canonical (raw string compare), so a malformed canonical rendered "out of sync"/"in sync" in the pane while the list badge said "parse error".

## Engine: `DiffRow` — diagnostics without an API break

A tuple **subclass** that iterates, compares, and hashes as the historical `(runtime, name, status)` 3-tuple while carrying a `.reason` attribute (plus `__getnewargs__` for pickle round-trips). Every existing positional consumer — CLI printers, MCP tools, web groupers, external callers of the public `diff_*` functions — keeps working unchanged; the Codex design gate weighed NamedTuple-widening vs a parallel detailed API, and the subclass beats both (no breaking change, no drift-prone second API). Reasons are captured at the existing catch sites (exception text embeds the source path; the agents/commands sites also gain the missing `logger.warning`), invalid-name rows (#1243) carry the `validate_name` message via the listing's new `(raw_name, reason)` pairs, and `diff_mcp_servers` distinguishes canonical-parse from broken-`.mcp.json` — which marks *every* canonical row, so the reason must name the file that actually broke.

## Wire: additive, sanitized at the boundary

List endpoints emit `runtimes[]` entries with an optional `reason`, sanitized by a shared `sanitize_diff_reason` (project-root prefix stripped from **embedded** message text — engine reasons are arbitrary strings, not bare paths — then the established HOME-collapse + secret-shape whole-replace + truncation). Per-name diff endpoints for agents/commands now **re-parse** the canonical and emit per-runtime parse-error rows with reason + runtime preview, closing the pane-vs-badge contradiction; all three per-name payloads (agents, commands, mcp-servers) gain `canonical_path` for the fix-it hint. All content reads in those routes go through a tolerant `read_text_lenient` (non-UTF-8 → U+FFFD per the #1233 contract; `PermissionError` → diagnosable "unreadable" rows instead of a 500 — per Codex impl review). The overview tile stays count-only per ADR-0009 §2 / #1228 — leaf-first diagnosis, intentionally preserved.

## UI + CLI + MCP: 5-surface parity

The diff pane and runtime-only detail render a diagnostic block under broken badges (escaped reason + "Fix the file at {path}, then click Refresh."); list-card badge tooltips carry the reason; `mm context diff` and `mem_context_diff` print a `— reason` suffix. en/ko keys under the parity gate; `?v` bumps context-gateway.js 68 / style.css 103.

## Reviews

Design gated by Codex before implementation (NEEDS-FIX → adopted: embedded-path sanitizer, `canonical_path` on all three payloads, exact mcp reason assignment, runtime-only vitest, lenient-read criteria; its per-name scope blocker was stale — #1232 already passes `scope=target_scope`, verified). Implementation reviewed by Codex (NEEDS-FIX → both Majors fixed in the follow-up commit: pickle support, tolerant reads).

## Tests

Engine reason pins (agents/commands/mcp, healthy-row `None` negatives, the 3-tuple/equality/pickle compat pin), route pins (sanitized reasons on list entries, per-name re-parse + `canonical_path` + no-abs-path-leak, unreadable-canonical 200, mcp canonical-vs-`.mcp.json` reasons), i18n key pin, and a new vitest (`ctx-parse-error-diagnostics.test.mjs`: reason+hint rendering, XSS escape, healthy/empty negatives, badge tooltip). Full dual gate green: 446 + 219 web pytest, 151 vitest.

Closes out the #1229 catalog. 

### DIFF
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -235,9 +235,14 @@ def _print_skills_diff(root: Path, *, scope: TargetScope = "project_shared") -> 
     if not rows:
         click.echo(f"  (no skills to compare in {scope})")
         return
-    for runtime, name, status in rows:
+    for row in rows:
+        runtime, name, status = row
         color = "green" if status == "in sync" else "yellow"
-        click.secho(f"  {runtime:15s}  {name}  [{status}]  scope={scope}", fg=color)
+        # Diagnostic reason (parse error / invalid name rows, #1229 U7) —
+        # raw engine text is fine for the local operator.
+        reason = getattr(row, "reason", None)
+        suffix = f"  — {reason}" if reason else ""
+        click.secho(f"  {runtime:15s}  {name}  [{status}]  scope={scope}{suffix}", fg=color)
 
 
 # ── Sub-agent sub-handlers (Phase 2) ─────────────────────────────────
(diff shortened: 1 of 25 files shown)


pandas-studio (approved):
LGTM

Issue #1249, opened 2026-06-11
test_session_tracing fails on machines with real session_trace config (Mem2MemConfig() reads unisolated ~/.memtomem/config.json)


memtomem:
## Symptom

On any machine whose real `~/.memtomem/config.json` contains

```json
{ "session_trace": { "enabled": true, "langfuse_enabled": true } }
```

four tests in `packages/memtomem/tests/test_session_tracing.py` fail (`TestConfigConstraints::test_langfuse_validator_keys_missing` plus the three `TestConfigSaveValidationAndRollback` rollback tests), e.g.:

```
>       assert app_mock.config.session_trace.langfuse_enabled is False
E       AssertionError: assert True is False
```

CI never sees this (clean HOME), so it only bites contributors who actually use session tracing. Verified with the same failure on both current `main` (01b5ad28) and 66f251ac — long-standing, not introduced by recent changes.

## Cause

The affected tests construct `Mem2MemConfig()` directly (e.g. `test_session_tracing.py:728`) without isolating the user config file, so the machine's real `~/.memtomem/config.json` leaks into the expected-defaults baseline. Other suites already solve this with the `set_home` helper (`tests/helpers.py`) or an explicit monkeypatched HOME — `test_session_tracing.py:402` even does it for one test — so the fix is mechanical: apply the same isolation to the config-constructing tests in this file.

Sibling of the fresh-HOME class of bugs previously caught in #576-#578 (isolated-HOME smoke).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1247, opened 2026-06-11
Context Gateway review: unverified findings backlog (59 items, carried from #1229)


memtomem:
Carries forward the **unverified finder backlog** from the Context Gateway full-stack review (#1229). The verified catalog there is fully resolved (23/23 across PRs #1227–#1245); these are the finder reports whose adversarial verification was interrupted by session limits and **never completed**.

## Read this first

- **Every item below is a hypothesis, not a confirmed bug.** They are raw finder-lens output (single-agent static reads, no refutation pass). The #1229 campaign's verification rounds refuted a substantial share of similar reports — treat each line as "worth 15 minutes of code reading", not as a defect.
- **The snapshot predates 19 merged PRs** (#1227–#1245, plus #1230–#1237). Several items name files that were heavily reworked since (`skills.py`, `settings.py`, `agents.py`/`commands.py` diff paths, `context_skills.py` routes, `settings_sync.py`), so some are likely already fixed or obsolete. Triage must re-derive each claim against current `main` before any fix work.
- Full finder output with evidence chains and suggested fixes lives in the maintainer-local session artifacts (`.agents-dev/log/ctx-gw-review-findings.json` → `unverified`, cross-referenced against the confirmed sets; not committed).

## Reported-critical (verify first)

- [x] **`mm context update` never removes files deleted upstream** — dest silently diverges from wiki HEAD while the lockfile pins HEAD and status reports ok (`context/install.py` family). Two sibling reports describe the same additive-only update shape from different angles (wiki asset update, dirty detection resurrecting user-deleted files) — verify together. — **fixed by #1250** (update/pinned reconcile against the lockfile `files` manifest; both sibling reports below).

### backend-install-migrate (10) — incl. the one **reported-critical**

- [x] `context/dirty.py` — Flat-layout installs: status reports 'missing' for assets that exist, and `mm context update` bypasses the dirty refusal then silently shadows the (possibly edited) flat file
- [x] `context/dirty.py` — Malformed installed_at string in a lockfile entry raises uncaught ValueError, crashing status/update/migrate classification
- [x] `context/install.py` — mm context update never removes files deleted upstream — dest silently diverges from wiki HEAD while lockfile pins HEAD and status reports ok — **fixed by #1250** (lockfile entries now carry a `files`/`files_commit` manifest; update reconciles upstream deletions instead of copying additively).
[...]

PR #1250, opened 2026-06-11
fix(context): wiki update deletion fidelity — reconcile upstream deletions, detect local ones (#1247 B1)

memtomem:
Part of #1247 — fix batch B1 (the reported-critical install/update family). Resolves backlog items **2/14** (additive-only update), **12** (silent resurrection of user-deleted files), and **10** (wiki-shipped `*.bak` install asymmetry). Item **15** (`installed_at` capture race) stays open on #1247 — the real fix is content digests, which is a cost decision worth its own design pass.

## What was broken

- **Update never removed upstream-deleted files** (reported-critical): `copy_tree_atomic` iterates src only, `_apply_update` pinned the new `wiki_commit`, and post-copy max-mtime capture classified stale files clean forever — dest silently diverged from wiki HEAD while `mm context status` reported `ok`. `install --all --force` funnels through the same copy, so the recovery verb couldn't repair the divergence either.
- **Deletions were invisible to dirty detection**: the walker only sees files that exist, so a user-deleted file classified clean and the next update silently resurrected it — the one local-edit class with a guaranteed false negative.
- **Wiki-shipped `*.bak` files were installed but untracked**: the copier didn't share the walker's suffix skip, so such files dodged dirty tracking and the manifest, and sat in the namespace `update --force` uses for preservation siblings.

## What this PR does

1. **Reconcile after copy** (`_reconcile_removed_files`): update and pinned re-extraction delete dest files absent from the copy source — but only when provably safe: the file's mtime predates the *pre-update* `installed_at` (old wiki bytes), or its `.bak` was just written by `--force`. With a valid manifest, files not recorded there are user-authored and always kept; anything of unknown provenance is kept with a warning (the `--all` confirm gap can race edits past a stale dirty report — when in doubt, never delete). Emptied directories are pruned; `files_removed` is surfaced on `UpdateResult`/`InstallResult` and in the CLI output.
2. **Lockfile file manifest**: entries now record `files` (sorted POSIX relpaths) + `files_commit`. The commit pairing is the staleness guard — `upsert_entry` preserves unknown keys, so a pre-manifest tool rewriting an entry leaves a stale list while the pin moves; consumers honor the manifest only when `files_commit == wiki_commit` and the shape validates (`manifest_from_entry` — `lock.json` can be git-tracked and hand-merged, so malformed shapes degrade to pre-manifest semantics, never crash). Schema documented in ADR-0008.
3. **Deletion-aware dirty detection**: with a valid manifest, `is_asset_dirty` reports missing files as local edits (`DirtyReport.missing_files`); update refuses without `--force`, and restoring a deletion becomes an explicit choice. Summaries cover both edit classes everywhere they render (`StaleInstallError`, `update --all` / `install --all` classification, status rows) so a missing-only report can't print as "0 file(s) modified locally". Unchanged-HEAD updates keep the no-op short-circuit — nothing is written, so nothing can resurrect; status is the diagnosis surface there (regression-pinned).
4. **Copier/walker symmetry**: `copy_tree_atomic` gains an opt-in `skip_suffixes`; the three wiki-install sites pass `DIRTY_SKIP_SUFFIXES` so the installed surface equals the tracked surface. Reconcile membership uses the copier's effective file set (`iter_installed_files(src)` / pin `ls-tree` via the new `WikiStore.asset_files_at_commit`), so a wiki file replaced by a symlink — which the copier skips by contract — can't strand its old regular copy. The skills staging mirror is deliberately unchanged (verified: its diff parity compares full trees).

## Verification

- 22 new tests across `test_context_update.py`, `test_dirty.py`, `test_lockfile.py`, `test_context_install_all.py`; the behavioral ones were run pre-fix and failed for the claimed reason (file persisting, clean-on-deletion, `0 file(s) modified locally`, stranded symlink replacement — the latter two reproduced independently by the review gate).
- Full suite: 6201 passed, 11 skipped (CI filter `-m "not ollama"`; `test_session_tracing` excluded locally per the known #1249 env leak). `ruff check` + `ruff format --check` clean; `mypy` clean on touched files.
- Codex gates: design gate (NEEDS-FIX → resolutions folded in), implementation review (NEEDS-FIX ×2 → fixed with regression tests), final re-verify **SHIP** with zero findings.

### DIFF
--- a/docs/adr/0008-wiki-layer.md
+++ b/docs/adr/0008-wiki-layer.md
@@ -92,7 +92,7 @@ be reconsidered in v2 if a real workflow demands it.
          │  mm context install <type> <name>   (copytree + lockfile pin)
          ▼
 <project>/.memtomem/                 ← project canonical (Invariant 1)
-├── lock.json                        ← { skills: { foo: { wiki_commit, installed_at } } }
+├── lock.json                        ← { skills: { foo: { wiki_commit, installed_at, files, files_commit } } }
 ├── skills/<name>/SKILL.md + overrides/...
 ├── agents/<name>/...
 └── commands/<name>/...
@@ -157,11 +157,13 @@ are out of scope for the install/upgrade lifecycle.
   "skills": {
     "foo": {
       "wiki_commit": "abc123def4567890abc123def4567890abc12345",
-      "installed_at": "2026-04-30T12:34:56Z"
+      "installed_at": "2026-04-30T12:34:56Z",
+      "files": ["SKILL.md", "scripts/run.py"],
+      "files_commit": "abc123def4567890abc123def4567890abc12345"
     }
   },
-  "agents":   { "bar": { "wiki_commit": "…", "installed_at": "…" } },
-  "commands": { "baz": { "wiki_commit": "…", "installed_at": "…" } }
+  "agents":   { "bar": { "wiki_commit": "…", "installed_at": "…", "files": ["…"], "files_commit": "…" } },
+  "commands": { "baz": { "wiki_commit": "…", "installed_at": "…", "files": ["…"], "files_commit": "…" } }
 }
 ```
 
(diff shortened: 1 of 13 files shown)


pandas-studio (approved):
LGTM

PR #1251, opened 2026-06-11
docs(adr): accept ADR-0006 and add Axis F — local-provenance bundle-import redaction (F.3)

memtomem:
## What

Promote ADR-0006 (Web UI folder/upload privacy redaction trust-boundary) from
`Proposed (deferred pending trigger)` to **Accepted**, and amend it with a new
**Axis F — bundle import**, decided as **F.3 (local-provenance-aware)**.

Docs-only. No code changes; implementation of F.3 is a follow-up PR.

## Why now

The ADR's public trigger #3 has fired: `mm web` now ships `--allow-remote-ui`
/ `_validate_bind` (`cli/web.py:228-244`, RFC #787), so the "Web UI = local
user" assumption the deferral leaned on no longer holds once a user opts into
remote UI. That public flag alone is sufficient to promote.

## What Axis F decides

JSON bundle import was never covered by axes A–E. Two ingresses —
`POST /api/export/import` (`web/routes/export.py`) and MCP `mem_import`
(`server/tools/export_import.py`) — share `tools/export_import.py:
import_chunks`, which embeds and upserts with no `privacy.enforce_write_guard`
call. The existing exemption (documented only as a comment in
`tests/test_web_invariants_registry.py`) justifies this by round-trip
fidelity: "archived chunks already passed redaction at original write time."
That only holds for bundles *this instance exported*, not foreign ones.

**Decision: F.3.** A bundle carrying a verifiable **local-provenance** marker
(this install's own export produced it) round-trips unchanged; bundles with an
absent/invalid marker are scanned per record and rejected on secret hits
unless `force_unsafe`, on both ingresses.

The ADR is explicit about what the marker does **not** prove: it attests
*origin*, not that every chunk passed redaction — `force_unsafe` writes,
legacy pre-guard rows, and content from the still-unguarded folder-index path
can all sit in a self-export. F.3 is therefore an explicit **local-provenance
round-trip exemption** (we trust re-importing our own export); the stronger
per-chunk-provenance alternative is noted and deferred. The marker mechanism
(HMAC vs. signature vs. manifest) is also deferred to the implementation PR —
this amendment fixes the policy.

## Also in this amendment

- **Trigger record** — records the public `--allow-remote-ui` trigger.
- **Implementation status** — corrects the record: only the **upload** surface
  is route-guarded today (`enforce_write_guard`, `surface="web_api_upload"`);
  the four folder-index surfaces still index through `IndexEngine`
  (`index_path` / `index_path_stream`, both funnelling to the private
  `_index_file` chokepoint) with no guard, so axes A–E are decided but only
  **partially built**.
- **Reading note** — scopes the original 2026-04-30 Background / outline /
  references as historical (some cited line numbers and method names — e.g.
  the never-existent `index_directory()` — have drifted).
- **Consequences / References** updated for Axis F, plus a stale-background fix
  (`mem_edit` gained a guard at `memory_crud.py:496` since the original draft;
  `mem_delete` writes no content, so needs none).

No `TRACKER.md` row: Axis F is decided here, so there is no open deferred
question to track.

### DIFF
--- a/docs/adr/0006-web-ui-folder-upload-redaction.md
+++ b/docs/adr/0006-web-ui-folder-upload-redaction.md
@@ -1,13 +1,25 @@
 # ADR-0006: Web UI folder/upload privacy redaction trust-boundary
 
-**Status:** Proposed (deferred pending trigger)
-**Date:** 2026-04-30
+**Status:** Accepted (trigger fired 2026-06-11; amended to add Axis F — bundle import)
+**Date:** 2026-04-30 (amended 2026-06-11)
 **Context:** Issue #585 — PR #575 follow-up review surfaced that
 `packages/memtomem/src/memtomem/privacy.py: DEFAULT_PATTERNS` is enforced
 only on the MCP `mem_add` / `mem_batch_add` paths. The Web UI's
 folder-index and upload surfaces accept content raw, bypassing the LTM
 trust boundary that CLAUDE.md asserts ("STM-bypass must not be
 safety-bypass").
+
+> **Reading note (2026-06-11 amendment).** This ADR layers a 2026-06-11
+> amendment over the original 2026-04-30 analysis. The **Background**, the
+> axis tables (**A–E**), and the **Implementation outline** describe the
+> original state and may cite line numbers or method names that have since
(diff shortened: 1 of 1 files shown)


pandas-studio (approved):
LGTM

PR #1253, opened 2026-06-11
fix(context): gate wiki-install ingress and settings rule promote behind Gate A (#1247 B2)

memtomem:
## Summary

Closes the last two ungated first-party write paths into the git-tracked
`project_shared` tree, per ADR-0011 §5 "two gates fire on every
project_shared write" (#1247 backlog ids **3** and **50**, both
confirmed-major in the issue triage):

- **`mm context install` / `update`** (single, `--all`, pinned restore)
  copied wiki bytes — a user-tier store where secrets are legal — straight
  into `<project>/.memtomem/<type>/<name>` with no privacy scan, and the
  `--force` path duplicated dirty (possibly secret-bearing) dest files into
  new unscanned `.bak` siblings.
- **`POST /settings-sync/rules/promote`** appended a private-tier hook rule
  to the shared canonical behind a consent gate but no content scan — the
  only private→shared surface without one.

## Fix shape

**Install/update (CLI):** Gate A runs before the first dest mutation, so a
refusal leaves zero residue — no dest bytes, no `.bak`, no lockfile drift
(upserts already trail copies).

- HEAD paths scan the copier's effective file set (`iter_installed_files`),
  so a wiki-shipped `*.bak` the copier skips cannot false-block (#1250
  parity preserved).
- The pinned path scans git objects at the pin via the new
  `WikiStore.read_asset_file_at_commit` (immutable bytes — no scan→write
  TOCTOU), decoded `errors="replace"` so an ASCII secret inside a non-UTF8
  blob still blocks, filtered by the new shared `_atomic.is_copy_skipped_rel`
  predicate (scan set == materialized set == copied set).
- `--force` scans each dirty file before its `.bak` is created, mirroring
  the version-create snapshot standard.
- Batch rows fail individually (red ✗, batch continues), exit 1; surfaces
  are attributed per real ingress (`cli_context_install`,
  `cli_context_update`, `cli_context_update_all`, `cli_context_install_all`).

**Promote (web):** scans the exact appended canonical fragment
`{event: [rule]}` — `body.event` is a free string that lands as a JSON key,
so a secret-shaped event key blocks too — before the consent gate, returning
422 with a secret-free message (`web_settings_rule_promote`). Scope is
hardcoded `project_shared`; the route's `target_scope` is the tier being
*read*. The web-invariants registry now pins the in-route scan
(`promote_target_rule` moved to `_REDACTION_PROTECTED`; the AST acceptor
recognizes `scan_text_content` as a chokepoint wrapper).

Block messages take a per-surface `remediation_hint` (default `None` keeps
existing sync/migrate messages byte-identical).

## Incidental fixes (same call paths, disclosed per repo convention)

- `copy_asset_at_commit` no longer materializes skip-listed rels (`.bak`,
  `.git`, `__pycache__`, `.DS_Store`) into its extraction tmpdir under
  `.memtomem/<type>/` — previously a crash/SIGKILL could strand unscanned
  wiki bytes as commit-able residue inside the git-tracked tree.
- `mm context update --all` now exits 1 when any row fails, mirroring
  `install --all` (previously always exit 0; the mid-loop-error test's exit
  pin updated accordingly).
- `commit_is_reachable` now enforces its documented contract: only full
  40-hex object ids count as pins. A symbolic ref (`main`) in a hand-edited
  lockfile satisfied `git cat-file` while breaking pin immutability — the
  ref could move between the Gate A scan and the extraction.

## Docs

ADR-0011 §5 dated rider (Gate A now covers wiki ingress + promote; Gate B
confirm intentionally absent on install/update — those verbs have no
`--scope` choice, dest is project_shared by construction), ADR-0008
Invariants 1–2, `configuration.md` + `getting-started.md` gate notes.

## Tests

23 new tests across 6 files, all validated to **fail against the unmodified
tree first** (DID NOT RAISE / 200 / exit 0 / git-show spy), then green
post-fix; the two scan-set parity tests pass pre-fix by design, pinning
#1250 behavior. Full matrix: install/update/pinned blocks with zero-residue
asserts, non-UTF8 blob block, batch row isolation (2 assets / 2 projects),
promote rule + event-key blocks, consent-gate fail-fast, CLI exit-1 mapping
with no secret echo, `is_copy_skipped_rel` unit + walker-parity,
reachability shape rejection.

Gates: full `pytest -m "not ollama"` 6253 passed (4 `test_session_tracing`
fails reproduce on a clean tree — pre-existing local-env issue, known #1249
isolation gap), vitest 151 passed, ruff check + format clean, mypy no new
errors. Codex review: design gate ×3 rounds + implementation ×2 rounds →
SHIP (0 open findings).

Refs #1247.

### DIFF
--- a/docs/adr/0008-wiki-layer.md
+++ b/docs/adr/0008-wiki-layer.md
@@ -40,10 +40,13 @@ Introduce four new surfaces, each governed by an invariant:
 
 `mm context install` snapshots a wiki artifact as a **directory tree**
 (`shutil.copytree` semantics) into the project — including any
-`overrides/` subdirectory. Fan-out (`generate_all_skills`,
-`generate_all_agents`, `generate_all_commands`) reads only from the
-project tree, never from the wiki. CI machines, archived projects, and
-machines without the wiki all run fan-out unchanged.
+`overrides/` subdirectory. Install runs the ADR-0011 Gate A privacy
+scan over the wiki bytes before the copy; a `project_shared` hit
+refuses with no dest or lockfile residue (#1247). Fan-out
+(`generate_all_skills`, `generate_all_agents`, `generate_all_commands`)
+reads only from the project tree, never from the wiki. CI machines,
+archived projects, and machines without the wiki all run fan-out
+unchanged.
 
 ### Invariant 2 — Explicit conflict surface for local edits
 
@@ -52,7 +55,10 @@ modified after install (mtime > `lockfile.installed_at`).
 
 Default behavior: refuse with a clear error. `--force` overwrites and
 leaves a `.bak` copy of each clobbered file. This mirrors ADR-0001's
-on_drop policy of never silently dropping data.
+on_drop policy of never silently dropping data. The `.bak` write is
+itself a `project_shared` write: `--force` scans each dirty file before
+its `.bak` lands, and a privacy hit refuses the whole update before any
+`.bak` or copy is written (#1247, ADR-0011 Gate A).
 
 ### Invariant 3 — Wiki is optional, project is authoritative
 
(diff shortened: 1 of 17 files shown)


pandas-studio (approved):
LGTM

PR #1254, opened 2026-06-11
fix(context): refuse mutations over corrupt lockfile / known-projects store (#1247 B3)

memtomem:
## Summary

#1247 backlog **B3 — id 16** (confirmed major): a corrupt or transiently unreadable `lock.json` / `known_projects.json` was silently re-baselined to an empty doc by the tolerant `load()`, and the very next mutation **persisted the reset** — wiping every sibling asset's install record (lockfile) or the whole registered-project list (known-projects). A corrupt lockfile also wedged the CLI: `mm context update` said *NotInstalled → run install*, `mm context install` said *AlreadyInstalled → run update*.

The codebase already raised on version mismatch for exactly this clobber hazard (`LockfileVersionError`, "silent reset would clobber a forward-compatible lockfile"); corrupt/unreadable was the one silent exception. This PR makes the posture consistent: **strict reads (every write path) refuse; diagnostic reads degrade; nothing re-baselines.**

## Changes

- **`context/lockfile.py`** — new `LockfileError` base + `LockfileCorruptError`. `load(strict=True)` (default) raises on unreadable (`OSError` ≠ missing), invalid JSON, invalid UTF-8 (`json.loads(bytes)` decodes before parsing — raises `UnicodeDecodeError`, not `JSONDecodeError`), and non-object top level. `strict=False` keeps the tolerant empty-doc fallback; a *missing* file stays the write-safe default in both modes. No rename-aside/quarantine — refuse + recovery hint only.
- **`context/status.py`** — `load_with_recovery` / `_tolerant_iter_entries` catch the `LockfileError` base: `mm context status` renders the error row + exits 1 on corrupt JSON the same way it already did for version mismatch (no traceback).
- **`cli/context_cmd.py`** — the three `except LockfileVersionError` arms widen to `LockfileError`: install/update/install `--all` produce a clean ClickException naming the lockfile, breaking the install/update wedge.
- **`context/install.py`** — the multi-project classify arm widens to `LockfileError` → per-project `state="error"` row; the install `--all` restore upsert can no longer wipe a corrupt project's lockfile.
- **`context/projects.py`** — new `KnownProjectsCorruptError`; `load(strict=True)` used by all three mutators (`add`, `remove_by_scope_id`, `update_entry_by_scope_id`). Version mismatch folds in: `_write` re-renders at the current version, so mutating a future-version file is the same hazard. Read-only discovery keeps the tolerant `[]`.
- **`web/routes/context_projects.py`** — known-projects POST/PATCH/DELETE catch the corrupt error → **500** with the hint (server-side state, not client error) instead of re-baselining (POST) or a misleading 404 (PATCH/DELETE).
- **`docs/adr/0008-wiki-layer.md`** — records the refuse-don't-rebaseline posture under "Lockfile schema".

Non-changes, verified: `remove_entry` / known-projects removers never persisted the reset (no write on no-match); `migrate.py`'s lockfile bookkeeping already wraps `remove_entry` in best-effort catch-and-warn; all in-tree `is_asset_dirty` callers pass `lock_entry=` explicitly.

## Tests

New tests across 6 files (parametrized; 28 new cases); **all pre-fix-fail validated by mutation** (strict branches reverted to tolerant → every strict-dependent test fails on its DID-NOT-RAISE / status-code / byte-preservation assertion, none on imports alone):

- `test_lockfile.py` — upsert over merge-conflict garbage refuses + bytes byte-identical; transient `PermissionError` during the in-lock read refuses + siblings survive; remove/read/iter raise; invalid-UTF-8; strict/tolerant load matrix; error-hierarchy pin.
- `test_context_projects.py` — `add` over invalid-JSON / invalid-UTF-8 / future-version / non-list-`projects` / non-object-row / row-without-root refuses + bytes preserved (parametrized); remove + all three update shapes refuse; tolerant degrade on non-list `projects` (the non-iterable case used to crash with TypeError).
- `test_web_routes_context_projects.py` — POST/PATCH(label-only, enabled-only, combined)/DELETE over corrupt + future-version + non-list-`projects` stores → 500, bytes preserved.
- `test_context_install.py` / `test_context_update.py` — the wedge pin: corrupt lockfile now raises `LockfileCorruptError` naming `lock.json` instead of `AlreadyInstalledError`/`NotInstalledError`.
- `test_context_status.py` — corrupt-JSON status degrades (error row, exit 1, no traceback), matching the existing version-mismatch behavior.

Full suite: 6276 passed; the only 4 failures are `test_session_tracing.py` items that fail identically on a clean origin/main tree (pre-existing #1249 isolation gap / known suite-combination flake — verified by stash + re-run). `ruff check` + `ruff format --check` clean; mypy advisory — 2 pre-existing arg-type errors in the route file, untouched by this change.

## Review

- Codex **design** round: NEEDS-FIX (no blockers) — both majors addressed: `update_entry_by_scope_id` made strict + PATCH route handling/tests; `UnicodeDecodeError` treated as corrupt in both stores. All five design open questions resolved as proposed (shared base class; folded version error for the projects store; HTTP 500; `read_entry` raises; migrate keeps catch-and-warn).
- Codex **implementation** round: NEEDS-FIX (no blockers) — prior majors confirmed closed; one new major (strict mode still re-baselined a version-1 doc whose `projects` member is not a list) fixed in the follow-up commit, extended to unrepresentable rows for the same persist-the-loss reason.

Refs #1247 (backlog id 16).

### DIFF
--- a/docs/adr/0008-wiki-layer.md
+++ b/docs/adr/0008-wiki-layer.md
@@ -198,6 +198,17 @@ through plain `dict` is sufficient). The `version` field is reserved for
 schema migrations; future fields (`skill_version`, `compat`, `mode`) can
 be added forward-compatibly.
 
+A lockfile that exists but cannot be read as a JSON object (unreadable,
+invalid JSON / invalid UTF-8, non-object top level) MUST refuse mutations
+rather than silently re-baseline: write paths load-then-write the whole
+doc, so a tolerant "fresh empty doc" fallback would be persisted, wiping
+every sibling entry's install record. Strict reads raise
+(`LockfileCorruptError`, sibling of the version-mismatch error);
+diagnostic surfaces (`mm context status`) may degrade tolerantly but
+never write. A *missing* lockfile is the normal pre-install state and
+stays a write-safe empty default. The same refuse-don't-rebaseline rule
+applies to `known_projects.json` mutations (`KnownProjectsCorruptError`).
+
 ## Vendor format matrix
 
 `OVERRIDE_FORMATS = { (asset_type, vendor): (alias, extension) }` lives
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -61,7 +61,7 @@ from memtomem.context.migrate import (
     migrate_scope,
 )
 from memtomem.context.projects import KnownProjectsStore
-from memtomem.context.lockfile import LockfileVersionError
+from memtomem.context.lockfile import LockfileError
 from memtomem.context.status import classify_status, load_with_recovery, scan_user_artifacts
 from memtomem.context.generator import (
     GENERATORS,
@@ -1454,7 +1454,7 @@ def install_cmd(
         raise click.ClickException(str(exc)) from exc
     except InvalidNameError as exc:
         raise click.ClickException(str(exc)) from exc
-    except LockfileVersionError as exc:
+    except LockfileError as exc:
         raise click.ClickException(str(exc)) from exc
     except PrivacyScanError as exc:
         # Gate A refusal (ADR-0011 §5, #1247) — exc.message carries the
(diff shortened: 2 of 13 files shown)


pandas-studio (approved):
LGTM

PR #1257, opened 2026-06-11
fix(context): harden sync engine — name dedupe, atomic project-memory writes, on_drop surface (#1247 B4)

memtomem:
## Summary

#1247 B4 (sync-engine batch) — fixes confirmed backlog findings **11** (major), **19** (major), **47** (minor) from the triage on #1247.

### id 11 — duplicate frontmatter names silently last-writer-win

`out_path` is a pure function of `(target, name)`, so two canonicals with different stems but the same `name:` both queued the same runtime file: the loser was silently overwritten, **both** writes were reported in `result.generated` (contradicting the in-code #1123 B3-1 "written exactly once" invariant), and `diff` masked the drop by indexing canonicals with the same last-wins order — the dropped artifact was invisible on every surface.

- Phase 1 now tracks the first canonical to claim each name; later claimants from a different source path get a typed **`duplicate_name`** skip (new `SkipCode`) naming both paths, plus a once-per-loser warning.
- Placement (per Codex design review): **after** the Gate A scan — a secret-bearing duplicate loser must still trip the `project_shared` all-or-nothing raise — and **after** the no-fanout check, so those pinned skip rows are unchanged. Before override resolution (the colliding name resolves the same override file via the winner, so no scan surface is lost).
- `diff_agents` / `diff_commands` switch to the matching **first-parsed-wins** precedence; last-wins there would report permanent phantom drift against the file sync now writes.
- Codex impl round caught the adjacent shadow case: a parse-failure **fallback-name** entry (unreadable/unparseable file) overwrote an earlier parsed entry with the same effective name — sync writes that name fine, so diff showed permanent phantom "parse error" drift. The fallback row now registers only when no parsed entry holds the name (warning log still fires); the reverse order keeps its parsed-wins recovery, now pinned as load-bearing. Both shadow pins mutation-validated (guard reverted → fail).

### id 19 — project-memory writes were not crash-safe

The CLAUDE.md/GEMINI.md/… fan-out and the **not-regenerable** canonical `.memtomem/context.md` write used bare `Path.write_text` on all six CLI/MCP sites (`cli/context_cmd.py` init/generate/sync, `server/tools/context.py` init/generate/sync), violating the `_atomic.py` invariant every other fan-out family already honors. All six now use `atomic_write_text` with explicit `0o644` (the helper's `0o600` default would turn a git-tracked CLAUDE.md user-private).

Scope disclosure: the two `context.md` sites are same-family adjacents of the reported fan-out sites — included here because a truncated `context.md` is strictly worse (not regenerable).

New **AST architectural guard** (`test_context_atomic_write_guard.py`) makes the docstring invariant enforced: it scans every gateway write surface for bare `.write_text`/`.write_bytes` **attribute references** (not just calls — three of the six sites passed the method uncalled to `asyncio.to_thread`, which a call-only scan walks right past). Future bare writes must use the helpers or register an `ALLOWED_BARE_WRITES` entry with a rationale.

### id 47 — web sync `on_drop` surface

`SyncRequest.on_drop` accepted any string (out-of-vocabulary values silently degraded to `"ignore"` in the engine), and `on_drop="error"`'s `StrictDropError` escaped as an opaque 500 even though earlier fan-out writes had already landed (the intentional #908 partial-write boundary).

- `on_drop` is validated against the engine's `ON_DROP_LEVELS` via a pydantic `field_validator` → FastAPI-native 422. Single vocabulary owner (CLI `click.Choice` and MCP `_validate_on_drop` already gate the same way); `commands.py` gains the `ON_DROP_LEVELS` re-export for sibling parity with `agents.py`.
- `StrictDropError` now carries `generated` (keyword-optional, default `None`, defensive copy — one-arg construction stays valid), and both sync routes translate it to **422** with a structured `{reason_code: "strict_drop", message, generated}` detail in the #1210 shape, so the client sees exactly which runtime files already changed. API-only reachability — the UI never sends `on_drop`.

## Known residual (routed to B10)

The web sync toast falls through to `sync_success` when `skipped` contains `duplicate_name` — same shape #1229 fixed for `target_conflict`. Needs a JS branch + en/ko i18n keys + `?v` bump; deliberately out of scope for this backend batch.

No ADR contract touched: ADR-0008 Invariant 4 unchanged; no ADR prescribes same-scope name uniqueness (ADR-0011 §419 covers the cross-scope story).

## Testing

- Engine: first-seen winner + per-target `duplicate_name` rows; winner re-claims across targets; **no-fanout precedence** and **Gate A precedence** (secret-bearing loser still aborts `project_shared`) pinned per Codex design round.
- Module mirrors (agents + commands): single generated tuple, winner content survives, skip row names both paths; post-sync `diff` reports `in sync` (no phantom drift).
- Mechanism-spy pins for all six atomic write sites (incl. mode `0o644`), CLI via `CliRunner` + MCP handlers.
- Routes: unknown `on_drop` → 422 (both routes), `error` → 422 with partial-write disclosure matching on-disk state, `warn` → 200 + `dropped`.
- **Pre-fix-fail verified** (src stashed): 8 failed + 6 errors + guard naming all six bare-write sites; surviving passes are contract-preservation pins.
- Full suite: **6303 passed**, 4 failed = `test_session_tracing` pre-existing (#1249; reproduced on clean main during B3).
- `ruff check` + `ruff format --check` green; `mypy` clean on touched modules.
- Codex: design round NEEDS-FIX → applied (dedupe placement vs Gate A/no-fanout, `StrictDropError` API compat); impl round NEEDS-FIX → applied (diff fallback shadow guard); verify-fix round **SHIP** (0 findings).

### DIFF
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import click
 
 from memtomem.context import versioning
+from memtomem.context._atomic import atomic_write_text
 from memtomem.context.agents import (
     AGENT_DIR_FILENAME,
     ON_DROP_LEVELS,
@@ -869,7 +870,11 @@ def init_cmd(
                         sections[key] = val
 
         ctx_path.parent.mkdir(parents=True, exist_ok=True)
-        ctx_path.write_text(sections_to_markdown(sections), encoding="utf-8")
+        # Atomic write (#1247 id 19): a crash mid-write must not truncate the
+        # canonical context.md — unlike the runtime fan-out targets it is NOT
+        # regenerable. 0o644: project file meant to be read (and committed)
+        # by other tools, not a 0o600 state file.
+        atomic_write_text(ctx_path, sections_to_markdown(sections), 0o644)
         click.secho(f"Created {CONTEXT_FILENAME}", fg="green")
         click.echo(f"  Sections: {', '.join(sections.keys())}")
         click.echo("  Edit this file, then run 'mm context generate' to sync.")
@@ -1005,7 +1010,9 @@ def generate_cmd(
                 content = gen.generate(sections)
                 out_path = root / gen.output_path
                 out_path.parent.mkdir(parents=True, exist_ok=True)
-                out_path.write_text(content, encoding="utf-8")
+                # Atomic write (#1247 id 19): crash mid-write must not
+                # truncate the user's CLAUDE.md / GEMINI.md / etc.
+                atomic_write_text(out_path, content, 0o644)
                 click.echo(f"  {name:10s}  {gen.output_path}")
     elif not inc:
         click.secho(f"{CONTEXT_FILENAME} not found. Run 'mm context init' first.", fg="red")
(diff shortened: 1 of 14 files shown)


pandas-studio (approved):
LGTM

PR #1258, opened 2026-06-11
fix(context): migrate fan-out divergence guard + EXDEV symlink parity (#1247 B5)

memtomem:
Part of the #1247 unverified-backlog fix campaign — **B5: migrate hardening** (backlog ids **6**, **7**; both confirmed against main in the 2026-06-11 re-triage).

## id 6 (major) — fan-out cleanup deleted runtime files with no divergence check or backup

`_remove_runtime_fanout_for` ran after every applied `mm context migrate <kind> <name> --to <scope>` and deleted `<runtime_root>/<name><suffix>` (agents/commands) or rmtree'd `<runtime_root>/<name>` (skills) across all four runtimes at the source scope, keyed by name only. Two unrecoverable loss paths:

1. **Runtime-side edits** made after the last sync (a state the dashboard explicitly models as "out of sync") — the divergence delta existed nowhere else.
2. **Hand-authored name collisions** — a file or hand-installed skill dir sync never wrote, deleted as a side effect of moving the canonical away. Heaviest case: the skill-dir rmtree. Always-foreign case: `~/.codex/prompts/<name>.md`, a reserved table row with **no** registered generator.

The dry-run preview never showed the deletion half; paths were reported only after apply.

### Fix

- **Divergence check before delete.** Each target is byte-compared against what sync would write, reconstructed from the canonical at its **post-move** location (the moved bytes are identical to what sync last read at the source scope; per-vendor overrides live inside the artifact dir, so they moved with it). Expected-bytes rule mirrors `_sync_atomic` Phase 2 exactly — override bytes verbatim when one resolves, else the generator render — i.e. the same comparison `diff_agents` / `diff_commands` / `_skill_effective_equal` already pin.
- **Snapshot-then-delete for diverged/unverifiable targets.** Files → sibling `<name>.<ext>.bak` (mirrors `_execute_cleanup_flat`; invisible to suffix-filtered discovery). Skill dirs → `<runtime_root>/.bak/<name>/` — deliberately *not* a skill-shaped sibling: `validate_name` accepts dots, so `<name>.bak` would surface as a phantom diff row, be **re-imported into canonical by extract** (#1229's round-trip failure mode), and be discoverable by the real runtimes; one level under a manifest-less `.bak/` parent, every single-level `<root>/*/SKILL.md` walk stays blind to it. Internal-pattern names were ruled out — the sync-time reaper deletes those.
- **Failed snapshot keeps the target** — deleting without a backup is exactly the loss this guard exists to prevent.
- **Left in place entirely:** symlinked targets (user hand-routing; generators only ever `os.replace` regular files — never deref/back-up/remove) and generator-less (kind, runtime) pairs (`~/.codex/prompts`: sync can never have written there, deletion was pure collateral).
- **Dry-run previews the deletion half** via `MigrateScopeResult.fanout_planned`, computed by the same `_existing_fanout_targets` selection the apply path consumes — preview and apply cannot disagree (Codex design-review fold). Apply surfaces snapshots via `fanout_backed_up` in both CLI and MCP output (hint-prose parity).

## id 7 (minor) — `_stage_move` EXDEV fallback dereferenced symlinks

The cross-FS staging fallback used stdlib-default `shutil.copytree` / `shutil.copy2`, dereferencing resolvable symlinks and materializing out-of-tree target bytes into staging — and from there into the (possibly git-tracked) destination tier — while the same-FS `os.rename` path moves links intact, violating the package's own no-deref mirror contract (`_atomic.copy_tree_atomic`). Gate A still blocks secret-class content behind links on both paths; the exposure was non-secret out-of-tree bytes landing in `.memtomem/`.

Fixed with `symlinks=True` / `follow_symlinks=False` — cross-FS now matches same-FS semantics; dangling links become non-fatal as a side effect.

## Review & tests

- **Codex design review** (pre-implementation): NEEDS-FIX → all folded — dry-run/apply parity for foreign/symlink skips, a skill-tree (`copytree`) backup-failure test alongside the file one, non-claude skill-runtime coverage. All five design questions (backup semantics, skills backup location, foreign-file policy, symlink policy, preview scope) resolved in line with the design.
- **Codex impl review**: SHIP, zero findings (round 1; reviewer independently re-ran the migrate suite and ruff on the changed files).
- **12 new tests**, pre-fix-fail verified: 10 fail on unfixed source; the 2 deliberate negative pins (in-sync targets must NOT produce `.bak` churn) validated by mutation — forcing the divergence check to always-match fails the diverged set, forcing always-diverged fails the negative pins. New symlink-creating tests carry `@pytest.mark.requires_symlinks`.
- Existing 81 migrate tests unchanged and green; full suite green except the known pre-existing `test_session_tracing` failures (#1249, reproduced on clean main).

Backlog bookkeeping: checks off ids 6 and 7 in #1247.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -2261,6 +2261,15 @@ def _print_migrate_scope_result(result: MigrateScopeResult, *, apply_: bool) -> 
     click.echo(f"  to   {result.to_scope}: {result.dst_path}")
 
     if not apply_:
+        if result.fanout_planned:
+            click.echo(
+                f"  will remove {len(result.fanout_planned)} stale runtime "
+                f"fan-out target(s) at scope='{result.from_scope}' (content "
+                f"that diverges from the canonical render is snapshotted to "
+                f"a .bak first):"
+            )
+            for path in result.fanout_planned:
+                click.echo(f"    - {path}")
         click.echo("\nRun with --apply to execute.")
         click.echo(
             f"After apply, run `mm context sync --scope {result.to_scope}` "
@@ -2278,6 +2287,15 @@ def _print_migrate_scope_result(result: MigrateScopeResult, *, apply_: bool) -> 
             f"target(s) at scope='{result.from_scope}':"
         )
         for path in result.fanout_cleaned:
+            click.echo(f"    - {path}")
+    if result.fanout_backed_up:
+        click.secho(
+            f"  {len(result.fanout_backed_up)} target(s) diverged from the "
+            f"canonical render — snapshotted before removal (review and "
+            f"delete manually):",
+            fg="yellow",
+        )
+        for path in result.fanout_backed_up:
             click.echo(f"    - {path}")
     click.echo(
         f"\nNext: run `mm context sync --scope {result.to_scope}` to "
--- a/packages/memtomem/src/memtomem/context/migrate.py
+++ b/packages/memtomem/src/memtomem/context/migrate.py
@@ -43,18 +43,25 @@ from typing import Any, Iterator, Literal
 import click
 
 from memtomem.config import TargetScope
+from memtomem.context import override as _override
 from memtomem.context._atomic import _file_lock, _lock_path_for
-from memtomem.context._names import InvalidNameError, validate_name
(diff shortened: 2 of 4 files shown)


pandas-studio (approved):
LGTM

PR #1259, opened 2026-06-11
fix(context): install/status hardening — flat-layout guard, untracked status rows, degraded-wiki probe (#1247 B6-7)

memtomem:
## Summary

Sixth fix batch for #1247 (**B6-7 — install/status small fixes**), closing backlog ids **0, 1, 4, 5, 8, 9**: two majors (flat-layout silent shadow, status blind to non-lockfile project_shared canonicals) and four minors (malformed `installed_at` crash, circular half-install hints, `install --all` preview/execute mismatch, status traceback on a degraded wiki).

### id 0 (major) — flat-layout installs: update silently shadowed, status misreported

The flat population (`.memtomem/<agents|commands>/<name>.md` + lock entry, no dir — legacy reverse-imports, or a user resolving a flat+dir collision by deleting the dir) is first-class in `migrate`'s truth table and ADR-0008 Invariant 2, but `update`/`status` bypassed it: `is_asset_dirty` probes dir-only → `missing_dest` sailed past the dirty refuse gate, the dir layout landed next to the flat file, and dir-wins fan-out silently stopped serving the (possibly edited) flat bytes. `status` called the row `missing` / "dest missing" while the flat file was being actively served.

- New `install._flat_layout_probe`: flat sibling present + dest dir absent → dirty-or-unprovable check (strict `mtime > installed_at`; missing/unparseable `installed_at` counts as dirty — *can't prove clean must protect*, design-gate M1). Skills excluded (no flat layout).
- `_apply_update` (single + `--all` execute) refuses dirty/unprovable flat without `--force`, pointing at `mm context migrate`; with `--force` (or a provably clean flat) it proceeds with a shadow warning. **No flat `.bak`, no flat deletion**: nothing overwrites the flat file (it stays on disk, merely shadowed) and consolidation stays `migrate`'s job.
- `_classify_for_all_update` / `_classify_for_install_all` emit matching `refuse` rows so batch previews show the same refusal the execute gate raises.
- `classify_status` missing-branch reason becomes ``flat layout; run `mm context migrate <kind> <name>` `` when the flat sibling exists.

The gate covers both `missing_dest` **and** `never_installed`-with-entry reports — without the latter, id 1's degrade (below) would have reopened the bypass for corrupt entries (Codex design gate M1).

### id 1 (minor) — malformed `installed_at` crashed classification

`dirty.py` guarded only `isinstance(str)` and then ran `datetime.fromisoformat` unguarded — a hand-edited `"yesterday"` crashed `status` / `update` / `update --all` / `install --all` / `migrate` with a raw ValueError traceback, defeating the tolerant-read posture. Now an unparseable string degrades to `never_installed` in **every** branch (parse moved before the dest probe, so malformed ≡ non-string), with a warning log. `migrate._classify_row` extends its guard with the same parse probe, demoting corrupt entries to the `skip_manual` path — deliberately **not** `except ValueError: return False` inside `_is_flat_file_dirty`, since `False` means *clean* and would approve overwriting user edits.

### id 4 (minor) — circular remediation hints on half-installs

Copy-succeeded-then-lockfile-write-failed leaves dest with no entry; install's refusal said "run `mm context update`" unconditionally, and update said "run `mm context install` first" — a closed circle with no in-product exit. The `AlreadyInstalledError` message now branches on `(has_lock, has_dest)`: the dest-only state explains the likely causes (interrupted install / hand-placed files) and the actual recovery (inspect dest, remove if disposable, re-run install). No auto-adopt — install can't tell a leftover from hand-placed content it must not clobber (pinned by `test_install_refuses_when_only_dest_present`).

### id 5 (minor) — `install --all` preview missed asset-missing-at-pin

`ProjectInstallClassification`'s docstring promises a state=`error` row for "asset path missing at the pinned commit", but the classifier never probed it — such rows previewed green as `install` and the user found out from the execute phase's `AssetNotFoundError`. The classifier now probes `wiki.asset_files_at_commit(pin, …)` (the extractor's own ls-tree helper, so the two can't drift) right after the reachability check; absence → `error` row, classify-time `CommitNotFoundError` race → `orphan`. The execute-side catch stays as race defense.

### id 8 (major) — status blind to non-lockfile project_shared canonicals

`mm context init --include …` and `mm context migrate --to project_shared` write the project_shared canonical with **no** lockfile entry, and `classify_status` sourced project_shared rows from lock.json only — so the default status view reported "0 asset(s) installed" while sync actively fanned those artifacts out. ADR-0016 §6 names `mm context status --scope <tier>` as the per-tier inspection surface; this aligns the code to the ADR. New `untracked` state (deliberately not `local-draft`: project_shared canonicals fan out, drafts don't), produced by a thin wrapper over the existing `_scan_draft_tier` walk with the lockfile-tracked names dropped (no double rows; the id-0 flat population stays a single `missing` row with the migrate hint). CLI gains the `○` glyph, an accurate header (`N asset(s) installed (+ M untracked)`), and summary/empty-state coverage.

### id 9 (minor) — status traceback on a degraded wiki

`classify_status`'s probe caught only `WikiNotFoundError`, but `WikiStore._git` wraps any git failure in a bare `RuntimeError` — a wiki cloned from an empty remote (`.git` present, no HEAD) or git missing from PATH escaped as a traceback from the read-only diagnostic verb, violating its own "always exits 0 for normal runs" contract. The probe now degrades on `(RuntimeError, OSError)` with a `wiki unusable: <detail>` row reason, and the CLI header distinguishes present-but-unusable (via the pure-stat `exists()`) from not-present. `classify_status`'s public signature is unchanged.

Known adjacent gap, deliberately out of scope (not in the #1247 backlog): `mm context update`'s `wiki.is_dirty()` still raises on a degraded wiki.

## Copy fixes

The aggregate `--all` refusal lines promised "each dirty file gets a .bak sibling", which is false for flat-refuse rows — reworded to cover both classes (dest-tree files get `.bak`; flat files are kept on disk but stop being served). (Codex design gate m2.)

## Codex impl gate: unprovable install records (2nd commit)

The impl review caught a hole the id-1 degrade would have opened: malformed `installed_at` + **existing** dest went from a ValueError crash to a silent overwrite (no dirty proof, no `.bak`) in `_apply_update`, both `--all` classifiers, and `install --all`'s skip-collapse (whose `--force` re-extracted with no preservation). The follow-up commit adds the "unprovable install record" gate: `never_installed` + `dest.is_dir()` (entry present, timestamp unusable) refuses everywhere a write could clobber, and `--force` preserves **every** current dest file as `.bak` — with no epoch there is no dirty subset to limit to. Gate A scans the full bak set; reconcile's `baked=` follows. dest-absent entries keep classifying `update`/`install` (fresh creation clobbers nothing). status's missing-row reason says `install record unusable (malformed installed_at)` when the dest exists.

## Verification

- Codex design gate: 1 round (NEEDS-FIX → M1 + m2 folded, all 5 design questions answered). Codex implementation gate: round 1 NEEDS-FIX (the unprovable-record hole above) → round 2 **SHIP, zero findings** (reviewer independently ran the six touched test files: 236 passed).
- 27 new test functions (+1 strengthened): 19 verified pre-fix-fail against `origin/main` (src reverted, tests kept), 5 verified pre-fix-fail against the first commit for the impl-gate fold, 4 negative/behavior-preservation pins — the force-path pin holds by design pre/post, and the two untracked-dedup pins were mutation-validated (tracked-set filter removed → both fail).
- `ruff check` + `ruff format --check` clean; `mypy` on the touched context modules clean.
- Full suite local: **6325 passed**, 11 skipped (`test_session_tracing` excluded — pre-existing #1249).

Part of #1247 (B6-7). Ids 15 (installed_at race) and the toast/frontend residuals remain with their assigned batches.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -1680,7 +1680,8 @@ def _run_update_all(
     if needs_force and not force:
         click.secho(
             f"\n{len(needs_force)} project(s) have local edits; "
-            f"pass --force to overwrite (each dirty file gets a .bak sibling). "
+            f"pass --force to proceed (modified files in the dest tree get a .bak "
+            f"sibling; flat-layout files are kept on disk but stop being served). "
             f"Refusing to write any project — re-run with --force or resolve manually.",
             fg="red",
             err=True,
@@ -1796,6 +1797,7 @@ _STATUS_GLYPH: dict[str, tuple[str, str]] = {
     "missing": ("!", "red"),
     "stale-pin": ("⚠", "red"),
     "local-draft": ("·", "magenta"),
+    "untracked": ("○", "yellow"),
 }
 
 # ADR-0011 §3 / ADR-0016 §7: project_local agents / commands / skills never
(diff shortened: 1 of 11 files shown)


pandas-studio (approved):
LGTM

Issue #1271, opened 2026-06-12
Context Gateway campaign: web dashboard UX completion (UX track)


memtomem:
Tracker for the UX track of the Context Gateway completion campaign (sibling mechanism-track umbrella is filed separately). Scope: web dashboard first; CLI taxonomy/help and getting-started docs are a later phase.

### Re-verification of the 2026-05-19 internal UX review (against main, 2026-06-12)

Recorded here so closed findings don't get re-fixed:

| Finding | Status | Evidence |
|---|---|---|
| Info-4 direction-aware tile pointers (ADR-0009 §2) | **fixed** | `pointer_missing_target`/`pointer_out_of_sync`/`pointer_missing_canonical` rendering shipped |
| Info-5 deep links `?section=&filter=&artifact=&runtime=` | **fixed** | `_ctxParseDeepLink`/`_ctxSetDeepLink` + deep-link banner |
| Korean i18n coverage | **fixed** | en.json = ko.json = 1166 keys, 339 ctx keys, parity test-enforced |
| Tier-switch consequences invisible | **largely fixed** | #943 write-block sweep + tier badge annotations + disabled-reason copy |
| Confirm dialog cancel i18n | **fixed** | `modal.cancel_btn`; remaining `cancelText` param folded into the a11y item |
| Sync All preview | **mostly fixed** (#1244) | aggregate create/overwrite counts shipped; per-type × per-runtime breakdown still missing |
| List endpoint `?target_scope` asymmetry | **fixed** | skills/commands/mcp-servers all accept it now |
| Error envelope inconsistency | **open (partial)** | overview classifies `error_kind`; mutator/list/settings paths still emit ≥4 wire shapes; JS reads only `.detail` |
| a11y (aria-live, SR loading text) | **open** | zero `aria-live`/`sr-only` in context-gateway.js |
| AbortController on lazy-load switches | **fixed** (#1317) | per-surface AbortControllers abort superseded fetches on scope/tier/section switch; `seq` counters kept as defense in depth |
| Impossible "No project scopes" empty state | **open** | renders for a state the code marks "should never happen" |
| Raw-UTC ISO tooltip on Last sync | **open** | `title` attr shows raw ISO |
| Draft-stash silent quota failure | **open** | `_ctxStashDraft` swallows errors |
| Add Project duplicate-add opacity | **open** | idempotent 200 + generic success toast |

### Work items

[...]

PR #1295, opened 2026-06-12
fix(web): empty scopes roster renders load error with retry (#1287)

### LINKED ISSUE #1287: web: replace impossible 'No project scopes' empty state + first-run hint
### Problem

The artifact list panel renders "No project scopes" (`settings.ctx.no_project_scopes`) for a state the code itself documents as impossible — the comment above it reads "Should never happen — server cwd always present". Reaching it means a projects load/commit failure, so showing an empty-state message there reads as data loss to the user.

### Scope

- Replace the impossible empty state with a load-error state + Retry CTA (re-trigger the projects fetch).
- Add a one-time first-entry hint when all tiles are empty and never-synced (e.g. "Create or import artifacts, then Sync All pushes them to your runtimes"), dismissible, localStorage-flagged.

### Acceptance criteria

- Forced empty-scopes payload renders error copy + a working Retry.
- First-run hint shows once; never after dismissal or once content exists.
- Keys in en+ko same PR; vitest pins both states.

Size: S/M (quick win). Part of #1271.

memtomem:
## Problem

A healthy `/api/context/projects` payload always contains the server-cwd scope, so an empty roster means the load failed — the code comment above the branch literally says "should never happen". Rendering the "No project scopes" empty state there described an impossible inventory and read as data loss to the user.

## Fix

- The empty-roster branch now renders a load-error state (`settings.ctx.scopes_load_failed`) with a **Retry** button that re-runs `loadCtxList` for the section.
- The thrown-failure catch path uses the same shared helper — it previously rendered `load_failed` copy with no recovery affordance; same surface, same fix.
- Legacy key `settings.ctx.no_project_scopes` removed from both locales; new keys land in en+ko.
- `context-gateway.js` cache-bust bumped `?v=71 → ?v=72`.

## Scope note (issue narrowing, disclosed on #1287)

The "one-time first-run hint" half of #1287 is intentionally **not** implemented: ADR-0009 §4 empty-state teaching copy already ships per-tile ("Nothing here yet — open {leaf} to create or import", plus state-driven pointers that teach Sync All once content exists). A second dashboard-level banner would duplicate that guidance. 

## Tests

- New vitest spec `ctx-scopes-load-error-retry.test.mjs`: error copy + Retry presence (legacy literal asserted absent), recovery on Retry click against a healthy payload. Mutation-validated (restoring the legacy line fails the spec).
- The spec polyfills jsdom's missing global `CSS` (real browsers always have `CSS.escape`; without it the recovery render false-fails into the catch path — jsdom-harness-only gap).
- tests-js suite green (34 files / 178 tests); `test_i18n.py` parity green (66 passed).

Fixes #1287. Part of the UX-track campaign #1271.

Note: bumps the same `?v` line as #1293/#1294 — whichever lands later resolves the one-line conflict by bumping one step further.


pandas-studio (approved):

memtomem:
Codex review (pre-merge gate over the three quick-win PRs) returned one Major on this PR: `context-portal.js` still rendered `t('settings.ctx.no_project_scopes')` in its own defensive empty-roster branch while this PR removes that key from both locales — a raw-key leak path. Fixed in the follow-up commit: `_ctxScopesLoadError` now takes the surface's reload as a callback, the portal branch uses it with `loadCtxProjects` as the retry, portal cache-bust bumped, and a portal regression spec pins the copy + raw-key absence + recovery. #1293 / #1294 came back clean.

PR #1399, opened 2026-06-16
fix(web): Context Gateway Sync All — isolate per-phase failures (don't abort siblings) (#1396)

### LINKED ISSUE #1396: Context Gateway: web "Sync All" aborts all remaining types on the first phase's recoverable privacy block
## Summary

In the Context Gateway web UI, **"Sync All" (전체 동기화) aborts every remaining artifact type as soon as one earlier phase hits a recoverable privacy block (HTTP 422)** — even though the other types have no privacy issue and could sync fine. A single secret-bearing artifact in an early phase leaves the whole project unsynced, with zero partial progress.

Found during a manual Playwright-driven QA pass of the Import/Sync flows (prod mode, isolated+seeded HOME).

## Repro

1. In a project's shared store (`.memtomem/`), have one secret-bearing skill (e.g. a `SKILL.md` containing `api_key: sk-…`) **plus** unrelated drift in other types (e.g. a command missing its runtime copy, an agent out of sync).
2. Context Gateway → Overview (Advanced) → tier `project_shared`.
3. Click **전체 동기화 (Sync All)**. The confirm correctly previews e.g. "create 6 / overwrite 2" across skills/commands/agents.
4. Confirm.

## Expected

The clean phases (commands, agents, …) sync; the secret-bearing skill is reported as a failed/blocked phase, but does **not** prevent the other types from syncing. This is exactly what the backend `POST /api/context/sync-all` route already does — its module docstring states *"Per-phase report, NOT cross-type all-or-nothing. A failed phase is [caught]…"* and `_run_phase` wraps each phase in `try/except`, so a privacy-blocked skills phase becomes `status: "failed"` while every later phase still runs (run-level `partial`).

## Actual

The run stops at the **first** phase. The status panel shows:

```
스킬: 실패 · 명령어: 실행 안 함 · 서브에이전트: 실행 안 함 · MCP 서버: 실행 안 함 · 훅: 실행 안 함
(Skills: FAILED · Commands/Subagents/MCP/Hooks: NOT RUN)
```

A hard error toast appears and **nothing** is synced — verified on disk: the unrelated command was never fanned out and all overview counts were unchanged (zero progress).

## Evidence

Network trace for the click — only one sync POST fires, then the loop stops:

```
POST /api/context/skills/sync        → 422 (reason_code=privacy_blocked)
# no commands/sync, agents/sync, mcp-servers/sync, or settings/sync follows
```

The Advanced "Sync All" is a **client-side sequential loop over the per-type routes**, not the backend `/api/context/sync-all` route (the backend route is never called from `context-gateway.js`).

## Root cause

`packages/memtomem/src/memtomem/web/static/context-gateway.js` (~L2465–2491):

```js
const types = ['skills', 'commands', 'agents', 'mcp-servers'];
for (const typ of types) {
  resp = await fetch(`/api/context/${typ}/sync`, …);
  if (!resp.ok) { failed = { phase: typ, reason: … }; break; }   // ← aborts ALL later phases
  succeeded.push(typ);
}
```

A recoverable, per-artifact 422 (privacy block) is treated identically to a fatal error: the loop `break`s, so every later type is skipped. Because `skills` is first, a single seeded secret blocks the entire run. The phase-isolating backend `POST /api/context/sync-all` (`context_sync_all.py`) is bypassed.

## Impact

Sync All is the primary "fix everything" affordance and the main bulk action a Simple-mode / first-run user reaches for. One secret-bearing artifact in an early phase silently prevents *all other artifact types* from syncing, with no partial progress and no per-phase breakdown of what could still have run.

## Suggested fix

Either route the UI's Sync All through the existing `POST /api/context/sync-all` (which already returns an isolated per-phase report), or make the client loop continue past a recoverable `privacy_blocked` / 422 (accumulate into the failed/skipped list instead of `break`) and surface a partial-success summary. At minimum, distinguish recoverable 422s from fatal 5xx before aborting the run.

memtomem:
## Summary

The Context Gateway **"Sync All"** fan-outs aborted every remaining artifact type as soon as one earlier phase hit a recoverable per-phase HTTP error (e.g. a `project_shared` privacy-block **422**). Because `skills` runs first, a single secret-bearing skill left the whole project unsynced with **zero partial progress**.

Both "Sync All" surfaces are client-side sequential loops over the per-type `POST /api/context/<type>/sync` routes, and both `break` on the first non-OK response:
- **Overview** Advanced "Sync All" (`context-gateway.js` ~L3102) — the repro in #1396.
- **Projects-card** "Sync All" (`context-gateway.js` ~L2478) — the same latent bug in the same feature; fixed here for parity.

## Fix

Mirror the backend `POST /api/context/sync-all` contract — *"Per-phase report, NOT cross-type all-or-nothing"*:

- On a per-phase **HTTP error response**, mark the phase `failed` and **continue** to the next type.
- Only a **transport failure** (`fetch` threw — the server is unreachable, so siblings would fail too) still aborts.
- The partial-failure toast now names **every** failed phase — derived from `phaseStates` (overview) / a `failedPhases` array (card). The reason stays the first failure's; the per-phase summary carries the rest.
- **Settings** stays gated behind a clean artifact run (unchanged — it often shares a root cause with artifact failures).

Why isolate *all* per-phase HTTP errors rather than just privacy blocks: the per-type 422 body is a string-only `{detail: ...}` (`SyncPhaseError`'s `reason_code` is consumed only by the server-side sync-all envelope, never serialized), so the client can't sniff "privacy block" from the wire. Isolating every per-phase HTTP error is both correct and the only robust option.

## Tests

- `tests-js/ctx-sync-all-progress.test.mjs`: new **first-phase privacy-block** case (asserts commands/agents/mcp-servers all POST after the skills 422); the mid-run-failure test now asserts the later type still syncs (`done`, not `not_run`).
- `tests/web/test_context_gateway_sync_all.py`: the two failure tests now assert siblings still fire + settings stays skipped.
- `tests/web/test_context_portal_card_sync.py`: new card-path isolation test.
- Cache-bust `context-gateway.js?v=93→94`.

Full vitest suite (317), the affected Playwright files (24), `test_i18n.py` (79), and `ruff` are green. Codex-reviewed → SHIP.

Closes #1396

### DIFF
--- a/packages/memtomem/src/memtomem/web/static/context-gateway.js
+++ b/packages/memtomem/src/memtomem/web/static/context-gateway.js
@@ -3112,18 +3112,12 @@ document.getElementById('ctx-sync-all-btn')?.addEventListener('click', async () 
         break;
       }
       if (!resp.ok) {
-        // A per-phase HTTP error response (e.g. a project_shared privacy-block
-        // 422, or a transient 5xx) is recoverable for the RUN: the server is up
-        // and the OTHER artifact types have no issue, so they must still sync.
-        // Mirror the backend ``/sync-all`` per-phase isolation ("Per-phase
-        // report, NOT cross-type all-or-nothing") instead of the all-or-nothing
-        // this client loop used to do — a single secret-bearing skill no longer
-        // strands every later type as ``not_run`` (#1396). Record the failed
-        // phase and CONTINUE; only a transport failure (``fetch`` threw, above)
-        // aborts, since a dead connection would just fail the siblings too.
-        // ``failed`` holds the FIRST failure — its reason drives the toast and
-        // gates the settings phase; the per-phase summary shows every ``failed``
-        // row, so the toast's failed-phase list is derived from ``phaseStates``.
+        // A per-phase HTTP error (e.g. a project_shared privacy-block 422) is
+        // recoverable for the run: the other types must still sync, so mirror
+        // the backend ``/sync-all`` per-phase isolation — CONTINUE rather than
+        // abort the whole fan-out (only a transport failure, above, aborts). The
+        // FIRST failure drives the toast reason + gates settings; the failed
+        // list comes from ``phaseStates`` (#1396).
         const reason = await _ctxErrorMessageFromResponse(resp, `Sync ${typ} failed`);
         if (!failed) failed = { phase: typ, reason };
         setPhase(typ, 'failed');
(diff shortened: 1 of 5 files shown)


pandas-studio (commented):
LGTM

Issue #1353, opened 2026-06-14
Context Gateway: first-time-user onboarding & IA — surface the canonical→runtime mental model


memtomem:
## Summary

The Context Gateway tab is structurally complete but conceptually opaque to first-time users. It exposes a four-axis model (artifact-type × tier × project × sync-state) without ever stating the one idea that makes it legible: **memtomem keeps one source-of-truth Store (`.memtomem/`) and pushes it one-way out to runtime tools (Claude Code, Codex, Kimi…)**. Because that model is never surfaced, every downstream label reads as undefined jargon and a new user cannot answer: *What is canonical? Sync or Import? Which project am I in? Where do errors live? Why enroll?*

Five independent audits (Nielsen heuristic, cognitive-load, terminology, IA, prior-art) converge on the same root cause and the same highest-severity findings:

- **Jargon at the riskiest moment** — the move/copy confirm reads *"This writes the canonical into the git-tracked project_shared tier"* (`en.json:415`), with raw `canonical` + raw slug `project_shared`, never defined in the UI.
- **Four overlapping status terms for two problems, no direction** — `status_in_sync` / `status_out_of_sync` / `status_missing_target` / `status_missing_canonical` (`en.json:497-500`), re-rendered as badges (`en.json:504-507`); nothing tells the user whether the fix is **Sync** (push) or **Import** (pull).
- **Direction never drawn** — the Overview header renders a root block (`context-gateway.js:1496`) and a runtimes block (`context-gateway.js:1500`) with no connector between them.
- **Dual project identity** — `_ctxScopeDisplayLabel` (`context-gateway.js:579`) swaps the whole label to "Server CWD" (`en.json:369`) for the cwd scope, so one folder shows two names.
- **No onboarding** — the Overview section (`index.html:565-580`) has no help banner, unlike the Index tab (`index.html:702-705`).

This is the **conceptual / IA layer** above the already-filed string-level defects. The plan: a mental-model primer + a shared glossary + a phased UI fix (Minimal now → Moderate next → validate before Bold).

## Mental-model primer (to surface in the Overview)

> memtomem keeps one **Store** (your master copies, in `.memtomem/`). **Sync** pushes them out to your **Runtimes** — Claude Code, Codex, Kimi. **Import** pulls existing ones back in. The flow is one-way: edit in the Store, then Sync.

```
  STORE (.memtomem/)  ── Sync (push) ─▶  Runtimes: Claude Code · Codex · Kimi
[...]

PR #1424, opened 2026-06-17
feat(web): flip Context Gateway Simple-as-default (ADR-0026 D-F), reversibly

memtomem:
## What

Flips the Context Gateway default to **Simple** (`_CTX_SIMPLE_DEFAULT = true`):
a first-time user lands in the Simple Overview (one-line verdict + per-type rows)
instead of the four-axis Advanced grid. **ADR-0026 D-F.**

Shipped as a **reversible experiment** rather than gated on the §Validation naive
user test — recruiting 6 naive participants proved impractical, and the change is
reversible: the Advanced toggle is the per-user rollback (persisted), and
reverting the one constant is the global rollback. **P2** (the irreversible
Push/Pull re-frame) stays deferred and still needs genuine naive evidence.

> **Stacked on #1423** (the Context Gateway walkthrough guide) — this PR's base is
> that branch, so the diff is just the flip. After #1423 merges I'll rebase onto
> main and retarget.

## Changes

- `context-gateway.js` — `_CTX_SIMPLE_DEFAULT` `false → true` (+ comment); cache-bust `?v=97→98`.
- Docs fan-out (same change, per the default-change discipline): `docs/guides/context-gateway.md` leads with the Simple default and frames the precise four-status table as the Advanced view; ADR-0026 D-F → **Shipped (reversible)**, §Validation now gates only P2.
- Tests: the two default-assertion specs now assert Simple; specs that exercise Advanced (the tier control bar, the keyboard-enter-Simple path) start from an explicit Advanced flag; the persistence spec pins that an explicit **Advanced** choice survives reload.

## Verification

- vitest `ctx-simple-mode` 11/11 · Playwright `test_context_gateway_simple_mode` 10/10 · harness + doc-guards 15/15 · ruff clean.
- `style.css` unchanged: the "scoping tighten" the staged plan anticipated is a no-op — the nav/control-bar deep-link trap was already `:has`-guarded in P1a and the tile grid lives only in the Overview (it cannot strand a user).

Refs #1353.

### DIFF
--- a/docs/adr/0026-context-gateway-onboarding-ia.md
+++ b/docs/adr/0026-context-gateway-onboarding-ia.md
@@ -1,15 +1,19 @@
 # ADR-0026: Context Gateway first-time-user onboarding & comprehension layer
 
-**Status:** Accepted & partially shipped — P0 and P1a/P1b are merged; the
-P1 Simple-as-default flip and all of P2 remain deferred pending the
-§Validation first-run user test. See §"Implementation status (as of
-2026-06-17)". D-F (staged rollout, Advanced default first) is followed.
+**Status:** Accepted & partially shipped — P0, P1a/P1b, and the P1
+Simple-as-default flip (D-F) are merged; all of P2 remains deferred. See
+§"Implementation status (as of 2026-06-18)". **D-F shipped 2026-06-18 as a
+reversible experiment** (Simple is now the default) rather than gated on the
+§Validation user test — 6 naive participants proved impractical, and the
+Advanced toggle (per-user) plus the one-line constant (global) are the
+rollback. The §Validation test + kit remain available if P2 (irreversible)
+ever needs naive evidence.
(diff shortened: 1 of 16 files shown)


memtomem:
Codex review (3-agent workflow) returned **NEEDS-FIX** on the first pass — behavior was covered (it re-ran the suites: Playwright + harness 13, vitest 11) but three doc/comment spots still contradicted the shipped Simple-as-default state. All fixed in the amended commit:

- **ADR Consequences** — "No behavioural change in P0/P1" → "P1's flip changes first-run *presentation* only, reversible."
- **ADR §Validation pass bars** — probes 1–3 reframed as **diagnostic/historical** (P0/P1 shipped without the study); the live gate is **P2 only** (probe 5).
- **style.css Simple-mode comment** — refreshed (Simple is the default; rows may carry inline Sync/Import). Comment-only — the CSS rules are unchanged.

pandas-studio (approved):

PR #1500, opened 2026-07-01
fix(index): surface redaction-blocked files from the debounce/flush queue path

memtomem:
## What

Follow-up to #1499 (ADR-0006 PR-A). The redaction gate that PR-A added at the
`IndexEngine._index_file` chokepoint is enforced correctly, but two callers of
`index_path` discard the resulting `IndexingStats`, so a secret-bearing file
they hand off is silently reported as indexed even though the gate skipped it.

## The gap

`cli/indexing.py`'s `_make_indexer` — the closure backing `mm index
--debounce-window` / `--flush`, i.e. the hook-driven path built specifically
for `PostToolUse[Write]` callers — called `IndexEngine.index_path` and threw
away the returned `IndexingStats` entirely. `index_path` aggregates per-file
failures (redaction blocks, parse/embedding errors) into
`IndexingStats.errors` instead of raising, so the closure never raised either:
`debounce.drain_ready` / `drain_all` saw a normal return, marked the path
`indexed`, and deleted it from the retry queue with **zero signal** that a
secret-bearing file was actually skipped by the gate.

This defeats ADR-0006 Axis B's "reject is observable, not silent" rationale for
the one entrypoint aimed at automated hook callers — who have no terminal to
notice a missed warning line.

`git blame` confirms `_make_indexer` predates PR-A (#548, 2026-04-29) and has
always swallowed per-file errors this way; the redaction gate is just the first
*security-relevant* failure to fall into a pre-existing hole. So the fix is
scoped generally, not redaction-only.

The LangGraph integration's `index()` tool had the identical discard — worse,
it returned only `total_files` / `indexed_chunks` / `duration_ms`, so an agent
calling it could never learn a file was blocked, not even after the fact.

## The fix

- **`cli/indexing.py`** — `_make_indexer` now raises `_IndexingStatsError` when
  `stats.errors` is non-empty. `drain_ready` / `drain_all`'s existing
  `except Exception` retry path (entry stays queued) and
  `_print_drain_result`'s generic `(path, message)` rendering apply unchanged —
  no other code needed to change.
- **`integrations/langgraph.py`** — `index()` now returns `blocked_files` /
  `blocked_paths` / `errors` alongside the existing stats, matching the shape
  `mem_index` and the web `IndexResponse` already expose.

`cli/shell.py`'s interactive `index` and `indexing/watcher.py`'s startup
backfill have a smaller version of the same gap (they log `blocked_files` as a
count but not `blocked_paths` or generic `stats.errors`). That's recorded in
ADR-0006 as a known, lower-severity partial gap tracked alongside PR-B rather
than fixed here, since some signal already exists there.

## Tests

- New e2e pin (`test_cli_index_noop_e2e.py`) drives a real `mm init` →
  `--debounce-window` enqueue → `--flush` round trip and asserts the blocked
  file surfaces as an `Errors` entry and **stays queued for retry** (plus the
  `--json` shape).
- New `test_langgraph.py` pin covers the `MemtomemStore.index()` return-dict
  change; the prior exact-dict assertion is updated for the new keys.

Also verified manually in an isolated `HOME`: `mm index --debounce-window …`
then `mm index --flush` now prints `Errors: 1` / `Remaining in queue: 1`
instead of `Indexed: 1` for a secret-bearing file.

## Docs

- `CHANGELOG.md` — new `[Unreleased]` › Security bullet.
- `docs/adr/0006-web-ui-folder-upload-redaction.md` — adds the hook-drain path
  to the PR-A entrypoint enumeration, a "gap found and closed same-day" note,
  and the `cli/shell.py` / `watcher.py` partial-gap record.
- `docs/guides/reference/core-memory-tools.md` — the `mem_index` reference now
  documents the redaction gate, the `Blocked (redaction)` output line, the
  CLI-only `--force-unsafe`, and the debounce-queue enforcement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,21 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   that already passed a write-ingress guard (`mem_add` / `mem_edit`, upload,
   chunk edit, …) is not re-scanned, so no existing write path regresses.
 
+- **`mm index --debounce-window` / `--flush` and the LangGraph `index()` tool
+  now surface secret-blocked files instead of silently discarding them
+  (ADR-0006 PR-A follow-up).** The debounce queue's indexer closure called
+  `IndexEngine.index_path` but never inspected the returned `IndexingStats`,
+  so a secret-bearing file correctly skipped by the redaction gate was
+  reported as `Indexed` and dropped from the queue with no retry — the file
+  was never stored, but the hook caller had no way to know. The drain result
+  now surfaces a redaction-blocked file (`IndexingStats.blocked_files`) as an
+  `Errors` entry and leaves the path queued for retry, matching `mm index`'s
+  direct-run behavior. Terminal non-security skips (too-large / binary files)
+  keep draining silently as before, so they don't accumulate in the queue. The
+  LangGraph integration's `index()` tool had the same reporting gap — an agent
+  calling it could never learn a file was blocked — and now returns
+  `blocked_files` / `blocked_paths` / `errors` alongside the existing stats.
+
 ## [0.3.2] — 2026-06-30
 
 A security release. The Web UI now validates `Host`/`Origin` on every `/api/*`
--- a/docs/adr/0006-web-ui-folder-upload-redaction.md
+++ b/docs/adr/0006-web-ui-folder-upload-redaction.md
@@ -425,7 +425,8 @@ Axes A–E are decided but only **partially built**:
   force_unsafe=…)` and raises `PrivacyRejection` on a hit without `force_unsafe`.
   Bulk entrypoints (`index_path` / `index_path_stream` — `trigger_index`,
   `reindex_all`, `memory-dirs/add` auto_index, `index_stream`, `mm reindex`,
-  `mem_index`, the watcher) catch it per file and aggregate into
+  `mem_index`, the watcher, the `mm index --debounce-window` / `--flush`
+  hook-drain path) catch it per file and aggregate into
   `IndexingStats.blocked_files` / `blocked_paths`, so one flagged file does not
   abort the run; single-file `index_file` lets it propagate so callers roll back
   or surface it rather than silently succeeding. Callers that already ran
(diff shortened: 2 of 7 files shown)


pandas-studio (approved):

memtomem:
Ran a Codex review pass on this branch. Verdict: **NEEDS-FIX** — no blockers, logic confirmed correct (Codex extracted the commit and ran the targeted tests: 3 passed). One Major finding, a doc-accuracy issue:

> `docs/guides/reference/core-memory-tools.md` — the debounce note claimed a blocked entry could be cleared with a manual `mm index --force-unsafe <path>`. That's inaccurate: the drain path never threads `--force-unsafe`, and the redaction gate fires *before* the content-hash skip (`indexing/engine.py:891`), so a direct force-unsafe index stores the content but does **not** dequeue the entry — the next `--flush` re-runs the gate on the still-secret-bearing file and it stays queued.

Fixed in the force-pushed amend (`8b3bae0a`): the doc now states the accurate remediation — remove/rotate the secret so the next `--flush` drains and clears the entry; a direct `--force-unsafe` index stores the content but leaves the queue entry, which keeps erroring until the file stops tripping the gate (or the entry is cleared from `~/.memtomem/index_debounce_queue.json`).

Verified the same overclaim does not appear in the CHANGELOG or ADR bullets (both say "stays queued for retry", which is accurate). `test_docs_guards.py` green.

memtomem:
Second independent review pass (Opus) on `8b3bae0a`. Verdict: **NEEDS-FIX** — one Major, everything else confirmed clean (LangGraph dict change, both test pins as genuine regressions, doc accuracy, exception hygiene / no secret bytes leaked).

**Major (valid — fixed):** raising on *any* non-empty `stats.errors` was too broad. `IndexEngine._index_file` puts **terminal, non-security** markers into `stats.errors` too — `file too large` (`engine.py:845-856`) and `binary file detected` (`engine.py:866-875`), both returned before the suffix filter. Those can never succeed on retry, so raising on them would pin such a file in the debounce queue **forever**, re-erroring on every subsequent `--flush` (a livelock + monotonic queue growth). The pre-fix silent-drop was actually correct for un-indexable assets; only the *security* block (redaction) should stay queued, because it self-clears once the secret is removed.

**Fix (force-pushed as `8f1a5f3f`):**
- `cli/indexing.py` — `_make_indexer` now raises only on `stats.blocked_files` (the redaction/security case), not on `stats.errors` generally. Terminal skips keep draining as before. Surfacing *transient* (embedding-backend) errors on the debounce path is a separate, deliberate call, left as-is.
- New regression test `test_debounce_flush_drains_terminal_skip_without_livelock` — a null-byte (binary) file enqueued via the hook path must **drain** (`Remaining in queue: 0`), not stick. Guards against re-broadening the raise.
- CHANGELOG + ADR-0006 updated to state the scoping (raise on `blocked_files`, not `stats.errors`) and why (terminal-skip livelock).

Re-verified: the redaction pin still passes (blocked file stays queued + surfaces), the new terminal-skip pin passes, `ruff` clean, docs guards + index/redaction/debounce suites green (51 passed locally). CI re-running on `8f1a5f3f`.

Issue #1508, opened 2026-07-02
test: lifespan tests load the developer's repo-root .env into os.environ — session-tracing validator tests fail on dev machines


memtomem:
## Symptom

Four session-tracing config-validation tests fail in **full-suite runs only** on a dev machine whose repo checkout has a root `.env` containing `LANGFUSE_PUBLIC_KEY` / `LANGFUSE_SECRET_KEY`. Each passes when its file runs alone, and CI is green (runners have no `.env`):

- `tests/test_session_tracing.py::TestConfigConstraints::test_langfuse_validator_keys_missing`
- `tests/test_session_tracing.py::TestConfigSaveValidationAndRollback::test_cli_config_set_validation_error`
- `tests/test_session_tracing.py::TestConfigSaveValidationAndRollback::test_web_api_patch_rollback[asyncio]`
- `tests/test_session_tracing.py::TestConfigSaveValidationAndRollback::test_mcp_config_rollback[asyncio]`

All four expect the `SessionTraceConfig` "requires both langfuse_public_key and langfuse_secret_key" validator to fire when no keys are configured; instead the operation succeeds:

```
E   AssertionError: assert 'Failed to persist config' in 'Set session_trace.langfuse_enabled = True (persisted to config.json)'
```

Observed on clean `origin/main @ cd0ae767` (surfaced while verifying #1506 / PR #1507 full runs, unrelated to that change).

## Minimal repro

Reproduces on any machine — fabricate the `.env` with fake keys at the repo root:

```bash
printf 'LANGFUSE_PUBLIC_KEY=pk-fake\nLANGFUSE_SECRET_KEY=sk-fake\n' > .env
uv run pytest packages/memtomem/tests/test_lazy_init_acceptance.py \
              packages/memtomem/tests/test_session_tracing.py -q
# 4 failed — remove either file from the invocation and everything passes
```

## Root cause (verified, three interacting parts)

1. **Lifespan tests source the developer's real `.env`.** `server/lifespan.py:108` calls `_load_dotenv()` (`lifespan.py:74-78`), which runs python-dotenv's `load_dotenv()` — resolving the repo-root `.env` from CWD. Any test that drives the real `app_lifespan` therefore imports the developer's personal env file **into `os.environ` of the whole pytest process**, outside any monkeypatch, permanently for the run. Verified polluters (empirical pair matrix, each paired with `test_session_tracing.py`):
[...]

PR #1540, opened 2026-07-02
fix(config): stop sub-configs from binding bare unprefixed env vars

### LINKED ISSUE #1522: config: 23 sub-config BaseSettings classes bind bare unprefixed env vars (API_KEY, ENABLED, LANGFUSE_*) outside the documented MEMTOMEM_ surface
## Summary

23 of the 24 `BaseSettings` classes in `packages/memtomem/src/memtomem/config.py` define no `model_config`, so pydantic-settings binds their fields to **bare, case-insensitive, unprefixed env var names**. Only the top-level `Mem2MemConfig` (`config.py:840`) carries `env_prefix="MEMTOMEM_"` + `env_nested_delimiter="__"` — the documented config surface. The sub-configs silently accept a second, undocumented env surface made of generic names.

Split out of #1508 per review — this is a config-contract decision, independent of (and much larger than) the test-isolation fix there.

## Verified behavior

```python
# LANGFUSE_PUBLIC_KEY=pk-test LANGFUSE_SECRET_KEY=sk-test
SessionTraceConfig(enabled=True, langfuse_enabled=True)   # validator passes; keys filled from bare env

# API_KEY=bare-env-leak
EmbeddingConfig().api_key                                  # -> 'bare-env-leak'
```

Both reproduced on `origin/main @ cd0ae767`. The nested path is affected too: `Mem2MemConfig()`'s `session_trace` default picks up the bare `LANGFUSE_*` vars (this is what broke the four validator tests in #1508).

## Why it matters

- **Generic names collide.** `API_KEY`, `ENABLED`, `MODEL`, `HOST`, `PROVIDER`, … are common shell exports and CI variables. Any of them silently overrides memtomem config — e.g. `ENABLED=true` in the shell flips `SessionTraceConfig.enabled` (and any other sub-config with an `enabled` field) with no `MEMTOMEM_` opt-in from the user.
- **Security-relevant fields are reachable.** `EmbeddingConfig.api_key`, `SessionTraceConfig.langfuse_secret_key`, `WebhookConfig` fields etc. can be populated from env names the user never associated with memtomem.
- **Validators can be defeated from outside the documented surface** — the #1508 failure mode: the langfuse keys-missing guard passes because bare env supplied keys.
- **Hermeticity**: every test constructing any sub-config directly is exposed to the developer's shell env (the test-side mitigation lands with #1508; the product-side surface remains).

## Affected classes

All classes at `config.py` lines 43–799 (`EmbeddingConfig`, `StorageConfig`, `SearchConfig`, `IndexingConfig`, `DecayConfig`, `MMRConfig`, `AccessConfig`, `NamespacePolicyRule`, `NamespaceConfig`, `RerankConfig`, `QueryExpansionConfig`, `ImportanceConfig`, `WebhookConfig`, `ConsolidationScheduleConfig`, `PolicyConfig`, `ContextWindowConfig`, `HealthWatchdogConfig`, `SchedulerConfig`, `LLMConfig`, `SessionSummaryConfig`, `HooksConfig`, `ContextGatewayConfig`, `SessionTraceConfig`) — 23 classes, `Mem2MemConfig` excluded.

## Options

- **A — convert sub-configs to `pydantic.BaseModel`.** Env binding then flows exclusively through `Mem2MemConfig`'s `MEMTOMEM_<SECTION>__<FIELD>` path. Cleanest contract; matches what the docs advertise. Risk: silently breaks anyone who (knowingly or not) relies on bare names today — needs a changelog note and probably a deprecation beat. Direct constructions in code/tests (`SessionTraceConfig(...)`, `EmbeddingConfig()`) keep working, minus the env side channel.
- **B — explicit `env_prefix` per sub-config** (e.g. `MEMTOMEM_SESSION_TRACE_`). Keeps standalone-BaseSettings ergonomics but invents a third env dialect alongside the nested-delimiter one; more drift surface, little gain over A.
- **C — document-and-keep** (declare bare names supported). Locks in the collision/validator-bypass behavior above; would also mean #1508's validator tests are asserting unsupported expectations. Not recommended.

Leaning A. Whatever the decision, `test_docs_guards.py`'s env allowlist and the config reference docs need to move in the same PR (docs-as-tests), and the #1508 test-isolation fix should land first so the suite is hermetic enough to validate the change.

## Environment

Verified on macOS, Python 3.13 venv, `origin/main @ cd0ae767`, pydantic-settings from the `[all]` extra.

memtomem:
## Summary

Closes #1522.

22 sub-config sections plus `NamespacePolicyRule` were `BaseSettings` classes with no `model_config`, so pydantic-settings bound their fields to **bare, case-insensitive, unprefixed env names** (`API_KEY`, `ENABLED`, `MODEL`, `LANGFUSE_*`, …) — a second, undocumented env surface next to the documented `MEMTOMEM_<SECTION>__<FIELD>` one. Generic shell exports silently overrode config, reached secret-bearing fields (`embedding.api_key`, `session_trace.langfuse_secret_key`), and defeated the langfuse keys-required validator from outside the documented surface (the #1508 failure mode).

## Changes

- **Convert the 23 classes off `BaseSettings` onto a shared strict `ConfigModel` base** (option A from the issue). Env binding now flows exclusively through `Mem2MemConfig`'s `env_prefix="MEMTOMEM_"` + `env_nested_delimiter="__"`. Verified: nested `MEMTOMEM_EMBEDDING__API_KEY` binding is unchanged with `BaseModel` submodels; since the prefixed name already won when both were set, users on the documented surface see zero change.
- **Sweep the `isinstance`/`issubclass` `BaseSettings` checks to `BaseModel`** (`_coerce` list path, `_dedup_key`, fragment-loader coercion, `_json_default`). Missing the `_json_default` one would have sent `namespace.rules` items through the `str()` fallback and silently corrupted `config.json` on save — pinned by a new round-trip test.
- **Deliberate, credentials-only Langfuse exception.** `SessionTraceConfig`'s field names happen to match the Langfuse SDK's standard env vars, so bare `LANGFUSE_*` setups worked today by accident. With `langfuse_enabled=true`, the keys-required validator now also accepts the SDK's own `LANGFUSE_PUBLIC_KEY` / `LANGFUSE_SECRET_KEY`; `get_langfuse_client` already omits empty kwargs and lets the SDK read its documented variables itself. The values are **never copied into the config object** (pinned: they cannot reach `config.json` or config-display surfaces), and `LANGFUSE_ENABLED` alone never turns tracing on — activation requires the explicit memtomem opt-in.
- **Strictness preserved** (follow-up review fix, second commit): `BaseSettings` defaults to `extra="forbid"` + `validate_default=True`, so switching to plain `BaseModel` would have silently swallowed unknown keys — an env typo like `MEMTOMEM_EMBEDDING__TYPO=bad` raised before #1522 and was ignored after the switch, and `NamespacePolicyRule.model_validate()` dropped stray keys in `config.d` fragment entries (both reproduced). The shared `ConfigModel` base carries both settings, so the only behavior change this PR ships is the env-surface removal; a new pin test locks the loud-failure contract.
- **Guard test**: only `Mem2MemConfig` may be a `BaseSettings` subclass, and every other section model must inherit `ConfigModel` — the bare-env surface cannot silently regrow, and a bare-`BaseModel` section cannot silently relax validation.
- **Docs**: `configuration.md` states the prefixed-only contract + the single documented exception near the Session Trace table; bare field-name shorthand in prose (`BASE_URL`, `PROVIDER`, `ENABLED=false`, …) replaced with full `MEMTOMEM_*` names in `configuration.md` / `llm-providers.md`; CHANGELOG entry with migration snippet.

## Test plan

- New: bare-env kill (standalone + nested `default_factory` path), prefixed nested binding, `namespace.rules` config.json round-trip, Langfuse SDK-env acceptance / partial-keys rejection / startup-path combination / no-persistence pin, `BaseSettings` guard.
- Existing 4 validator tests updated to the new error message (now names both surfaces).
- Full suite: 7442 passed, 11 skipped (`-m "not ollama"`); ruff check + format clean; mypy clean on `config.py`.

Design review: gated through Codex (design pass + diff review); the review's docs findings and startup-path test suggestion are incorporated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,27 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   its warning and aggregates non-redaction errors to one line per directory.
   `mm index` output is unchanged.
 
+### Changed
+
+- **Config sections no longer bind bare, unprefixed environment variables**
+  (#1522). The 22 sub-config sections (`embedding`, `storage`, `llm`,
+  `session_trace`, …) plus `NamespacePolicyRule` were `BaseSettings` classes
+  without an `env_prefix`, so
+  generic shell exports like `API_KEY`, `ENABLED`, `MODEL`, or `HOST` silently
+  overrode memtomem configuration — including secret-bearing fields
+  (`embedding.api_key`, `session_trace.langfuse_secret_key`) and validator
+  guards — outside the documented `MEMTOMEM_` surface. They are now plain
+  pydantic models: environment binding flows exclusively through
+  `MEMTOMEM_<SECTION>__<FIELD>`.
+
+  **Migration**: if you relied on a bare name, add the documented prefix —
+  e.g. `API_KEY=sk-…` → `MEMTOMEM_EMBEDDING__API_KEY=sk-…` (or
+  `MEMTOMEM_LLM__API_KEY=sk-…`). One deliberate exception: with
+  `session_trace.langfuse_enabled=true`, the Langfuse SDK's own
+  `LANGFUSE_PUBLIC_KEY` / `LANGFUSE_SECRET_KEY` (and `LANGFUSE_HOST`) still
+  count as credentials — they are read by the SDK itself and are never copied
+  into memtomem config. `LANGFUSE_ENABLED` alone never turns tracing on.
+
 ## [0.3.2] — 2026-06-30
 
 A security release. The Web UI now validates `Host`/`Origin` on every `/api/*`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,7 +112,9 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   (`embedding.api_key`, `session_trace.langfuse_secret_key`) and validator
   guards — outside the documented `MEMTOMEM_` surface. They are now plain
   pydantic models: environment binding flows exclusively through
-  `MEMTOMEM_<SECTION>__<FIELD>`.
+  `MEMTOMEM_<SECTION>__<FIELD>`. Validation strictness is unchanged — unknown
+  keys (an env typo like `MEMTOMEM_EMBEDDING__TYPO`, or a stray key in
+  `config.json`/`config.d`) still fail loudly, exactly as before.
 
   **Migration**: if you relied on a bare name, add the documented prefix —
   e.g. `API_KEY=sk-…` → `MEMTOMEM_EMBEDDING__API_KEY=sk-…` (or
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -1,6 +1,6 @@
 # Configuration Reference
 
-memtomem reads configuration from environment variables. All variables use the `MEMTOMEM_` prefix, with nested sections separated by `__` (double underscore).
+memtomem reads configuration from environment variables. All variables use the `MEMTOMEM_` prefix, with nested sections separated by `__` (double underscore). Unprefixed names are never read, with one documented exception: the Langfuse SDK credentials described in [Session Trace](#session-trace).
 
 ```bash
 # Example: switch from Ollama to OpenAI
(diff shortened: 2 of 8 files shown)


pandas-studio (approved):

memtomem:
Pushed a follow-up review fix (`de4b100c`) after a second Codex pass on the approved head returned NEEDS-FIX with one Major: the `BaseSettings`→`BaseModel` switch silently relaxed validation. `BaseSettings` defaults to `extra="forbid"` + `validate_default=True`; plain `BaseModel` ignores unknown keys, so after `f998999b` an env typo like `MEMTOMEM_EMBEDDING__TYPO=bad` was silently swallowed (it raised on `main`), and `NamespacePolicyRule.model_validate()` dropped stray keys in `config.d` fragment entries. Both reproduced on the branch head before the fix.

The fix introduces a shared `ConfigModel` base (`extra="forbid"`, `validate_default=True`) that all 23 section/entry classes inherit, strengthens the guard test (every section model must inherit `ConfigModel`), and adds a pin test for the loud-failure contract. Full suite 7442 passed / ruff clean. PR body updated accordingly — the approval was on `f998999b`, so this needs a re-approve.

PR #1544, opened 2026-07-02
fix(wiki): store subprocess boundary hygiene (git show + URL end-of-options)

memtomem:
## Summary

T1-10 from the 2026-07-02 ctx/wiki review: two subprocess-boundary hygiene gaps in `wiki/store.py`.

**(a) `read_asset_file_at_commit` bypassed the `_git` error boundary.**
It ran a raw `subprocess.run(check=True)`, so a failing `git show` — the rel absent at the commit for a racy caller, a GC'd object, or a missing git binary — escaped as a raw `CalledProcessError` / `OSError` instead of the normalized, credential-redacted `RuntimeError` every other store subprocess produces. That skips every caller's `except RuntimeError` handler (CLI traceback, web 500), and both the Gate A privacy scan and `install --all` digest reconciliation call this method directly. It now routes through `_git_bytes` (the byte-exact twin introduced in #1521), which preserves the exact-bytes contract the Gate A scan depends on.

**(b) User-supplied URLs could be parsed as git options.**
`clone` / `remote add` / `remote set-url` passed the URL without an end-of-options `--`, so a `-`-prefixed value was consumed by git's option parser — `--upload-pack=<cmd>` on clone runs an arbitrary command. All three legs now pin the URL as positional after `--`.

## Tests

New tests were RED before the fix, green after:

- `TestReadAssetFileAtCommit`: happy-path bytes pin, `git show` failure → normalized `RuntimeError` (not `CalledProcessError`), `OSError` → `RuntimeError` (mirrors the existing `remote_url` pin).
- `TestInitFromUrl::test_dash_prefixed_url_is_positional`: real-git clone from a `-`-named bare repo succeeds (treated as a path, not an option).
- `TestSetRemote::test_dash_prefixed_url_is_positional`: both `remote add` and `remote set-url` legs store a `-`-prefixed URL verbatim and round-trip it.

`uv run pytest -m "not ollama"`: 7481 passed. `ruff check` + `ruff format --check`: clean. Codex review: SHIP, no findings.

### DIFF
--- a/packages/memtomem/src/memtomem/wiki/store.py
+++ b/packages/memtomem/src/memtomem/wiki/store.py
@@ -217,11 +217,12 @@ def _git(
 def _git_bytes(args: list[str], cwd: Path) -> subprocess.CompletedProcess[bytes]:
     """Run ``git <args>`` in ``cwd`` returning raw *bytes* stdout.
 
-    Twin of :func:`_git` for path-carrying porcelain output (``ls-tree -z``):
-    ``text=True`` decodes with the host's preferred locale encoding, which
-    corrupts non-ASCII pathnames on non-UTF-8 hosts — the caller decodes
-    explicitly instead. Failure classification and credential redaction
-    mirror :func:`_git`.
+    Twin of :func:`_git` for byte-exact output — path-carrying porcelain
+    (``ls-tree -z``) and blob content (``show <rev>:<path>``): ``text=True``
+    decodes with the host's preferred locale encoding, which corrupts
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):

pandas-studio (approved):
Reviewed the subprocess-boundary change: read_asset_file_at_commit now crosses the _git_bytes redaction/normalization boundary (no raw CalledProcessError/OSError escaping the RuntimeError contract), and clone/remote add/set-url pin the user URL as positional after --. RED-to-green tests cover both legs; full suite and ruff green; Codex SHIP twice (pre- and post-commit).

Issue #1608, opened 2026-07-04
server: instructions describe multi-agent tools unavailable in default core mode


memtomem:
**Priority: P0 · Size: M · Workstream: WS1 (MCP contract correctness)** — from the 2026-07 improvement review (internal plan: `memtomem-docs` → `memtomem/planning/improvement-plan-2026-07.md`).

## Problem

`server/instructions.py` is a static string surfaced to every MCP client on `initialize`. Its multi-agent workflow section (lines 24–62) tells clients to call `mem_agent_register`, `mem_session_start`, `mem_agent_search`, `mem_agent_share`, and `mem_batch_add` directly — but the shipping default `MEMTOMEM_TOOL_MODE=core` prunes the surface to 9 tools, none of which are these. The instructions never mention `mem_do` routing or `MEMTOMEM_TOOL_MODE`, so a client following the documented path verbatim calls tools that don't exist. The documented multi-agent workflow is a dead-end in the default configuration.

## Proposed fix

Replace the static `INSTRUCTIONS` constant with a `build_instructions(tool_mode)` builder and select the text by `MEMTOMEM_TOOL_MODE` before `FastMCP(...)` is constructed (`server/__init__.py` reads the env var at module scope, so it is available at construction time):

- **core**: route the multi-agent workflow through `mem_do(action="agent_register", ...)` etc., and mention `MEMTOMEM_TOOL_MODE=standard|full` for opting up.
- **standard/full**: keep direct tool names where they are actually exposed.
- Sub-item: state how to discover `mem_do` actions (`action="help"`), which also fixes discoverability of register-only actions like `mem_ingest` that are never individually exposed even in `full` mode.

## Files

- `packages/memtomem/src/memtomem/server/instructions.py`
- `packages/memtomem/src/memtomem/server/__init__.py`
- `packages/memtomem/tests/test_server_instructions.py`

## Verification

Extend the existing pin test into a per-mode parity check: every tool name that appears in a mode's instructions must actually be registered in that mode.

PR #1627, opened 2026-07-04
fix(server): prune tool modes via public API + pin the exact per-mode surface

### LINKED ISSUE #1609: server: tool-mode pruning depends on FastMCP private _tool_manager internals
**Priority: P1 · Size: S · Workstream: WS1 (MCP contract correctness)** — from the 2026-07 improvement review (internal plan: `memtomem-docs` → `memtomem/planning/improvement-plan-2026-07.md`).

## Problem

Tool-mode pruning reaches into FastMCP private internals: `server/__init__.py:190-192` iterates `mcp._tool_manager._tools` and calls `mcp._tool_manager.remove_tool(name)`. A FastMCP version bump that renames or restructures `_tool_manager` would break core/standard pruning **silently** — the server would start in effective `full` mode with no error.

## Proposed fix

- Add `tests/test_tool_mode_pruning.py` asserting the **exact** exposed tool set per `MEMTOMEM_TOOL_MODE` (core = the 9 documented tools; standard = core + the pack set; full = everything). This doubles as an upstream-change canary.
- Wrap the removal in a small helper with a defensive check that **raises** (not skips) if the private attribute shape changes, plus a comment pinning the coupling.

## Files

- `packages/memtomem/src/memtomem/server/__init__.py`
- `packages/memtomem/tests/test_tool_mode_pruning.py` (new)

## Dependencies

Land with or immediately after the mode-aware instructions fix (same file: `server/__init__.py`).

memtomem:
## Problem

Tool-mode pruning reached into FastMCP private internals — iterating `mcp._tool_manager._tools` and calling `mcp._tool_manager.remove_tool` (`server/__init__.py`). A future `mcp` release renaming or restructuring `_tool_manager` would break core/standard pruning **silently**: the server would start in effective `full` mode, exposing every tool, with no error.

## Change

- Removal now uses the **public** `FastMCP.remove_tool` (available mcp>=1.27.2, the pinned floor).
- Enumeration routes through a guarded `_registered_tool_names()` helper that **raises** if `_tool_manager.list_tools` or `remove_tool` change shape — pruning can no longer fail open to full mode.
- The two coupling points are documented as such and pinned by the new test.

## Test — `tests/test_tool_mode_pruning.py`

- In-process: asserts both coupling points (`_tool_manager.list_tools`, `remove_tool`) exist and are shaped as expected — the upstream-change canary.
- Subprocess `tools/list` probes under each `MEMTOMEM_TOOL_MODE` assert the **exact** advertised tool set, computed independently from `_CORE_TOOLS` + the `_STANDARD_PACKS` registry slice (intersected with the pre-prune snapshot so register-only actions like `ingest` don't create phantom expectations). A pruning regression (wrong tools removed) and a silent FastMCP-API break (nothing removed) both fail here.
- A focused guard asserts the multi-agent/session tools the instructions route through `mem_do` are pruned in the default core mode (the #1608/#1609 class).

## Verification

- `uv run pytest packages/memtomem/tests/test_tool_mode_pruning.py` — 6 passed.
- `test_server_instructions.py` + `test_docs_guards.py` + `test_meta_tool.py` — 120 passed.
- `ruff check` + `ruff format --check` clean.

Built on `origin/main` after #1608 (mode-aware instructions) merged, since both touch `server/__init__.py`.

Closes #1609

### DIFF
--- a/packages/memtomem/src/memtomem/server/__init__.py
+++ b/packages/memtomem/src/memtomem/server/__init__.py
@@ -187,10 +187,34 @@ _STANDARD_PACKS = frozenset(
     }
 )
 
+
+def _registered_tool_names() -> frozenset[str]:
+    """Enumerate registered tool names via FastMCP's public-ish surface.
+
+    FastMCP exposes ``remove_tool`` publicly but has no *synchronous*
+    public tool listing (``FastMCP.list_tools`` is async and yields wire
+    objects), so we call the tool manager's synchronous ``list_tools``.
+    Both that method and ``FastMCP.remove_tool`` (used below) are the
+    coupling points to FastMCP internals — pinned by
+    ``tests/test_tool_mode_pruning.py``. Guard the shape explicitly and
+    raise if it changes: a silent failure here would ship every tool in
+    ``core`` mode instead of pruning to 9 (#1609).
+    """
+    manager = getattr(mcp, "_tool_manager", None)
+    lister = getattr(manager, "list_tools", None)
+    if manager is None or not callable(lister):
+        raise RuntimeError(
+            "FastMCP tool-manager API changed: cannot enumerate registered tools "
+            "for MEMTOMEM_TOOL_MODE pruning. Pin the mcp dependency or update the "
+            "tool-mode pruning in memtomem/server/__init__.py."
+        )
+    return frozenset(t.name for t in lister())
+
+
 # Snapshot of every tool registered by the imports above, taken BEFORE
 # pruning — the only place the full surface is still observable once a
 # non-full mode has pruned ``mcp``. Used by the per-mode pin tests.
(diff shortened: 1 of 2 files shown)


memtomem:
Codex review: the subprocess pin test used blocking `readline()`, so a server that stayed alive but went silent would hang CI instead of tripping the 20s deadline. Fixed by `select.select`-ing on the stdout fd with the remaining timeout before each read (test is already POSIX-only).

pandas-studio (approved):

memtomem:
Codex re-review: replaced buffered `readline()` with nonblocking `os.read` + a newline-frame buffer — `select` readiness only guarantees *some* bytes, so `readline` could still block on a partial line. Now no read path can outlive the 20s deadline.

Issue #1631, opened 2026-07-04
cli: split mm init -y from --non-interactive (deprecation window)


memtomem:
Follow-up to #1616, which shipped the non-breaking first step (loud stderr notice + explicit `--help` text when `-y` implies `--preset minimal`).

## Decision recorded in #1616

`-y` will be **narrowed to confirmation-skipping only**, matching every other destructive command in the tree (`reset`, `uninstall`, `upgrade`, `gc`, `tags`). Scripted wizard-skipping should be spelled `--non-interactive` (optionally with an explicit `--preset`).

## Plan

1. **Now (shipped in #1616):** notice + help text announce the upcoming change; `-y` behavior unchanged.
2. **Next minor:** keep `-y` ≡ `--non-interactive` but upgrade the notice to a deprecation warning naming the target version.
3. **Following minor (or the next major, whichever ships first):** `-y` on `mm init` no longer implies `--non-interactive`; a bare `mm init -y` in a TTY runs the wizard (there is no confirmation prompt to skip today, so `-y` becomes a no-op accepted for cross-command muscle memory). `--non-interactive` remains the scripted entry point.

## Files

- `packages/memtomem/src/memtomem/cli/init_cmd.py` (option binding + notice text)
- `packages/memtomem/tests/test_init_cmd.py` (`TestNonInteractivePresetNotice`)
- docs touch-up where `mm init -y` is referenced in guides

memtomem:
Stage 2 shipped in #1639 (merged as 26e6ebf0): `-y` and `--non-interactive` are now separate flags; every `-y` use emits a stderr deprecation warning naming the v0.5.0 target, and docs/error paths teach `--non-interactive`. Remaining scope for this issue is stage 3 — flip `-y` to an accepted no-op — release-gated on a 0.4.x warning window.

Issue #1619, opened 2026-07-04
cli/docs: polish batch — help text, wording, mode glossary


memtomem:
**Priority: P2 · Size: S (batch) · Workstream: WS4 (CLI & docs consistency)** — from the 2026-07 improvement review (internal plan: `memtomem-docs` → `memtomem/planning/improvement-plan-2026-07.md`).

Small polish items batched into one issue:

- [ ] `mm search --format` (`cli/search.py:45-50`): add help text + `show_default` — the `context`/`smart` formats are currently undiscoverable from `--help`.
- [ ] `mm web --open` help wording (`cli/web.py:539`): "Run with opening the browser" → grammatical fix ("Open the browser after launch").
- [ ] Docs glossary disambiguating the three overlapping "mode" vocabularies: Simple/Advanced (UI toggle) vs `--mode prod/dev` (`MEMTOMEM_WEB__MODE`) vs core/standard/full (`MEMTOMEM_TOOL_MODE`).
- [ ] MMR silently disabled without dense retrieval (`search/pipeline.py:923`): surface one INFO line (or a `mem_status` note) so BM25-only deployments know diversity re-ranking is off.
- [ ] Record a removal timeline for the deprecated `mem_context_migrate` alias (`server/__init__.py:133`).

Issue #1616, opened 2026-07-04
cli: mm init -y silently means --non-interactive + --preset minimal


memtomem:
**Priority: P1 · Size: S · Workstream: WS4 (CLI & docs consistency)** — from the 2026-07 improvement review (internal plan: `memtomem-docs` → `memtomem/planning/improvement-plan-2026-07.md`).

## Problem

For every destructive command in the tree, `-y`/`--yes` means "skip the confirmation prompt" (`reset`, `uninstall`, `upgrade`, `gc`, `tags`, …). But `mm init -y` binds `-y` to `--non-interactive` (`init_cmd.py:2955`), which **silently applies `--preset minimal`** (`init_cmd.py:3006`). A user who has internalized "-y = yes to prompts" gets a materially different init than the wizard would have produced, with no signal.

## Proposed fix (non-breaking first step)

- Print a loud one-line notice when `-y` triggers non-interactive mode: which preset is being applied and how to get the wizard back.
- Make the `--help` text for `-y` explicit about the preset side-effect.
- Decide in this issue whether to later split the flags (`-y` = confirm-skip only; `--non-interactive` explicit) with a deprecation window.

Both en and ko strings where user-facing (preserve the 1594-key i18n parity where applicable).

## Files

- `packages/memtomem/src/memtomem/cli/init_cmd.py`
- test + docs touch-up

PR #1639, opened 2026-07-04
cli: split mm init -y from --non-interactive with v0.5.0 deprecation warning (#1631)

memtomem:
Stage 2 of the `-y` deprecation window recorded in #1616 and tracked in #1631 (stage 1 — the loud stderr notice — shipped in #1616).

## What changed

- **`-y` and `--non-interactive` are now separate flags** on `mm init`. `--non-interactive` keeps the wizard-skip semantics (including the implicit `--preset minimal` and its stderr note); `-y` (param `legacy_yes`) is a documented deprecated alias that still implies `--non-interactive`.
- **Every parsed `-y` use emits a stderr deprecation warning naming v0.5.0** — including `-y --preset X` (those scripts also break at stage 3) and usage-error exits (the warning is emitted before `--preset`/`--advanced` validation). The `--non-interactive` spelling never triggers it. Splitting the option is what makes spelling-precise warning possible: as a single click option the two spellings were indistinguishable.
- **Failure paths stop teaching the deprecated alias**: the missing-extras error now says `mm init --non-interactive`, the non-TTY guard says `Pass --non-interactive (optionally with --preset <name>)` instead of `Pass --preset <name> or -y` (`--preset` alone still prompts and cancels without writing config under piped stdin), and `_y_refuse_hint` teaches `--non-interactive`.
- **CHANGELOG**: new "Deprecations" entry with the v0.5.0 behavior-change target, following the timeline convention set in #1619.
- **Docs** (README, getting-started, configuration, PyPI README): examples switched to `--non-interactive`; getting-started documents the alias status and corrects the non-TTY paragraph (previously implied `--preset`/`--advanced` alone were scripted entry points).

## Tests

`TestNonInteractivePresetNotice` extended from 3 to 8 tests: deprecation warning presence/wording for `-y` (bare, with preset, on usage-error exit), absence for `--non-interactive`, `-y` ↔ `--non-interactive` config parity pin (stage-3 safety), help-text pins, and the two reworded error paths. Verified end-to-end with isolated-HOME smoke runs (`-y` vs `--non-interactive` configs byte-identical).

## Not in this PR

Stage 3 (the actual flip: `-y` stops implying `--non-interactive` and becomes an accepted no-op) is release-gated on a 0.4.x warning window — #1631 stays open.

Refs #1631

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,13 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   `mem_context_memory_migrate` (or the `mem_do` action
   `context_memory_migrate`); the `mem_do` action alias `context_migrate`
   follows the same timeline.
+- **`mm init -y` as a wizard-skip alias** (deprecated in #1616, flag split +
+  warning in #1631) — `-y` and `--non-interactive` are now separate flags;
+  `-y` still implies `--non-interactive` but every use emits a stderr
+  deprecation warning. **Behavior-change target: v0.5.0**, when `-y` on
+  `mm init` becomes an accepted no-op (init has no confirmation prompt to
+  skip). Scripts should use `--non-interactive`, ideally with an explicit
+  `--preset`.
 
 ### Added
 
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ The interactive picker starts with three presets — **Minimal** (BM25, no downl
 For automation / CI:
 
 ```bash
-mm init -y                            # minimal preset, same as before
-mm init --preset korean -y            # Korean-optimized bundle, no prompts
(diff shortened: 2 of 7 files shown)


pandas-studio (approved):

memtomem:
CI note: the red `test (windows-latest)` cell is **not from this PR** — the same two failures (`test_project_shared_decline_json_exit_zero`, `test_cancelled_confirm_json`, both `JSONDecodeError ... (char 1)`) are present on every main run since f7006c8b (#1630) and are masked by the cell's `continue-on-error`. Bisect + verified root cause (click 8.4.2 `_readline_prompt` `WIN` branch ignores the `err=True` stdout redirect under `CliRunner`) filed as #1640. Gating cells (ubuntu, macOS) and all other checks are green.

PR #1675, opened 2026-07-05
cli: add empty-state hint for mm search / mm recall on zero results

### LINKED ISSUE #1666: cli: friendly empty-state hint for mm search / mm recall
## Good first issue — friendly empty-state hint for `mm search` / `mm recall`

When a one-shot `mm search` or `mm recall` returns zero results, the CLI gives
no guidance:
- `mm search` (table) prints just a header + separator + `0 BM25 + 0 dense → 0 results`
- `mm search --format plain` prints **nothing at all**
- `mm recall` prints just a header and `0 chunk(s)`

The interactive REPL already does the friendly thing — it prints a yellow
`No results found.` (`packages/memtomem/src/memtomem/cli/shell.py`). This issue
brings the one-shot commands in line.

### Where
- `packages/memtomem/src/memtomem/cli/search.py` — the format branches around the
  `plain` / `table` output (search `def` is at the top of the file)
- `packages/memtomem/src/memtomem/cli/memory.py` — the `recall` command output
- reference behaviour: `cli/shell.py` (`No results found.`)

### What to do
On zero results, emit a short hint, e.g.:
> `No results found — run \`mm status\` to confirm your index has chunks.`

### ⚠️ Hard constraint — do not break the `--json` contract
`mm` JSON output is **byte-clean** (a machine consumer parses stdout; see
`cli/_prompts.py` and #1640). Therefore:
- the hint MUST be written to **stderr** (`click.echo(..., err=True)`), and
- it MUST be gated to the human formats (`table` / `plain`) only — the `json`
  branch must keep emitting exactly `[]` on stdout, unchanged.

Add (or extend) a test asserting that `--format json` stdout is byte-identical
with and without results present (i.e. `[]` on empty).

### Acceptance criteria
- Zero-result `mm search` (table + plain) and `mm recall` print the hint to stderr
- `--format json` stdout unchanged (`[]`), pinned by a test
- `uv run pytest -m "not ollama"` green; `ruff` passes

### Size
1–2 files + a test. Slightly beyond a pure-test starter because of the JSON
nuance, but still self-contained and within the `good first issue` bar.

---
**Claiming:** please comment here before you start. See
[`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) for setup. Contributors
welcome.

Mubashir78:
Print a yellow 'No results found. See `mm status` to confirm your index
has chunks.' hint to stderr when search or recall returns nothing,
matching the interactive REPL behavior.

- gated to human-readable formats (table, plain, context, smart) only
- --format json stdout stays byte-clean [] on empty results
- adds tests for all formats and the JSON invariance constraint

Closes #1666

### DIFF
--- a/packages/memtomem/src/memtomem/cli/memory.py
+++ b/packages/memtomem/src/memtomem/cli/memory.py
@@ -371,7 +371,7 @@ async def _recall(
             project_context_root=project_context_root,
         )
 
-    if not chunks and fmt != "json":
+    if not chunks and fmt in ("table", "plain"):
         click.secho(
             "No results found. See `mm status` to confirm your index has chunks.",
             fg="yellow", err=True,
--- a/packages/memtomem/src/memtomem/cli/memory.py
+++ b/packages/memtomem/src/memtomem/cli/memory.py
@@ -371,6 +371,12 @@ async def _recall(
             project_context_root=project_context_root,
         )
 
+    if not chunks and fmt != "json":
+        click.secho(
+            "No results found. See `mm status` to confirm your index has chunks.",
+            fg="yellow", err=True,
+        )
+
     if fmt == "json":
         out = [
             {
--- a/packages/memtomem/src/memtomem/cli/memory.py
+++ b/packages/memtomem/src/memtomem/cli/memory.py
@@ -374,7 +374,8 @@ async def _recall(
     if not chunks and fmt in ("table", "plain"):
         click.secho(
             "No results found. See `mm status` to confirm your index has chunks.",
-            fg="yellow", err=True,
+            fg="yellow",
+            err=True,
         )
 
     if fmt == "json":
--- a/packages/memtomem/src/memtomem/cli/search.py
+++ b/packages/memtomem/src/memtomem/cli/search.py
@@ -114,7 +114,8 @@ async def _search(
     if not results and fmt in ("table", "plain"):
         click.secho(
             "No results found. See `mm status` to confirm your index has chunks.",
-            fg="yellow", err=True,
+            fg="yellow",
+            err=True,
         )
 
     if fmt == "context":
--- a/packages/memtomem/src/memtomem/cli/search.py
+++ b/packages/memtomem/src/memtomem/cli/search.py
@@ -111,6 +111,12 @@ async def _search(
             project_context_root=project_context_root,
         )
 
+    if not results and fmt != "json":
+        click.secho(
+            "No results found. See `mm status` to confirm your index has chunks.",
+            fg="yellow", err=True,
+        )
+
     if fmt == "context":
         if not results:
             return
--- a/packages/memtomem/src/memtomem/cli/search.py
+++ b/packages/memtomem/src/memtomem/cli/search.py
@@ -111,7 +111,7 @@ async def _search(
             project_context_root=project_context_root,
         )
 
-    if not results and fmt != "json":
+    if not results and fmt in ("table", "plain"):
         click.secho(
             "No results found. See `mm status` to confirm your index has chunks.",
             fg="yellow", err=True,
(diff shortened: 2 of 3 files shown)


memtomem (commented):
Thanks for the PR, and welcome! 🎉 This is a clean, well-scoped change with tests for every format and the JSON-invariance constraint — exactly the shape we like. A couple of things before it can merge:

## Please address

**1. Scope the hint to `table` / `plain` only.** The `fmt != "json"` guard also fires for `--format context` and `--format smart`, which are our *prompt-ready* output modes meant to be fed into an LLM's context (see the format docs in `cli/search.py`). stdout stays byte-clean, but any wrapper that merges stderr into stdout would leak "No results found…" straight into the model context. Suggest gating on the human-display formats explicitly, e.g.:

```python
if not results and fmt in ("table", "plain"):
    click.secho(
        "No results found. See `mm status` to confirm your index has chunks.",
        fg="yellow", err=True,
    )
```

Same for `_recall` in `cli/memory.py` (there the only non-json formats are `table`/`plain`, so it's effectively equivalent — but the explicit allowlist keeps the two call sites reading the same and won't silently start emitting on a future prompt-format addition). Then drop `context`/`smart` from the parametrize list in `test_cli_empty_state.py`.

## Non-blocking note

- The new tests read `result.stderr` directly, which requires Click ≥ 8.2 (where `mix_stderr` was removed and stderr is always captured separately); `pyproject.toml` still declares `click>=8.1`. This matches an existing assumption elsewhere in the suite and our lockfile pins 8.4.2, so CI is green — no action needed here. We'll bump the declared floor to `click>=8.2` repo-wide in a separate change.

## Merge blocker (outside the diff)

- The **CLA check is failing** — please sign the CLA (the CLAAssistant bot comment has the link). We can't merge until that's green.

Once the CLA is signed and the hint is scoped to `table`/`plain`, this is good to go. Thanks again!

Mubashir78:

memtomem:
Thanks for this — it's a clean, well-scoped implementation of #1666. 🙌 The hint is written to **stderr**, gated to the human-readable formats, and the `--json` stdout stays a byte-clean `[]`, exactly as the issue's hard constraint requires. The added `test_cli_empty_state.py` (including the JSON-invariance test) is in the right place and covers the contract nicely.

Two small things before we can merge:

1. **Please rebase onto the latest `main`.** The branch is currently a few commits behind, so GitHub shows it as "out-of-date" and the full CI suite hasn't run against current `main`. A rebase will also let the required checks run green.
   ```bash
   git remote add upstream https://github.com/memtomem/memtomem.git   # if not already
   git fetch upstream
   git rebase upstream/main
   git push --force-with-lease
   ```
   (Heads up: the real change here is just the 3 files — `cli/search.py`, `cli/memory.py`, and the new test. If a diff view ever shows doc/lockfile changes, that's only the rebase gap; it disappears once you're on top of `main`.)

2. **Optional nit:** the PR description says the hint is gated to `(table, plain, context, smart)`, but the code gates to `table`/`plain` only — which is correct for #1666. Could you tweak the description to match, just so it's not misleading for the next reader?

Really nice first contribution — thank you! Once it's rebased and CI is green I'll approve.

Mubashir78:
Rebased onto latest main and updated the PR description to match the actual gating (table/plain only). Thanks for the thorough review!

pandas-studio (approved):

memtomem:
Good news — after the rebase, **every check passes except `lint`** ✅, and the one lint failure is just a `ruff format` nit (the linter rules themselves pass). Quick fix:

Run the formatter and push:
```bash
uv run ruff format packages/memtomem/src/memtomem/cli/
git commit -am "style: ruff format empty-state hint"
git push
```

It only needs to split this one call in **both** `cli/search.py` and `cli/memory.py` (ruff's magic-trailing-comma rule wants one argument per line):
```diff
             "No results found. See `mm status` to confirm your index has chunks.",
-            fg="yellow", err=True,
+            fg="yellow",
+            err=True,
```

Once that's pushed and `lint` goes green, this is good to merge. Thanks for the quick rebase! 🙌

Mubashir78:
Fixed the ruff format nit, fg/yellow and err=True are now on separate lines in both search.py and memory.py. CI should be green now.

Issue #1667, opened 2026-07-05
cli: add --help usage examples (epilog) to mm search and mm add


memtomem:
## Good first issue — add `--help` usage examples to `mm search` and `mm add`

None of the `mm` subcommands set Click's `epilog=`, so `mm search --help` /
`mm add --help` show only the raw option list — no "Examples:" block. Adding a
couple of copy-paste examples materially improves first-run discoverability, and
the examples already exist verbatim in `docs/guides/getting-started.md`, so the
risk is low.

### Where
- `packages/memtomem/src/memtomem/cli/search.py` — the `search` command
- `packages/memtomem/src/memtomem/cli/memory.py` — the `add` command
- (source for the example text: `docs/guides/getting-started.md`)

### What to do
Add an `epilog="Examples:\n\n  ..."` to the `@click.command()` decorator for
**`search`** and **`add`** with 2–3 realistic, copy-paste examples each, e.g.:

```
mm search "deployment checklist"
mm search "redis" --format json

mm add "Redis LRU→LFU reduced cache misses by 40%" --tags "redis,performance"
```

Keep the examples consistent with the actual flags each command accepts (check
the command's existing options).

### Scope (please keep it tight)
Limit this to **`search` and `add`** only — a repo-wide `epilog` sweep across
every command would exceed the `good first issue` bar (≤3 files). We can do the
rest as a follow-up.

### Acceptance criteria
- `mm search --help` and `mm add --help` show an "Examples:" block
- examples use only real, accepted flags
- `uv run ruff check` / `ruff format --check` pass; existing tests still green

### Size
~2 files, additive (help text only — no behaviour or public API change).

---
**Claiming:** please comment here before you start. See
[`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) for setup. First-time
contributors very welcome.

JHON12091986:
/claim #1667

I'll add usage examples to the `--help` output for `mm search` and `mm add` using Click's `epilog=` parameter. This will make the CLI more user-friendly by showing practical examples directly in the help text.

**Approach:**
- Add `epilog=` to the `@click.command()` decorator for both commands
- Include 1-2 common usage examples for each command
- Keep examples concise and clear

ETA: 1-2 hours.

JHON12091986:
PR creado: https://github.com/memtomem/memtomem/pull/1668

Ready for review.

JHON12091986:
PR creado: https://github.com/memtomem/memtomem/pull/1669

[...]

PR #1678, opened 2026-07-06
test: add unit tests for strip_llm_response (#1664)

### LINKED ISSUE #1664: test: add unit tests for strip_llm_response
## Good first issue — add unit tests for `strip_llm_response`

`strip_llm_response` peels a markdown code fence (with or without a language
tag) off raw LLM output so downstream parsers see only the payload. It is a
small, pure, dependency-free function — a great first contribution for getting
familiar with the test harness.

**It currently has no direct unit tests** (it is only exercised indirectly via
`extract_tags_llm`).

### Where
- Function: `packages/memtomem/src/memtomem/llm/utils.py` (`strip_llm_response`)
- New test file: `packages/memtomem/tests/test_llm_utils.py`

### What to do
Add a focused unit test covering at least:
- plain text with no fence → returned unchanged (minus surrounding whitespace)
- fenced block **with** a language tag (e.g. ` ```json ... ``` `) → inner payload only
- fenced block **without** a closing ` ``` ` → still strips the opening fence line
- empty string → `""`
- whitespace-only string → `""`
- text that merely contains backticks mid-line (not a fence) → **not** stripped

### Template to copy
`packages/memtomem/tests/test_embedding_aliases.py` is a clean example of a
dependency-free pure-function test (no fixtures needed — just import and assert).

### Acceptance criteria
- New `test_llm_utils.py` with the cases above, all passing
- `uv run pytest -m "not ollama" packages/memtomem/tests/test_llm_utils.py` green
- `uv run ruff check` / `ruff format --check` pass

### Size
1 new test file, no production changes. Scoped well within the `good first
issue` bar (≤3 files, no public API change).

---
**Claiming:** please comment here before you start so we don't double up. See
[`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) for dev setup (`uv sync`,
`uv run pytest -m "not ollama"`). First-time contributors very welcome.

JHON12091986:
Closes #1664


JHON12091986:
@memtomem I've applied ruff format to fix the linting error. The PR should pass CI now. Ready for review.

memtomem:
Thanks for digging into these — appreciate the effort! 🙏 I think a couple of wires got crossed between your two PRs, so let me lay it out clearly and it should be a quick fix.

**What this PR actually contains:** the diff here is the `strip_llm_response` unit test — i.e. the work for **#1664**, not the `search --help` epilog (#1667) that the title/description mention. That's totally fine; we can turn this into the #1664 PR. Two changes are needed:

1. **Move the test into the collected test tree.** It's currently at repo-root `tests/test_llm_utils.py`, but this repo only collects tests under `packages/memtomem/tests` (`[tool.pytest.ini_options] testpaths = ["packages/memtomem/tests"]`), and root `tests/` is also in `.gitignore`. So as placed, **the test is never run** — CI passes only because it isn't collected. Please move it to:
   ```
   packages/memtomem/tests/test_llm_utils.py
   ```
   (This is also the path #1664 asks for.)

2. **One assertion is off.** In `test_strip_llm_response_multiple_code_blocks`, the input is `print('Hello')` (single quotes) but the expected output is `print("Hello")` (double quotes). `strip_llm_response` only peels the outer code fence — it never rewrites the payload — so the quotes stay single. Once the file is in the collected path, this assertion will fail. Please make the expected value match the input:
   ```python
   assert strip_llm_response(text) == "print('Hello')\n```\n```json\n{\"key\": \"value\"}"
   ```
   Also add a trailing newline at end of file so `ruff format` is happy.

Would you also update the **title/description** to reference #1664 (the strip tests)? Then this becomes a clean, single-purpose PR. After that, `uv run pytest -m "not ollama"` should pass locally — thanks again!

pandas-studio:
Thanks for the contribution, and welcome! 🎉 Unit tests for `strip_llm_response` are a genuinely useful addition, and most of these are correct. A few things need sorting out before this can land:

**1. The test file isn't collected by CI.**
It's added at the repo root (`tests/test_llm_utils.py`), but our pytest config only collects `packages/memtomem/tests` (`testpaths` in `pyproject.toml`). That's why every check shows green — the file is never actually run. Please move it to `packages/memtomem/tests/test_llm_utils.py`.

**2. One assertion fails once the test is actually run.**
`test_strip_llm_response_multiple_code_blocks` expects `print("Hello")` (double quotes), but `strip_llm_response` only strips code fences — it doesn't normalize quotes, so it returns `print('Hello')` (single quotes, unchanged). Running the file directly gives **7 passed, 1 failed**. The fix is to expect the single-quote form:

```python
assert strip_llm_response(text) == "print('Hello')\n```\n```json\n{\"key\": \"value\"}"
```

**3. Missing trailing newline.**
Please add a newline at end of file so `ruff format` passes.

**4. Title/description don't match the diff.**
The title mentions the `search --help` epilog and restoring comments, but the diff only adds this test file — it looks like the epilog changes may have been lost in a rebase/force-push. Could you either update the title/description to reflect what's actually here (e.g. `test: add unit tests for strip_llm_response`), or restore the intended changes?

Once the file is moved, the assertion fixed, and the trailing newline added, this should be good to go — happy to re-review. Thanks again!

memtomem:
Quick note to avoid crossed wires with #1667: the live diff in this PR is the `strip_llm_response` unit test, which is the work for **#1664** — not the `search --help` epilog (#1667) that the title/description reference. To keep things clear, we're treating this PR as the #1664 test contribution, and #1667 (the epilog) has been picked up separately.

No need to close anything — the test work here is salvageable. To land it as the #1664 PR, please:
1. Move the file to `packages/memtomem/tests/test_llm_utils.py` (repo-root `tests/` isn't collected by CI, which is why the checks are green without the test actually running).
2. Fix `test_strip_llm_response_multiple_code_blocks` — the expected value should be `print('Hello')` (single quotes); `strip_llm_response` only peels the code fence and never rewrites the payload.
3. Add a trailing newline and update the title/description to reference #1664.

Thanks again for the contribution!

memtomem:
Thanks for sticking with this — and apologies, our earlier reviews sent you round and round without giving you the full picture. Let me fix that here: everything below is verified against the code, so this should be the last round.

First, the important one, and it's not your fault that it slipped past CI:

## 1. The latest commit empties two source files

`aa29ef64` ("feat(cli): add --help usage examples...") truncates both files to zero bytes:

| commit | `cli/memory.py` | `cli/search.py` |
|---|---|---|
| `c08f8258` | 16,658 B | 6,915 B |
| `9b4956d1` (merge main) | 16,657 B | 6,915 B |
| `aa29ef64` (current head) | **0 B** | **0 B** |

That's where the `633-` in the diff comes from. As it stands, `import memtomem.cli` fails with `ImportError: cannot import name 'add'` — the CLI doesn't start at all. This didn't show up as a red check only because the PR currently conflicts with `main`, so CI never ran. It looks like a rebase or merge went sideways rather than anything you intended.

## 2. Please drop the `--help` epilog work from this PR

This is on us for not being clearer earlier. #1667 (the `search`/`add` epilog) has been picked up separately and is now live in **#1754**, so epilog changes here would collide with it. Let's scope this PR to **#1664** (the `strip_llm_response` tests) only — that work is genuinely useful and worth landing on its own.

## 3. The test import points at a module that doesn't exist

`memtomem.cli.utils` isn't a real module — `strip_llm_response` lives in `memtomem/llm/utils.py`. We asked you to move the file but never checked the import, so `pytest --collect-only` exits 2 with `ModuleNotFoundError`. Sorry about that.

## 4. One more assertion, which we should have caught in round one

With the import fixed, the suite is **4 passed, 1 failed**. `test_strip_llm_response_only_backticks` expects `"```"` but the function returns `""`. Looking at the implementation, a lone fence is simultaneously the first line and the last line, so both get peeled and nothing remains.

---

## Suggested final file

Here's the whole thing, ready to paste over `packages/memtomem/tests/test_llm_utils.py`. I ran it locally: **5 passed**, and `ruff check` + `ruff format --check` both clean.

```python
from memtomem.llm.utils import strip_llm_response


def test_strip_llm_response_simple():
    text = "```python\nprint('Hello')\n```"
    assert strip_llm_response(text) == "print('Hello')"


def test_strip_llm_response_multiple_code_blocks():
    # strip_llm_response only peels the outer fence; it never rewrites the payload.
    text = '```python\nprint(\'Hello\')\n```\n```json\n{"key": "value"}\n```'
    assert strip_llm_response(text) == 'print(\'Hello\')\n```\n```json\n{"key": "value"}'


def test_strip_llm_response_no_code_block():
    text = "Hello world"
    assert strip_llm_response(text) == "Hello world"


def test_strip_llm_response_empty():
    assert strip_llm_response("") == ""


def test_strip_llm_response_only_backticks():
    # A lone fence is both the first and last line, so nothing is left.
    text = "```"
    assert strip_llm_response(text) == ""
```

This also drops the unused `import pytest` (`ruff check` flags it as F401), adds the trailing newline, and uses the blank-line spacing `ruff format` wants.

## Getting the branch back to a clean state

```bash
git fetch origin
git checkout feat/add-help-examples

# restore the two files that aa29ef64 emptied, and drop the epilog work
git checkout origin/main -- packages/memtomem/src/memtomem/cli/memory.py \
                            packages/memtomem/src/memtomem/cli/search.py

# paste the file above over the test, then:
git add -A
git commit -m "test: add unit tests for strip_llm_response (#1664)"
git rebase origin/main
git push --force-with-lease
```

Then `git diff origin/main` should show exactly one added file and nothing else. Could you also retitle to something like `test: add unit tests for strip_llm_response` and change the body to `Closes #1664`? Right now it says `Closes #1667`, which would close the wrong issue on merge.

Verify with:

```bash
uv run pytest packages/memtomem/tests/test_llm_utils.py -v
```

---

One alternative, entirely your call: if you'd rather not spend another round on the git surgery, say the word and we'll open a PR with your tests, credited to you as author, and close this one. The tests are your work either way — we're happy to land them whichever route is less hassle for you. Thanks again for your patience with this one.

JHON12091986:
@memtomem I have followed your instructions:

- Restored the two files that were emptied
- Created the test file with the correct path and assertions
- Updated the commit message to reference #1664

The PR is now focused only on the unit tests for strip_llm_response.

Please review when you have a moment.

JHON12091986:
@memtomem I have followed your instructions. The PR is now focused only on the unit tests for strip_llm_response. Please review when you have a moment.

JHON12091986:
@memtomem I have pushed the changes to my fork. The PR should now be up to date with the correct test file.

Please review when you have a moment.

memtomem:
Nice work — this is very close now. 🎉 The destructive commit is gone, the conflict is resolved, and CI finally ran for real (it never did before, because the branch was conflicting). The test file itself is good: **5 passed**, `ruff check` clean.

Two small things left, and neither is really your fault.

## 1. A UTF-8 BOM snuck into the file

`lint` is red. The only reason is three invisible bytes (`EF BB BF`) at the very start of the file — a byte-order mark, almost certainly added by the editor when you pasted:

```diff
-﻿from memtomem.llm.utils import strip_llm_response
+from memtomem.llm.utils import strip_llm_response
```

`ruff check` and `pytest` don't care, but `ruff format --check` does. Strip it and the file is byte-for-byte accepted (I verified: "1 file already formatted").

To remove it, either re-save the file as **UTF-8 without BOM** in your editor, or run:

```bash
python3 -c "
p = 'packages/memtomem/tests/test_llm_utils.py'
d = open(p, 'rb').read()
if d.startswith(b'\xef\xbb\xbf'):
    open(p, 'wb').write(d[3:])
"
```

Heads-up: our `CLAUDE.md` shows the lint command scoped to `packages/memtomem/src`, but CI actually lints `packages/memtomem/tests` too — so the local command as documented wouldn't have caught this. That's a docs bug on our side and we'll fix it. To reproduce what CI runs:

```bash
uv run ruff check packages/memtomem/src packages/memtomem/tests tools
uv run ruff format --check packages/memtomem/src packages/memtomem/tests tools
```

## 2. The title and body still point at the wrong issue

This one matters more than the BOM. The body still says `Closes #1667`, so merging this PR would **auto-close #1667** — which is a different issue that @elxecutor is actively working on in #1754. Could you update:

- **Title** → `test: add unit tests for strip_llm_response`
- **Body** → `Closes #1664` (and drop the mentions of the `--help` epilog / restored comments, which no longer describe this diff)

---

Once those two are done, lint goes green and this is ready to merge. Thanks for hanging in there through the back-and-forth — the tests themselves are solid.

pandas-studio (approved):
Everything from the last round is resolved — the BOM is gone (verified byte-level), the title/body now correctly reference #1664, and all 18 CI checks are green. I also ran the test file locally against the current `main` tip: **5 passed**.

Approving. Thanks for hanging in there through the back-and-forth — welcome aboard as a memtomem contributor! 🎉

PR #1718, opened 2026-07-12
ci: gate releases and publish reproducible SBOMs

memtomem:
## Summary
- gate TestPyPI/PyPI publication on exact successful main CI plus tag/package/lock/changelog parity
- validate built wheel/sdist metadata and isolated base, all-extras, and sdist installs before OIDC publication
- add production-only CycloneDX 1.5 core/all SBOM workflow with manual v0.3.6 backfill and no-overwrite asset semantics

## Security properties
- publish job retains only id-token: write
- preflight uses actions: read and contents: read
- SBOM job uses contents: write and no id-token
- existing release assets are never clobbered

## Validation
- 42 focused tests passed
- actionlint passed
- real v0.3.6 wheel/sdist metadata validation passed
- real v0.3.6 core/all CycloneDX export and validation passed

### DIFF
--- a/.github/workflows/release-sbom.yml
+++ b/.github/workflows/release-sbom.yml
@@ -52,6 +52,10 @@ jobs:
           RELEASE_TAG: ${{ inputs.release_tag }}
         run: |
           set -euo pipefail
+          if [[ "$GITHUB_EVENT_NAME" == "workflow_dispatch" && "$GITHUB_REF" != "refs/heads/main" ]]; then
+            echo "manual SBOM backfills must run from the protected main branch" >&2
+            exit 1
+          fi
           if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
             echo "SBOM backfill requires an exact production tag, got: $RELEASE_TAG" >&2
             exit 1
--- a/.github/workflows/release-sbom.yml
+++ b/.github/workflows/release-sbom.yml
@@ -56,7 +56,7 @@ jobs:
             echo "SBOM backfill requires an exact production tag, got: $RELEASE_TAG" >&2
             exit 1
           fi
-          python3 tooling/tools/release_preflight.py contract \
+          uv run --no-project --python 3.12 python tooling/tools/release_preflight.py contract \
             --tag "$RELEASE_TAG" \
             --repo-root release \
             --github-output "$GITHUB_OUTPUT"
@@ -77,7 +77,7 @@ jobs:
           VERSION: ${{ steps.contract.outputs.version }}
         run: |
           set -euo pipefail
-          python3 tooling/tools/release_sbom.py \
+          uv run --no-project --python 3.12 python tooling/tools/release_sbom.py \
             --core-raw "$RUNNER_TEMP/memtomem-core.raw.json" \
             --all-raw "$RUNNER_TEMP/memtomem-all.raw.json" \
             --core-out "$RUNNER_TEMP/memtomem-${VERSION}-core.cdx.json" \
--- a/.github/workflows/release-sbom.yml
+++ b/.github/workflows/release-sbom.yml
@@ -1,0 +1,118 @@
+name: Release SBOM
+
+on:
+  workflow_call:
+    inputs:
+      release_tag:
+        description: Immutable production tag to describe
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        description: Existing production tag to backfill
+        required: true
+        type: string
+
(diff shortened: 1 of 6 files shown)


tsdata (approved):

memtomem:
## Review

Verified against the PR branch and repo state. Well-designed fail-closed release gate — **ship-able**. No hard release-blocking bug; one latent maintainability risk worth addressing.

### Overview
- **`preflight` job** (new, read-only): tag/pyproject/uv.lock/CHANGELOG parity → commit is an ancestor of `main` → **waits for the exact main-push CI run to succeed** → wheel/sdist metadata validation → isolated base / all-extras / sdist install checks.
- **`publish` job**: `needs: preflight`, revalidates artifacts right before publish; keeps `id-token: write` only and the `pypi` environment approval gate intact.
- **`release-sbom.yml`** (new reusable): builds/normalizes/validates CycloneDX 1.5 core + all-extras SBOMs; **no-clobber semantics** — an existing asset passes only if byte-identical, otherwise refuses to overwrite.

Confirmed: `ci.yml` triggers on `push: branches:[main]` so the gate actually matches real `event=push, headBranch=main` runs ✓ · `mm --version` works ✓ · CHANGELOG regex matches `## [0.3.6] — …` ✓ · hardcoded floors match current `pyproject.toml` ✓ · `tools/` is covered by CI lint ✓.

### Main finding (medium) — hardcoded security floors drift from pyproject

`tools/release_preflight.py` checks `_REQUIRED_DIRECT` (and the `urllib3>=2.7.0;` prefix) via **exact-string membership**:

```python
missing_direct = {req for req in _REQUIRED_DIRECT if req.lower() not in requirements}
```

Today the set matches `pyproject.toml` exactly. But when a floor is bumped (e.g. a Dependabot security update to `cryptography>=48.0.2`), the published metadata carries the **higher, still-safe** floor while the hardcoded `cryptography>=48.0.1` is no longer present → the release is **hard-blocked** with a misleading `misses direct floors` message (the dep *is* present, just at a higher floor). The `urllib3>=2.7.0;` case is worse: a floor bump breaks the `startswith("urllib3>=2.7.0;")` extras detection entirely. This repo bumps deps frequently (#1712–16), so it's an easy path to hit.

It fails closed (won't ship a bad artifact — safe direction), but it stops legitimate releases. Suggested fix (pick one):
1. Parse the floors from `pyproject.toml` and compare, or
2. Compare `>=` lower bounds with `packaging` instead of exact equality, or
3. At minimum add a unit test asserting `_REQUIRED_DIRECT == pyproject floors`, so drift fails in **CI (fast, clear)** rather than at **release (slow, confusing)**. A source-pin comment (mirroring the `privacy.py` SHA-pin convention) would also help.

### Minor / confirm

- **Behavioral tightening (confirm intent):** `preflight` now applies to test tags too, so **TestPyPI dry-runs** also require (a) the commit on `main`, (b) a green main-push CI, and (c) CHANGELOG/lock/pyproject parity. This matches the documented 2-PR flow (release PR → merge to main → tag), but dry-runs from an un-merged branch commit are no longer possible. OK if intended.
- **stdout capture (low):** in publish's *Revalidate*, `version=$(… contract …)` assumes the script/uv emit nothing else to stdout. uv logs to stderr so it's fine in practice; `| tail -n1` would harden it.
- **SBOM reproducibility (low):** `normalize` strips only `serialNumber` / `metadata.timestamp`. Deterministic today thanks to the pinned uv (0.11.16); a future uv bump will change `metadata.tools` etc. and make a backfill `cmp` refuse — safe-by-design, but worth a comment documenting that expectation.
- **CI conclusion handling (low):** `wait_for_exact_main_ci` treats `neutral`/`skipped` as neither success nor terminal-failure, so it polls to timeout (~40 min) and fails with `timed out` instead of a clear "CI concluded skipped". `ci.yml` has no path filters today so unlikely; a clearer early-exit would aid diagnosis.

### Strengths
- All actions SHA-pinned; least-privilege per job (preflight=read, publish=id-token only, sbom=contents:write & no id-token).
- Fail-closed API handling (3-consecutive-error cap, timeout, `_TERMINAL_FAILURES` early failure) with injectable `fetch_runs`/`monotonic`/`sleep` for solid, deterministic tests.
- No-clobber asset semantics, isolated install verification (base/all/sdist + `uv pip check` + import smoke).
- `pypi` environment approval gate preserved **after** preflight, so the human-approval model is kept and strengthened; publish rebuilds + revalidates, keeping consistency with the preflight-validated artifact.

**Verdict: SHIP** — the medium floor-drift item can be a follow-up, but given this repo's dependency-bump cadence I'd recommend guarding it here with one unit test (option 3).

PR #1719, opened 2026-07-12
ci: require security supply-chain gates

memtomem:
## Summary
- scan both uv.lock and tests-js/package-lock.json with pinned OSV Scanner as a blocking job
- validate reviewed owner/reason/expiry exceptions and emit scanner-native config
- guard every external action with immutable SHA/digest pins and enforce reviewed plugin-to-core mappings
- add a stable security-gates aggregate check for branch protection

## Stack
Depends on #1718. After #1718 merges, this PR will be retargeted to main.

## Validation
- 49 focused release/security tests passed
- combined Python + JavaScript OSV scan: No issues found
- Ruff and actionlint passed


tsdata (approved):

memtomem:
## Review

Verified against the PR branch. `release-sbom.yml` exists on the branch HEAD (it's stacked on #1718), while this PR's own diff is the 5 files below; plugin mapping (`0.2.1`→core `0.3.5`), the `package-lock.json` path, and `tools/osv-exceptions.toml` all exist on the branch. Solid supply-chain hardening — **ship-able, contingent on merge ordering**.

### Overview
Promotes supply-chain gates from advisory → **blocking**:
- **`python-deps`**: scans `uv.lock` + `tests-js/package-lock.json` in **one OSV invocation** (prevents a partial-blocking gap). Exceptions use a canonical `owner/reason/id/ignoreUntil` schema; CI converts it to OSV Scanner's native TOML.
- **`bandit`**: now blocks on *new* findings vs the committed baseline.
- **`security-gates`** (new aggregate, `if: always()` + each `needs.*.result == success`): a stable required-check name for branch protection.
- **`check_osv_exceptions.py`** + `test_supply_chain_contracts.py`: exception schema validator plus guards for workflow SHA-pinning, uv-pin drift, and plugin document coherence.

### Strengths
- **Strong supply-chain guards**: full-SHA/digest pin enforcement across every workflow `uses:`, a single-and-matching `UV_VERSION` assertion across `ci`/`release`/`release-sbom`, and a three-way lock of `plugin.json` / `marketplace.json` / `.mcp.json`.
- **Fail-closed exception tooling**: rejects unknown/missing keys, blank strings, duplicate ids, `datetime`≠`date` (datetime is a `date` subclass, so it's rejected first), quoted dates, and expired (`<= today`) entries — with thorough tests.
- **OSV rc distinction preserved**: `rc==1` (advisories) blocks, `rc>1` (scan error, e.g. OSV.dev unreachable) is handled separately so a short/empty summary isn't misread as "clean". `pipefail` + `tee || rc=$?` is correct.
- `security-gates` is the canonical aggregate-required-check pattern; `permissions: {}` is least-privilege.

### Findings / risks

**1. (medium — merge ordering) The #1718 dependency is load-bearing.**
`test_uv_toolchain_pin_matches_ci_release_and_sbom_workflows` reads `release-sbom.yml` (a #1718 artifact). It is **not** in this PR's diff, so if #1719 lands on `main` before #1718, that test fails on `main` with a FileNotFoundError. The PR body manages this ("after #1718 merges, retargeted to main") — just keep that discipline (no out-of-order admin-merge). Consistent with the current `BLOCKED` merge state.

**2. (low — design) `security-gates` aggregates only the 4 security jobs.**
It does not `needs` the functional `test`/`lint`/`typecheck` jobs. If intended as an *additive* gate, fine — but branch protection must keep this **alongside** the existing required checks, not as a replacement, or functional tests stop being required. That config change lives outside this PR (repo settings) and should land with it.

**3. (low — maintenance coupling) Plugin pin vs core version.**
`_PLUGIN_CORE_MAP = {"0.2.1": "0.3.5"}` freezes `.mcp.json`'s `memtomem==0.3.5`, while core is currently **0.3.6**. If the plugin intentionally lags core, fine — but the new guard now means any future core-pin bump in the plugin must update `_PLUGIN_CORE_MAP` in lockstep. Worth confirming the lag is intentional (this PR codifies the existing pin, it didn't introduce it).

**4. (low — ops) Time-boxed exceptions expire against real UTC `today`.**
`ignore_until <= today` means a merged exception can flip `main` red overnight with **no code change**. That's intentional security hygiene (forces re-review), but worth calling out so it isn't surprising.

**5. (nit) `render_native` emits TOML string values via `json.dumps`.**
Safe for ASCII advisory ids / English reasons, but JSON and TOML string escaping diverge on some unicode/control cases. Harmless in practice; a dedicated TOML escaper (or a `tomllib` round-trip) would be bulletproof.

**6. (nit) OSV now blocks on the dev-only `tests-js/package-lock.json`.**
A dev/test-tool advisory can block `main` — slightly aggressive, but the reviewed-exception path mitigates it. Acceptable.

### Verdict
**SHIP** — conditional on **merging #1718 first, then retargeting** (finding 1). Code quality, test coverage, and the fail-closed design are all strong; please confirm findings 2 (keep branch protection additive) and 3 (plugin-pin lag intended) before merge.

PR #1720, opened 2026-07-12
release: plugin 0.2.2 pins core 0.3.6

memtomem:
## Summary
- bump the Claude plugin manifest and marketplace entry from 0.2.1 to 0.2.2
- pin the plugin MCP server to the independently verified memtomem==0.3.6 release
- update the reviewed plugin-to-core mapping guard

The marketplace top-level metadata version remains 1.0.0.

## Stack
Depends on #1719, which depends on #1718. Retarget to main after the dependencies merge.

## Validation
- 34 plugin mapping and docs guard tests passed
- all three JSON documents parse successfully

PR #1721, opened 2026-07-12
fix(cli): make historical file rescan fail closed

memtomem:
## Summary
- replace Path.rglob with an explicit lstat-based walker that surfaces directory enumeration failures
- reject managed-base symlinks, skip nested symlinks, enforce resolved containment, and keep duplicate roots idempotent
- preserve the existing read-only JSON contract while making root/enumeration/read/containment errors exit non-zero

## Stack
Depends on #1720 -> #1719 -> #1718. Retarget to main as dependencies merge.

## Validation
- 23 rescan tests passed
- covers all three managed directories, duplicate roots, enumeration/read failures, base and nested symlinks, no secret content in output, and unchanged file bytes
- Ruff and diff checks passed

### DIFF
--- a/packages/memtomem/src/memtomem/cli/mem_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/mem_cmd.py
@@ -22,7 +22,9 @@ from __future__ import annotations
 
 import asyncio
 import json
+import stat
 from pathlib import Path
+from collections.abc import Iterator
 from typing import cast, get_args
 
 import click
@@ -36,6 +38,18 @@ from memtomem.memory_scope import resolve_memory_scope_dir
 
 
 _MEMORY_SCOPE_CHOICES = list(get_args(TargetScope))
+
+
+def _iter_audit_files_fail_closed(root: Path) -> Iterator[Path]:
+    """Yield regular non-symlink files and propagate every walk/stat error."""
+    for entry in sorted(root.iterdir()):
+        entry_stat = entry.lstat()
+        if stat.S_ISLNK(entry_stat.st_mode):
+            continue
+        if stat.S_ISDIR(entry_stat.st_mode):
+            yield from _iter_audit_files_fail_closed(entry)
+        elif stat.S_ISREG(entry_stat.st_mode):
+            yield entry
 
 
 @click.group("mem")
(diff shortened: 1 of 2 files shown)


memtomem:
## Review — `fix(cli): make historical file rescan fail closed`

### Overview
Replaces the `rescan-files` historical walker (`Path.rglob`) with an explicit `lstat`-based walker across the `_imported`/`_fetched`/`sessions` managed dirs:

- **fail-closed**: directory enumeration / `stat` / resolve failures now surface in `errors` and exit non-zero instead of being silently skipped. (The old `rglob` silently dropped unreadable subdirectories, so a partially-unreadable tree looked "clean" — this is the core fix.)
- **symlink isolation**: a managed base that is itself a symlink is rejected as `base_symlink` (target not scanned); nested symlinks (dir or file) are excluded from recursion/yield.
- **containment**: each file `resolve(strict=True)` + `is_relative_to(base_resolved)`.
- read-only / JSON contract preserved.

Direction and implementation are both sound. Skipping symlink dirs *before* recursion (`S_ISLNK` check) means there is no symlink-loop infinite-recursion risk, and it closes the old `rglob` `**`-follows-symlinked-dir path that could read out-of-scope content through a base symlink.

### Correctness — verified good
- `list(_iter_audit_files_fail_closed(base))` is fully materialized inside the `try`, so an `OSError` raised mid-recursion propagates to the caller and is labeled via `exc.filename`. ✅
- The walker only yields non-symlink regular files reached through real directories, so `resolve()` cannot escape `base` — `containment_failed` is effectively defense-in-depth against a TOCTOU race (dir swapped to a symlink between walk and resolve). Not dead code; intentional belt-and-suspenders. ✅
- All new error kinds feed `errors` → `if violations or errors: raise SystemExit(1)`; non-JSON output renders them generically as `ERROR {path} ({error})`. ✅
- `base_symlink` continues without scanning, so no target content is read — the `_SECRET_CONTENT not in output` assertion backs the fail-closed / no-leak contract. ✅

### Findings (all low severity — non-blocking)

- **[minor] Errors are not deduped across duplicate roots.** `seen` dedups scanned files, but base-level errors (`enumeration_failed`, `base_symlink`) and the errors appended *before* the `seen` check (`read_failed` from `resolve()`, `containment_failed`) are outside that dedup. `all_index_roots()` can yield duplicate roots (the test configures `[tmp_path, tmp_path]`, which is why `seen` exists), so a duplicate root + an enumeration failure emits the **same error twice**. `test_all_managed_roots_...` uses an error-free case, so it doesn't catch this. Cosmetic only (exit code stays 1). Skipping an already-seen base up front would tidy it.

- **[nit] Import order.** `from collections.abc import Iterator` is placed after `from pathlib import Path`; isort ordering would put `collections.abc` first. Not a lint failure (ruff `select` doesn't enable the `I` rules), so it's consistent with "Ruff passed" — pure style.

- **[note] Traversal order changed.** Global `sorted(rglob("*"))` → per-level `sorted(iterdir())` DFS. Still deterministic and `seen` dedup is identity-based so results are unchanged, but the *output order* of `violations`/`errors` differs. Harmless unless a downstream consumer asserts ordering.

### Test coverage — strong
Six new cases: happy path, duplicate roots, enumeration failure, read failure, base symlink, nested symlinks (dir + file), plus byte-invariance and no-secret-in-output. `@pytest.mark.requires_symlinks` is registered and auto-skips on Windows. The `_configure` helper extraction is a nice cleanup.

### Verdict: **SHIP** ✅
Core fail-closed logic is correct, symlink isolation and loop protection are solid, and the tests are thorough. The findings above are all cosmetic and do not block. If anything is worth a follow-up, it's the duplicate-root error dedup.

One process note: `base` is a release branch (`release/plugin-0.2.2`) stacked on #1720→#1719→#1718 — retarget to `--base main` as the dependencies merge and mind the merge order.

tsdata (approved):

PR #1725, opened 2026-07-12
security(web): add disk-backed upload quarantine

memtomem:
## Summary

- parse multipart uploads explicitly with a 32-file / zero-field boundary
- copy parsed files into an owner-only disk-backed quarantine before privacy adjudication
- enforce per-file and aggregate limits during quarantine copying
- atomically promote accepted files without overwriting concurrent same-name uploads
- clean parser spools and quarantine artifacts on limit, decode, privacy, indexing, cancellation, and unexpected failures
- preserve generic client errors and server-side diagnostic logging

Index failures intentionally keep an already promoted file and report a generic per-file error, matching the reviewed persistence semantics. ADR-0029 remains unchanged.

## Verification

- `uv run ruff check packages/memtomem/src/memtomem/web/upload_quarantine.py packages/memtomem/src/memtomem/web/routes/system.py packages/memtomem/tests/web/test_upload_quarantine.py packages/memtomem/tests/test_web_routes.py`
- `uv run ruff format --check packages/memtomem/src/memtomem/web/upload_quarantine.py packages/memtomem/src/memtomem/web/routes/system.py packages/memtomem/tests/web/test_upload_quarantine.py packages/memtomem/tests/test_web_routes.py`
- `uv run mypy packages/memtomem/src/memtomem/web/upload_quarantine.py`
- `uv run pytest packages/memtomem/tests/web/test_upload_quarantine.py packages/memtomem/tests/test_web_routes.py -q` (223 passed)
- `git diff --check`

Closes #1722

### DIFF
--- a/packages/memtomem/src/memtomem/web/routes/system.py
+++ b/packages/memtomem/src/memtomem/web/routes/system.py
@@ -21,7 +21,7 @@ import sys
 from pathlib import Path
 from typing import Any
 
-from fastapi import APIRouter, Depends, File, HTTPException, Request, UploadFile
+from fastapi import APIRouter, Depends, HTTPException, Request
 from fastapi.responses import JSONResponse, StreamingResponse
 
 from memtomem.config import (
@@ -1474,15 +1474,34 @@ async def preview_namespace(
 
 
 _ALLOWED_UPLOAD_EXTS = {".md", ".txt", ".json", ".yaml", ".yml", ".toml"}
-_MAX_UPLOAD_FILES = 32
-_MAX_UPLOAD_BYTES = 100 * 1024 * 1024
-_MAX_UPLOAD_AGGREGATE_BYTES = 200 * 1024 * 1024
-_UPLOAD_CHUNK_BYTES = 1024 * 1024
 
 
-@router.post("/upload", response_model=UploadResponse, dependencies=[Depends(require_configured)])
+@router.post(
+    "/upload",
+    response_model=UploadResponse,
+    dependencies=[Depends(require_configured)],
+    openapi_extra={
+        "requestBody": {
+            "required": True,
+            "content": {
+                "multipart/form-data": {
+                    "schema": {
+                        "type": "object",
+                        "required": ["files"],
+                        "properties": {
+                            "files": {
+                                "type": "array",
+                                "items": {"type": "string", "format": "binary"},
+                            }
+                        },
+                    }
+                }
+            },
+        }
+    },
+)
 async def upload_files(
-    files: list[UploadFile] = File(...),
+    request: Request,
     force_unsafe: bool = False,
     index_engine=Depends(get_index_engine),
 ) -> UploadResponse:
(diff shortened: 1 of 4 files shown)


tsdata (approved):

memtomem:
## Code Review — disk-backed upload quarantine

### Overview
Replaces in-memory upload buffering in `/api/upload` with an explicit multipart parse + disk-backed quarantine flow. Files stream into an owner-only `.quarantine-*` temp dir (inside `~/.memtomem/uploads`, `0o700`), are privacy-adjudicated there, and only accepted files are atomically hard-linked into the uploads dir. Adds a bounded parser (32 files / 0 fields), per-file + aggregate byte limits, and cleanup on every failure path.

Reviewed against the branch and Starlette 1.3.1 internals.

### Strengths
- **Security model is sound.** Unadjudicated content never lands in the indexable uploads dir — it lives only in a hidden `0o700` quarantine subdir until `enforce_write_guard` passes. Confirmed `~/.memtomem/uploads` is not a watched/auto-discovered `memory_dir`, so quarantine `.part` files can't be picked up before promotion.
- **Path-traversal safe.** Quarantine names are index-derived (`0000.part`); final names go through `Path(...).name` twice.
- **Atomic no-overwrite promotion.** `os.link` + `O_EXCL` + random-suffix fallback correctly handles concurrent same-name uploads, and non-linkable filesystems (`EXDEV/EPERM/ENOTSUP`) fail closed rather than corrupting.
- **Cancellation-safe cleanup.** `_close_form` and `shutil.rmtree` run synchronously in `finally`, so `CancelledError` can't interrupt cleanup mid-way. The outer handler re-raises `CancelledError` and translates everything else to a generic 500 — no internals leak.
- **Error classification** maps parser failures to correct statuses (413/400/422) with server-side `logger.exception` diagnostics.
- **Excellent test coverage** — exact/+1 boundaries for per-file, aggregate, and file-count limits; every cleanup path; owner-only mode assertions; OpenAPI contract; body-limit middleware.

### Minor findings (all low severity — none blocking)
- **`_ClosingMultiPartParser` depends on Starlette's private `_files_to_close_on_error`.** Legitimate defense-in-depth (base `parse()` only closes spools on `MultiPartException`/`OSError`). The text-field test exercises the override's `except` branch, so a Starlette rename surfaces as CI failure rather than silently — but a one-line comment pinning the Starlette-internal assumption would be worth adding.
- **413-vs-400 classification via message substring** (`"Too many files" in str(exc)`) is fragile against upstream wording changes, but guarded by the 33-file → 413 test. Acceptable.
- **Redundant `prepare_upload_dir`/`os.chmod`** — called once in `quarantine_uploads` and again per accepted file in `promote_no_overwrite` (up to 32 extra dir `chmod`s). Harmless.
- **Aggregate limit checked only after each full file copy**, so one file can push on-disk bytes up to `MAX_UPLOAD_BYTES` past the aggregate limit before rejection — bounded harmlessly by the 201 MiB body-limit middleware.
- **`promote_no_overwrite`: a spurious `source.unlink()` failure discards an already-valid promotion** and re-raises. Extremely unlikely and errs toward the safe direction.

### Verdict
**SHIP.** Correct, well-tested, and a genuine security improvement over in-memory buffering. No correctness or security defects found. The only note worth acting on is a short comment documenting the Starlette-internal dependency in `_ClosingMultiPartParser`.

pandas-studio:
## Post-merge review (requested via /review of #1722; this PR implements it)

### Overview

Replaces the in-memory upload path in `POST /api/upload` with an explicit `MultiPartParser` (`max_files=32`, `max_fields=0`) and a disk-backed quarantine under the upload directory. Per-file and aggregate limits are enforced while copying into quarantine; privacy adjudication runs only after the batch is fully quarantined; accepted files are promoted via `os.link` with no-overwrite semantics; parser spools and quarantine artifacts are cleaned on limit, decode, privacy, indexing, cancellation, and unexpected failures. `openapi_extra` preserves the previous multipart schema contract, and a dedicated test module covers boundaries, lifecycle cleanup, permissions, concurrency, and the OpenAPI contract.

### What's done well

- **Correct ordering per the issue contract**: count limit at parse time (before handler-level reads — the 33-file test asserts the upload dir is never even created), size limits during quarantine copy (before any in-memory batch can form), privacy guard only after quarantine. Peak memory is now one file at a time (`read_bytes` per iteration) instead of the whole batch.
- **Solid filesystem hygiene**: `mkdtemp` (0o700, unpredictable name) + `O_WRONLY|O_CREAT|O_EXCL` at 0o600 resists symlink games; `Path(filename).name` strips traversal; promotion by hard link never replaces an existing path, with `EXDEV/EPERM/ENOTSUP` surfaced as an explicit failure rather than a silent copy fallback. Concurrent same-name uploads are actually tested with `asyncio.gather`.
- **Cleanup paths are honest**: `_ClosingMultiPartParser` covers the spool leak Starlette doesn't handle on cancellation; `_close_form` is deliberately synchronous so cancellation can't interrupt cleanup at an await point; `CancelledError` is re-raised rather than converted to a 500.
- **No detail leakage**: client-facing errors stay generic (asserted in tests, e.g. the "sensitive server detail" → generic-500 test), diagnostics go to server logs; the index-failure path preserves the reviewed keep-promoted-file semantics and is pinned by a test.
- The private-attr access (`_files_to_close_on_error`) and the "Too many files" substring match are version-coupled, but both are exercised by tests (the `max_fields=0` path and the 33-file 413 assertion respectively), so a Starlette bump that breaks either fails loudly, as the inline comment claims.

### Findings (all minor; PR is already merged — candidates for follow-ups)

1. **Orphaned quarantine dirs after a hard crash.** A `kill -9`/power loss between `mkdtemp` and the `finally` leaves `.quarantine-*` directories in `~/.memtomem/uploads` forever — there is no startup sweep or reaper. Suggest a follow-up: remove stale `.quarantine-*` **directories** (age-gated) at app lifespan startup. Note the reaper must match directories only: a user can legitimately upload a file named `.quarantine-x.md`, which promotion would happily place in the upload dir.
2. **Aggregate limit is checked only after each file finishes copying.** Transient quarantine usage can reach ~`MAX_UPLOAD_AGGREGATE_BYTES + MAX_UPLOAD_BYTES`, on top of Starlette's parser spools (bounded by the 201 MiB request middleware) — worst case roughly 2× the request size on disk at once. Passing the remaining aggregate budget into `_copy_to_quarantine` would fail earlier and halve the transient footprint. Not a correctness bug.
3. **`except BaseException` in the handler converts `KeyboardInterrupt`/`SystemExit`/`GeneratorExit` into HTTP 500** (only `CancelledError` is re-raised). Rare in a request handler, but `if not isinstance(exc, Exception): raise` would be the safer shape.
4. **Test collection is import-based and silently zero when run directly.** `__test__ = False` + re-import into `test_web_routes.py` works today (verified: the PR run collected them), but (a) `pytest tests/web/test_upload_quarantine.py` collects 0 tests without warning, and (b) it rides on pytest's `collect_imported_tests` default, which is now a configurable knob — a future default flip would un-collect these silently. Consider pinning `collect_imported_tests = true` in config, or eventually lifting the shared app/client fixtures into a conftest so the module can stand alone.
5. Nit: `os.chmod(destination, 0o600)` after `os.link` is redundant — the quarantine file is already 0o600 and hard links share the inode. Harmless.

`force_unsafe=true` upload bypass coverage lives in the pre-existing `test_web_routes.py` cases (unchanged by this PR), so the acceptance list from #1722 is fully covered between the two files.

**Verdict: solid hardening, correctly ordered, well-tested. Findings above are follow-up material, none block.**

PR #1729, opened 2026-07-12
feat(web): harden startup readiness and search filters

memtomem:
## Summary
- classify storage startup failures and clean up partially initialized components
- add readiness and missing-state 503 contracts, watcher reconfiguration, and truthful initial-index outcomes
- move exact source/type/date filtering into backend search and align export namespace stats
- repair hash/history navigation and add actual-lifespan golden QA coverage

## Validation
- Python suite: 8,440 passed; 34 HOME-writing tests passed separately outside the sandbox
- JavaScript: 69 files, 461 tests passed
- ruff check and format check passed
- mypy passed for 12 changed source modules
- git diff --check passed
- isolated actual-lifespan flow verified: readiness, registration/indexing, exact search, export parity, clean shutdown

## Browser note
- Playwright golden test is included and auto-skips when Chromium is unavailable; local Chromium was not installed.

### DIFF
--- a/packages/memtomem/src/memtomem/cli/web.py
+++ b/packages/memtomem/src/memtomem/cli/web.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import atexit
 import contextlib
+import http.client
 import json
 import os
 import signal
@@ -252,7 +253,11 @@ def _run_foreground(config: _WebRunConfig) -> None:
         f"Starting memtomem Web UI at http://{config.host}:{config.port} (mode={resolved_mode})"
     )
 
-    async def after_started(server: uvicorn.Server, timeout: float) -> None:
+    async def after_started(
+        server: uvicorn.Server,
+        serve_task: asyncio.Task[None],
+        timeout: float,
+    ) -> None:
         if not config.open_browser:
             return
 
@@ -265,9 +270,19 @@ def _run_foreground(config: _WebRunConfig) -> None:
         else:
             deadline = time.monotonic() + timeout
         while not server.started:
+            if serve_task.done():
+                return
             if time.monotonic() >= deadline:
                 click.secho(
                     "Warning: Web server did not start within the timeout period; not opening browser.",
+                    fg="yellow",
+                )
+                return
+            await asyncio.sleep(0.1)
+        while not await asyncio.to_thread(_readiness_ok, config.host, config.port):
+            if serve_task.done() or time.monotonic() >= deadline:
+                click.secho(
+                    "Warning: Web backend did not become ready; not opening browser.",
                     fg="yellow",
                 )
                 return
@@ -294,10 +309,13 @@ def _run_foreground(config: _WebRunConfig) -> None:
         )
         web_server = uvicorn.Server(web_config)
 
-        await asyncio.gather(
-            web_server.serve(),
-            after_started(web_server, timeout=float(config.timeout)),
+        serve_task = asyncio.create_task(web_server.serve())
+        opener_task = asyncio.create_task(
+            after_started(web_server, serve_task, timeout=float(config.timeout))
         )
+        await asyncio.gather(serve_task, opener_task)
+        if not web_server.started:
+            raise click.ClickException("Web UI failed during startup. See the startup log above.")
 
     with _web_pid_lock(config.port):
         asyncio.run(start_server())
@@ -330,6 +348,36 @@ def _wait_for_tcp(host: str, port: int, *, timeout: float, child: subprocess.Pop
                 return True
         except OSError:
             time.sleep(0.1)
+    return False
+
+
+def _readiness_ok(host: str, port: int) -> bool:
+    conn = http.client.HTTPConnection(_connect_host(host), port, timeout=0.5)
+    try:
+        conn.request("GET", "/api/readiness")
+        response = conn.getresponse()
+        response.read()
+        return response.status == 200
+    except (OSError, http.client.HTTPException):
+        return False
+    finally:
+        conn.close()
+
+
+def _wait_for_readiness(
+    host: str,
+    port: int,
+    *,
+    timeout: float,
+    child: subprocess.Popen[bytes],
+) -> bool:
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        if child.poll() is not None:
+            return False
+        if _readiness_ok(host, port):
+            return True
+        time.sleep(0.1)
     return False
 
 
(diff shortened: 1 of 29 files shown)


tsdata (approved):

memtomem (commented):
## Review — startup readiness & search filters

Large but cohesive hardening pass. No blocker-class correctness bug found; the atomic `app.state` publish, readiness gating, and error classification/redaction are all solid. Notes below are mostly low/medium.

### Correctness

- **`_matches_metadata` naive-vs-aware `created_at` (medium).** In `search/pipeline.py`, `chunk.created_at < metadata_filter.created_from` compares against an always-tz-aware bound (the route forces UTC). If any indexed chunk has a **naive** `created_at` (legacy/migrated rows deserialized from an offset-less ISO string), this raises `TypeError` and the search 500s — but only on the new date-filter path, so the golden test (aware `now` only) won't catch it. Suggest normalizing naive → UTC in `_matches_metadata`, or skipping the comparison when tzinfo differs. The SQL leg is unaffected; this is only the post-fusion filter.

- **SQL `created_at` is a lexicographic string compare (low).** `_metadata_filter_sql` compares the stored ISO string against `created_from.isoformat()`. Correct **only** because stored timestamps are UTC (`+00:00`, `timespec="seconds"`) and `app.js` sends `toISOString()` normalized to the same shape. An external API caller passing a non-UTC-derived value with microseconds, or a DB with any non-UTC stored offset, can mis-order. Worth a one-line comment pinning the "stored timestamps are UTC isoformat" assumption next to the fragment.

- **Context-window expansion re-applies the full metadata filter (low, design).** After `_expand_context`, `fused` is re-filtered by `_matches_metadata`. `source_exact` neighbors pass (same file), but a **chunk_type** filter strips adjacent neighbors of a different type — defeating context expansion for that filter. If surrounding context should still surface under a type/date filter, consider re-applying only `source_exact` post-expansion.

- **`activateTab` redirect branch duplicated + stale comment (nit).** The advanced-only handler runs the identical `activateTab(_visibleMainTabs()[0]||'home', {historyMode:'replace'})` in both `if` and `else`. The `popstate` comment still says *"Staying put is the safe floor"* though it now redirects to the first visible tab — update or drop it.

- **`popstate` + `hashchange` double-fire (nit).** Back/forward across a hash change fires both listeners; both call `activateTab(..., {historyMode:'none'})`. Idempotent so harmless, but redundant. The new `hashchange` listener is only needed for manual URL edits — could early-return when the active tab already matches.

### Style / simplification

- **`_filter_only_search` if/else duplication.** `recall_chunks` now defaults `metadata_filter=None` (it's in the Protocol), so the two near-identical calls collapse to one passing `metadata_filter=metadata_filter`. Elsewhere the `**metadata_kwargs` splat handles the same backward-compat concern — pick one pattern.

- Startup rollback ordering verified correct: `SearchPipeline.close()` closes the reranker but not the LLM, so the `if search_pipeline is None: close(reranker)` guard avoids a double-close and the shared `llm` closes exactly once in both rollback and `close_components`. 👍

### Test coverage

- **`FileWatcher.reconfigure()` has no non-browser test.** The desired/current set-diff, the `unschedule` path, and especially the **rollback-on-exception** branch aren't exercised by the Python suite — the only end-to-end exercise is the `@pytest.mark.browser` golden, which auto-skips in CI without Chromium. A small unit test (fake observer; assert schedule/unschedule + rollback re-raise) would lock in nontrivial new logic.
- Startup classification, the 503 contract, exact-filter threading, and `index_status` are all well covered.

### Security

- Redaction is handled correctly: `_classify_startup_error` never embeds the original exception text (test asserts `"boom" not in str`), and `add_memory_dir` logs the raw error but returns a generic `"Initial indexing failed"` (test asserts no path leak). `/api/readiness` is `no-store` and exposes only a coarse `reason_code`. Consistent with the repo's canonical-path-leak conventions.

### Conventions

- Search-pipeline stage order (CLAUDE.md invariant) preserved — the metadata filter runs inside retrieval SQL and as a post-fusion filter, not as a reordered stage. ✅
- `app.js?v=150` / `sources-memory-dirs.js?v=15` bumped with the parity test updated; en/ko locale keys added in pairs. ✅

**Verdict: ship-able.** Recommend addressing the naive-datetime guard (medium) and the `reconfigure()` unit test (low) before merge; the rest are optional cleanups.

memtomem:
Addressed all review notes in 73bb217a:

- normalized legacy naive chunk timestamps to UTC before metadata-bound comparisons
- documented the UTC ISO storage assumption for SQL lexical date filtering
- preserved cross-type/date context neighbors while retaining the exact-source boundary
- simplified filter-only recall and cleaned up hash/history navigation
- added deterministic retrieval-debug initialization for the Browser CI failure
- added FileWatcher reconfigure reconciliation and rollback tests

Validation: 268 browser tests passed, 461 JS tests passed, 73 search/context tests passed, 22 focused contract/watcher/storage tests passed, Ruff/format/mypy/diff checks passed.

memtomem (commented):
## Re-review — after `73bb217a` (fix: address startup QA review feedback)

Every point from the prior review is resolved correctly, and the watcher rollback fix goes further than flagged. Traced each change against PR head. **Verdict: ship-able, nothing blocking remains.**

### Findings status

- **`_matches_metadata` naive/aware `created_at` (was medium)** — ✅ Fixed. Normalizes to UTC (naive → `replace(tzinfo=UTC)`, aware → `astimezone(UTC)`) with a comment on the storage contract; regression test `test_post_filter_treats_legacy_naive_timestamp_as_utc`.
- **SQL `created_at` lexical-compare assumption (was low)** — ✅ Comment added pinning "stored as UTC ISO-8601, bounds normalized to UTC."
- **Context expansion stripped different-type neighbors (was low/design)** — ✅ Post-expansion now re-applies **only** `source_exact`, keeping type/date-divergent context neighbors; `test_chunk_type_filter_keeps_different_type_context_neighbors` asserts the before/after windows survive.
- **`activateTab` duplicated if/else + stale comment (nit)** — ✅ Collapsed to one call; comment updated.
- **`popstate` + `hashchange` double-fire (nit)** — ✅ `hashchange` now guards `tab !== activeTab`, so it no-ops after `popstate` (traversal fires popstate→hashchange, order is safe).
- **`_filter_only_search` if/else duplication (style)** — ✅ Single `recall_chunks(..., metadata_filter=...)` call.
- **`FileWatcher.reconfigure()` untested (low)** — ✅ Two unit tests: add/remove reconciliation + rollback-on-removal-failure.

### Bonus correctness improvement (beyond the review)
The `reconfigure()` rollback was previously asymmetric — it only un-did *added* watches, leaving *removed* roots un-watched if a later `unschedule` threw. Now it splits `unschedule` from the `pop` (a failed unschedule leaves the entry intact so the dict stays truthful) and re-schedules `removed` roots on rollback. Traced the rollback test: the error fires during `unschedule(old_watch)` before the pop/append, so `removed` stays empty, `old_root` remains watched, only the added `new_watch` is unscheduled, and `_config` is left unmutated — assertions hold. Rollback is now symmetric.

### Tangential change
`STATE.showRetrievalDebug: false` is wired (read at the retrieval-debug toggle + `show-retrieval-debug` body-class gate), not dead state; the smoke-matrix switch to `wait_for_function` is a flake fix for the debug-meta `display:none` gate. Default behavior unchanged.

### Remaining (optional, non-blocking)
- **Dense-search perf under highly selective metadata filters.** The metadata clause sits in the outer JOIN after the vec `LIMIT`, so a very selective filter can force the adaptive over-fetch to escalate `inner_k` toward a full `chunks_vec` scan. Bounded and pre-existing in shape — just worth awareness for slow filtered vector search on large DBs.

**SHIP.** All feedback addressed with tests.

Issue #1587, opened 2026-07-03
mcp/web: cross-process CRUD read-modify-write is still unserialized (follow-up to #1570)


memtomem:
## Summary

The per-file lock added in #1570 (`AppContext.get_memory_file_lock`) serializes concurrent memory-CRUD **within one MCP server process only**. Three markdown-mutating surfaces remain unserialized against it and share the same stale-line-range / lost-update / rollback-erase shape #1570 fixed for the MCP tools:

## Where

1. **Web API** — `packages/memtomem/src/memtomem/web/routes/chunks.py:120` (PATCH via `replace_chunk_body`) and `chunks.py:181` (DELETE via `remove_lines`), plus web add at `web/routes/system.py`. `mm web` stores components on `app.state` (`web/app.py`), not `AppContext`, so the `AppContext`-scoped lock does not cover concurrent web requests, nor web-vs-MCP when both run.
2. **`memory-migrate`** — `mem_context_memory_migrate` / `mm context memory-migrate` take the engine sidecar `_file_lock` before `shutil.move()` + `update_chunks_scope_for_source()`, not `get_memory_file_lock`. So a CRUD edit can re-fetch "fresh" under the async lock and migrate can still move the file / update the DB mid-span. `_locked_chunk`'s re-key only covers a migrate that finished *before* the edit locked.
3. **Cross-process in general** — a second MCP server process, or the CLI (`mm mem add`) racing the MCP server, are serialized by neither lock for the read→rewrite step (only the `index_file` step takes the sidecar lock).

## Why it's separate from #1570

Closing this properly needs a **cross-process CRUD critical section**: hold the sidecar `_file_lock` across the whole read → rewrite → re-index span, which requires refactoring `IndexEngine.index_file` so it can run with the sidecar lock *already held* (skip the nested re-acquire) instead of self-deadlocking, and acquiring the sidecar off the event loop with a bounded timeout so a blocking `portalocker` acquire doesn't stall the loop. That's a materially larger change than #1570's in-process lock and touches the indexing/lock boundary, so it was scoped out.

## Suggested direction

- Extract the lock registry into shared components/app state so web + MCP share one in-process lock namespace.
- Add an `index_file(..., lock_held=True)` (or an internal `_index_file_locked`) path so CRUD can hold the sidecar `_file_lock` across the RMW without the nested acquire deadlocking.
- Make `memory-migrate` and the web writers participate in the same lock discipline (source + target sidecars in canonical order for migrate).

[...]

PR #1732, opened 2026-07-13
feat: add pinned context and retrieval quality baseline

memtomem:
## Summary

- establish a public, synthetic v0.3.8 retrieval corpus contract with 48 files, 192 chunks, and 100 Korean/English queries
- add privacy/integrity auditing, reproducible retrieval baseline metadata, and CI regression checks
- add a LangGraph `BaseStore`-compatible persistent adapter
- add file-backed Pinned Context with scope precedence, privacy gates, CLI commands, and MCP actions
- add review-first memory formation candidates plus assertion/relation persistence
- document nine user-facing scenarios that communicate memtomem strengths
- update the documented full MCP surface from 87 to 95 tools

## Verification

- `uv run python tools/retrieval-eval/audit_public_corpus.py`
  - 48 files / 192 chunks / 100 queries
  - 0 secret hits / 0 disallowed emails / 0 global IP literals
- `uv run python tools/retrieval-eval/check_baseline.py`
- `uv run ruff check packages/memtomem/src packages/memtomem/tests tools/retrieval-eval`
- `uv run ruff format --check packages/memtomem/src packages/memtomem/tests tools/retrieval-eval`
- `uv lock --check`
- `git diff --check`
- regression suite excluding browser/LLM integration and `test_tools_logic.py`: **8390 passed, 11 skipped**

The full run exposed the pre-existing `_fake_ctx.get_memory_file_lock` test-stub regression tracked separately in #1731; no product-code workaround is included here.

## Scope boundaries

- no `memtomem-stm` dependency introduced
- core MCP mode remains at 9 exposed tools; new actions remain registry-routed
- existing private/local `docs/plans/` and `docs/qa/` files are not included


tsdata (approved):

memtomem:
## Review — NEEDS-FIX

Read the full diff (42 files, +4,917) and cross-checked suspect paths against source. Two blockers, one structural race, plus medium/minor items below.

### Blockers

**B1. `MemtomemStore` import regression on installs without the new `langgraph` extra**

`integrations/langgraph.py` now does a module-level `from memtomem.integrations.langgraph_store import MemtomemBaseStore`, and `langgraph_store.py` raises `ImportError` at import time when the `langgraph` package is missing. The existing `langgraph.py` adapter had **zero** dependency on the langgraph package, so after this PR every existing `from memtomem.integrations.langgraph import MemtomemStore` breaks unless the user installs the new extra. CI can't catch this because it syncs `[all]`, which now includes `langgraph`, and there is no minimal-install job (the "minimal-install smoke" mentioned in the `langgraph_store.py` docstring does not exist in `ci.yml`).

Fix: re-export lazily via PEP 562 module `__getattr__` (or drop the re-export and document importing `MemtomemBaseStore` from `memtomem.integrations.langgraph_store` directly). Also add `pytest.importorskip("langgraph")` to `test_langgraph_basestore.py`.

**B2. `mm review approve` appends to the daily memory file without the #1587 L2 sidecar lock**

`review_cmd._decide` (memory destination) calls `append_entry(target, ...)` + `index_file(target, already_scanned=True)` with no lock. `mm add` writes the same `YYYY-MM-DD.md` while holding `async_file_lock(_lock_path_for(target))` across append + reindex and passes `lock_held=True` (`cli/memory.py:265`) — its comment spells out the exact hazard: a concurrent MCP `mem_edit`/`mem_delete` rollback in the server process can erase the appended entry. This new CLI path is exposed to the same scenario. Wrap append + reindex in the sidecar lock and pass `lock_held=True`, mirroring `mm add`.

### Major

**B3. Approve is write-first, decide-second → duplicate writes under concurrent approval**

Both CLI `_decide` and MCP `mem_candidate_review` perform the durable write first and only then run `decide_memory_candidate` (`UPDATE ... WHERE status='pending'`). Two concurrent approvals both pass the pending pre-check, both write (duplicate appended entries), then one loses the decide and reports an error — after the side effect already landed. Invert the order: use the atomic pending→approved UPDATE as the claim, then write, and revert the claim if the write fails (matches the existing idempotency claim/release convention).

### Medium

- `mem_candidate_review` detects `mem_add` failure by substring match (`"blocked" in ... or "error" in ...`). Candidate content containing the word "error" causes a false failure; a reworded error message silently approves without a write. Use a structured result or exception instead.
- The `fact` pattern in `formation._KIND_PATTERNS` (`\b(fact|uses?|is|are|runs?)\b|사실|사용|이다|입니다`) matches nearly every English/Korean declarative sentence, so almost all session events become candidates at hardcoded confidence 0.8 — the review queue will be mostly noise. Narrow the pattern or differentiate confidence by kind.
- Docs drift: `mcp-clients.md` heading says "Available MCP Tools (95)" but the table still enumerates the old 87 (none of the 8 new tools appear); `reference.md` per-category counts and the package README's "All 95 tools with parameters" detail list are also not updated.
- `FormationMixin.get_memory_candidate` finds one row by listing all four statuses at `limit=10_000` (each triggering the expiry UPDATE). Use `SELECT ... WHERE id = ?`; candidates beyond 10k are currently invisible.

### Minor

- `add_assertion`: `INSERT OR IGNORE` into `canonical_entities` means an existing `(canonical_name, entity_type)` under a different id leaves the new assertion pointing at a nonexistent entity — invisible to `query_assertions`. Needs lookup-or-create.
- `MemtomemBaseStore._records()` raises on one corrupt JSON file, breaking every list/search, while `PinnedContextStore._read_scope` skips corrupt files — pick one policy. `_embedding_cache` also grows unboundedly.
- `json.dumps` without `ensure_ascii=False` in `review scan` / formation tools escapes Korean content; inconsistent with sibling calls.
- `pinned_cmd._with_store`: if `PinnedContextStore(...)` raises after the manual `__aenter__`, the components context leaks. Prefer `AsyncExitStack`.

### Positives

- Security posture is solid: id validation blocks traversal, base64-encoded namespace segments, `yaml.safe_load/safe_dump`, parameterized SQL throughout, `enforce_write_guard` on every new write ingress with the project_shared `force_unsafe` hard-refusal preserved, and the corpus audit replaces the realistic AWS key example with a `<SYNTHETIC_...>` placeholder and pins it.
- Schema changes are additive-only; the `get_session_events` `id` addition is backward-compatible.
- Benchmark design is careful: pinned corpus/query hashes, blocking quality floors with opt-in fingerprint/latency checks, and indexing completeness (48/192/0-blocked) as a calibration validity condition.

### Process notes

- This bundles four independent features + a benchmark + docs in one PR. Recommend splitting into (a) benchmark + CI gate, (b) BaseStore adapter, (c) Pinned Context, (d) review-first formation, (e) docs — each lands reviewable and revertable on its own.
- `docs/guides/use-cases.md` is a Korean-language marketing doc; committed docs are English-only, and both new guides describe features first landing in this PR, so they belong in the private docs repo until the features stabilize.
- The retrieval-floor gate is blocking inside the required golden-path job, but the baseline was calibrated on macOS arm64 while CI runs ubuntu. The 15% margin may hold, but be ready to demote to advisory if it flakes.

memtomem:
Addressed the review findings in `97a0b6dd`:

- kept `MemtomemStore` importable without the optional LangGraph dependency via lazy `__getattr__`; added a minimal-install subprocess regression test and `importorskip` for BaseStore tests
- added atomic `pending -> writing` candidate claims, retry release, and finalization so concurrent approvals cannot duplicate writes
- wrapped CLI memory approval append + reindex in the #1587 sidecar lock and passed `lock_held=True`
- replaced MCP error-string matching with the structured `_mem_add_core` `(message, stats)` result
- narrowed generic fact extraction and assigned per-kind confidence values
- replaced candidate 10k scans with direct id lookup; added expiry-aware claims
- resolved canonical entities by `(canonical_name, entity_type)` before assertion insertion
- made BaseStore reads skip corrupt records and bounded the embedding LRU cache to 1,024 entries
- emitted Korean JSON without ASCII escaping and moved Pinned CLI lifecycle to `AsyncExitStack`
- documented all eight new MCP tools and converted the public use-case guide to English

Verification: 205 relevant CLI/server/integration tests passed; Bandit 1.9.4, Ruff, format check, focused mypy, and diff check passed. The broad local suite process was terminated twice by the local runner at the same point, so the fresh GitHub CI matrix is the final broad regression check.

memtomem:
## Re-review of 97a0b6dd — SHIP (pending required test jobs)

All 9 findings from the previous review are addressed; each fix was verified against source, not just the diff.

### Verified fixed

- **B1 (import regression)**: the eager import in `integrations/langgraph.py` is replaced with a PEP 562 `__getattr__` lazy re-export + `TYPE_CHECKING` import. The new `test_langgraph_optional_import.py` proves `MemtomemStore` imports in a subprocess with `langgraph` import-blocked, and `test_langgraph_basestore.py` gained `pytest.importorskip("langgraph")`. The minimal-install contract is now pinned by a test.
- **B2 (sidecar lock)**: the `review approve` memory path now holds `async_file_lock(_lock_path_for(target), timeout=_CRUD_SIDECAR_LOCK_BUDGET_S)` across append + `index_file(..., lock_held=True)` — same #1587 pattern as `mm add`.
- **B3 (write-first race)**: replaced with a proper two-phase claim — atomic `pending→'writing'` UPDATE (with expiry check) before the write, `finalize` (`'writing'→'approved'`) after, and `release` back to pending on failure, with `asyncio.CancelledError` handled explicitly. Applied to both CLI and MCP paths; claim atomicity and release-retry are tested.
- **String-matched failure detection**: `mem_candidate_review` now calls `_mem_add_core` directly; verified its `tuple[str, IndexingStats | None]` contract — the `stats is None` failure check is structurally sound.
- **Over-matching `fact` pattern**: narrowed to explicit signals (`fact:`, `runs on|depends on|uses X for`, Korean equivalents) with per-kind confidence (0.95–0.75); negative test confirms generic declaratives ("The service is available", "서비스입니다") no longer become candidates.
- **Docs drift**: Pinned Context / Formation rows added to the tool tables in `mcp-clients.md` and `reference.md`; the package README now links to the complete table instead of claiming per-tool parameter docs. `use-cases.md` rewritten in English with the marketing tone removed.
- **`get_memory_candidate`**: direct `WHERE id = ?` lookup with targeted expiry; the 10k invisibility window is gone (tested).
- **`add_assertion` entity loss**: entity id is re-resolved by `(canonical_name, entity_type)` after the `INSERT OR IGNORE`; reuse is tested.
- Also landed: unified corrupt-JSON skip policy (+test), LRU-bounded embedding cache, `ensure_ascii=False`, `AsyncExitStack` in `pinned_cmd`, and `index is True` handling in `_projection`.

### Remaining

- **[Minor — suggest a follow-up issue]** A crash between claim and finalize/release strands a candidate in `'writing'` permanently: the expiry sweep and reject both touch only `'pending'`. It is visible via `mm review list --status writing` but there is no recovery command. A stale-`'writing'` sweep (e.g. older than N minutes → back to `pending`) would close the gap.
- Process notes stand unchanged: the PR still bundles four independent features (revert granularity), and the blocking retrieval-floor gate in the required golden-path job was calibrated on macOS arm64 while CI runs ubuntu — worth watching for flakes in the first days after merge.
- Merge gate: `test (ubuntu/macos)` and both windows shards were still pending at review time — merge on green as usual.

PR #1735, opened 2026-07-13
fix: recover stale memory formation claims

memtomem:
## Summary

- timestamp approval claims when candidates enter `writing`
- backfill pre-upgrade NULL `writing` timestamps at migration time so they receive one full grace period
- atomically recover only claims older than a timezone-aware cutoff with a SQLite-compatible `BEGIN IMMEDIATE` select/update transaction
- serialize recovery against finalize so only one status transition can succeed
- quarantine a durable write that loses finalization as `write_uncertain`, preventing blind duplicate re-approval and explicitly telling CLI/MCP operators that the write already persisted
- persist candidate status transitions with actor, reason, and timestamp for auditability
- expose recovery through `mm review recover` and `mem_candidate_recover` with a conservative 15-minute default
- add the forward nullable-column migration while retaining the existing schema fence version, consistent with prior additive nullable migrations
- update the 96-tool documentation

## Safety properties

- fresh `writing` claims are never reclaimed
- pre-upgrade NULL claims become recoverable only after a new grace period
- recovery and finalize are mutually exclusive at the SQLite transaction/status predicate
- a completed write that loses its claim cannot be blindly re-approved
- invalid/naive cutoffs and empty actor identities are rejected
- recovery is oldest-first with a bounded 1-1000 row limit
- no SQLite `RETURNING` dependency is introduced

## Verification

- 224 focused storage/schema/CLI/server tests passed
- added NULL-backfill, oldest-first/limit, recovery-finalize, write-uncertain/reapproval, MCP response, and MCP validation tests
- Ruff and format check passed
- focused mypy passed
- Bandit 1.9.4 passed
- full registry reports 96 tools and includes `mem_candidate_recover`
- `git diff --check` passed

Closes #1734

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,11 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 - **Interrupted approval recovery** — `mm review recover` and
   `mem_candidate_recover` atomically return stale `writing` candidates to the
   pending queue after a conservative threshold and persist an audited status
-  transition without reclaiming fresh approvals.
+  transition without reclaiming fresh approvals. Pre-upgrade NULL claim
+  timestamps receive a one-time grace-period backfill; recovery supports
+  older SQLite versions without `UPDATE ... RETURNING`, and a completed write
+  that loses finalization is quarantined as `write_uncertain` to prevent blind
+  duplicate approval.
 
 ## [0.3.8] — 2026-07-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   evidence-linked candidates from exact session events and write durable
   memory only after approval. Additive temporal assertion tables preserve
   direction and multiple `supersedes`/`contradicts`/`supports` edges.
+- **Interrupted approval recovery** — `mm review recover` and
+  `mem_candidate_recover` atomically return stale `writing` candidates to the
+  pending queue after a conservative threshold and persist an audited status
+  transition without reclaiming fresh approvals.
 
 ## [0.3.8] — 2026-07-13
 
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -897,7 +897,7 @@ The keys may come either from the config surface above or from the Langfuse SDK'
 
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `MEMTOMEM_TOOL_MODE` | `core` | Which MCP tools are exposed: `core` (9 tools), `standard` (38 incl. `mem_do`), `full` (95) |
+| `MEMTOMEM_TOOL_MODE` | `core` | Which MCP tools are exposed: `core` (9 tools), `standard` (38 incl. `mem_do`), `full` (96) |
 
 In `core` mode, use `mem_do(action="...", params={...})` to access any of the 70+ non-core actions. Fewer tools means less context usage for AI agents.
 
--- a/docs/guides/mcp-clients.md
+++ b/docs/guides/mcp-clients.md
@@ -381,7 +381,7 @@ mm status
 uses, so the output is identical. Useful as a sanity check between
 `mm init` and the first editor-side call.
 
-### Available MCP Tools (95)
+### Available MCP Tools (96)
 
 | Category | Tools |
 |----------|-------|
@@ -416,12 +416,12 @@ uses, so the output is identical. Useful as a sanity check between
 | **Evaluation** | `mem_eval` |
 | **Context** | `mem_context_detect`, `mem_context_init`, `mem_context_generate`, `mem_context_diff`, `mem_context_sync`, `mem_context_memory_migrate`, `mem_context_artifact_migrate`, `mem_context_artifact_transfer`, `mem_context_version`, `mem_context_promote` — cross-runtime artifact sync (`mm context`). Parameters: [Context tool reference](reference.md#context-tool-parameters); workflow: [Context Gateway](context-gateway.md) |
 | **Pinned Context** | `mem_pinned_list`, `mem_pinned_get`, `mem_pinned_set`, `mem_pinned_delete`, `mem_context_compose` |
-| **Formation** | `mem_formation_scan`, `mem_candidate_list`, `mem_candidate_review` |
+| **Formation** | `mem_formation_scan`, `mem_candidate_list`, `mem_candidate_review`, `mem_candidate_recover` |
 
 
 \* Requires `MEMTOMEM_TOOL_MODE=full`. In `core` or `standard` mode, use `mm config` (CLI) or the Web UI Settings tab instead.
 
-> **Tool mode**: Set `MEMTOMEM_TOOL_MODE` to `core` (9 tools, default), `standard` (core + common packs + `mem_do`), or `full` (all 95 tools individually) to control how many tools are exposed. In `core` mode, use `mem_do(action="...", params={...})` to access any non-core action. Fewer tools = less context usage for AI agents.
+> **Tool mode**: Set `MEMTOMEM_TOOL_MODE` to `core` (9 tools, default), `standard` (core + common packs + `mem_do`), or `full` (all 96 tools individually) to control how many tools are exposed. In `core` mode, use `mem_do(action="...", params={...})` to access any non-core action. Fewer tools = less context usage for AI agents.
 
 ### STM Proxy Tools (optional, separate package)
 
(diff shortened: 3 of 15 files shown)


memtomem:
# Review — recover stale memory formation claims

Solid PR overall, and the conventions all check out:

- The migration follows the existing `try/except "duplicate column"` idempotent pattern, and the sweep predicate gets a matching `(status, claim_started_at)` index.
- Both SELECTs and `_CANDIDATE_KEYS` are widened together (appended at the end — row-widening safe).
- `PRAGMA foreign_keys=ON` is set in the backend, so `ON DELETE CASCADE` on the audit table actually works.
- Tool classification is consistent: `formation` is not in `_STANDARD_PACKS`, so `mem_candidate_recover` is full-only + `mem_do`-routable like its siblings, and the documented standard=38 count stays correct.
- The 95→96 docs sweep is complete (no stale "95" left; the `#available-mcp-tools-96` anchor matches the new heading slug).
- Validation is translated per surface (CLI `ClickException` / MCP `ok:false` JSON / storage `ValueError`).

## Findings (by severity)

### 1. `writing` rows stranded before the upgrade are unrecoverable — the original #1734 scenario (medium-high)

The sweep requires `claim_started_at IS NOT NULL AND claim_started_at <= ?`, but rows stranded **before this migration** have `claim_started_at = NULL`. An operator following the new docs (`mm review list --status writing` → `mm review recover`) on an existing DB gets `recovered: 0` and nothing happens — the exact rows this PR exists to fix stay stuck.

Treating NULL as unconditionally stale isn't safe (an old-binary process that opened the DB before the fence check could still hold an in-flight NULL claim). Suggested fix: a **one-time backfill at migration time**:

```sql
UPDATE memory_candidates SET claim_started_at = <now>
 WHERE status='writing' AND claim_started_at IS NULL
```

Pre-existing stranded rows become recoverable 15 minutes after upgrade, and any still-live old-binary claim gets a fresh timestamp, so it stays protected.

### 2. When finalize loses to recovery, the durable write persists but the candidate returns to `pending` → duplicate write on re-approval (medium)

The state transition is mutually exclusive, but the durable write itself is not transactional with the status. If the sweep fires mid-write (easy to trigger — `--stale-after-minutes` allows 1), the write completes, finalize matches 0 rows, and the candidate is `pending` again; re-approval writes the same content twice. Acceptance criterion 3 ("no duplicate durable writes") is only protected by the time threshold.

- The CLI raises `ClickException("Candidate claim was lost before finalization")`, but the MCP path returns `{"ok": false, "status": "approved", "write": {...}}` — `ok:false` with `status:"approved"` and a write payload is misleading. Please add an explicit branch like the CLI's, e.g. "claim was recovered concurrently; durable write already persisted — inspect before re-approving".
- Neither surface tells the operator that the write already landed on disk; both should.

### 3. `UPDATE … RETURNING` — first use in the codebase, requires SQLite ≥ 3.35 (low, portability)

No prior `RETURNING` in the repo. Pythons linked against system SQLite on Debian 11 (3.34) or RHEL 8 (3.26) will die with `OperationalError`. Either document the minimum SQLite version, or use a two-step `SELECT id … LIMIT` → `UPDATE … WHERE id IN (…) AND status='writing'` + rowcount check inside the same transaction (same atomicity).

### 4. `SCHEMA_VERSION` 1→2 fence bump is stricter than precedent (discussion)

The fence contract is "bump only when an older binary must not run under the new schema", and the earlier nullable/default column additions on `chunks` shipped without a bump. This change doesn't corrupt anything under an old binary (audit-record gaps and a leftover `claim_started_at` on approved rows at worst), while the bump hard-blocks rollback to 0.3.8 on first run. If the bump is deliberate for audit-trail completeness, fine — just confirming intent.

## Minor

- The audit table has no pruning and candidates are never deleted (expiry only flips status), so CASCADE effectively never fires and transitions accumulate forever. Harmless at current scale; worth knowing.
- Per-surface actor defaults (`cli-operator` / `mcp-operator` / `operator`) read as intentional provenance markers — fine.

## Test coverage

Core paths are well covered (fresh-claim protection, recovery↔finalize exclusion, tz/actor validation, legacy-table migration, CLI/registry exposure). Missing:

- `limit` and oldest-first ordering behavior
- an e2e for finding 2 (what happens when a recovered candidate is re-approved) — add alongside the fix
- the MCP tool's `ok:false` validation branches

## Verdict

**NEEDS-FIX (minor)** — mergeable quality, but finding 1 (NULL backfill) defeats the PR's purpose on existing DBs and should land before merge; finding 2's MCP response branch is a few-line fix in the same PR. Findings 3 and 4 can be follow-up issues/comments.

memtomem:
Addressed all review findings in `17e3038b`:

- migration now stamps pre-upgrade `writing` rows with NULL `claim_started_at` at upgrade time, giving live old processes and stranded rows the same fresh 15-minute grace period
- removed `UPDATE ... RETURNING`; recovery now uses `BEGIN IMMEDIATE`, oldest-first SELECT, per-id guarded UPDATE, transition inserts, and one commit/rollback span (compatible with older SQLite)
- reverted the schema fence bump because this is an additive nullable column/audit table, consistent with existing migration precedent
- if a durable write completes after recovery won the state race, both CLI and MCP explicitly report that the write already persisted; the candidate is quarantined as `write_uncertain`, so a normal re-approval cannot duplicate it
- MCP no longer returns `ok:false` with `status:approved`; it returns `status:write_uncertain`, `durable_write_persisted:true`, the write details, and inspection guidance
- added tests for NULL backfill, oldest-first + limit, recovery/finalize exclusion, write-uncertain reapproval prevention, MCP persisted-write response, and all MCP validation branches

Verification: 224 focused tests passed; Ruff, format, focused mypy, Bandit 1.9.4, and diff check passed.

The transition table remains intentionally append-only: normal candidates produce only a small bounded number of lifecycle rows, and preserving the operator/reason history is the audit feature rather than disposable telemetry.

memtomem:
# Re-review — 17e3038b (fix: harden interrupted claim recovery)

All four findings from the previous review are addressed in this commit, each with matching tests:

| Previous finding | Resolution | Verified |
|---|---|---|
| 1. Pre-upgrade NULL claims unrecoverable | One-time migration backfill stamps NULL `writing` rows with *now*, granting a full grace period | ✅ Idempotent (touches NULL only), legacy-row backfill test added |
| 2. Duplicate-write risk + misleading MCP response on finalize-lost | New `write_uncertain` quarantine status; CLI warning includes the write location, MCP returns `durable_write_persisted: true` + "do not re-approve"; `claim_memory_candidate` requiring `pending` blocks blind re-approval | ✅ Both surfaces branch explicitly; storage quarantine, re-approval block, and MCP response tests added |
| 3. `UPDATE … RETURNING` (SQLite ≥ 3.35) | Rewritten as `BEGIN IMMEDIATE` + SELECT + per-row UPDATE with rollback on exception | ✅ Matches the existing "BEGIN IMMEDIATE before SELECT" precedent in `orphan_gc.py`; no awaits inside the transaction, so no event-loop interleaving |
| 4. `SCHEMA_VERSION` fence bump | Reverted to 1, consistent with prior additive nullable migrations | ✅ |

Additional improvements noticed:

- `ORDER BY claim_started_at, id` — the id tiebreak prevents same-second ordering nondeterminism; oldest-first + limit behavior is now tested.
- The MCP `ok:false` validation branches are tested — all three coverage gaps from the previous review are closed.
- Quarantine actors (`cli-finalizer` / `mcp-finalizer`) keep the per-surface provenance convention, and the `pending → write_uncertain` transition is audit-recorded.
- `block.source_path` exists on the pinned block model, so the CLI's write-location warning is sound.

## Remaining findings

### `write_uncertain` has no CLI/MCP exit path (low — fine as a follow-up)

There is no official way to resolve a quarantined candidate:

- `decide_memory_candidate` / `claim_memory_candidate` both require `status='pending'`, and the CLI/MCP review paths short-circuit on "not pending".
- The expiry sweep also only targets `pending`, so quarantined rows persist indefinitely.

The docs say "resolve the candidate manually", but the only actual means is raw SQL. Suggest allowing `mm review reject` from `write_uncertain` (safe — it writes nothing), or documenting the concrete resolution step. A follow-up issue is fine.

### Residual race — acknowledged and acceptable (nit)

If another reviewer claims the recovered candidate before the quarantine lands, the quarantine matches 0 rows. Both surfaces report this explicitly ("state changed again; inspect the review queue"), so it never fails silently. Without a global lock this window is unavoidable; the current handling is sufficient.

## Verdict

**SHIP** — both blockers (NULL backfill, finalize-lost handling) and both recommendations (RETURNING, fence) are resolved in ways consistent with existing codebase precedent, with tests matching each fix. The missing `write_uncertain` exit path is not a merge blocker; track it as a follow-up issue.

tsdata (approved):

PR #1736, opened 2026-07-13
feat: add STM core capability bridge

memtomem:
## Summary
- advertise schema-versioned `context_compose` and `candidate_propose` capabilities through `mem_do(action="version")`
- expose an explicit `candidate_propose` meta-tool action backed by the existing review queue
- preserve the existing `context_compose` action as the pinned-first composition contract consumed by memtomem-stm

## Safety and compatibility
- proposals are privacy-scanned and bounded to 2,000 characters
- source metadata is bounded and idempotency keys are stored only through a SHA-256 fingerprint
- repeated proposals return the existing pending candidate
- proposal submission never writes durable memory; approval remains the only promotion path
- all additions are capability-gated and additive for older clients

## Validation
- full core suite: 8495 passed, 11 skipped
- meta/formation/pinned contract suite: 42 passed
- tool registration and logic suite: 98 passed
- changed core modules mypy passed
- repository-wide mypy still reports 14 pre-existing errors in unrelated web/CLI files
- git diff --check passed

## Related
- enables the capability-gated integration merged in memtomem-stm#691

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,8 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 - **STM capability bridge** — `mem_do(action="version")` advertises versioned
   `context_compose` and `candidate_propose` contracts. External proposals are
   privacy-gated, idempotently queued as pending review candidates, and never
-  write durable memory before approval. Their synthetic sessions use the
-  reserved `formation` namespace and are excluded from normal retrieval scope.
+  write durable memory before approval. Proposal content remains only in the
+  review queue and is not indexed as retrievable chunks.
 
 ## [0.3.8] — 2026-07-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,8 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 - **STM capability bridge** — `mem_do(action="version")` advertises versioned
   `context_compose` and `candidate_propose` contracts. External proposals are
   privacy-gated, idempotently queued as pending review candidates, and never
-  write durable memory before approval.
+  write durable memory before approval. Their synthetic sessions use the
+  reserved `formation` namespace and are excluded from normal retrieval scope.
 
 ## [0.3.8] — 2026-07-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   older SQLite versions without `UPDATE ... RETURNING`, and a completed write
   that loses finalization is quarantined as `write_uncertain` to prevent blind
   duplicate approval.
+- **STM capability bridge** — `mem_do(action="version")` advertises versioned
+  `context_compose` and `candidate_propose` contracts. External proposals are
+  privacy-gated, idempotently queued as pending review candidates, and never
+  write durable memory before approval.
 
 ## [0.3.8] — 2026-07-13
 
--- a/packages/memtomem/src/memtomem/formation.py
+++ b/packages/memtomem/src/memtomem/formation.py
@@ -86,3 +86,71 @@ async def scan_session_candidates(storage: Any, session_id: str) -> list[dict[st
         if await storage.add_memory_candidate(candidate):
             created.append(candidate)
     return created
+
+
+async def propose_memory_candidate(
+    storage: Any,
+    content: str,
+    *,
+    source: str,
+    source_ref: str,
+    idempotency_key: str,
+) -> tuple[dict[str, Any], bool]:
+    """Queue one explicit external proposal for review without promoting it."""
+    body = content.strip()
+    if not body:
+        raise ValueError("content cannot be empty")
+    if len(body) > 2000:
+        raise ValueError("content exceeds 2000 characters")
+    if len(source) > 128 or len(source_ref) > 512 or len(idempotency_key) > 256:
+        raise ValueError("proposal metadata exceeds size limit")
+    if not source.strip() or not idempotency_key.strip():
+        raise ValueError("source and idempotency_key are required")
+    if privacy.scan(body):
+        raise ValueError("content contains sensitive data")
+
+    classification = _classify(body)
+    kind, operation, destination, confidence = classification or (
+        "proposed",
+        "add",
+        "memory",
+        1.0,
+    )
+    now = datetime.now(timezone.utc)
+    fingerprint = hashlib.sha256(
+        f"external\0{source.strip()}\0{idempotency_key.strip()}".encode()
+    ).hexdigest()
+    external_session_id = f"external:{source.strip()}:{fingerprint[:24]}"
+    await storage.create_session(
+        external_session_id,
+        source.strip(),
+        "formation",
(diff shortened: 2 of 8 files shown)


memtomem (commented):
## Review

Direction and structure look good — validation → classification → session creation → candidate insert is clear, and the new path follows the existing `scan_session_candidates` conventions (fingerprint, 30-day expiry, privacy gate). I verified against the head SHA that all imports already exist in `formation.py`, `create_session` is idempotent on id conflict (`ON CONFLICT(id) DO NOTHING`), and dedupe is actually enforced by the `UNIQUE(session_id, fingerprint)` schema constraint. Capability gating is consistent with what memtomem-stm#691 consumes.

### Correctness — recommend fixing before merge

**1. Phantom candidate: the duplicate fallback can return a `candidate_id` that does not exist in storage** (`formation.py`, tail of `propose_memory_candidate`)

When the insert is IGNOREd and `list_memory_candidates(status="pending", limit=1000)` does not find the original, `duplicate or candidate` returns the **fresh, never-persisted** candidate. The tool response then contains a non-existent `candidate_id`, a hardcoded `"status": "pending"`, and `duplicate: true`. This path is reachable:

- The original candidate was already **approved / rejected / expired** — the row remains, so the UNIQUE constraint keeps blocking the insert, but its status is no longer `pending`, so the lookup misses. Notably, after the 30-day expiry, re-proposing with the same idempotency key returns phantom responses **forever**.
- More than 1,000 pending candidates exist and the original falls outside the `ORDER BY created_at` top 1000.

Suggestion: add a status-agnostic storage lookup such as `get_memory_candidate_by_fingerprint(session_id, fingerprint)`, and have the tool return the candidate's actual status instead of hardcoding `"pending"`. As soon as an STM caller follows up on the returned `candidate_id`, this bug becomes user-visible. This also removes the O(N) pending scan on every duplicate.

**2. Idempotency fingerprint excludes content**

The fingerprint is computed from `source + idempotency_key` only, so re-sending the same key with **different content** silently returns the earlier candidate as a duplicate. Standard idempotency contracts reject key reuse with a mismatched payload. At minimum, detect the content mismatch and raise, or include the returned candidate's content (or a content hash) in the response so callers can detect it. Fits naturally in the same change as item 1.

### Minor

- **`source_ref` is not privacy-scanned** — only the body goes through `privacy.scan`; `source_ref` (up to 512 chars) flows unscanned into evidence and session metadata. Given this is an external-input path, consider scanning it too (or documenting the exclusion).
- **Unclassified content gets confidence 1.0** — the `_classify` fallback is `("proposed", ..., 1.0)`, while heuristic classifications max out at 0.95. Unclassified external text would sort above everything in the review queue; a conservative value (e.g. 0.5) seems more appropriate.
- **Session namespace `"formation"`** — `create_session(external_session_id, source, "formation")` introduces a new namespace label. If intentional, a one-line note on how it relates to existing namespace conventions would help.

### Test coverage

The two new tests are valid (I traced their pass paths statically against the schema). Missing cases:

- **Re-proposal after decision/expiry** — this is exactly the test that would catch finding 1; please add it alongside the fix.
- Rejection paths for >2000 chars, empty content, and oversized metadata.
- A tool-layer test for the `mem_candidate_propose` JSON response shape (current coverage is function-level only).

### Security

The safety claims in the PR body match the code (privacy gate, 2,000-char bound, SHA-256 fingerprint, approval-only promotion), with the `source_ref` scan gap noted above. The approve path reuses the existing claim/release/finalize state machine, so the promotion boundary holds.

memtomem:
Addressed all review findings in `87fec5ae`:

- added status-independent `get_memory_candidate_by_fingerprint(session_id, fingerprint)` lookup, removing the O(N)/1,000-row pending scan and phantom-ID fallback
- duplicate proposals now return the persisted candidate and its actual status, including rejected/expired states
- idempotency-key reuse with different content now fails explicitly
- privacy-scanned `source_ref` as well as content
- lowered unclassified external proposal confidence from 1.0 to 0.5
- documented the reserved `formation` namespace as excluded from normal retrieval scope
- added decided/expired re-proposal, payload mismatch, empty/oversized/sensitive metadata, and tool JSON response-shape tests

Validation:
- formation/meta/pinned contract suite: 51 passed
- sessions/tool registration/tool logic suite: 135 passed
- Ruff and changed-module mypy passed
- git diff --check passed

memtomem (commented):
## Re-review (head `87fec5a`)

The follow-up commit (`fix: harden candidate proposal idempotency`) addresses all findings from the previous review. Verified against the head snapshot:

### Previous findings — all resolved

- **Phantom candidate (blocking #1) — fixed properly.** The status-agnostic `get_memory_candidate_by_fingerprint(session_id, fingerprint)` lookup replaces the pending-only scan, and the `RuntimeError` guard on a missing row is a correct defensive-unreachable (with FK satisfied by the prior `create_session`, the only `OR IGNORE` trigger left is the UNIQUE constraint). The tool now returns the candidate's actual status instead of hardcoding `"pending"`. Both regression tests I asked for are here (re-proposal after rejection → `"rejected"`, after expiry → `"expired"`).
- **Content mismatch on key reuse (blocking #2) — fixed** with an explicit `ValueError`, plus a test.
- **`source_ref` privacy scan** — added, with a sensitive-`source_ref` case in the parametrized validation test.
- **Fallback confidence 1.0 → 0.5** — fixed and pinned in the tool-layer test.
- **Tool-layer response-shape test** — added; asserts the exact key set.
- **Boundary validation tests** — added (empty content, 2001 chars, oversized source/source_ref).

The new storage method itself is sound: the pending-expiry refresh mirrors the existing `get_memory_candidate` convention, and `UNIQUE(session_id, fingerprint)` guarantees at most one row for the `fetchone()`.

### New finding — introduced by this iteration

**The new CHANGELOG sentence is not backed by code:** *"Their synthetic sessions use the reserved `formation` namespace and are excluded from normal retrieval scope."*

I grepped the full head snapshot — nothing reserves the `formation` namespace or excludes it from retrieval. Proposal content is indeed unsearchable, but only because it never becomes chunks (candidates live solely in `memory_candidates`), not because of any namespace mechanism. In fact, `mem_add(namespace="formation")` would index chunks that *are* retrievable, and the synthetic sessions still appear in `list_sessions` (no namespace filter). Suggest rewording to something like *"proposal content is stored only in the review queue and is never indexed as chunks"* and dropping *"reserved"*.

### Remaining nits (non-blocking)

- A sensitive pattern in `source_ref` still reports `"content contains sensitive data"` — `"content or source_ref"` would aid debugging.
- The validation parametrize is missing `idempotency_key` > 256 chars and the empty `source`/`key` ("required") paths.

### CI

lint, typecheck, bandit, OSV, and golden-path pass; the main test jobs (ubuntu/macos/windows shards, browser, js-unit) are still **pending** — please wait for green before merging.

### Verdict

Both blocking findings are cleanly resolved with regression tests. With the one-line CHANGELOG fix, this looks good to merge (pending CI).

tsdata (approved):

memtomem:
Closed the remaining re-review items in `296b7d3e`:

- corrected the CHANGELOG contract: proposals remain only in the review queue and are not indexed as retrievable chunks; no unsupported reserved-namespace claim
- clarified the privacy validation error to cover `content or source_ref`
- added validation coverage for oversized idempotency keys and empty source/key

Validation: 54 contract tests passed; Ruff, changed-module mypy, and `git diff --check` passed.

PR #1738, opened 2026-07-13
fix: resolve uncertain memory candidates safely

memtomem:
## Summary

- allow an inspected `write_uncertain` candidate to transition atomically to `rejected` without another durable write
- require a non-empty reviewer and resolution reason
- expose the resolution through the existing `mm review reject` and `mem_candidate_review(decision="reject")` surfaces
- preserve normal `pending -> rejected` behavior while keeping uncertain re-approval blocked
- record `write_uncertain -> rejected` in the candidate transition audit
- document the concrete operator inspection and resolution workflow

## Safety properties

- resolution writes no memory or Pinned Context content
- only one concurrent resolver can match `status="write_uncertain"`
- a second resolution attempt fails closed
- direct re-approval remains unavailable
- actor and reason are persisted in the audit transition

## Verification

- 227 focused storage/schema/CLI/server tests passed
- direct CLI and MCP resolution paths tested
- missing reviewer/reason and concurrent second-attempt behavior tested
- Ruff and format check passed
- focused mypy passed
- Bandit 1.9.4 passed
- `git diff --check` passed

Closes #1737

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,9 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   timestamps receive a one-time grace-period backfill; recovery supports
   older SQLite versions without `UPDATE ... RETURNING`, and a completed write
   that loses finalization is quarantined as `write_uncertain` to prevent blind
-  duplicate approval.
+  duplicate approval. Operators can close that quarantine through the normal
+  CLI/MCP reject action with a required reviewer and reason; resolution writes
+  no additional durable memory and records an audited transition.
 
 ## [0.3.8] — 2026-07-13
 
--- a/docs/guides/pinned-context.md
+++ b/docs/guides/pinned-context.md
@@ -52,8 +52,17 @@ approval finalization, does not touch fresh claims, and records a persistent
 If a recovered claim's original process later reports that its durable write
 already completed, memtomem moves the candidate to `write_uncertain` and
 reports the destination. Do not approve it again: inspect the Markdown or
-Pinned Context block first, then resolve the candidate manually. This
-quarantine prevents a blind retry from duplicating the persisted content.
+Pinned Context block first, then close the quarantine without another write:
+
+```bash
+mm review reject CANDIDATE_ID \
+  --reviewer alice \
+  --reason "confirmed the durable entry already exists"
+```
+
+The same `mem_candidate_review` MCP action accepts `decision="reject"` with a
+non-empty reviewer and reason. The atomic `write_uncertain → rejected`
+transition is audit-recorded; direct re-approval remains blocked.
 
 ## LangGraph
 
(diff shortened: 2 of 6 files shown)


tsdata (approved):

memtomem:
# Review — resolve uncertain memory candidates safely (#1737)

Implements the `write_uncertain` exit path flagged in the #1735 re-review. Routing the resolution through the existing reject surfaces (`mm review reject` / `mem_candidate_review(decision="reject")`) instead of a dedicated verb is the simpler contract of the two options the issue proposed — good choice. No additional durable write, reviewer/reason required, transition audit-recorded.

## Correctness checks (all pass)

- **Atomicity/concurrency**: the surfaces read status first (a TOCTOU window), but the actual transition is decided by the storage-level `UPDATE … WHERE status='write_uncertain'` predicate — a concurrent second attempt matches 0 rows and fails closed (CLI `ClickException` / MCP `state_changed`). Pinned by tests.
- **Re-approval stays blocked**: the approve path now checks `status != "pending"` explicitly (the MCP top-level check was decomposed into per-branch checks), with the storage `claim_memory_candidate` pending predicate as second defense. The existing quarantine test already pins the storage layer.
- **Normal reject unchanged**: `pending → rejected` still goes through `decide_memory_candidate`. The only behavioral deltas from decomposing the MCP top-level check are the new `write_uncertain` branch and improved error messages (now including the actual status).
- **Triple validation**: CLI (`ClickException`) / MCP (`ok:false`) / storage (`ValueError`) — consistent with the per-surface error-translation convention and the recover implementation.
- **Audit**: the transition INSERT shares the transaction with the UPDATE and commits together — same structure as the sibling methods.
- **Test/lazy-import compatibility**: the CLI test monkeypatches `memtomem.cli._bootstrap.cli_components`, and `_decide` imports it lazily inside the function, so the patch genuinely applies.
- **Docs**: `pinned-context.md` replaces "resolve manually" with concrete commands; the example flags (`--reviewer`, `--reason`) match the real options.

All seven acceptance criteria in #1737 are met.

## Minor (non-blocking)

- **`decision_reason` is overwritten on resolve**: the quarantine warning ("write already persisted at <location>") is replaced by the operator's resolve reason. The original warning survives in the `pending → write_uncertain` audit transition, so nothing is lost — but post-incident investigation needs to look at transitions, worth knowing.
- **No surface-level re-approval-block test**: "approve on write_uncertain → not-pending error" is only pinned at the storage layer (claim refusal). One line each for the CLI/MCP approve branches' new error messages would complete the regression guard; low risk, optional.
- The MCP mock test's minimal `{"id","status"}` candidate dict is exactly valid — the `write_uncertain` branch returns before any destination access.

## Verdict

**SHIP** — the implementation matches the issue's design intent (safe one-way resolution, no `write_uncertain → pending`), and atomicity/audit/validation follow the same patterns as the sibling code. The two minor items are a few lines if you want them in this PR, or fine to leave.

PR #1739, opened 2026-07-13
feat: add native Claude and Codex plugins

memtomem:
## Summary
- generate Claude and Codex workflow assets from one reviewed contract
- simplify the safe Claude base plugin and move hooks into a separate opt-in automation plugin
- add a native Codex plugin marketplace package with six policy-aware skills
- document both runtimes and guard versions, manifests, generated assets, and hook safety

## Verification
- 55 targeted plugin, docs, supply-chain, and hook tests passed
- Ruff checks passed
- Claude strict manifest and marketplace validation passed
- Codex plugin and all six skill validators passed
- isolated Claude and Codex marketplace installs passed
- generated asset drift check and git diff check passed

## Notes
- OpenCode integration is intentionally excluded from this PR
- unrelated local docs/plans and docs/qa files are not included


memtomem:
# Review — feat: add native Claude and Codex plugins

## Overview

This PR restructures plugin assets around a single reviewed contract (`packages/memtomem-plugin-assets/contract.toml`) rendered by `tools/render_plugin_assets.py`, slims the Claude base plugin (0.2.4 → 0.3.0) down to a safe read/explicit-write surface, moves hooks into a separate opt-in `memtomem-automation` plugin with a Python stdin-JSON dispatcher, and adds a native Codex marketplace package. The direction is right: replacing the shell hooks' `${prompt}` interpolation (a shell-injection vector) with an argv-based dispatcher — and proving `$(touch ...)` is inert in a test — is a clear security win, as are `disable-model-invocation: true` on write skills and keeping `mem_do` out of shipped skills.

## 🔴 Blockers

### 1. `plugins/memtomem/.mcp.json` is gitignored and missing from the PR

Verified in the checkout:

```
$ git check-ignore -v plugins/memtomem/.mcp.json
.gitignore:21:.mcp.json    plugins/memtomem/.mcp.json
```

`.gitignore` line 21 ignores all `.mcp.json`, with a negation only for `packages/memtomem-claude-plugin/.mcp.json` (line 25). The new Codex `.mcp.json` exists only in the working tree and was silently dropped from the commit — it is not among the 47 changed files. Consequences on a fresh checkout (CI):

- `test_codex_plugin_matches_contract_and_marketplace` fails with `FileNotFoundError`, and `test_generated_plugin_assets_are_in_sync` (renderer `--check`) fails on drift — the currently-pending test jobs should go red on all 3 OSes.
- If merged anyway, the shipped Codex plugin's `plugin.json` references a nonexistent `"mcpServers": "./.mcp.json"` — a broken plugin with no MCP server.

**Fix**: add `!plugins/memtomem/.mcp.json` to `.gitignore` and `git add -f plugins/memtomem/.mcp.json`. The "isolated Codex marketplace install passed" verification ran against the local tree where the file exists, so it couldn't catch this.

### 2. `fake_mm` fixture is POSIX-only — expected to fail the gating Windows shards

The fixture in `test_plugin_assets.py` writes a shebang script named `mm`, `chmod(0o755)`, and prepends it to `PATH`. On Windows: (a) `shutil.which("mm")` won't find an extensionless file via PATHEXT lookup, and (b) even if found, a shebang script isn't executable via CreateProcess (WinError 193). The dispatcher fails open, so `test_automation_prompt_uses_json_stdin_and_argv_safely` (expects empty SessionStart stdout + search context) and `test_automation_indexes_only_supported_write_paths_and_flushes` (expects index calls) fail on `windows-test-shard`. The repo's existing pattern is `@pytest.mark.skipif(os.name == "nt", ...)` (e.g. `test_wiki_store.py`), but since real Windows users get an `mm.exe` shim from `uv tool install` and the dispatcher genuinely runs there, prefer adding an `mm.bat` wrapper to the fixture to keep Windows coverage instead of skipping.

## 🟡 Recommended before merge

### 3. No drift guard for `CORE_VERSION`

Single-sourcing versions is the point of the contract, yet `hook_dispatch.py:CORE_VERSION = "0.3.8"` and the two docs sites (`claude-code.md` and the automation README's `uv tool install 'memtomem==0.3.8'`) are not tied to `contract.toml [core] version` by any test — and the old `_PLUGIN_CORE_MAP` guard was removed. A core-only bump next release that leaves the dispatcher/docs behind is exactly the failure mode the docs-as-tests guards exist for. Suggest: parse `CORE_VERSION` out of the dispatcher source in `test_plugin_assets.py` and assert it equals the contract version, plus grep both docs for `memtomem==<contract version>`. (The release checklist's "plugin bump sites" also change with this PR and should be updated.)

### 4. `_session_start` double-probes and can exceed its hook timeout

`_compatible_mm(refresh=True)` runs one probe (3 s cap), then on incompatibility `_session_start` calls `_probe_mm()` **again** (`hook_dispatch.py` lines 576–579). Worst case is 3+3 s plus interpreter startup against a 5 s SessionStart hook timeout — so the diagnostic context is most likely to be lost precisely when `mm` is slow or hung. Restructure so `_compatible_mm` returns `(executable, version)` (or reuse the probe result) and probe once.

### 5. Incompatible-`mm` cache is effectively a no-op

The non-refresh path of `_compatible_mm()` treats a cached non-matching version as a miss and re-probes + rewrites the cache on every call. A user with an older `mm` installed pays up to a 3 s probe on **every prompt and every Write/Edit**. Cache a "checked, incompatible" state and only re-probe at SessionStart.

### 6. Prompt text lands in `hook.log` on failure

`_run` logs `args!r` on failure, so a failing search writes up to 500 chars of the user's raw prompt in plaintext to `hook.log`. Per the project's logging convention (no content in validation logs), log only the subcommand name.

## 🟢 Minor

- **`python3` on Windows**: python.org installs expose `python`/`py`, not `python3`, so the automation hooks may fail to launch for Windows users (fail-open, but a one-line support note in the README would help).
- **Docs vs behavior**: the table says "After a file write **succeeds**" but the dispatcher queues the path without checking `tool_response`; and "Prompts under 20 characters are skipped" while the actual condition is `<= 20`.
- **Generated phrasing**: template sentences like "If it does not provide a usable content to remember" read awkwardly and ship in user-visible skills.
- **Marketplace source format change** (`git-subdir` object → `"./packages/..."` relative path): schema-valid and strict validation passed, but worth one empirical install via GitHub (`/plugin marketplace add memtomem/memtomem`) before merge, since `git-subdir` was a deliberate choice in #1680.

## What's good

- The shell-injection fix is real and tested (injection attempt proven inert). Hook schema usage (`args` exec form, `statusMessage`, `CLAUDE_PLUGIN_DATA`) is all valid per current Claude Code docs — verified separately.
- Contract → render → `--check` drift gate fits the repo's docs-as-tests culture.
- Consistent safe defaults: no session-lifecycle automation, explicit-only write skills, `mem_do` not exposed.
- Supply-chain tests extended to both runtimes and both marketplaces.

## Verdict

**NEEDS-FIX** — the architecture is approvable, but #1 is both a CI failure and a shipped-artifact breakage, and #2 blocks the gating Windows shards. #3 completes what this PR sets out to do and belongs in the same change.

memtomem:
Addressed the review in d87fb2fa:

- track plugins/memtomem/.mcp.json explicitly and verified it from a Git archive
- make the fake mm fixture cross-platform with an mm.bat wrapper on Windows
- guard dispatcher and documentation core versions against contract.toml
- reduce SessionStart to one version probe and cache incompatible/missing results for the session
- remove prompt arguments and stderr content from hook logs
- launch hooks through uv instead of a platform-specific python3 alias
- align timeout margins, docs wording, and generated skill phrasing

Verification: 57 targeted tests passed; Ruff, renderer drift, Claude strict validation, Codex plugin validation, and tracked-only archive validation passed.

tsdata (approved):

memtomem:
# Re-review after d87fb2fa — fix: harden cross-platform plugin assets

## Summary

**Both blockers and all four recommendations from the previous review are resolved in d87fb2fa**, each with an accompanying regression test. All required CI checks are now green, including both Windows shards.

## Previous findings — resolution check

### 🔴 Blockers — both resolved

- **#1 gitignored `.mcp.json`**: `.gitignore` gained a commented `!plugins/memtomem/.mcp.json` negation and the file is now tracked (47 → 49 changed files). Contents match the renderer output (`memtomem==0.3.8` pin + `MEMTOMEM_TOOL_MODE=core`), and the two tests that failed on a fresh checkout now pass empirically on ubuntu/macos.
- **#2 POSIX-only `fake_mm` fixture**: rewritten as a Python script behind platform wrappers (`mm.bat` on Windows, `#!/bin/sh` exec elsewhere), with `.BAT` prepended to `PATHEXT` so `shutil.which("mm")` resolves. Coverage kept on Windows instead of skipping — and both `windows-test-shard` jobs are green, which empirically settles the `.bat` argument-forwarding question too.

### 🟡 Recommendations — all four resolved

- **#3 `CORE_VERSION` drift guard**: new `test_core_version_is_single_sourced_across_automation_assets` extracts `CORE_VERSION` from the dispatcher source, asserts it equals `contract.toml [core] version`, and greps both docs sites for `memtomem==<version>`.
- **#4 SessionStart double probe**: restructured into `_mm_probe() -> (executable, version)`; one probe per event. Hook timeouts raised to 8/12 s, leaving headroom for `uv` startup over the inner subprocess caps (probe 3, search 4, index 8, flush 9).
- **#5 no-op incompatible cache**: cache now records `checked: true` and the non-refresh path returns the cached negative result without re-probing. Pinned by the new assertion `_calls(log) == [["--version"]]` after a UserPromptSubmit against an incompatible `mm`. Old-format caches fall through to a re-probe — migration-safe.
- **#6 prompt text in `hook.log`**: `_run` now logs only the subcommand plus exception type / return code; `test_automation_failure_log_does_not_store_prompt` verifies the prompt never reaches the log on the failure path.

### 🟢 Minor items

`python3` → `uv run --no-project python` (uv is already a hard requirement via `uvx`, and this fixes the Windows alias concern), docs wording ("20 characters or fewer", "After a Write/Edit tool call"), and the generated phrasing ("If the request does not clearly specify …") — all fixed in both the renderer templates and the generated files, so the drift gate stays intact.

## New observations (non-blocking)

- **Mid-session install not detected**: with negative results cached, installing `mm` mid-session keeps hooks skipping until the next SessionStart. Consider adding "then start a new session" to the SessionStart install-guidance message — a one-line follow-up.
- **`compat.json` concurrent writes**: `write_text` is not atomic, but a torn cache fails open via `JSONDecodeError` → re-probe. No real risk.
- (Carried over) Worth one empirical `/plugin marketplace add memtomem/memtomem` install via GitHub after merge, since the marketplace `source` format changed from `git-subdir` to a relative path.

## Verdict

**SHIP** — all six findings fixed with tests, and every required check is green (ubuntu, macos, both Windows shards, lint, typecheck, browser, security gates). No objection to merging.

PR #1740, opened 2026-07-13
feat: harden context compose schema v2

memtomem:
## Summary

- promote `context_compose` to capability schema 2
- preserve `namespace` and `context_window` throughout composed retrieval
- exclude all sources under active pinned roots before ranking truncation and context expansion
- include each retrieved item's namespace in the compose response

## Compatibility contract

Consumers must only activate composed retrieval when `mem_version` advertises `context_compose.schema_version >= 2`. Older or non-advertising cores continue to use legacy search, so core and STM can be released in either order.

`candidate_propose` remains at schema 1.

This PR is feature-only. Package and plugin version pins remain at 0.3.8;
the coordinated 0.3.9 pin updates and changelog dating belong in a separate
release PR.

## Pinned-source exclusion

The exclusion is implemented inside the search pipeline and is opt-in for composed retrieval. It uses normalized pinned-root membership before top-k truncation and context expansion, covering shadowed files, other-agent files, malformed pinned files, and multi-chunk sources.

We intentionally do not exclude pinned directories during indexing. Legacy `mem_search` must remain able to find pinned content, and indexing-time exclusion would change that existing behavior.

## Verification

- 227 focused tests passed after rebasing onto #1739 and applying review feedback, including context compose, filter-only and post-expansion exclusion, namespace architecture, search pipeline, scope threading, plugin assets, docs guards, and supply-chain contracts
- Ruff, mypy, lock checks, release preflight, and wheel/sdist artifact checks passed before the rebase
- `git diff --check` passed

The prior full-suite attempt in this environment stopped at 32% with native process exit 133/SIGTRAP and no pytest failure traceback; the affected and newly merged integration surfaces are covered by the focused suite above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes will be documented in this file.
 Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
-## [0.3.9] — 2026-07-13
+## [Unreleased]
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes will be documented in this file.
 Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
-## [Unreleased]
+## [0.3.9] — 2026-07-13
 
 ### Added
 
@@ -35,6 +35,14 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   privacy-gated, idempotently queued as pending review candidates, and never
   write durable memory before approval. Proposal content remains only in the
   review queue and is not indexed as retrievable chunks.
+
+### Fixed
+
+- **Context composition preserves retrieval scope without duplicating pinned
+  files** — schema 2 accepts namespace and context-window inputs, returns each
+  retrieved chunk's namespace, and excludes every in-scope `pinned/` source
+  before ranking truncation and context expansion. Legacy `mem_search` remains
+  unchanged.
 
 ## [0.3.8] — 2026-07-13
 
--- a/docs/guides/pinned-context.md
+++ b/docs/guides/pinned-context.md
@@ -20,6 +20,13 @@ One block is limited to 2,000 characters, all pinned blocks to 6,000, and a
 default composed bundle to 12,000. A block is never cut in the middle; omitted
 ids are returned explicitly.
 
+`mem_context_compose` schema 2 accepts the same optional `namespace` and
+`context_window` retrieval controls used by search. Pinned files remain
+searchable through legacy `mem_search`, but the composed path excludes every
+file under the active user and project `pinned/` roots from its retrieved leg,
+including shadowed, agent-specific, and malformed blocks. This prevents pinned
+content from appearing a second time as an ordinary search result.
+
 ## Review-first candidates
 
 Candidate generation is explicit and never writes long-term memory by itself.
--- a/docs/guides/reference/operations.md
+++ b/docs/guides/reference/operations.md
@@ -151,7 +151,7 @@ Running both `memtomem-server` (MCP) and `memtomem-web` simultaneously is suppor
 
 ## STM: Proactive Memory Surfacing (Optional)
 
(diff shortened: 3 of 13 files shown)


tsdata (approved):

memtomem:
## Review — context compose schema v2

**Verdict: NEEDS-FIX.** The core implementation (pipeline exclusion, cache keying, capability gating) is correct and well built, but CI is red on a one-line omission, and the release prep bundled into this PR is incomplete in a way that would silently disable the feature for plugin users.

### What's solid

- **Exclusion covers every retrieval leg**: BM25, dense (query expansion rewrites the query before retrieval, so expanded legs are covered), the session-summary rescue leg, the empty-query filter-only path, and a re-apply after context-window expansion — all before rerank-pool and top-k truncation, matching the PR description.
- **Cache safety**: `_cache_key` includes the normalized exclusion roots, so the composed path and legacy `mem_search` never share a cache slot. `_normalize_source_roots` sorts, dedups, and resolves, keeping the key deterministic.
- **Path comparison is symlink-consistent**: indexing stores `source_file` resolved-absolute, and `_source_is_excluded` re-resolves both sides; `is_relative_to` avoids prefix false-positives (pinned in the test: `pinned-other/` does not match).
- **Compatibility contract holds**: new params are optional, `candidate_propose` stays at schema 1, and the capability advertisement ships in the same PR as the behavior — so `schema_version >= 2` gating on the consumer side is sound.
- Search-time opt-in exclusion (rather than indexing-time) correctly preserves legacy `mem_search` visibility of pinned content.

### Blocking

1. **CI red — namespace architectural guard** (required check, will fail on all 3 OS):

   ```
   test_validate_namespace_architectural_guard.py::test_no_unclassified_ns_surfaces
     - pinned.py::mem_context_compose(namespace)
   ```

   The new `namespace=` MCP surface is unclassified. It's a read filter (never round-trips into storage), so the right fix is adding `("pinned.py", "mem_context_compose", "namespace")` to `DEFERRED_NS_SURFACES` next to the existing `mem_search` deferral, with the read-filter rationale.

2. **Release prep is half-applied and mixed into a feature PR.** The version bump touches only `pyproject.toml` + `uv.lock`, but the core version is pinned in more places, all still at 0.3.8:
   - `packages/memtomem-plugin-assets/contract.toml` (`core.version`)
   - `packages/memtomem-claude-automation-plugin/bin/hook_dispatch.py` (`CORE_VERSION`)
   - `packages/memtomem-claude-automation-plugin/README.md`, `docs/guides/integrations/claude-code.md`, `docs/guides/use-cases.md`

   The plugin-asset guards only check contract.toml's *internal* consistency, not against the package `__version__`, so this drift passes tests. If 0.3.9 ships as-is, the plugin `.mcp.json` keeps installing `memtomem==0.3.8`, which never advertises schema 2 — the feature stays inactive for plugin users. Recommendation: strip the version bump + CHANGELOG dating out into a separate release PR that bumps all core pins together, and keep this PR feature-only (entries under `[Unreleased]`).

### Minor

- `_filter_only_search` enumerates `candidate_limit = max(top_k*5, 100)` newest-first *before* exclusion; a large recently-updated pinned set can eat the enumeration budget. Edge case — worth a comment.
- `bm25_candidates` / `dense_candidates` stats now count post-exclusion; fine if read as "candidates entering fusion", but it changes the meaning.
- The `[Unreleased]` CHANGELOG section was removed entirely; keeping an empty one reduces merge conflicts for subsequent PRs.
- Coverage gaps: filter-only-path exclusion and post-expansion re-exclusion are untested, and the compose test mocks the pipeline (no end-to-end through `mem_context_compose`). The filter-only case is cheap to add.
- Docs changes (`pinned-context.md`, `operations.md`) were verified against the source and are accurate.

memtomem:
Applied all review feedback in `af81ce18`:

- classified `mem_context_compose(namespace)` as a deferred read-filter surface
- removed the partial 0.3.9 version bump and restored `[Unreleased]`; coordinated package/plugin pins remain for a separate release PR
- documented the bounded filter-only enumeration behavior when excluded sources consume the candidate budget
- documented that BM25/dense stats count eligible post-exclusion fusion inputs
- added filter-only exclusion coverage
- added post-context-expansion re-exclusion coverage
- added a tool-level `mem_context_compose` scope/response test

Verification: 227 focused tests passed, plus Ruff, mypy, lock, and diff checks.

PR #1741, opened 2026-07-13
feat: add safe OpenCode plugin

memtomem:
## Summary

- add `opencode-memtomem@0.1.0` as a configuration-only OpenCode plugin
- register the exact-pinned memtomem 0.3.8 MCP server, six commands, and three read-only skills
- merge conservative permissions without replacing user MCP, commands, skills, or memtomem-specific rules
- add generated-asset contracts, docs, CI/OSV coverage, Dependabot, and npm trusted-publishing workflow

## Safety

- read tools are allowed
- add/index remain `ask`
- `memtomem_mem_do` is denied by default
- no event hooks or automatic indexing are installed

## Validation

- `59 passed` across plugin contract, supply-chain, and docs tests
- `npm test`: 4 OpenCode package tests passed
- `npm run pack:check`: package dry-run passed with zero runtime dependencies
- OpenCode 1.17.18 live smoke: 6 commands and 3 skills loaded
- OpenCode MCP smoke: `memtomem==0.3.8` connected successfully
- `git diff --check` passed


tsdata (approved):

memtomem:
## Review — OpenCode plugin

**Verdict: NEEDS-FIX.** The design and supply-chain handling are release-quality, but required lint is red (format-only), and the permission merge has a real behavioral inconsistency that contradicts the plugin's core safety claim — worth fixing before publish.

### What's solid

- **Supply-chain hygiene is exemplary**: zero runtime dependencies (pinned by test), dev-only lockfile added to the blocking OSV scan, Dependabot group, new CI job wired into the `security-gates` fan-in, and npm trusted publishing (OIDC) with provenance, an `npm` environment gate, tag↔package.json version match, and a tag-on-main ancestry check — consistent with the repo's PyPI release patterns.
- **Generated-asset discipline holds**: `generated.ts` and the skills render from the single `contract.toml` source with in-sync check coverage, and the cross-runtime leak guards were extended to OpenCode output.
- **User-config precedence**: an existing `mcp.memtomem`, same-name commands, same-name user skills, and memtomem-specific permission rules win over plugin defaults, and re-application is idempotent — all pinned by tests.
- The MCP command is a fixed array (`uvx --from memtomem==0.3.8 memtomem-server`), so there is no injection surface, and the skill bodies gate writes/indexing behind explicit confirmation.

### Blocking

1. **CI red — `ruff format` (required lint check)** on the three touched Python files:

   ```
   Would reformat: packages/memtomem/tests/test_plugin_assets.py
   Would reformat: packages/memtomem/tests/test_supply_chain_contracts.py
   Would reformat: tools/render_plugin_assets.py
   ```

   `uv run ruff format` fixes it.

2. **`mergePermissions` treats object-form global rules inconsistently, weakening an explicit user deny.** The scalar branch preserves a global `"deny"`/`"ask"` (deny does not open memtomem reads). But the object form `{"*": "deny"}` lands in the `generic` bucket and then `SAFE_PERMISSIONS` is layered on top, adding `memtomem_mem_search: "allow"` etc. — the plugin weakens an explicit global deny, and the same setting behaves differently depending on spelling. That contradicts the README's "without weakening explicit rules". When the object's `"*"` value is `deny`/`ask`, handle it like the scalar branch, and add object-form `{"*": "deny"}` / `{"*": "ask"}` tests (currently only the three scalars and object-allow are covered).

### Suggestions (non-blocking)

- **Two path-boundary bugs in `hasUserSkill`**: (a) `absolute.endsWith(name)` also matches suffix paths like `~/skills/my-memtomem-search`, wrongly skipping the plugin skill — use `basename(absolute) === name`. (b) `absolute.startsWith(packageRoot)` is a prefix check that also matches `<packageRoot>-sibling` — use `startsWith(packageRoot + sep)` or a `path.relative`-based check. Same path-boundary principle the repo already enforces on the Python side.
- **Hardcoded compatibility string drift**: `_opencode_skill` in `render_plugin_assets.py` embeds the literal `"OpenCode >=1.17.18 <2"` instead of deriving it from the contract's `version_range`; a future range bump silently stales the skill frontmatter. Also, the contract's `minimum_version` field appears unconsumed by any test or asset — dead or unverified.
- **`_opencode_body` rewrites a fixed list of 5 tool names**: a future workflow mentioning `mem_list`/`mem_stats`/`mem_read` would leak unprefixed, and the leak guard only asserts the `mem_search` variant. Derive the replacement list from the contract or generalize the guard.
- **Docs pin the plugin version untested**: `opencode plugin add opencode-memtomem@0.1.0` appears literally in three docs; package.json↔contract is tested but the docs pins are not — a drift candidate on the next bump. A grep guard like the existing core-version docs guard would cover it.
- **mcp-clients.md TOC numbering**: an unnumbered "OpenCode" entry sits between sections 7 and 8 of a numbered list — renumber or match the convention.

memtomem:
Addressed all blocking items and follow-up suggestions in `a7e4421f`:

- applied repository-wide Ruff formatting to the touched Python files
- made object-form `{"*":"deny"}` / `{"*":"ask"}` preserve the same safety boundary as scalar forms, with regression tests
- fixed configured-skill path component and package-boundary checks
- derived skill compatibility and tool-name rewriting from the shared contract
- added docs-version and compatibility drift guards
- numbered OpenCode consistently and updated affected anchors/references

Validation: 59 related Python tests passed, 5 OpenCode package tests passed, full Python Ruff format/check passed, package dry-run passed, and generated assets are in sync.

memtomem:
Follow-up: the refreshed OSV database flagged five newly published Pillow 12.2.0 advisories (`PYSEC-2026-2253` through `2257`). Commit `436705f3` updates only Pillow to the fixed 12.3.0 in `uv.lock`; `uv lock --check` and the related test/lint suite pass locally.

Issue #1681, opened 2026-07-09
config.json overrides bypass cross-field model validators (setattr, no validate_assignment)


pandas-studio:
## Summary

`load_config_overrides` applies every `config.json` value onto an
already-constructed config model via plain `setattr`
(`config.py:1246-1247`). Because `ConfigModel` does **not** set
`validate_assignment=True` (`config.py:62`), Pydantic's `mode="after"`
cross-field model validators never re-run for override-loaded values. A
state the validator was written to reject loads silently through the
`config.json` path.

## Reproduction

```python
import os
os.environ.pop("LANGFUSE_PUBLIC_KEY", None)
os.environ.pop("LANGFUSE_SECRET_KEY", None)
from memtomem.config import Mem2MemConfig, SessionTraceConfig

# Direct construction is guarded (as designed):
SessionTraceConfig(enabled=True, langfuse_enabled=True)
# -> ValidationError: langfuse_enabled=true requires public/secret key

# The config.json override path is NOT guarded:
c = Mem2MemConfig()
setattr(c.session_trace, "enabled", True)
setattr(c.session_trace, "langfuse_enabled", True)
# -> no error; (True, True) accepted. validate_assignment == False
```

Equivalent real-world trigger: a `~/.memtomem/config.json` with

```json
{ "session_trace": { "enabled": true, "langfuse_enabled": true } }
```

and no `langfuse_public_key`/`secret_key` (in config or env) loads
cleanly, whereas the `_require_keys_when_enabled` validator
(`config.py:845-871`) was written specifically to reject it.

## Impact

All four `@model_validator(mode="after")` cross-field checks in
`config.py` are bypassed on the override path:

| Line | Validator | Rejected state that now slips through |
|------|-----------|----------------------------------------|
| 311  | `IndexingConfig.check_chunk_token_range` | `min_chunk_tokens > max_chunk_tokens`, `target > max` |
| 565  | `RerankConfig._check_pool_bounds` | `max_pool < min_pool` |
| 845  | `SessionTraceConfig._require_keys_when_enabled` | `langfuse_enabled=true` with no keys |
| 873  | `SessionTraceConfig._require_langfuse_package_when_enabled` | `langfuse_enabled=true` without the `langfuse` package |

`load_config_overrides` applies **any** `config.json` key that the section
`hasattr`s (not just `MUTABLE_FIELDS`), so the entire cross-field surface
[...]

PR #1745, opened 2026-07-13
fix(langgraph): preserve programmatic config precedence

memtomem:
## Summary

- resolve ambient config before applying `MemtomemStore(config_overrides=...)`
- let the component factory skip reloading ambient config for already-resolved callers
- add regression coverage proving isolated SQLite paths cannot be replaced by `~/.memtomem` settings

## Why

`create_components(config)` reloaded `config.d`, persisted config, and environment values after `MemtomemStore` applied its constructor overrides. Integrators that explicitly selected an isolated database could therefore write to the ambient user database instead.

## Validation

- `uv run pytest -q packages/memtomem/tests/test_langgraph.py packages/memtomem/tests/test_component_factory_config.py packages/memtomem/tests/test_multi_agent_integration.py packages/memtomem/tests/test_lazy_init_acceptance.py` (87 passed)
- targeted Ruff checks passed
- `git diff --check` passed
- broader suite reached 3,286 passed / 1 skipped before manual interruption; the displayed multiprocessing failure was caused by the Ctrl-C interruption

### DIFF
--- a/packages/memtomem/src/memtomem/integrations/langgraph.py
+++ b/packages/memtomem/src/memtomem/integrations/langgraph.py
@@ -84,12 +84,19 @@ class MemtomemStore:
     async def _ensure_init(self) -> Components:
         """Initialize components on first call; return the cached instance."""
         if self._components is None:
-            from memtomem.config import Mem2MemConfig
+            from memtomem.config import (
+                Mem2MemConfig,
+                load_config_d,
+                load_config_overrides,
+            )
             from memtomem.server.component_factory import create_components
 
             config = Mem2MemConfig()
+            load_config_d(config)
+            load_config_overrides(config)
 
-            # Apply overrides. Unknown sections / keys raise immediately:
+            # Apply programmatic overrides after ambient config. Unknown
+            # sections / keys raise immediately:
             # ``config_overrides`` is a programmatic constructor argument, so
             # a typo like ``{"storge": ...}`` or ``{"storage":
             # {"sqlite_pat": ...}}`` would otherwise fall back to the default
@@ -117,7 +124,11 @@ class MemtomemStore:
                         )
                     setattr(section_obj, key, value)
 
-            self._components = await create_components(config)
+            # Ambient config was resolved above so the constructor arguments
+            # remain the final, highest-precedence layer. Loading it again in
+            # the factory would overwrite an isolated sqlite_path or
+            # memory_dirs with ~/.memtomem settings.
+            self._components = await create_components(config, load_ambient_config=False)
         return self._components
 
     async def close(self) -> None:
--- a/packages/memtomem/src/memtomem/server/component_factory.py
+++ b/packages/memtomem/src/memtomem/server/component_factory.py
@@ -60,13 +60,24 @@ class Components:
     embedding_broken: dict | None = None
 
 
-async def create_components(config: Mem2MemConfig | None = None) -> Components:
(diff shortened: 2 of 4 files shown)


tsdata (approved):

memtomem:
## Code review — SHIP ✅

**Overview.** `MemtomemStore(config_overrides=...)` applied programmatic overrides and then `create_components()` re-loaded `config.d`/persisted config on top, silently clobbering them — so an integrator's isolated DB could get redirected to the ambient `~/.memtomem` store. This adds a `load_ambient_config` kwarg to `create_components` and reorders `_ensure_init` to resolve ambient config first, then apply the constructor overrides as the final (highest-precedence) layer.

Verdict: correct, correctly scoped, backward-compatible. No blockers.

### Correctness ✅
- **Precedence is right.** Final order is `Mem2MemConfig()` (defaults + env) → `load_config_d` → `load_config_overrides` (env wins over `config.json`) → programmatic `setattr` → `create_components(..., load_ambient_config=False)`. The constructor arg is now the top layer, as intended.
- **No double load.** The caller loads ambient once and the factory skips it, so ambient config is resolved exactly once (no perf regression).
- **Monkeypatch seams hold.** In-function `from memtomem.config import ...` resolves module attributes at call time, so `monkeypatch.setattr(_cfg, "load_config_d", ...)` lands. `create_storage` is called unqualified at module scope in the factory, so the `monkeypatch.setattr(_factory, "create_storage", ...)` guard is valid.

### Scope / conventions ✅
- **Backward-compatible.** `load_ambient_config: bool = True` default leaves the other six callers (`web/app.py`, `server/context.py`, `cli/_bootstrap.py`, conftest, etc.) untouched — all paths that legitimately want ambient config.
- **No sibling bug.** `integrations/langgraph_store.py` also loads ambient config, but it's a separate `BaseStore` that only does so when `root is None` (to derive a default root); explicit `root`/`embedding` kwargs take precedence, so it isn't affected. The fix is scoped precisely to the one buggy path.
- Keyword-only arg, clear docstring, good rationale comments.

### Test coverage ✅
- Factory level (`test_component_factory_config.py`): asserts the ambient loaders are not called when `load_ambient_config=False` (raising monkeypatch).
- Integration level (`test_programmatic_override_wins_over_ambient_config`): a real regression test — ambient tries to set `ambient_db`, and the test asserts `isolated_db` wins and `load_ambient_config=False` is threaded through end-to-end.
- Existing `_fake_create` stubs updated to accept the new kwarg — nothing left stale.

### Security ✅
Effectively a security fix: prevents an integrator's explicitly-isolated DB / memory_dirs from being silently redirected to the user's ambient store (wrong-store writes / data leakage). Right direction.

### Minor (non-blocking)
- The programmatic override path uses raw `setattr` (`langgraph.py`), which bypasses cross-field validators — same pattern as `feedback_config_override_setattr_bypasses_validators` (#1681). Pre-existing, not introduced here; a possible follow-up is to reuse the section re-validate + assign-back pattern from `load_config_overrides` on the programmatic path too.
- `test_factory_can_skip_already_resolved_ambient_config` proves "factory reached + loaders not called" rather than directly asserting `sqlite_path` was preserved; the stronger guarantee lives in the integration test, so this is acceptable.

_Reviewed by reading the diff against current `main` (this checkout is on `main`, so tests weren't executed locally); the static analysis matches the PR's reported 87-passed run. CI green is the only gate left._

PR #1747, opened 2026-07-13
feat: preserve context windows in compose schema 3

memtomem:
## Summary
- add context-compose schema 3 as an additive superset of the released schema 2 contract
- serialize budgeted adjacent chunks with source and namespace metadata
- preserve matched-hit priority, whole-chunk budgeting, and original before/after order
- advertise schema 3 and document why schema 2 remains immutable

## Compatibility
Core 0.3.9 already released schema 2 without adjacent chunks in the response. Reusing schema 2 would make identical capability advertisements behave differently by release. Existing clients can ignore the optional context field; schema-aware clients can require schema 3 for visible windows.

## Validation
- 159 focused tests passed across pinned context, capability, CLI help, and docs
- ruff check and format passed for all 301 source files
- mypy passed for the changed pinned module; repository-wide advisory mypy retains 14 unrelated baseline errors
- real stdio core-to-STM schema 3 smoke passed
- full local suite reaches a platform SIGTRAP on macOS under both Python 3.12 and 3.13; Ubuntu CI remains the authoritative full gate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes will be documented in this file.
 Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
+## [Unreleased]
+
+### Added
+
+- **Context Compose schema 3** — composed retrieval results can include
+  de-duplicated adjacent chunks with source and namespace metadata after all
+  matched hits claim their schema 2 budget; remaining context is selected
+  globally by distance. Schema 2 remains
+  the immutable scoped-composition contract released in 0.3.9; clients can use
+  the integer capability to distinguish visible context-window support.
+
 ## [0.3.9] — 2026-07-13
 
 ### Added
--- a/docs/guides/pinned-context.md
+++ b/docs/guides/pinned-context.md
@@ -21,11 +21,20 @@ default composed bundle to 12,000. A block is never cut in the middle; omitted
 ids are returned explicitly.
 
 `mem_context_compose` schema 2 accepts the same optional `namespace` and
-`context_window` retrieval controls used by search. Pinned files remain
-searchable through legacy `mem_search`, but the composed path excludes every
-file under the active user and project `pinned/` roots from its retrieved leg,
+`context_window` retrieval controls used by search. Schema 3 additionally
+returns adjacent chunks under each retrieved item's optional `context` object.
+Matched hits retain schema 2 budget priority; de-duplicated adjacent chunks use
+only the remaining composed bundle character budget, selected globally by
+distance with hit rank and before/after as deterministic tie-breakers. Pinned
+files remain searchable through legacy `mem_search`, but the composed path
+excludes every file under the active user and project `pinned/` roots from its retrieved leg,
 including shadowed, agent-specific, and malformed blocks. This prevents pinned
 content from appearing a second time as an ordinary search result.
+
+Schema 3 is a new capability instead of an in-place schema 2 change because
+schema 2 shipped in memtomem 0.3.9 without adjacent chunks in its response.
+Keeping that released contract immutable lets clients distinguish scoped
+composition from composition that also preserves visible context windows.
 
 ## Review-first candidates
 
--- a/docs/guides/reference/operations.md
+++ b/docs/guides/reference/operations.md
@@ -151,7 +151,15 @@ Running both `memtomem-server` (MCP) and `memtomem-web` simultaneously is suppor
 
 ## STM: Proactive Memory Surfacing (Optional)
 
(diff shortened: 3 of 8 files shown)


memtomem (commented):
## Review — context compose schema 3

The neighbor-budgeting logic is correct (verified against both new tests): nearest-first, before/after interleaved, `reversed(selected_before)` restores source order, whole-chunk budgeting with no mid-chunk cuts. Docs, capability bump, and tests are consistent. Three things to consider before merge:

### 1. (Medium) No de-duplication across matched hits and adjacent context

The `retrieved` loop never tracks chunk ids. If two adjacent chunks from the same file both surface as matched hits in `top_k`, hit B can appear both as its own `retrieved` item **and** inside hit A's `context.after` — serialized twice and charged against `max_chars` twice. Schema 2 couldn't hit this because it carried no context. Since "preserve context windows" makes overlapping windows plausible, consider a `seen` id set that suppresses an adjacent chunk already emitted as a matched hit (and shared neighbors between two hits).

### 2. (Medium) With `context_window` set, schema 3 can drop matched hits schema 2 would keep

The PR frames schema 3 as purely additive ("existing clients can ignore the optional context field"), but adjacent chunks consume the shared budget, so **`retrieved` and `used_chars` themselves change** for the same input. Under 0.3.9 schema 2, `result.context` was ignored, so matched hits monopolized the budget; under schema 3 a top hit's neighbors can crowd out a lower-ranked matched hit. Because the capability is bumped unconditionally to 3, there's no per-request way to ask for schema-2 budgeting — a schema-2 client that sends `context_window` (which the docs say schema 2 accepts) gets a changed core response it can't opt out of. The behavior is intended and documented ("counts against the budget"), but it's additive only when `context_window` is unset; worth making that boundary explicit in the compatibility note.

### 3. (Low) Unrelated CLI help tests

The `mem`/`pinned`/`review` subcommand assertions and `test_reviewed_resume_command_help` (`--scope`, `--max-chars`, `--reviewer`) in `test_cli.py` are unrelated to schema 3. Harmless, but they mix an extra change into this PR.

Overall the logic, tests, and docs are solid — #1 is the only finding that could produce visible output changes in practice.

memtomem:
Addressed all three review points in 37c43377.

- Context composition now reserves the schema-2 budget for all matched hits before adding any schema-3 neighbors.
- Adjacent chunks are de-duplicated by chunk id across matched hits and all context windows, so each chunk is serialized and charged at most once.
- Remaining context budget is allocated globally by distance, with hit rank and before/after as deterministic tie-breakers; selected before chunks are still restored to source order.
- The compatibility docs now state this budget policy explicitly.
- The unrelated CLI help assertions were removed from this PR.

Regression coverage includes the case where a high-ranked hit references the next matched hit and a shared neighbor: both hits remain present, the shared chunk is attached only at its nearest distance, and used_chars counts it once. Validation: 36 focused tests passed, changed-module mypy passed, ruff check/format passed, and git diff --check passed.

memtomem (commented):
## Re-review — both prior findings resolved

The updated two-phase compose addresses both Medium findings from the previous pass. Verified against all three tests.

### #1 De-duplication — resolved

`emitted_chunk_ids` now tracks matched-hit ids plus already-emitted context-chunk ids, so an adjacent chunk that is itself a matched hit (or a neighbor shared between two hits) is skipped instead of serialized twice / double-charged. The new `test_compose_schema_three_preserves_hits_and_deduplicates_context` covers exactly that case (hit B in hit A's `window_after`, shared neighbor): traced to `retrieved=[first, second]`, no `context` on `first`, `second.after=[shared]`, `used_chars=10`.

### #2 Matched-hit priority — resolved

Phase 1 claims the full matched-hit budget in rank order before phase 2 spends any remaining capacity on context. As a result the `retrieved` hit set is now identical to schema 2 for the same input — a top hit's neighbors can no longer crowd out a lower-ranked matched hit. The CHANGELOG/docs wording was corrected to match ("after all matched hits claim their schema 2 budget", "selected globally by distance"). `used_chars` still grows by the context content, but that's an informational field a schema-2 client ignores.

### #3 CLI test scope — mostly addressed

The unrelated `test_reviewed_resume_command_help` parametrization was dropped; only the `mem`/`pinned`/`review` subcommand-presence assertions remain, which are legitimate (if incidental) coverage.

### Minor

- The `context_windows` 6-tuple annotation is correct but verbose; a small `NamedTuple`/dataclass would read better. Optional.
- Global distance allocation is O(context_window × hits) — fine.

Logic, tests, and docs are consistent and the regressions are covered. LGTM.

memtomem:
Applied the optional readability suggestion in f31d79b5: the positional 6-tuple is now a private slots dataclass with named fields and per-instance selection lists. Selection order, de-duplication, budget accounting, and wire output are unchanged. Validation: 36 focused tests passed; changed-module mypy, ruff check/format, and diff check passed.

memtomem (commented):
## Re-review — dataclass refactor, logic unchanged

This revision is a pure readability refactor addressing the earlier optional note: the verbose `tuple[Any, dict, list, list, tuple, tuple]` selection tuple is replaced by a `_ContextWindowSelection` dataclass (`@dataclass(slots=True)`, named fields, `field(default_factory=list)` for the mutable selections).

- **Logic is identical** — global distance allocation, `emitted_chunk_ids` de-duplication, phase-1 matched-hit budget priority, and `reversed(selected_before)` source-order restoration all unchanged; only tuple-unpacking became attribute access.
- `slots=True` + `default_factory=list` is correct on 3.12 with no mutable-default pitfall.
- The three schema-3 tests are unchanged and still cover budgeting, hit-preservation-under-budget, and dedup.

No remaining concerns. LGTM.

tsdata (approved):

PR #1748, opened 2026-07-13
feat: add bilingual retrieval benchmark

memtomem:
## Summary

- fix per-request RRF weight cache isolation and correct the sensitivity measurement
- add a frozen 120-query bilingual holdout with separate English, Korean, and cross-language tracks
- add portable qrels, quality gates, CI coverage, and preregistered RRF tuning
- compare language-specific embeddings, BGE-M3, and multilingual reranking
- document the one-run experiment, measured Apple M4 Max environment, limitations, and follow-up work

## Validation

- 65 targeted tests passed
- retrieval v2 baseline passed for all 120 English, Korean, and cross-language queries
- public corpus audit passed: 48 files, 192 chunks, no email, IP, or secret hits
- git diff check passed

## Experiment status

The BGE-M3 and reranker measurements are recorded as provisional evidence. This PR does not change the product default embedding or enable reranking by default. A later change should repeat each profile 5-10 times and measure variance, cold-start time, memory use, and reranker pool sizes before proposing a production default change.

## Known limitations

- the committed model comparison is a single run on an Apple M4 Max with 64 GB RAM
- a broader non-LLM suite previously stopped at about 32 percent with process SIGTRAP rather than a test assertion
- mypy still reports 14 pre-existing errors in files outside this change


tsdata (approved):

memtomem:
## Code review — head `7b0a8ed`

Reviewed against the current head (rebased onto latest `main`, +3 commits since the initial pass). Focus is the delta: English-doc swap, variance-aware floors, and the directional-gate fix. Verified locally against the committed artifacts.

**CI:** all required checks green including `golden-path` (which runs the v2 gate on ubuntu); platform test matrix (macOS/ubuntu/Windows) still pending with no failures.

### Overview (5 commits)

| commit | change |
| --- | --- |
| `59f983f` | RRF per-request cache isolation fix + v2 bilingual benchmark + CI gate |
| `42a5d64` | k-sweep follow-up methodology docs |
| `14e8cfb` | translate model-comparison report to English (KO → EN) |
| `02d720e` | variance-aware floors (per-track spread) |
| `7b0a8ed` | directional-gate fix (per-metric spread + ceilings) |

### Correctness — verified ✓

**Key fix in `7b0a8ed` — a real directionality bug.** `hard_negative_hits@10` is a *lower-is-better* metric, but the original baseline (and `02d720e`) gated it as a **floor** (lower bound) — i.e. "must retrieve at least N hard negatives to pass", the wrong direction: an *improvement* (fewer hard negatives) would fail the gate. `7b0a8ed` reclassifies it via `LOWER_IS_BETTER_SUFFIXES` into a **ceiling** (`round(mean * 1.10 + spread, 6)`). Confirmed the floor/ceiling partition is mutually exclusive — zero `hard_negative_hits` keys leaked into `quality_floors`.

**Per-metric spread precision.** `02d720e` subtracted a per-track `max_run_spread` from every floor, over-loosening stable metrics. `7b0a8ed` adds per-metric `run_spreads` (max−min per run) in `_combine_track_runs` and subtracts each metric's own spread, so a volatile slice no longer weakens stable metrics. Verified e.g. `ko|negation|hard_negative_hits@10` ceiling = `0.88` (per-metric spread 0), where the blunt per-track approach would have produced `0.967`.

- `_quality_bounds` matches its unit test (floor `0.395`, ceiling `1.5`).
- `check_baseline_v2.compare()` gates floors as `observed < floor → fail` and ceilings as `observed > ceiling → fail`, both handling `observed is None`.
- **Code↔data parity holds**: recomputing `_quality_bounds(track)` from the committed aggregates reproduces the committed `quality_floors`/`quality_ceilings` exactly across all three tracks; each track carries per-metric `run_spreads` (10-run regeneration).
- The pipeline cache-isolation fix (`59f983f`) remains correct: the default path (`rrf_weights=None`) produces a byte-identical cache key to before, so no cache churn.

### Test coverage — strengthened ✓

Three new tests: `run_spreads`+bounds formula, directional ceiling failure, and — notably — `test_v2_committed_quality_bounds_match_generation_formula`, which pins the committed baseline to the generation formula so future hand-edits or regenerations can't silently drift. 65 → 68 passing.

### Conventions ✓

The KO→EN doc swap satisfies the "committed docs in English" convention (no dangling references; the README distinguishes the v1 `round(mean*0.85, 2)` rule from the v2 spread-aware rule). CHANGELOG and the report's test count are kept in sync.

### Minor observations (non-blocking)

- **Korean per-metric spreads are ≤0.005 in this regeneration**, so the effective floor margin is back near ~10%. The mechanism is sound (10-run calibration captures variance), but if tie-break variance resurfaces on a specific slice the cushion is thin. Acceptable given the calibration design.
- **`same_language_precision@10` is treated as higher-is-better (floored).** For the cross-language track "more same-language results = better" isn't unambiguous; the floor is conservative (won't fail on improvement), so this is a semantics note, not a defect.
- `compare()` hard-indexes `baseline["quality_ceilings"][track]`; a track added without a ceilings entry would `KeyError`. Consistent with the existing `quality_floors` pattern.

### Verdict: **SHIP** ✅

The directional-gate fix resolves a latent wrong-direction gate present in both the original baseline and the interim floor commit; per-metric spreads plus the parity test make the gate more precise and self-validating. `golden-path` (ubuntu) confirms the new gate passes cross-platform. Mergeable once the pending platform matrix goes green; the three observations are optional follow-ups.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #1750, opened 2026-07-14
docs: align v0.3.10 contracts and retrieval evaluation

memtomem:
## Summary
- align public docs, MCP client setup, CLI catalog, and runtime help with v0.3.10
- harden generated MCP server commands across source, persistent, and uvx installs
- add the 19-profile retrieval k-sweep runner, raw results, report, and benchmark-v2 guard coverage
- correct OpenCode setup, pinned-context, search, operations, and optional-extra documentation

## Validation
- 382 targeted tests passed
- 7 web-targeted tests passed, 1 browser-dependent test skipped
- MyPy passed for changed source files
- git diff --check passed
- k-sweep JSON selections and profile counts validated

## Notes
- Defaults remain top_k=10, candidate_limit=50, RRF k=60, reranking disabled
- Local docs/plans and docs/qa artifacts are intentionally excluded


tsdata (approved):

memtomem:
## Code review — v0.3.10 contract alignment + retrieval k-sweep

Reviewed the diff against `origin/main` (single commit), plus a second pass from our Codex reviewer. Verdict: **NEEDS-FIX** — the core `_mcp_server_command()` refactor and its tests are solid, but two of the regenerated MCP-setup surfaces now emit PATH-dependent commands that break for the supported project-venv (`uv add`) install path.

### Strengths
- `_mcp_server_command()` is a clean improvement over the old blanket `uvx --from memtomem` branch, which silently dropped optional extras. The new branching is exhaustive and each shape is right: source/project → `uv run --directory <dir>`, uvx origin → `uvx --isolated --from memtomem[all]==<version>`, everything else → `<interpreter> -m memtomem.server` (preserving the resolved env/extras). The docstring rationale for keeping the raw interpreter path (avoiding venv-shim/symlink env loss) is appreciated.
- Test coverage is strong: per-origin parametrization, a real `subprocess` launch of `-m memtomem.server --help`, and docs-as-tests guards enforcing the version pin, the 96-tool MCP table, the extras table, and the CLI command list. Benchmark `k`-param validation and the `assess` gate are guarded too.
- `search.py` docstring correction (no per-call `rerank_pool`; pool derives from `rerank.oversample/min_pool/max_pool`) matches the actual config contract. Version pins (`memtomem[all]==0.3.10`) are consistent with pyproject.

### Major
- **`web/static/context-portal.js:147` (+165, +181)** — The Claude/Antigravity/Codex install guides now register bare `memtomem-server` (`"args": []`). This assumes the binary is on the global PATH, which holds for `uv tool install` but **not** for `uv add` project installs, where `memtomem-server` lives in the project `.venv/bin`. A user who launched the web UI via `uv run mm web` and copies this block will register an MCP command an external client (Claude Desktop / Codex CLI) can't resolve. The prior `uvx --from memtomem` form worked regardless of install shape — this is a correctness regression for a supported path. Prefer an install-aware command, or point users at `mm init --mcp ...`, which already persists the correct `uv run --directory ...` / interpreter command.
- **`server/__init__.py:425/430`** — `_print_direct_stdio_help()` prescribes an exact-pinned `uvx` command in the "Configure your MCP client with" block, but the copyable `Example:` line below it registers bare `memtomem-server`. The two blocks now contradict each other, and the copyable one can break when the server was launched from an activated project venv or under a different PATH. Make the example match the displayed `uvx` command, or clearly split persistent-install vs no-install examples.

### Minor / Nit
- **`tests/test_docs_guards.py:733`** — `test_public_docs_do_not_use_floating_bare_uvx_server` only rejects two literal spellings and only scans Markdown, so quoting/spacing variants evade it and the runtime-help surfaces (`_print_direct_stdio_help()`, `context-portal.js`) are out of scope — which is exactly why the Major inconsistency above isn't caught by CI. Consider parsing `uvx` command tokens / a broader regex and adding assertions for both runtime surfaces.
- The version-pin guard couples every public-doc `memtomem[all]==X.Y.Z` to pyproject `version`. Intentional (docs-as-tests), but it means every release must bump all doc pins together — worth a line in the release runbook.

### Validation
- Focused Python tests: 358 pass (Codex) / 382 pass targeted (local). Ruff check + format pass. MyPy pass on changed source. Context Portal browser test skipped (fixture unavailable).

Nice work on the `_mcp_server_command()` consolidation and the guard coverage — the remaining asks are about not letting the human-facing setup guides drift back to PATH-fragile commands.

memtomem:
Addressed the NEEDS-FIX findings in f648ff87.

- Context Portal Claude, Antigravity, and Codex guides now use exact-pinned uvx commands with memtomem[all]==0.3.10, so they do not depend on the external client inheriting a project venv PATH.
- Direct stdio help now uses the same exact-pinned uvx command in both the configuration block and the copyable Claude example.
- Expanded the floating uvx guard and added explicit coverage for the JS runtime setup surface and direct-help output.

Validation: 64 targeted tests passed, including the Chromium Context Portal tests; Ruff check/format and git diff --check passed.

memtomem:
## Re-review — after `f648ff87` (`fix: make MCP setup commands path independent`)

The follow-up commit resolves all three findings from the prior review. **Verdict: SHIP.**

### Prior findings → resolution
- **Major 1 — `context-portal.js` bare `memtomem-server`** — Resolved. All three guides (Claude terminal, Antigravity JSON, Codex TOML) now emit `uvx --isolated --from 'memtomem[all]==0.3.10' memtomem-server`. Since uvx fetches from PyPI, the command is launchable regardless of how the web UI itself was installed, so the project-venv (`uv add`) regression is gone. The rationale is now captured in a code comment.
- **Major 2 — `_print_direct_stdio_help` inconsistency** — Resolved. The copyable `Example:` line now matches the `command: uvx` block above it (both exact-pinned uvx), so the two no longer contradict.
- **Minor — weak docs guard** — Resolved, and then some. `test_public_docs_do_not_use_floating_bare_uvx_server` now uses a `floating_from` regex that catches both the CLI (`--from memtomem memtomem-server`) and JSON-array (`"--from", "memtomem", ...`) spellings without false-positiving the pinned form. Two new guards extend coverage to the runtime surfaces that were previously unguarded: `test_runtime_setup_surfaces_use_exact_pinned_uvx` (asserts the JS pin equals pyproject `version` and no bare command remains) and a `test_server_cli.py` assertion on the direct-stdio help example. The regression can now be caught in CI.

### Validation
- Ran the changed guard surfaces locally: `test_docs_guards.py::TestRegistryAndInstallDocs` + `test_server_cli.py` → **24 passed**.
- `test_context_portal_runtimes` needles were updated to the pinned uvx form; expected green where the chromium fixture is available.

No remaining blockers. The fix not only closes the two setup-command regressions but strengthens the guard net around them.

PR #1760, opened 2026-07-15
fix: read the memory index as Markdown (mm memory doctor false orphans + broken_link blind spot)

memtomem:
Fixes defects 1 and 2 of #1757 — the reporting half. `--fix`'s write scope is
deliberately unchanged; relaxing it amends ADR-0020 §1 and follows separately.

## The defect

`mm memory doctor`'s index parser was anchored at line start and applied with
`.match()`, so a `MEMORY.md` line yielded at most one entry and a prose-prefixed
bullet (`- NS: [a](x.md)`) yielded none. Everything it could not see counted as
absent from the index, which broke two checks in opposite and compounding ways:

- **`index_orphan` false positives.** On the 364-file memory dir in the issue,
  whose index packed 228 links onto 74 lines, 63 were parsed and 301 orphans
  reported against 136 real ones.
- **`broken_link` false negatives.** A link at position ≥2, or on a
  prose-prefixed line, was never classified — so a dead pointer there reported
  clean. A safety check that passes when it shouldn't is the worse half.

## What changed

**The index is read as Markdown, by a Markdown parser.** The first attempt kept
the hand-rolled grammar and tried to detect its own mis-reads. Review broke that
three times over, each round finding a new way a destination can lie about the
path it resolves to — a paren inside it, a backslash escape, a character
reference, code spans delimited by backtick runs, nested brackets in a label.
Each was a live pointer the doctor would call dead, and `--fix` deletes on
exactly that verdict. The enumeration was of Markdown itself, so it was never
going to converge; `markdown-it-py` resolves all of it because that is what a
CommonMark parser is.

That is a new declared dependency, but not a new install: it already ships to
every user via `mcp[cli]` → `rich`. Link normalization is turned off — it
percent-encodes for the web, and `한글노트.md` is a filename, not a URL.

**The document is read as a document.** Pointers come only from genuine list
items in the token stream, so a bullet inside a ```` ```markdown ```` fence is an
*example* of an index line rather than an index line — a distinction no line-wise
reader can make, and one that had `--fix` editing code blocks. Reference
definitions resolve for the same reason (a line read alone can't see
`[live]: live.md` three lines down, and a live reference sibling going unseen let
its dead inline neighbour take the whole line).

**Each pointer is judged on its own.** Readability is per entry, so one
destination we won't resolve doesn't silence the pointer beside it. Only the
refusal to delete is per line — deletion takes the whole line, so doubt anywhere
on it is disqualifying.

**What stays flagged is the genuinely undecidable**: a destination carrying a
query, percent-escape, scheme or space may name a different file than its text
does. Those are reported as the new warn-severity `ambiguous_index_line`, feed
neither the broken-link nor the listed set, and are never deleted. Filenames that
merely need Markdown escaping now resolve normally instead of being quarantined.

**`--fix`'s write scope is frozen** at what it could already delete: candidates
must still match the legacy single-entry shape, and multi-entry, ambiguous,
multi-line and out-of-shape lines are all refused. Reading more of the index is a
Tier 1 change; deleting more of it is an ADR-0020 §1 change. Net effect here:
deletions are a strict subset of the previous behaviour, and two data-loss paths
that were reachable in shipped code are closed.

**The guards are re-asked under the lock.** They had been an admission check at
analysis time only, so an agent changing the file *elsewhere* — defining a
reference the candidate line already cited — turned an unchanged single-entry
line into one carrying a live sibling, and the splice took both.

## Verification

- 8615 passed (`-m "not ollama"`), ruff clean, `uv.lock` in sync.
- Every counterexample above is pinned as a parametrized table
  (`TestDeceivingLines`), asserting the resolutions rather than the workarounds.
  None were imagined up front; each came from an adversarial pass over the
  version before it.
- Each guard is mutation-verified: reverting it reds its own pins, and each
  mutation asserts it actually applied before the run.
- Driven end to end against a real `MEMORY.md` in an isolated `HOME`: fenced
  examples are neither link-checked nor edited; escaped, entity-encoded and
  parenthesised filenames resolve; a dead link beside an unreadable one is still
  reported; every refusal leaves the file byte-identical.

Refs #1757


memtomem:
## Review

Verified locally against the PR head (`fix/1757-index-parser-multi-entry`): all 117 tests in `test_memory_doctor.py` pass, and the one suspicious spot below was reproduced with a script rather than read off the diff.

### 🟡 Medium — unresolved link syntax sharing a line with a good link is never reported

The syntax-only ambiguity reporting in `_analyze_index_file` uses `parsed.ambiguous_lines - entry_lines`, and that set difference excludes every line that carries at least one entry. So:

```markdown
- [A](a.md) — and [B](b.md
```

lands in `ambiguous_lines` (the parser-level test `test_unresolved_link_syntax_beside_a_good_link_is_flagged` pins that much), but the analysis emits **no finding at all**: the `a.md` entry is `unreadable=False`, so the entry path doesn't report it, and the line is in `entry_lines`, so the syntax path skips it. Reproduced locally — entry-based items `[]`, syntax-only lines `frozenset()`.

This contradicts the finding summary's own claim ("hold link syntax that resolved to no link — they are … reported"), and it is the exact counterexample to this PR's core principle: an unread pointer `[B](b.md` hiding behind the company it keeps. If `a.md` is live, the line is never a `--fix` candidate either, so nothing surfaces it anywhere. (`--fix` safety is unaffected — the line is still refused if it ever becomes a candidate.)

**Suggestion:** carry unresolved-syntax lines in `ParsedIndex` as a set distinct from unreadable-target lines (e.g. `unresolved_syntax_lines`) and report them regardless of whether the line also has entries. The root cause is that two different doubts are folded into one `ambiguous_lines` set, making the set difference an inaccurate proxy for "syntax-only".

### 🔵 Low / design question — one ambiguous candidate aborts the whole `--fix` run

`_missing_target_entries` still runs `classify_link` on the *literal* text of `unreadable` entries. The analysis half deliberately skips classification for them ("an unreadable target is evidence of nothing"), but the fix half reads the same target literally, makes it a candidate, and `_unfixable_lines` then raises `UsageError` — aborting the entire run. One `live.md?view=1` line blocks fixing genuinely dead links in the same file and in every directory after it.

The PR body and `test_unresolvable_target_line_is_refused` pin this as intended, and it is fail-closed — but excluding `unreadable` entries from candidacy altogether would be more consistent with the analysis-side principle (the line can never be deleted anyway) and would let the rest of the fix proceed. It would also remove the asymmetry where the same input is a *warn* in the report but a *hard error* under `--fix`. This is not an ADR-0020 scope question (nothing new becomes deletable; an abort becomes a skip), so it could land here or as a follow-up. Relatedly, the docs row's "`--fix` refuses to touch these lines" slightly understates that the whole run can abort.

### 🔵 Low — smaller items

- **The fix path parses the same text three times**: `_missing_target_entries` (parse 1) → `_run_fix`'s `parse_memory_index(text)` (parse 2, identical text) → the fresh parse under the lock (parse 3, necessary). The first two could share one `ParsedIndex`. Negligible at the 200-line budget; just a cleanup.
- **`markdown-it-py>=3.0` is uncapped** while 4.x exists; the instance-attribute override of `normalizeLink`/`validateLink` is a documented pattern but a plausible surface for a major-version break. Consider `<5` (matching the `langgraph>=1.0,<2` style) — though `TestDeceivingLines` effectively acts as a contract suite, so going uncapped is defensible too.
- **`_unfixable_lines` hardcodes 6-space output indentation** inside the helper; formatting naturally belongs to the caller (`_run_fix`). Style nit.

### What's good

- **The approach change is right.** The judgment that enumerating CommonMark's divergences in a hand-rolled grammar would never converge is borne out by the three-review-round history recorded in `TestDeceivingLines`, and the dependency is a declaration rather than a new install (`mcp[cli]` → `rich` → `markdown-it-py`), with the reasoning left in the pyproject comment. Disabling `normalizeLink` (keeping `한글노트.md` out of percent-encoding) is exactly the right detail.
- **Read-widened / write-frozen separation is clean.** `_LEGACY_ENTRY_SHAPE_RE` freezes `--fix` at precisely the pre-PR shape and `TestFixScopeFrozen` pins it — the "deletions are a strict subset" claim checks out in code and tests.
- **Guards re-asked under the lock** (`_unfixable_reason(e, parsed=fresh)` in `_apply_fix`): the observation that an admission-time check can't see a reference definition added elsewhere is correct, and `test_revalidate_line_that_left_fix_scope_is_spared` covers it.
- **Per-entry verdicts vs per-line refusal** is the right split (the Medium above is this principle not quite reaching the syntax-only path), and the block-context work — fenced examples never edited, `_own_inlines` preventing child-adoption, the loose-list map trap — is each individually pinned.
- **Conventions:** `uv.lock` synced in the same commit, public doc updated in the same PR, 100-char lines, and the noise-budget regression pin (34/189 lines flagged by an early version) is a nice guard for the finding's usefulness.

### Test coverage — thorough; small gaps

- **Ordered-list and `+` bullets**: the CommonMark parser newly reads `1. [a](x.md)` / `+ [a](x.md)` (the legacy regex only matched `[-*]`). Neither the new reporting surface nor the legacy-shape refusal under `--fix` is pinned for them.
- **Undefined reference link** `- [Live][nope]`: stays literal text with no `](`, so it is silently prose — no entry, no flag. Shortcut references are indistinguishable from bracketed prose, so not flagging is probably right, but worth pinning as intended.
- **A CRLF index fixture** to confirm `splitlines()` line numbers stay aligned with markdown-it source maps.

### Verdict

Approvable, with the Medium item recommended before merge — it is a small fix (split the set, one test) and it closes a counterexample to the PR's own stated principle. The Low items can follow up.

tsdata (approved):

memtomem:
Thanks — the Medium was real and reproduced before I touched anything. Fixed in
24de9fa5, along with the test gaps and the small items. One deferred, with
reasoning.

## 🟡 Medium — fixed

Confirmed exactly as described: `- [A](a.md) — and [B](b.md` produced **no
finding at all**. Your diagnosis of the root cause was the useful part — one set
holding two unrelated doubts, so the report had to subtract one to guess at the
other.

`ParsedIndex` now tracks `unresolved_syntax_lines` apart from the entries, and
`ambiguous_lines` — the only place the union is right, because deletion takes the
whole line — became a derived property. The syntax path reports regardless of
whether the line also carries entries.

The part worth recording: **the parser-level pin passed throughout**. It asserted
`ambiguous_lines`, which was correct the whole time; the defect lived one layer
up, in the set difference. Only a pin on the *report* could catch it, so that's
where `test_unread_pointer_is_reported_beside_a_good_link` goes. Same lesson the
per-entry readability fix taught two commits earlier, and I didn't generalise it
then.

## Test gaps — both pinned

- **Ordered / `+` markers** (`TestListMarkers`): confirmed the CommonMark parser
  reads `1.` and `+` where the old `^\s*[-*]\s` grammar ignored them, and that
  `--fix` still refuses them for want of the legacy shape. Both halves now
  parametrized — the read is the fix (they were invisible before: unchecked *and*
  counted as orphans), the refusal is the frozen write scope.
- **Undefined shortcut reference**: pinned as intended. `- [Live][nope]` stays
  prose because it's indistinguishable from ordinary bracketed hook text
  (`see [note] below`) — guessing it meant a pointer would flag hooks across a
  healthy index.
- **CRLF fixture**: `TestSpliceRoundTrip` already covers CRLF byte-identity end to
  end through the splice, which is where a source-map/`splitlines()` drift would
  actually bite. I'd rather not add a second fixture asserting the same property
  one layer down.

## 🔵 Low — taken

- **Triple parse**: down to two (the analysis parse is now shared with
  `_missing_target_entries` via an optional `parsed=`). The third — under the
  lock, on fresh bytes — has to stay: that's the point of it.
- **`markdown-it-py` cap**: added `<5`. The repo leaves core deps uncapped, but
  those don't override library internals; we swap `normalizeLink`/`validateLink`
  on the instance, and a major reworking them would break at **user** runtime, not
  in CI, which resolves through `uv.lock`. Comment says to re-run
  `TestDeceivingLines` when raising it.
- **Indent in `_unfixable_lines`**: now returns `(line_no, raw, reason)`; the
  caller renders. It also stopped being *wrong* in the process — the refusal
  reason said "holds a target this command will not resolve on a guess" for lines
  refused over unresolved syntax, which the split surfaced.
- **Docs row**: says the run aborts, per your last point.

## 🔵 Low — deferred, deliberately

**Excluding `unreadable` entries from `--fix` candidacy (abort → skip).** Agreed
on the substance: it's inconsistent that the analysis half calls an unreadable
target "evidence of nothing" while the fix half classifies the same literal text
and lets it abort the run.

But that change *is* PR3. The [design comment on #1757](https://github.com/memtomem/memtomem/issues/1757#issuecomment-4977702900)
splits this into: relax the reader (here), amend ADR-0020 §1 (PR2), then replace
the blanket refusal with skip-and-report (PR3). Landing the abort→skip half here
would ship the new behaviour before the contract that describes it, and PR3 has
to rework this exact code path anyway — the per-line partition replaces
`_missing_target_entries` + `_unfixable_lines` wholesale, and the skipped lines
need JSON/status semantics that don't exist yet (a skipped-only run currently
computes `status: "clean"`). Doing it twice would be the expensive way.

You're right that it's not an ADR scope question. It's a sequencing one: the
abort is loud and fail-closed, so it can wait one PR for the surface that makes
it a skip.

memtomem:
## Correction — the CRLF gap was real, pinned in 24bbe85a

I dismissed that one above with "`TestSpliceRoundTrip` already covers CRLF … I'd
rather not assert the same property one layer down." Then I checked instead of
asserting, and the claim doesn't hold: that class is **handed** line numbers, it
doesn't derive them. A parse that mis-numbered a CRLF index would splice the
wrong line and every one of its cases would still pass.

Which is precisely what you flagged — "confirm `splitlines()` line numbers stay
aligned with markdown-it source maps" — and I answered it with a test that can't
see the alignment. The two counters do agree today (both treat `\r\n` as one
break); nothing said so, so now `test_crlf_line_numbers_stay_aligned` does.

Worth naming because it's the third time in this PR: verifying a property by hand
and calling it covered is what let the unread-pointer bug through in the first
place. The reflex to reach for it is apparently strongest right when I'm
explaining why a suggestion is unnecessary.

memtomem:
## Re-review — `24de9fa5` + `24bbe85a`

Follow-up to my earlier review. Verified locally at the PR head (`24bbe85a`, matches `origin`): 127 tests in `test_memory_doctor.py` pass, `ruff check` + `ruff format --check` clean on the changed files, and the previously-lost case was re-run through the same repro script that demonstrated the gap.

### Every finding from the previous round is addressed

- **🟡 Medium — mixed-line unresolved syntax unreported → fixed, correctly.** The root cause I pointed at (two different doubts folded into one `ambiguous_lines` set, making `ambiguous_lines - entry_lines` an inaccurate proxy) is what got fixed, not the symptom: `unresolved_syntax_lines` is now its own field, unreadable targets stay on the entries, and `ambiguous_lines` survives as a derived property (the union) for `--fix`'s line-level refusal. The analysis reports `unresolved_syntax_lines` directly, so `[B](b.md` beside a live `[A](a.md)` now surfaces. Repro confirms: the line that previously produced no finding is now reported. The report-level pin (`test_unread_pointer_is_reported_beside_a_good_link`) records honestly that the parser-level pin passed throughout and only a pin on the *report* could catch this — that's the right lesson to leave in the file.
- **🔵 Abort-the-whole-run design question → resolved by documenting the decision.** The docs row now states plainly that one ambiguous line aborts the whole `--fix` run and to repair such lines first. Keeping fail-closed and saying so is a legitimate resolution.
- **🔵 Triple parse → fixed.** `_run_fix` parses once and threads `parsed` into `_missing_target_entries` (as an optional parameter, so other callers are untouched). The fresh parse under the lock — the one that must not be shared — is retained.
- **🔵 Dependency cap → `>=3.0,<5`,** with a comment that makes the *reason* for capping this dep and not the others explicit (instance-attribute overrides of `normalizeLink`/`validateLink` are internals; a breaking major would surface at user runtime, not in CI, which resolves through `uv.lock`). `uv.lock` synced in the same commit.
- **🔵 Formatting leak → fixed.** `_unfixable_lines` returns `(line_no, raw, reason)` tuples; rendering lives in `_run_fix`. Bonus: splitting the ambiguity check in `_unfixable_reason` also gave the two doubts distinct refusal messages ("also holds link syntax that resolved to no link" vs "will not resolve on a guess"), which is a small UX improvement over the merged wording.
- **Coverage gaps → all three closed.** `TestListMarkers` pins both halves of the new surface (all four markers are read; ordered stays refused by the legacy shape under `--fix`, with an actual `--apply` run for each). The undefined shortcut reference is pinned as intended prose with the reasoning why flagging it would be worse. And `24bbe85a` pins CRLF alignment between markdown-it source maps and the `splitlines`-based splice unit — including the observation that `TestSpliceRoundTrip` couldn't have caught a divergence because it is handed line numbers rather than deriving them.

### Remaining nits — cosmetic, non-blocking

- **Lazy-continuation edge:** unresolved syntax on a wrapped line is recorded at the item's first line (`inline.map[0] + 1`), so for `- [A](a.md) hook\n  and [B](b.md` the finding quotes line 1, which doesn't visibly contain the `](` fragment. Verified locally that the line is flagged (and `multiline_lines` keeps `--fix` away regardless) — the report just points one line above the fragment. Not worth machinery to fix; a passing mention in the finding's items format someday, at most.
- **`ambiguous_lines` is now consumed only by tests** — production code uses the two split sets. Fine as a documented convenience for the line-level union, just noting it has become test-facing API.

### Verdict

All findings closed, each at the root cause rather than the symptom, and each closure pinned by a test that would have caught the original defect. LGTM.

PR #1764, opened 2026-07-15
docs: amend ADR-0020 — per-line eligibility for mm memory doctor --fix (#1757)

pandas-studio:
## Summary

Docs-only amendment to ADR-0020, recording the design pass from #1757 ("Design decision: line-splice contract preserved, span removal rejected") in the write contract **before** the code changes — the same contract-first pattern as the original ADR. Part 2 of the 4-PR split recorded on #1757 (part 1 was the parser fix, #1760; parts 3–4 follow).

#1760 widened the index parser to read every markdown link on a bullet line, so "one line = one entry" is no longer a parser-level given — it is a property `--fix` must verify per line before splicing. This amendment restates the contract in those terms:

- **§1** — eligibility becomes *strict grammar* (single-line bullet item, no unresolved link syntax, no unreadable target — the last two are exactly `ambiguous_index_line`) **+ all-links-dead**. The unit of deletion stays the whole line. A line failing either test is **skipped and reported for manual repair**, replacing the interim whole-run refusal (#1758) under which one non-conforming line blocked fixing the rest of the file.
- **§5** — two apply-time clauses: a fresh line qualifies only if *all* its links still classify `missing_target` (one resurrected target spares the whole line), and count-bounded candidate matching is defined in **physical line occurrences** (entries collapsed by line number), failing closed on a multiplicity mismatch.
- **Considered & rejected** — span-based entry removal, with the reasons of record: the span is not well-defined (stripping separators means parsing free-form agent prose, the reconstruction §2 forbids), nested hook links make span ownership unanswerable from syntax, and the blast radius covers §§1/2/5 including the "agent edited the line → line spared" property.
- **§3** — `ambiguous_index_line` recorded as reported-only; **§2** wording aligned to §1 eligibility; the testability consequence extended to the per-line-partition pins; header/context amended per the ADR-0008 convention.

Shipped `--fix` keeps the frozen fail-closed semantics (#1758, write scope pinned through #1760) until part 3 implements this contract. `docs/guides/reference/organization-maintenance.md` §5 deliberately keeps its current "refuses" wording until then (noted in Consequences).

One deviation from the #1757 design sketch, on purpose: the sketch stated the ambiguity test as a scan for markdown-significant characters outside the parsed links, guarding a link *regex* that truncated `[Live](notes_(v2).md)`. #1760 shipped a real CommonMark parse that subsumes that scan (balanced parentheses read cleanly and are pinned as non-ambiguous in `test_memory_doctor.py`), so the amendment records the parse-based test and explains the evolution inline — otherwise the ADR would mislead part 3 into reintroducing the obsolete regex failure.

## Verification

- Docs-only diff: one file, `docs/adr/0020-memory-index-write-contract.md`.
- Every behavioral claim checked against the decision of record (#1757 design comment) and shipped code (`memory_doctor_cmd.py`: `_unfixable_reason`, `_LEGACY_ENTRY_SHAPE_RE`, the whole-run refusal, `_apply_fix`'s count-bounded budget, `ambiguous_index_line`; parser tests for parenthesized destinations).
- `uv run pytest packages/memtomem/tests/test_memory_doctor.py packages/memtomem/tests/test_docs_guards.py -m "not ollama"` — 201 passed.
- `ruff check` + `ruff format --check` over src/tests/tools — clean.

Refs #1757 (does not close it — parts 3–4 remain).

### DIFF
--- a/docs/adr/0020-memory-index-write-contract.md
+++ b/docs/adr/0020-memory-index-write-contract.md
@@ -1,7 +1,9 @@
 # ADR-0020: Index-file write contract for `mm memory doctor --fix`
 
-**Status:** Accepted
-**Date:** 2026-06-01
+**Status:** Accepted (amended 2026-07-15: §1 line eligibility restated as
+strict-grammar + all-links-dead, per-line skip replaces the blanket refusal;
+§5 gains two apply-time clauses; span-based entry removal rejected — #1757)
+**Date:** 2026-06-01 (amended 2026-07-15)
 **Context:** `mm memory doctor` (shipped in #1170, documented in #1171,
 `docs/guides/reference/organization-maintenance.md` §5) is **read-only by design**: it reports the
 3-way drift between disk, the agent index/TOC file (`MEMORY.md` for a
@@ -23,20 +25,93 @@ canonical-artifact scope hierarchy) and reuses the round-trip-preservation
 discipline of ADR-0008 (lockfile) and ADR-0005 (force-reindex metadata
 contract): a writer must preserve everything it did not deliberately change.
 
+**Amendment (2026-07-15, #1757).** Real indexes were found packing several
+entries — and prose around them — onto one line, and the original parser saw
+only the first link per line (#1757: false orphans, a `broken_link` blind
+spot, and whole-line deletion of lines that also carried live entries). #1760
+widened `parse_memory_index` to read every markdown link on a bullet line, so
+"one line = one entry" is no longer a parser-level given — it is a property
+`--fix` must **verify per line** before splicing. This amendment restates §1's
+eligibility rule in those terms, adds two apply-time clauses to §5, and
+records the rejection of span-based entry removal. The splice mechanism (§2)
+is unchanged. The design pass behind it is recorded on #1757 ("Design
+decision: line-splice contract preserved, span removal rejected").
+
 ## Decision
 
 ### 1. Subtractive-only, provably-dead-only
 
-`--fix` may **only delete** pointer lines the doctor classifies as
-`broken_link` with link-class `missing_target` — a `- [title](target) — hook`
-line whose `target` resolves *inside* the memory root but points at a file
-that does not exist on disk. It never adds, reorders, re-wraps, reformats, or
-rewrites any line, and never touches the DB.
+`--fix` may **only delete** whole pointer lines whose every link is provably
+dead. It never adds, reorders, re-wraps, reformats, or rewrites any line, and
+never touches the DB. A link is *dead* when the doctor classifies it
+`broken_link` with link-class `missing_target`: its target resolves *inside*
+the memory root but points at a file that does not exist on disk.
+
+As accepted, eligibility was stated for the one-entry-per-line shape the
+harness itself specifies for `MEMORY.md`: a `- [title](target) — hook` line
+whose single target is dead. **Amended 2026-07-15 (#1757):** the parser now
+reads every markdown link on a bullet line, so eligibility is restated per
+line.
+
+`--fix` considers exactly the lines carrying **at least one `missing_target`
+link** — the drift it exists to close. Every other line is simply left alone
+and is not a `--fix` concern at all: a line of live pointers, prose, or a
+heading is not "skipped", it was never a candidate. Among candidate lines,
+the unit of deletion stays the **whole line**, and a candidate qualifies for
+deletion **iff both** hold:
+
+- **Strict grammar — the parse must account for the whole line.** The line is
+  a single-line bullet (`-`/`*`) pointer entry as the CommonMark parser
+  (#1760) reads it: one or more markdown links plus inert prose. Three shapes
+  fail that test, and all three fail **closed** (the line is never eligible
+  for deletion):
+  - a list item that continues past its first line — deleting the line would
+    strand the continuation as loose prose;
+  - link syntax on the line that resolved to no link (an unclosed `[B](b.md`)
+    — the line meant a pointer the grammar could not read;
+  - a target the doctor will not resolve on a guess — a destination that is
+    not a plain relative path (it carries a `?` query, a `%` escape, a `:`
+    scheme, or whitespace) may name a file other than its literal text does,
+    so its pointer is never *provably* dead.
+
+  The last two are exactly the lines the doctor already reports as
+  `ambiguous_index_line` at `warn`: something on them could not be read, and
+  a splice must not act on a doubt. (#1757's design sketch stated this
+  ambiguity test as a scan for markdown-significant characters left outside
+  the parsed links, guarding a link *regex* that truncated
+  `[Live](notes_(v2).md)` at the inner paren. #1760 shipped a real CommonMark
+  parse instead, which subsumes that scan — balanced parentheses in a
+  destination or in hook prose read cleanly and are *not* ambiguous. The
+  contract is the parse-based test above, not the character scan.)
+- **All links dead.** *Every* link on the line classifies `missing_target`.
+  One live (or ambiguous, or out-of-root) link spares the whole line: splicing
+  it would delete a pointer memtomem cannot prove dead, and carving the dead
+  entry out of the line is span surgery this ADR rejects (see Considered &
+  rejected).
+
+A **candidate** that fails either test is **skipped and reported for manual
+repair** — a dead pointer memtomem found but will not remove itself — while
+qualifying candidates in the same file are still fixed. This is a per-line
+partition: it replaces the interim whole-run refusal (#1758, frozen through
+#1760) under which one non-conforming line blocked fixing the rest of the
+file.
+
+Skips are part of the report, not a silent omission, so the output must make
+three outcomes distinguishable in both the human and `--json` surfaces:
+nothing to do (no candidates), every candidate handled, and some candidate
+left behind. A run that skips a candidate must **not** present as clean —
+`--json`'s status has to separate "no dead pointers" from "dead pointers this
+tool refuses to touch", and a partially-fixed file must name the lines it
+left. The same applies to a candidate dropped at apply time by §5's
+re-validation (a resurrected target, a multiplicity mismatch): it is reported,
+not silently absent. The concrete payload shape is the implementing PR's to
+design; what this contract fixes is that skipped candidates are always
+visible and never counted as success.
 
 Rationale: the agent owns curation; memtomem may only remove references it can
-*prove* are dead. Deletion of a provably-dead pointer is the one curation move
-that cannot conflict with the agent's intent — the target file is gone, so no
-correct version of the TOC keeps the line.
+*prove* are dead. Deletion of a line holding *only* provably-dead pointers is
+the one curation move that cannot conflict with the agent's intent — every
+target on it is gone, so no correct version of the TOC keeps the line.
 
 ### 2. Byte-exact preservation of everything else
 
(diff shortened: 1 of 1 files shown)


tsdata (approved):

pandas-studio (commented):
## Review — docs: amend ADR-0020 (per-line eligibility for `mm memory doctor --fix`)

### Overview

Docs-only amendment recording the #1757 design pass in ADR-0020 before the part-3 code lands, same contract-first pattern as the original ADR. §1 eligibility becomes strict-grammar + all-links-dead with per-line skip replacing the interim whole-run refusal (#1758); §5 gains the all-links-still-dead and multiplicity-guard apply-time clauses; span-based entry removal is recorded as rejected with the reasons from the #1757 design comment.

### Verification (every behavioral claim checked against source)

- **§1 strict-grammar failures match the shipped predicates.** Multiline items (`ParsedIndex.multiline_lines`), unresolved link syntax (`unresolved_syntax_lines`), and unreadable targets (`IndexEntry.unreadable` via `_target_is_unreadable`) are exactly the guards in `_unfixable_reason` (`memory_doctor_cmd.py:1120-1155`). The claim that "the last two are exactly `ambiguous_index_line` at `warn`" is correct — `ambiguous_items` is built from precisely those two sets and reported at `severity="warn"` (`memory_doctor_cmd.py:831-877`).
- **The interim whole-run refusal is real and correctly framed.** `_run_fix` raises `click.UsageError` on both dry-run and `--apply` when any candidate line is out of scope (`memory_doctor_cmd.py:1316-1325`), and `_LEGACY_ENTRY_SHAPE_RE` pins the pre-#1757 write scope with a comment deferring the relaxation to this amendment — the Consequences "frozen until part 3" statement matches shipped code.
- **The design-sketch deviation is verified.** `[Live](notes_(v2).md)` is pinned as a live, non-ambiguous entry in `test_memory_doctor.py:150` and `:779-804`, so recording the parse-based test instead of the character-scan is right — the ADR would otherwise steer part 3 back into the obsolete regex failure.
- **§5 multiplicity guard**: the amendment's fail-closed-on-count-mismatch is a deliberate tightening over shipped `_apply_fix` (which removes up-to-budget copies and whose docstring claims which-copy-survives is irrelevant). The amendment's rationale is actually a correction: removing one of two byte-identical lines that sit in *different sections* does not yield byte-identical results, so "which copy" is not irrelevant. Part 3 should also fix that docstring claim (`memory_doctor_cmd.py:1219-1225`) when it implements the guard.
- Guide `organization-maintenance.md` §5 keeps the "refuses" wording as the Consequences note says; the #1757 comment "Design decision: line-splice contract preserved, span removal rejected" exists; header follows the ADR-0008 amendment convention; `pytest test_memory_doctor.py test_docs_guards.py` re-run locally — 201 passed.

### Suggestions (minor, both doc-accuracy)

1. **§1 unreadable-target enumeration omits whitespace.** The code's predicate is `_PLAIN_RELATIVE_TARGET_RE = r"^[^\s?%:]+$"` and the doctor's own warn summary says "query, percent-escape, scheme **or space**", but the amendment's parenthetical lists only "`?` query, `%` escape, `:` scheme". Since the same paragraph asserts these failures are *exactly* `ambiguous_index_line`, and this text is the contract part 3 implements, add whitespace to the enumeration (reachable via a `<...>` destination, e.g. `[a](<my file.md>)`).
2. **Consequences now disagrees with itself on whether `--fix` shipped.** The untouched first bullet still reads "When the Tier 2 `--fix` ships, … §5 gains its remediation entry" (future tense), while the new amendment bullet correctly says "shipped `--fix` keeps the frozen fail-closed semantics". A parenthetical on the first bullet ("shipped in #1758; §5 entry landed") would remove the contradiction without rewriting accepted text.

Neither blocks merge — both are one-line touch-ups that could also ride along with part 3.

### Risks

None beyond docs drift: no code, no behavior change, and the amendment explicitly scopes shipped behavior as unchanged until part 3. The main hazard for a contract-first ADR is the contract mis-describing the code it builds on, and the checks above found only the whitespace-enumeration gap (suggestion 1).

**Verdict: LGTM** — accurate against source and the decision of record, follows the ADR-0008 amendment convention, and the two minor suggestions are non-blocking.

pandas-studio:
## Review follow-up — 3 of 4 findings applied

Codex reviewed the amendment (NEEDS-FIX: 1 major, 1 minor) alongside the earlier review's two minor points. All four were verified against source; three are now applied in the amended commit.

**Applied**

1. **§1 candidate domain (Codex, major).** "A line that fails either test is skipped and reported" read literally made *every* live-only line a skip — they all fail "all links dead". §1 now names the candidate set explicitly: `--fix` considers exactly the lines carrying at least one `missing_target`, and a non-candidate "is not skipped, it was never a candidate". Eligibility (strict grammar + all-links-dead) is stated over candidates only.
2. **Skip/status visibility (Codex, major).** Shipped `_emit_fix_json` derives status from removal count alone (`clean` / `fixed` / `would-fix`), so under the per-line partition a file with skipped candidates and zero removals would report `clean`. §1 now requires the three outcomes (no candidates / all candidates handled / some candidate left behind) to be distinguishable in both surfaces, requires a skipping run not to present as clean, and extends the same rule to candidates dropped at apply time by §5's re-validation. The payload shape stays part 3's to design — the contract only fixes that skips are visible and never counted as success.
3. **Unreadable-target enumeration (earlier review, minor).** The parenthetical listed `?` / `%` / `:` but the shipped predicate is `_PLAIN_RELATIVE_TARGET_RE = r"^[^\s?%:]+$"` and the doctor's own warn text says "query, percent-escape, scheme **or space**". Now stated as "not a plain relative path (it carries a `?` query, a `%` escape, a `:` scheme, or whitespace)".

**Not applied (deliberate)**

4. **Context/Consequences tense (Codex minor + earlier review's second point).** The untouched Context ("read-only by design", "does not write index files today", "pins the contract before any code lands") and the first Consequences bullet ("When the Tier 2 `--fix` ships") predate the shipped writer and now read as stale. Correct, but rewriting accepted, non-amended prose widens this PR beyond the recorded decision; an ADR's accepted text is also a record of what was true when it was decided. Left for a follow-up that can restate them historically in one pass.

**Also noted for part 3:** `_apply_fix`'s docstring claims which byte-identical copy survives is irrelevant (`memory_doctor_cmd.py:1219-1225`). The §5 multiplicity guard is a deliberate tightening precisely because that claim is wrong when identical lines sit in different sections — the docstring should be corrected when the guard lands.

**Verification:** `pytest test_memory_doctor.py test_docs_guards.py -m "not ollama"` — 201 passed; `ruff check` clean; diff is still one file. CI was green before the amend, including the `test (ubuntu-latest)` rerun — its first failure was `test_lock_budget_bounds_whole_call_not_per_target` (a timing assertion in `test_context_settings.py`, unrelated to docs) which passed on rerun and passes on main.

pandas-studio (commented):
## Follow-up review — head `18e5ada2` (delta vs. previously reviewed `5205eafe`)

### What changed

- **Suggestion 1 (whitespace in the unreadable-target enumeration) — resolved.** The bullet now reads "a destination that is not a plain relative path (it carries a `?` query, a `%` escape, a `:` scheme, or whitespace)", which matches `_PLAIN_RELATIVE_TARGET_RE = r"^[^\s?%:]+$"` exactly — both the plain-relative-path framing and the four-character class. The "exactly `ambiguous_index_line`" claim now holds against source.
- **New candidate framing in §1.** "`--fix` considers exactly the lines carrying at least one `missing_target` link … every other line was never a candidate" — a good clarification that scopes "skipped" to something `--fix` actually found and refused, rather than every line it didn't touch.
- **New observability paragraph in §1.** Three distinguishable outcomes (nothing to do / all handled / candidate left behind), skips never presenting as clean, apply-time drops (§5 re-validation) reported rather than silently absent, payload shape left to the implementing PR. This closes a real gap in the previous revision — without it, part 3 could have satisfied the letter of "skipped and reported" with a human-only message while `--json` consumers saw a clean run. No source-verifiable claims here (it constrains part 3, and the ADR says so explicitly).

### Still open (non-blocking, re-raising from the previous review)

- **Consequences bullet 1 still reads future-tense** ("When the Tier 2 `--fix` ships, … §5 gains its remediation entry") while the amendment bullet in the same section says "shipped `--fix`". Fine to leave for part 3 if the point-in-time reading is intentional, but the section currently disagrees with itself.

### One definitional edge for part 3 to settle (observation, no change requested)

The new candidate definition ("carries at least one `missing_target` link") and the shipped candidate collector can diverge on an **unreadable-only** line. §1 defines *dead* via the doctor's classification, and the doctor never classifies an unreadable entry (`memory_doctor_cmd.py:835-837` — it goes to `ambiguous_index_line`, not `broken_link`), so a line whose only link is `[a](<my file.md>)` with the file missing is *not a candidate* under the ADR. But `_missing_target_entries` (`:1092-1109`) classifies every entry without checking `e.unreadable`, so the same line *is* selected as a candidate today and would surface in `--fix`'s skip report. Either resolution is safe (the line is reported via `ambiguous_index_line` in the same run regardless); the implementing PR should just pick one deliberately and pin it, since the two readings put the line on different sides of the new three-outcome `--json` status.

**Verdict: LGTM** — the delta strengthens the contract (observability paragraph) and fixes the one accuracy gap; the remaining items are a cosmetic tense mismatch and a definitional edge that part 3 owns.

pandas-studio:
## Second review round — SHIP, no findings

Codex re-reviewed head `18e5ada2` after the three fixes landed: **SHIP**, zero blocker/major/minor/nit findings. It independently verified the candidate semantics against source (report path excluding `unreadable` at `memory_doctor_cmd.py:834-838`; `_missing_target_entries` at `:1092-1109`; the safety gate at `:1144-1154`) and confirmed the observability paragraph is precise enough to implement against while correctly leaving field names to part 3.

### The unreadable-only definitional edge — resolved as written, with a part-3 obligation

Both review rounds landed on the same edge, and it is worth recording since two readings put a line on different sides of the new three-outcome status:

- **The ADR settles it already.** §1 defines *dead* via the doctor's classification, and the doctor never classifies an unreadable entry (`:834-837` continues to `ambiguous_index_line` before `classify_link` runs). So a line whose only link is unreadable — `[a](<my file.md>)` with the file missing — is **not a candidate**, and never enters the skip report. It is still reported in the same run, as `ambiguous_index_line` at `warn`.
- **If a readable dead pointer shares that line**, the line *is* a candidate, fails strict grammar (an unreadable target on the line), and is reported as a skipped candidate. Both paths are consistent and safe.
- **Part 3 obligation:** shipped `_missing_target_entries` (`:1104-1109`) classifies every entry *without* checking `e.unreadable`, so it selects unreadable-only lines as candidates — the opposite of what this contract now says. Today that divergence is invisible (`_unfixable_reason` refuses those lines anyway and the whole run refuses), but under the per-line partition it decides whether such a line shows up as a skipped candidate or only as `ambiguous_index_line`. **Part 3 must not reuse that helper unchanged for candidate discovery.**

### Status

CI is green on `18e5ada2` (the earlier `test (ubuntu-latest)` red was `test_lock_budget_bounds_whole_call_not_per_target`, a timing assertion in `test_context_settings.py` unrelated to this docs change; it passed on rerun and passes on main). Approved by @tsdata. Ready to merge on your go-ahead.

Deferred by decision, not oversight: the pre-ship tense in the untouched Context and the first Consequences bullet — rewriting accepted, non-amended prose widens this PR beyond the recorded decision.

pandas-studio (commented):
## Re-review — head unchanged

Re-reviewed at head `18e5ada2` — identical to the revision covered by my 2026-07-15 12:53 UTC review, so that review's findings stand in full: **LGTM**, with the two non-blocking items still open (the future-tense Consequences bullet 1, and the unreadable-only-line candidate-definition edge for the part-3 PR to settle).

Freshly checked for this pass:

- All CI checks are green at this head (required and advisory), and the PR is `MERGEABLE` / `CLEAN` against `main`.
- Note for the merger: the standing approval (12:22 UTC) predates the 12:39 UTC force-push that produced the current head — it was granted against the previous revision, which lacked the observability paragraph and the candidate framing. Worth a re-glance from the approver if the process cares about approving the bytes being merged.

PR #1765, opened 2026-07-15
feat: per-line eligibility for mm memory doctor --fix (#1757)

pandas-studio:
## Summary

Implements ADR-0020's per-line eligibility for `mm memory doctor --fix`. The
amendment (#1764) landed contract-first and left `--fix` on the frozen
fail-closed semantics of #1758; this is the code that catches up, plus the docs
the ADR's own consequence says land with it.

**Eligibility (§1).** A *candidate* is a line carrying at least one
`missing_target` link. It is deleted whole iff the parse accounts for the whole
line (a single-line `-`/`*` bullet) **and** every link on it is provably dead.
A candidate failing either test is **skipped, named with its reason, and left
for a human** — while the eligible lines in the same file are still fixed. One
non-conforming line no longer blocks the run.

What that changes in practice:

| Line | Before | Now |
|---|---|---|
| `- NS: [Dead](gone.md)` (prose prefix) | whole run refused | removed |
| `- [D1](gone1.md) · [D2](gone2.md)` (all dead) | whole run refused | removed |
| `- [Dead](gone.md) · [Live](exists.md)` | whole run refused | line skipped + reported, rest of file fixed |
| `1. [Dead](gone.md)` (ordered marker) | whole run refused | line skipped + reported |
| `- [Q](live.md?view=1)` (unreadable only) | whole run refused | not a candidate at all |

**Unreadable targets are never candidates.** Their literal miss classifies
`missing_target`, but it is not proof of death — the query may belong to the
filename. Tier 1 already excludes them from `broken_link`; `--fix` now excludes
them at the candidate stage, so a line holding only unreadable links is not a
`--fix` concern rather than a pointer it "declined to remove".

**Multiplicity guard (§5).** Both sides now count **physical line occurrences**
over the same population, and a mismatch fails closed on every copy instead of
picking one. The old docstring's claim that the surviving copy is "irrelevant,
since the spliced result is byte-identical either way" was wrong: what is at
stake is *which copy the agent meant to keep*, not the bytes.

**The `--apply` report describes the file it leaves on disk.** It is built from
the locked fresh read: every dead pointer still in the index is named —
including one the agent wrote inside the race window (spared, since no
analysis-time count bounds it, but not hidden) — and a candidate the agent
deleted meanwhile is not, since it is not there to repair and its analysis-time
line number now belongs to another line. Reporting from the analysis snapshot
instead could call one line both removed *and* skipped. So `clean` means the
index holds no dead pointers, not that the run lost track of one — which is why
the apply path takes the lock even when analysis found nothing to do.

**Surfaces (§1: skips are never counted as success).** `--json` gains `partial`
/ `would-partial` alongside `clean` / `would-fix` / `fixed`, plus per-file
`skipped` records (`{line, text, reason}`) and `summary.skipped`. A run that
leaves a dead pointer exits `1`, in dry-run and `--apply` alike.

## Docs

- `docs/guides/reference/organization-maintenance.md` §5 — the "one entry per
  line, or it refuses" rule and the link-level counting are rewritten; the
  `ambiguous_index_line` row no longer says one such line aborts the whole run;
  the exit code, `--json` statuses, and the report rule are documented.
- `docs/adr/0020-memory-index-write-contract.md` — §5 records the reporting rule
  the amendment left to this PR; the pre-ship tense in Context/Consequences is
  settled.

## Verification

- `uv run pytest -m "not ollama"` — 8657 passed. `test_memory_doctor.py` goes
  159 (was 147).
- `ruff check` + `ruff format --check` over `src`/`tests`/`tools` — clean.
  `mypy` — no new errors (18, same as `main`).
- Every new pin was mutation-tested: reverting the behavior it guards turns it
  red, and the tests that encoded the old contract were inverted rather than
  deleted.
- Driven end-to-end against a real index in an isolated `HOME` (dry-run,
  `--apply`, re-run): only the targeted lines disappear, all 10 surviving lines
  stay byte-identical, the fenced example and unreadable line survive, and a
  second `--apply` is a no-op.

Contract-adjacent decisions in this PR (exit code, `--json` shape, the report
rule) were approved by the maintainer before landing; the diff was reviewed
adversarially until it came back clean.

## Known gap (pre-existing, not this PR)

An index `--fix` cannot read or decode is silently skipped, so it can report
`clean` about a file it never opened. That predates this change and needs a new
error result kind + `--json` status; filing separately rather than widening this
PR.

Refs #1757.


tsdata (approved):

pandas-studio:
## Review — per-line eligibility for `mm memory doctor --fix`

### Overview

Replaces #1758's whole-run fail-closed refusal with ADR-0020 §1's per-line partition: a candidate line (≥1 `missing_target` link) is deleted whole iff it is a single-line `-`/`*` bullet the parse fully accounts for **and** every link on it is provably dead; otherwise it is skipped with a named reason while the rest of the file is still fixed. Unreadable targets are excluded at the candidate stage. §5's multiplicity guard now counts physical line occurrences on both sides and fails closed on every copy of a mismatched raw. The `--apply` report is rebuilt from the locked fresh read, `--json` gains `partial`/`would-partial` + per-file `skipped`, and any skip exits `1`. ADR-0020 and the §5 reference doc are updated in the same PR, as the amendment required.

### Correctness

- **The `_apply_fix` four-way branch is sound.** I walked the cases: untouched file (single and duplicate raws), count drift in both directions, eligibility swap between byte-identical copies, eligibility growth with stable counts, agent add/edit/delete in the race window. Each lands in the intended bucket, and each has a test. The two populations the guard compares (`_pointer_lines` on both sides) are genuinely the same — the bug class the old link-level counting invited is closed.
- **Reporting from the fresh partition (not the snapshot) is the right call.** The removed/skipped-disjointness and stale-coordinate tests pin the two failure modes that motivated it. The apply-on-clean lock-and-re-read (so `clean` is a checked claim, not a T1 memory) is a nice touch and is also pinned.
- **No double reporting**: the loop only iterates fresh-*eligible* raws, which are disjoint from the partition's skips, so each physical line lands in exactly one report bucket.
- One deliberately conservative edge worth having said out loud (the docstring does): when the agent *deletes* one of N eligible identical copies, the count mismatch fails closed on the survivors even though removing them would match the promised net. Safe, re-run clears it — fine, just noting it's intentional asymmetry.

### Minor

- **`_apply_fix` docstring interleaving** (`memory_doctor_cmd.py:1294`): the dedented "**The report describes the file this call leaves on disk.**" block is spliced *between* step 2's *Multiplicity* and *Eligibility* sub-paragraphs, so *Eligibility* (still indented as part of step 2) now reads as orphaned under an unrelated top-level block. Looks like a formatting accident — moving the report block after the *Eligibility* paragraph (or after step 4) restores the numbered flow.
- **Unreachable defensive branch** (`memory_doctor_cmd.py`, `_line_skip_reason`): `"no longer holds a pointer"` can't fire — both callers derive `line_no` from candidates of the same `parsed`, so `line_entries` is never empty. Harmless as defense, but no test can reach it; a comment saying it's defensive (or an assert) would keep a future reader from hunting for the caller that triggers it.
- **`click.secho(summary, fg="green") if applied else click.echo(summary)`** (`_emit_fix_human`): a conditional expression as a statement for side effects — a plain `if/else` matches the file's idiom.
- **`raise click.exceptions.Exit(1)`**: works (and is what `ctx.exit` raises), but `sys.exit(1)` / `ctx.exit(1)` is the more conventional spelling. Non-blocking.
- **Perf, negligible**: `_line_skip_reason` re-runs `classify_link` (disk stats) on links already classified by `_missing_target_entries`, and its per-line entry scan is O(lines × entries). Both are bounded by the 200-line index budget — fine as is.

### Compatibility / risk

- **`--json` consumers**: `summary` gains `skipped`, file objects gain `skipped`, and `status` gains two values — a strict consumer switching on `status` may fall through on `partial`. Documented in §5 and pinned by tests; PR body says the surface was maintainer-approved. Just make sure any in-repo/plugin consumers of `--fix --json` are updated (I found none in this repo).
- **Exit-code change**: the old refusal was a `UsageError` (exit 2); a skip is now exit 1 in dry-run and apply. Scripts keying on `!= 0` are unaffected; scripts keying on `== 2` would be — documented.
- **Security**: subtractive-only scope unchanged; `outside_root` siblings spare the line (tested against a `../../etc/passwd` rider), and unreadable targets can no longer be deleted on a literal-text guess. No new surface.

### Tests & docs

- Coverage is excellent: 12 new tests including the race-window monkeypatch harness (`edit_then_apply`), duplicate/mismatch matrices, JSON shape + exit codes, and doc-drift pins (`test_reference_documents_per_line_skip_not_whole_run_refusal` asserting the *old* wording is gone is exactly the right pin). Old-contract tests were inverted rather than deleted, per the PR body's mutation-testing claim.
- ADR-0020 §5 reporting rule and the reference doc's §5/`ambiguous_index_line` rows match the shipped behavior I read in the code — no drift found.

### Verdict

Approve-quality. The findings above are all minor (docstring formatting, one unreachable branch, two style nits); none blocks merge.

PR #1770, opened 2026-07-15
fix: an unreadable index is an error, not a crash (Tier 1) or clean (--fix) (#1769)

pandas-studio:
## Summary

An index file `mm memory doctor` cannot **decode** was handled twice, and both ways were wrong (#1769):

- **Tier 1 crashed.** `_analyze_index_file` guarded its read with `except FileNotFoundError` / `except OSError`, but `UnicodeDecodeError` subclasses `ValueError` — it escaped as an unhandled traceback, corrupting `--json` output too.
- **Tier 2 said `clean` about a file it never opened.** `_run_fix` caught the decode error and `continue`d, dropping the file from the results — `--fix --apply --json` reported `{"status": "clean", "files": []}` and exited 0 while the index still held a live `missing_target` pointer. ADR-0020 §5 defines `clean` as "the index holds no dead pointers"; an unread file supports no such claim.

Both predate #1765, which surfaced them and deliberately left this gap open.

## Changes

**Tier 1** — the read guard becomes `except (OSError, UnicodeDecodeError)`, routing decode failures into the existing `index_missing` warn ("could not read …"), the same treatment as a missing file. Bare `ValueError` stays uncaught so parse bugs cannot masquerade as an unreadable file. The checks table already documented `index_missing` as "absent or unreadable" — no vocabulary change.

**Tier 2** — an unreadable index becomes a **per-file `error`**:

- `FixFileResult` gains `error: str | None` (always emitted in `--json`, `null` when the file was read).
- `--json` `status` gains a sixth value, **`error`**, which outranks `partial` / `fixed` / `clean` whenever any file could not be read — a skip is a complete account of what remains, an unread file is no account at all. One word covers dry-run and `--apply` alike (`applied` already carries the run mode).
- `summary` gains an `errors` count; the run exits `1`.
- The remaining dirs are still read, reported, and fixed — one bad index does not suppress the others.
- The locked fresh read inside `_apply_fix` gets the same conversion (an agent rewriting the file to non-UTF-8 between analysis and the lock), via a narrow `_IndexUnreadable`. **Only the read converts**: lock, `stat`, and write failures keep propagating, so an exception after the atomic replace can never be misreported as `removed=[]` — the §5 audit guarantee (every removed line is reported) survives. Two tests pin this propagation boundary.
- Error presence is checked with `is not None`, never truthiness — `str(OSError())` is empty, and a blank message under a truthiness check would resurrect the silent drop this PR fixes (messages fall back to the exception class name).

**Docs** — `organization-maintenance.md` §5 documents the `error` status, per-file `error` field, `errors` summary count, and the exit rule; ADR-0020 gets an append-only 2026-07-16 amendment recording the decision and the read-only conversion boundary.

## Tests

- Tier 1: undecodable index → `index_missing` warn + valid `--json` payload, exit 0 (regression for the traceback); OSError-arm pin (previously untested).
- Tier 2: undecodable index → `status: "error"`, exit 1, bytes untouched, in dry-run and `--apply` (regression for the false `clean`); mixed bad+healthy dirs → `error` precedence while the healthy index is still fixed on disk; apply-time corruption between T1 and the lock; `PermissionError` parity; blank-message `OSError()` regression; lock/write failures propagate un-converted (mutation-validated: widening the catch turns both pins red).
- Existing status/summary pins extended (`errors` key, `error` in the documented status list).

`uv run pytest -m "not ollama"`: 8666 passed. Ruff clean on `src`/`tests`/`tools`. Manually reproduced the issue's exact console transcript in an isolated `HOME` before/after.

Closes #1769


tsdata (approved):

pandas-studio:
## Review

### Overview

Fixes the two wrong handlings of an undecodable index file from #1769: Tier 1 crashed (`UnicodeDecodeError` is a `ValueError`, so it escaped the `OSError` guard), and `--fix` silently dropped the file and reported `clean`/exit 0. Tier 1 now routes decode failures into the existing `index_missing` warn; `--fix` reports a per-file `error`, adds an `error` JSON status that outranks `partial`/`fixed`/`clean`, counts `errors` in `summary`, and exits 1 — while still reading, reporting, and fixing the remaining dirs.

### Correctness — verified

- **Exception taxonomy is right.** `except (OSError, UnicodeDecodeError)` catches exactly the two read-failure shapes; bare `ValueError` stays uncaught so parse bugs can't masquerade as unreadable files. Tier 1 reads with an explicit `encoding="utf-8"`, so the new behavior is deterministic across platforms (no locale-dependent decode on Windows).
- **The conversion boundary is narrow and correctly placed.** Inside `_apply_fix`, only the locked fresh read raises `_IndexUnreadable`; lock, `stat`, and write failures propagate, so a post-replace exception can never be misreported as `removed=[]`. The two propagation-boundary tests pin this from both sides.
- **The monkeypatch targets are valid.** `_apply_fix` imports `_file_lock`/`atomic_write_text` function-locally from `memtomem.context._atomic` (memory_doctor_cmd.py:1361), so patching the source module attributes genuinely intercepts the calls — the propagation tests test what they claim to.
- **`is not None` presence checks are consistent** everywhere the new field is read (exit code, human emit, JSON `_reportable`, status), and `_read_error_message` guarantees a non-empty message — the blank-`OSError()` regression test locks this in.
- **Status precedence and exit code** match the docs and ADR amendment: `error` wins over everything, exit 1 on any error or skip, and the mixed bad+healthy test proves the healthy index is still fixed on disk.

Ran the 16 new/extended tests locally — all pass. CI required gates are green (OS test matrix still in progress at review time).

### Suggestions (minor)

- **Tier 1 doesn't reuse `_read_error_message`.** `_analyze_index_file` formats `f"could not read {index_file_name}: {exc}"`, so a blank-message `OSError()` yields a summary ending in `": "`. Unlike the `--fix` path this is cosmetic — the finding's presence never depends on message truthiness — but the helper is in the same module and would make the two paths symmetric. Fine as a follow-up or never.
- **`summary["files"]` semantics widened** from "files with removed/skipped" to "files with anything to report (incl. errors)". This is the right call and it's pinned by tests, just noting it as a (documented) shape change for any script consuming the field.

### Docs / conventions

- ADR-0020 amendment is append-only and records both the decision and the read-only conversion-boundary rationale; `organization-maintenance.md` §5 documents the sixth status, the per-file `error` field, the `errors` count, and the exit rule — and the docs-as-tests pin (`test_reference_documents_the_fix_status_values`) now enforces `error` stays documented.
- The pre-existing checks-table wording ("absent or unreadable") already covered the Tier-1 behavior — verified, no vocabulary drift.
- Tests avoid `chmod`-style POSIX-only permission setups in favor of monkeypatched `PermissionError`, so the suite stays Windows-safe.

### Verdict

**LGTM.** The fix is tightly scoped, the failure-accounting semantics are principled (an unread file supports no claim), the conversion boundary preserves the §5 audit guarantee, and the test coverage — including mutation-validated propagation pins and the blank-message regression — is thorough. The one suggestion above is optional.

PR #1772, opened 2026-07-16
fix: name the exception class when a Tier 1 index read fails blank (#1771)

pandas-studio:
## Summary

`_analyze_index_file` interpolated the exception straight into its `index_missing` summary, and `str(OSError())` is the empty string — so a blank-message I/O failure rendered as a line that trails off where the reason belongs:

```
! could not read MEMORY.md:
```

`_read_error_message` already solves exactly this for Tier 2's `--fix` path (#1770): it falls back to the exception class name. This routes Tier 1's read failure through the same helper rather than keeping two spellings of the same rendering:

```
! could not read MEMORY.md: OSError
```

## Changes

- `_analyze_index_file`'s `(OSError, UnicodeDecodeError)` arm calls `_read_error_message(exc)`. The `FileNotFoundError` arm above it has its own message and is untouched.
- The helper **moves** out of the Tier 2 section up to the shared report-model section, since both tiers now use it.
- Its docstring drops the Tier-2-only framing. The two callers differ in what's at stake, and the docstring now says so: for `--fix` an empty message would vanish under a truthiness check and resurrect the false `clean` of #1769; in Tier 1 nothing but readability is at stake — the finding is appended unconditionally, and its severity, count and exit-code contribution never consult the message. **Tier 2's presence check stays `error is not None` regardless** — the helper is not what makes it safe, and this PR doesn't make it depend on one.

No behavior change beyond the rendered string; no status, severity, exit-code or schema change.

## Tests

`TestCli::test_blank_oserror_message_names_the_exception_class` — monkeypatches `Path.read_text` to raise a bare `OSError()` and asserts the summary is exactly `"could not read MEMORY.md: OSError"`, not `"could not read MEMORY.md: "`. Mutation-validated: reverting the arm to interpolate `exc` directly turns the pin red.

The sibling `PermissionError` pin (`test_oserror_index_read_is_index_missing_warn`) and Tier 2's `test_fix_blank_oserror_message_is_still_an_error` cover the non-blank and Tier 2 sides unchanged.

`uv run pytest -m "not ollama"`: 8669 passed. Ruff clean on `src`/`tests`/`tools`. Verified against a real isolated `HOME` that the rendered finding reads `could not read MEMORY.md: OSError`.

Raised in review of #1770 as an explicitly optional follow-up.

Closes #1771

### DIFF
--- a/packages/memtomem/src/memtomem/cli/memory_doctor_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/memory_doctor_cmd.py
@@ -534,6 +534,19 @@ class DirReport:
         }
 
 
+def _read_error_message(exc: Exception) -> str:
+    """Render a read failure for a report — never empty.
+
+    ``str(OSError())`` is ``""``, so interpolating the exception directly can
+    leave a message that trails off (``"could not read MEMORY.md: "``) or, in
+    Tier 2, one that vanishes under a truthiness check and turns a real failure
+    back into "clean" (#1769). Both read paths route through here: the message
+    falls back to the exception class name, and Tier 2's presence check stays
+    ``error is not None`` regardless.
+    """
+    return str(exc) or type(exc).__name__
+
+
 # ── Read-only config + engine plumbing ──────────────────────────────
 
 
@@ -812,7 +825,7 @@ def _analyze_index_file(
             Finding(
                 check="index_missing",
                 severity="warn",
-                summary=f"could not read {index_file_name}: {exc}",
+                summary=f"could not read {index_file_name}: {_read_error_message(exc)}",
             )
         )
         return
@@ -1077,17 +1090,6 @@ def _emit_json(reports: list[DirReport]) -> None:
 #      each candidate against fresh content so concurrent agent edits survive.
 
 
-def _read_error_message(exc: Exception) -> str:
-    """Render a read failure for the report — never empty.
-
-    ``str(OSError())`` is ``""``; an empty message would vanish under any
-    truthiness check downstream and turn a real failure back into "clean"
-    (#1769) — so presence is ``error is not None`` everywhere, and the message
-    itself falls back to the exception class name.
-    """
-    return str(exc) or type(exc).__name__
-
-
 class _IndexUnreadable(Exception):
     """The locked fresh read inside :func:`_apply_fix` could not decode the index.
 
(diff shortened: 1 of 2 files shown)


tsdata (approved):

pandas-studio:
## Review

**Verdict: LGTM — ready to merge.** Every claim in the PR body was re-verified empirically; no blockers, and nothing rising even to nit level.

### Overview

A +43/−13 change: Tier 1's `_analyze_index_file` `(OSError, UnicodeDecodeError)` arm stops interpolating the exception directly and routes through `_read_error_message`, the helper Tier 2's `--fix` path already uses (#1770). The helper moves from the Tier 2 section up to the shared report-model section, with its docstring rewritten for both callers. No behavior change beyond the rendered string.

### Verified (measured, not just read)

- **Correctness / remaining bug shape**: grepped every exception-interpolation site in `memory_doctor_cmd.py` — no instance of the blank-`str(OSError())` shape remains. The one surviving `error=str(exc)` (line 1513) takes an `_IndexUnreadable`, which is constructed at exactly one site (line 1377) from `_read_error_message(exc)` output, so it is always non-empty. Not a bug.
- **Pin quality**: the new test asserts **exact equality** (`== "could not read MEMORY.md: OSError"`), not a substring — it pins precisely the trailing-off regression. The `resolve()`-scoped `read_text` patch matches the sibling test's pattern.
- **Mutation claim reproduced**: reverting the arm to `{exc}` turns the new pin red (`1 failed`); restoring goes green. The PR body's mutation-validation claim holds.
- **All three pins pass** together: the new blank-OSError pin, the sibling `PermissionError` pin, and Tier 2's `test_fix_blank_oserror_message_is_still_an_error` (0.54s).
- **CI**: all 18 checks green, including every required one (lint, test on 3 OSes, golden-path, notebooks).

### Code quality

- The relocated docstring accurately states the asymmetric stakes (Tier 2: the #1769 false-clean; Tier 1: readability only), and the `FixFileResult.error` comment keeps the invariant honest — presence stays `error is not None`; the helper is not what makes Tier 2 safe.
- Leaving the `FileNotFoundError` arm untouched (it has its own message) and keeping bare `ValueError` uncaught both preserve the existing design intent from #1769/#1770.
- Helper placement at the end of the report-model section is right — it renders for reports.

If forced to find something: the docstring's "Both read paths" reads as "both tiers," while there are technically three call sites (Tier 1, Tier 2's T1 read, and the locked fresh read) — not worth changing.

PR #1773, opened 2026-07-16
feat: report dangling wikilinks in the memory index, at info severity (#1762)

pandas-studio:
## Summary

`mm memory doctor` link-checked markdown pointers and ignored wikilinks, so a `[[deleted-memo]]` left behind by a removed file was invisible. This adds `dangling_wikilink` — **info-severity, index file only**.

Closes #1762.

## The decision

#1762 asked whether to check wikilinks at all, and left resolution, severity and scope open. The answers, and why:

- **Severity: info.** The doctor cannot distinguish a *forward reference* — which the memory convention allows, marking a memory worth writing later — from a *stale link* to a deleted memo. Reporting either at error severity would fail CI on documented, intended behaviour. Info surfaces the residue without gating; the summary names both readings so the report doesn't pretend to know which one it found.
- **Scope: index file only.** Memo bodies stay unread, so the read-only/perf contract in the module docstring is untouched.
- **Resolution: `[[name]]` → `name.md`** in the memory root, following the importers' convention (`indexing/importers.py`). An `|alias` names the display text, never the file; an Obsidian `#section` suffix is dropped.
- **`--fix` is unchanged.** Wikilinks are not entries, so they never feed `broken_link`, the listed set, or eligibility. A line whose markdown links are all dead stays deletable even if it carries a wikilink — considered and deliberately left, pinned by `test_wikilink_does_not_shield_a_dead_line_from_fix`.

## The parser guard

Both routes to a wikilink read text markdown-it has already **decoded**, and decoding is where a wikilink can be conjured from source that wrote none: `\[\[future]]` renders `[[future]]`, and a `[\[memo\]](gone.md)` label renders `[memo]` — the same shape a real `[[memo]](note)` leaves behind. Left unconfirmed, that demotes an ordinary pointer: its dead target then reported by nothing and offered to `--fix` by nothing.

So every match is confirmed against the raw source, and the label route additionally requires `[[inner]](` — it only fires on a real CommonMark link, so its source necessarily reads `[[inner]](destination)`. Escaping brackets is how an author says "not a wikilink"; honouring that also fixes a pre-existing case of the same demotion (`- [\[memo\]](gone.md) and `[[memo]]``), now pinned.

Two residual imprecisions are **accepted and pinned as such** (`test_raw_confirmation_is_inline_wide`, reasoning in the `_wikilink_in` docstring): escaped prose beside a code span quoting the same name raises an advisory finding, and an entity inside a target (`[[memo&amp;x]]`) is not collected. Both are info-severity-only — raised or missed, never an error, never a `--fix` candidate — and closing them means hand-rolling CommonMark's code-span and escape rules over raw source, the parsing-by-pattern this module exists to avoid.

## Docs

The forward-reference convention had no home in the repo's docs — it lived in a test comment. `docs/guides/reference/organization-maintenance.md` now states it under **Dangling wikilinks**, which is what makes the severity choice legible rather than arbitrary. The checks table gains the `dangling_wikilink` row, and the `broken_link` row's "wikilinks are not link-checked" note now points at the new check.

## Verification

- `uv run pytest -m "not ollama"`: 8685 passed, 11 skipped.
- Lint over the CI surface (`src`, `tests`, `tools`): `ruff check` + `ruff format --check` clean. `mypy` advisory: no new errors (4 before, 4 after).
- Real-data smoke over the configured memory dirs — the check fired on a genuine dangling wikilink (`· 1 wikilink(s) in MEMORY.md with no memory file …`), rendered with the info glyph, exit code unaffected, and present in `--json`.

## Follow-up filed

Codex review surfaced a **pre-existing** data-loss shape in the label route, reproduced byte-identically on `main`, so it is out of scope here (this PR neither introduces it nor widens it): filed as #1774.


tsdata (approved):

pandas-studio:
## Review — `dangling_wikilink` (#1762)

Verified against the PR head (local checkout is byte-identical to `origin/feat/1762-dangling-wikilink`). `pytest packages/memtomem/tests/test_memory_doctor.py`: 187 passed. `ruff check` over `src`/`tests`/`tools`: clean.

**Verdict: no blockers.** The design decisions are sound and, unusually, legible — the severity choice is defended in the docs rather than asserted in a commit message. Notes below are all minor.

### What holds up

- **The parser guard is the right call, and it's the part that would have bitten.** Reading the *rendered* label and confirming against `token.content` closes a real demotion: `[\[memo\]](gone.md)` renders `[memo]`, and an unconfirmed match would have stopped link-checking an ordinary pointer *and* dropped it from `--fix` — a silent loss with no finding to show for it. `test_escaped_label_pointer_stays_a_fix_candidate` pins the consequence the parser-level assertion can't see, which is the assertion that actually matters.
- **`followed_by_paren` asymmetry is well-reasoned.** Tightening only the route where a demotion costs something, and leaving the text route inline-wide, spends precision where the failure is expensive. The two accepted imprecisions are pinned as accepted (`test_raw_confirmation_is_inline_wide`) rather than left to be rediscovered as bugs.
- **`_WIKILINK_RE` genuinely mirrors `chunking/markdown.py:15`** (byte-identical pattern) — the "own copy, not an import of chunking internals" claim checks out.
- **Blast radius is contained.** `_read_inline`'s 2-tuple → 3-tuple change has no callers outside this module (grepped `src` + `tests`); `wikilinks` defaults to `()` on `ParsedIndex`, so no construction site breaks.
- Spot-checked the resolution edges: `[[#section-only]]` and `[[  ]]` skip on the empty-name guard, `[[https://example.com]]` classifies `url` and drops out, `[[nested/other|alias]]` resolves on the target. No crashes, no false errors.

### Minor

1. **Repeated same-name wikilinks on one line emit byte-identical items.** `- [[memo]] and [[memo]]` yields `wikilinks == ((1, 'memo'), (1, 'memo'))` → two identical `L1 [missing_target] [[memo]] → memo.md` lines and a count of 2. This is *consistent* with `broken_link`'s stated "per entry, not per line" philosophy (two identical dead pointers on a line report twice there too), so it's defensible — but two identical strings in an items list read as a report bug to anyone who didn't read that comment. Consider deduplicating on `(line_no, name)`, or leaving it and saying so.

2. **`outside_root` wikilinks are reported, but the docs row doesn't say so.** `[[../../etc/passwd]]` classifies `outside_root` and surfaces as `L1 [outside_root] …`, while the checks-table row describes only "no `name.md` in the memory root". The behaviour is right (info, not error — nothing follows a wikilink, so an escape isn't the hazard it is for a pointer); the row just doesn't prepare the reader for the `[outside_root]` prefix it will see.

3. **`[[name.md]]` diverges from the importers' convention the PR cites.** `_convert_obsidian_syntax` appends `.md` unconditionally (`[[memo.md]]` → `memo.md.md`); the doctor's `name if name.endswith(".md") else name + ".md"` is the more forgiving read. The doctor's behaviour is better — worth a half-line noting it's deliberately lenient rather than an exact mirror, since the PR body claims the importers' convention wholesale. Same for spaces: importers `%20`-escape into the link destination, the doctor resolves `[[Memo Name]]` → `Memo Name.md`, which is the correct filesystem read for a syntax that was never URL-encoded.

4. **Module-docstring reflow.** `orphans, dangling wikilinks, budget and cold candidates are advisory (a\npartially-indexed dir is\na legitimate steady state)` breaks mid-parenthetical across three lines. Cosmetic; ruff won't care.

### On the `--fix` interaction

`test_wikilink_does_not_shield_a_dead_line_from_fix` pins that `- [Dead](gone.md) — supersedes [[forward-ref]]` stays eligible, so `--fix` silently takes the forward reference with the line. Considered-and-left is the right disposition — a wikilink isn't an entry, and shielding would make any dead line un-fixable by adding a cross-reference — but this is the one place a user could lose information they meant to keep. `--fix` prints every removed line, so it's auditable; that's what makes it acceptable, and it may be worth one clause in the **Dangling wikilinks** doc section saying so out loud.

PR #1775, opened 2026-07-16
fix: settle wikilink labels per name, not per inline

pandas-studio:
Fixes the `--fix` data-loss path filed as #1774, found by Codex review during #1773.

## The bug

The label route confirmed a wikilink-shaped title by asking whether the *inline* writes `[[inner]](` anywhere — not whether *that occurrence* does. On

```markdown
- [[memo]](note) and [\[memo\]](README.md) and [Dead](missing.md)
```

the genuine wikilink vouched for the escaped-bracket pointer beside it: both decode to the label `[memo]`, both demoted, the live `README.md` pointer gone from the entries — and the line, now reading as all-dead, deleted whole by `--fix --apply`. That is the one consequence ADR-0020 promises `--fix` never has: the deletion proof was wrong.

## The fix — a per-name census, fail closed on ambiguity

`_settle_wikilink_labels` replaces the membership test. Per inline, per label name:

- **No raw `[[name]](` at all** → every same-named label was escaped or encoded; all stay pointers (post-#1773 behavior).
- **Every raw occurrence attributable** — raw occurrences minus those a *claim source* accounts for still cover every same-named label, and nothing on the inline is opaque → all are wikilinks; demote all (today's behavior).
- **Anything else** → **contested**: every same-named label stays an entry with the new `IndexEntry.unreadable_reason = "contested wikilink label"` (the old bool is now a derived property), flowing through the existing `ambiguous_index_line` idiom — warn, excluded from `broken_link` / listed / `--fix` candidacy, line skipped with a reported reason. Contested label occurrences collect no wikilink: `dangling_wikilink` must not report a guess.

*Claim sources* are the places a raw needle can sit whole without being a wikilink: code spans, `html_inline`, link `href`/`title` attributes, and out-of-link text (a text-resident `[[name]]` can only complete its needle inside its own token — no other token's raw starts with `(`). *Opacity* covers everything else via a token-type whitelist (an `image` or unknown token fails closed) plus bracket accounting: brackets left in text after whole `[[…]]` matches are scrubbed are the strands a needle split across tokens always leaves (a reference link consuming its middle strands `[other][` / `](x)`), and a non-label-shaped label's brackets get the stricter unscrubbed test. Every error path — decoding drift, unknown tokens, stranded brackets — can only over-subtract, which lands in *contested* (warn), never in a demotion.

## Review

Design gated through five adversarial Codex rounds before implementation, each closing a vouching path the previous round left open: inline-wide counting → text claims → token whitelist → attribute claims → bracket accounting; a sixth round re-gated the text-scrub refinement together with the diff (verdict: SHIP). All of these are pinned as preservation tests — through the parser, the candidate partition, and the real `--fix --apply` write path (dry-run advertises nothing, apply leaves the file byte-for-byte).

Deliberate fail-closed noise is pinned too: a genuine wikilink beside an image or bracketed prose goes contested (warn-only); flipping one of those pins back to a demotion is reopening the deletion path, not polishing.

## Verification

- 8709 passed (`-m "not ollama"`), ruff clean; mypy advisory count unchanged vs main.
- Real-directory smoke: `mm memory doctor` over every configured memory dir reports the same findings as main — the #1761 index shape (`[Topic](t.md) — built → [[a]](PR#42 merged)·[[b]](미커밋)`, present in a real `MEMORY.md`) still demotes cleanly, pinned as a test row. `--fix` dry-run: no candidates.

ADR-0020 §1 gains the contested label as a fourth unreadable shape (amended in this PR, per convention); the reference guide's `broken_link` / `ambiguous_index_line` rows updated, with the remediation Codex suggested (rename the bracketed label to keep a pointer; un-escaping makes it a wikilink instead).

Closes #1774


tsdata (approved):

pandas-studio:
## Review — fix: settle wikilink labels per name, not per inline

### Overview

Replaces the label route's inline-wide `[[name]](` membership test with a per-name census (`_settle_wikilink_labels`): raw occurrences of the needle, minus those a *claim source* (code span, `html_inline`, link `href`/`title`, out-of-link text) accounts for, must cover every same-named label — and the inline must contain nothing *opaque* (non-whitelisted token types, stranded brackets) — before any demotion. Anything short of that is a new third verdict, **contested**, carried per entry as `IndexEntry.unreadable_reason` (the old `unreadable` bool becomes a derived property) and flowing through the existing `ambiguous_index_line` warn path: not link-checked, not listed, never a `--fix` candidate, line skipped with a reason. This closes the #1774 data-loss path where a genuine `[[memo]](note)` vouched for an escaped `[\[memo\]](README.md)` beside it and the line was spliced with the live pointer still on it.

### Verification performed

- `pytest packages/memtomem/tests/test_memory_doctor.py`: **211 passed** on the branch.
- Probed five adversarial shapes *not* pinned in the test file (2 genuine + 1 escaped same-name; 2 escaped + 1 genuine; a genuine wikilink whose own destination carries a needle; an autolink carrying `[[memo]](y)` in its URL; a genuine wikilink wrapped in emphasis). Every one resolved fail-closed (contested/warn) or correctly demoted — no fail-open path found.
- Checked the one place the census could fail *open*: `claimed` subtracts needle counts from **decoded** claim sources against a **raw** count, so a claim source that decodes differently than the raw span could under-claim and over-demote. It can't here: `_markdown_parser` sets `md.normalizeLink = lambda url: url`, so `href` is verbatim, and escape/entity decoding only ever *adds* claims relative to raw (over-subtraction → contested, the allowed error). The "link destination carries the needle" test row pins this dependency — if link normalization is ever turned back on, that row fails.

### Correctness

- The three-verdict arithmetic (`raw_count == 0` → pointer; `not opaque and raw_count - claimed >= label_count` → wikilink; else contested) is sound: every identified error path (decoding drift, unknown token types, stranded brackets, double-claiming) lands in contested, never in a demotion. The direction of every inequality is the safe one.
- Bracket accounting is the clever part and holds up: a needle split across token boundaries (the reference-link smuggle) necessarily strands `[`/`]` fragments in text that the whole-`[[…]]` scrub can't remove, and the label-shaped link's own decoded `[inner]` is the single exempted bracket use. The stricter unscrubbed test on non-label-shaped labels closes the `[[[a]]](x)` shape.
- Contested label occurrences correctly collect no `dangling_wikilink` (no reporting on a guess), while independently-confirmed text-route wikilinks still pass through beside a contested label.
- `ambiguous_index_line` summary now counts **lines** (`ambiguous_line_nos`) instead of items — a real fix; previously a line with two odd targets counted twice. Pinned by `test_contested_label_is_warn_not_broken_nor_dangling`.
- The `unreadable_reason` precedence (target-shape doubt wins over contested) only affects the printed tag, not eligibility — both routes converge on the same exclusions. Fine.

### Test coverage

Excellent — the strongest part of the PR. The matrix covers all four seams: the parser (parametrized contested / still-demoted / fail-closed-noise rows), the candidate partition (`_partition_candidates`), the report surface (finding items + severity + summary), and the real write path (`--fix` dry-run advertises nothing, `--apply` leaves the file byte-for-byte). Two touches worth calling out:

- `FAIL_CLOSED_NOISE` pins the *deliberate* regressions (genuine wikilink beside an image or bracketed prose now warns instead of demoting) with a comment saying that flipping one back is reopening the deletion path — exactly the right guard against future "noise reduction".
- The #1761 real-index shape (`[[a]](PR#42 머지)·[[b]](미커밋)`) is pinned as still-quiet, so the fix provably doesn't regress the case the wikilink-label route exists for.

### Nits (non-blocking)

- `unreadable_reason` is a free-form string doing double duty as display tag and discriminant; the two literals appear in `parse_memory_index` and are matched nowhere else, so this is fine today, but a `Literal["unreadable target", "contested wikilink label"]` alias would let mypy catch a future typo'd third value for free.
- `_CENSUS_SAFE_TYPES` lists `s_open`/`s_close`, which the `"commonmark"` preset never emits (strikethrough is a GFM extension). Harmless future-proofing, but a one-word comment would stop a reader from hunting for where strikethrough got enabled.

### Docs

ADR-0020 amendment (fourth unreadable shape), the guide's `broken_link`/`ambiguous_index_line` rows, and the module docstring all match the shipped behavior — I checked the remediation wording and the summary message text against the source strings.

### Verdict

**SHIP.** No correctness findings; the fail-closed invariant holds under adversarial probing beyond the pinned rows, and the promised harm is pinned through the real write path.

PR #1779, opened 2026-07-16
fix(search): defer closing retired rerankers until in-flight searches release them (#1777)

memtomem:
## Summary

Hot reload replaces `SearchPipeline._reranker` and immediately closes the old
instance while an in-flight search may still hold it via the per-call snapshot
from #1776 — Stage 3b can then invoke the already-closed reranker
(fastembed/local silently reload the released model; cohere re-creates an
`httpx.AsyncClient` on the orphaned instance that nothing ever closes).
Closes #1777.

- The `(reranker, config)` pair now lives on a `_RerankerEntry` generation
  with an in-flight lease count. `search()` leases the generation for the
  whole ranked-search body; acquire/release are synchronous ops between
  awaits on one event loop, so a plain counter suffices (no lock).
- New `SearchPipeline.swap_reranker()` is the single hot-swap API: it
  publishes the new generation before any await, closes the old instance
  immediately at zero leases, and otherwise retires it; the last lease
  release schedules the close on `_bg_tasks`.
- Both web swap sites — `_sync_reranker` (disk hot reload) and the PATCH
  handler — now delegate to `swap_reranker`, collapsing the old
  "keep the two paths in lockstep" comments into one call site each. The
  never-installed-reranker close paths (validate failure, stale drop) are
  unchanged: those instances were never visible to a search.
- `pipeline.close()` drains retired-but-leased generations before gathering
  `_bg_tasks`, so a straggler release cannot schedule a second close.
- The canonical swallow-and-log close helper moved to
  `search/reranker/base.py` (`close_reranker_safely`) since the pipeline
  cannot import from `web/`; `hot_reload._close_reranker_safely` delegates.

## Behavior change

- When a search is in flight, disabling/replacing the reranker (disk edit or
  `PATCH /api/config`) now defers closing the old instance until that search
  releases it, instead of closing immediately. Swaps with no search in flight
  close synchronously before the swap call returns, exactly as before.
- Direct `_reranker`/`_rerank_config` assignment (property shims) keeps the
  pre-#1777 contract: fresh generation, old one abandoned unclosed — the
  assigner owns its lifecycle.

## Testing

- New `TestRerankerSwapLeasing` (pipeline level) with a close-aware fake that
  refuses to rerank after close: headline defer-until-release scenario,
  zero-lease synchronous close, double swap while leased, lease release on
  the exception and cache-hit early-return paths, `close()` draining exactly
  once. The fix was neutered (unconditional close in `swap_reranker`) to
  confirm the new tests go red.
- The real flows the issue named: `apply_runtime_config_changes` with a
  search blocked in retrieval (old reranker survives Stage 3b, closes exactly
  once after release) and PATCH-path parity. The existing
  closed-immediately assertions stay green via the zero-lease synchronous
  close; the `SimpleNamespace`/`AsyncMock` pipeline fakes in the web tests
  became real `SearchPipeline` instances since the swap now drives the real
  lease machinery.
- Existing `test_hot_reload_mid_search_uses_consistent_snapshot` passes
  unmodified (the property shims absorb its direct attribute flips).
- Full suite: 8,779 passed, 11 skipped; ruff clean; mypy unchanged from
  baseline (18 pre-existing errors).

## Follow-up

Provider-level defense-in-depth ("a closed reranker refuses to resurrect")
is split out per one-change-per-PR: #1778.

### DIFF
--- a/packages/memtomem/src/memtomem/search/pipeline.py
+++ b/packages/memtomem/src/memtomem/search/pipeline.py
@@ -36,7 +36,9 @@ applies in this branch too via ``recall_chunks``.
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import logging
+from collections.abc import Iterator
 from datetime import UTC, datetime
 from fnmatch import fnmatch
 from pathlib import Path
@@ -59,6 +61,7 @@ from memtomem.config import (
 )
 from memtomem.models import ContextInfo, NamespaceFilter, ScopeFilter, SearchResult
 from memtomem.search.fusion import reciprocal_rank_fusion
+from memtomem.search.reranker.base import close_reranker_safely
 from memtomem.storage.base import SearchMetadataFilter
 from memtomem.storage.sqlite_helpers import norm_path
 
@@ -109,6 +112,22 @@ def _bg_task_error_cb(task: asyncio.Task) -> None:
     exc = task.exception()
     if exc:
         logger.warning("Background task %s failed: %s", task.get_name(), exc)
+
+
+class _RerankerEntry:
+    """One (reranker, config) generation with an in-flight lease count (#1777).
+
+    Identity-hashed on purpose: retired generations live in a set keyed by
+    the entry object itself, so two generations wrapping the same values
+    never collide.
+    """
+
+    __slots__ = ("reranker", "config", "leases")
+
+    def __init__(self, reranker: object | None, config: RerankConfig | None) -> None:
+        self.reranker = reranker
+        self.config = config
+        self.leases = 0
 
 
 def match_source_filter(filter_str: str, source_path: str) -> bool:
@@ -303,8 +322,8 @@ class SearchPipeline:
         self._decay_config = decay_config or DecayConfig()
         self._mmr_config = mmr_config or MMRConfig()
         self._access_config = access_config or AccessConfig()
-        self._reranker = reranker
-        self._rerank_config = rerank_config
+        self._rerank_entry = _RerankerEntry(reranker, rerank_config)
+        self._retired_rerank_entries: set[_RerankerEntry] = set()
         self._expansion_config = expansion_config
         self._importance_config = importance_config
         self._context_window_config = context_window_config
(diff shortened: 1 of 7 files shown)


pandas-studio (commented):
## Review — PR #1779

### Overview

Replaces the per-call `(reranker, config)` snapshot from #1776 with a leased `_RerankerEntry` generation: `search()` leases the generation for its whole body, and `swap_reranker()` — now the single hot-swap API for both web paths — closes the old instance synchronously at zero leases or defers the close to the last lease release. This correctly fixes the #1777 use-after-close: the snapshot kept the *reference* consistent but nothing kept the *instance alive*.

I traced the interleavings rather than just reading the happy path; the core accounting is sound:

- **Once-only deferred close**: the last release pops the entry from `_retired_rerank_entries` synchronously before `create_task`, and acquire/release/pop all sit between awaits on one loop — no window for a double schedule. `close()` empties the set before gathering `_bg_tasks`, so a straggler release after drain finds nothing to schedule, and an already-scheduled deferred close is gathered. Both directions hold.
- **Publish-before-await** in `swap_reranker` is real: the new entry is assigned before any `await`, so a hanging `close()` can't leave the pipeline stale (preserving the property the old PATCH-handler comment pinned).
- **Per-entry counter suffices**: production swap sites always pass a freshly `create_reranker`-built instance (or `None`), so each generation's instance is unique and its lease count fully covers it — verified in `system.py` (`pending_reranker` is always fresh) and `_sync_reranker`.
- Lease release on the exception path, the cache-hit early return, and cancellation all go through the context manager's `finally`.

Tests are strong: close-aware fake with a rerank-after-close tripwire, the fix was neutered to prove the new tests go red, and the real `apply_runtime_config_changes` / PATCH flows are exercised (with the `SimpleNamespace` fakes upgraded to real pipelines — a nice side effect).

### Finding (plausible, low reachability)

**Same-instance swap splits lease accounting across generations.** `swap_reranker` early-returns when `old.reranker is reranker`, leaving the old (possibly leased) generation un-retired while the new generation wraps the same instance at `leases == 0`. A subsequent swap then sees zero leases on the *current* entry and closes the instance synchronously while the older generation's search is still mid-Stage-3b with it — the original #1777 symptom, resurrected through the identity-swap path:

1. Search S leases gen1 = (R, C1).
2. `swap_reranker(R, C2)` → gen2 = (R, C2), gen1 abandoned (leased, not retired).
3. `swap_reranker(None, None)` → gen2 has 0 leases → R closed while S still uses it.

Neither production call site can hit this today (both always construct a fresh instance or `None`), so this is an API-contract hazard rather than a live bug. Since `swap_reranker` is documented as *the* single hot-swap API, I'd either (a) note in the docstring that `reranker` must be a fresh instance never previously installed, or (b) make the identity case safe, e.g. retire the old generation too (its close via `close_reranker_safely` on the shared instance would then need care) or transfer `old.leases` onto the new entry when the instance is identical — the transfer variant is a 2-line guard. Fine as a follow-up; the property-shim path has the same shape but is explicitly documented as assigner-owns-lifecycle.

### Nits (non-blocking)

- `pipeline.close()` closes the *active* reranker with a bare `await self._reranker.close()` (raises on a flaky teardown) while retired generations go through the swallow-and-log helper. Pre-existing asymmetry, but now that `close_reranker_safely` lives in `search/reranker/base.py` it's a one-line unification if you want it.
- `close_reranker_safely`'s log message says "replaced reranker"; the `close()` drain path closes *retired* ones at shutdown — trivially inaccurate wording, not worth blocking on.
- `_RerankerEntry.reranker: object | None` keeps the loose typing instead of the `Reranker` protocol — consistent with the surrounding code, just noting it's a conscious continuation.

### Verdict

LGTM with the same-instance hazard noted above as an optional follow-up. The behavior change is clearly called out, the moved helper is a consequent change (pipeline can't import from `web/`), and the test additions pin exactly the interleavings that matter.

pandas-studio:
Thanks for the interleaving-level review.

**Same-instance swap finding — confirmed, addressed as (a) in 846d8d63.** Your trace is right: the `old.reranker is reranker` early return abandons the leased generation, and a later swap sees zero leases on the new generation and closes the shared instance under the in-flight search. Since neither production call site can pass a previously-installed instance (both always hand over a fresh `create_reranker` product or `None`), I pinned the fresh-instance contract in the `swap_reranker` docstring rather than generalizing.

One correction on the suggested 2-line transfer variant: it double-counts. Leases are held on the entry object the search actually leased, so the release decrements the *old* generation's counter — the count transferred onto the new entry would never be decremented, and that generation could then never reach zero leases, deferring its close to the `pipeline.close()` drain instead of the last release. Making identity re-install actually safe needs a per-instance (not per-generation) refcount, which is more than two lines and has no caller today — so contract-over-mechanism, per YAGNI. If a caller with identity re-install ever shows up, the docstring now says exactly what to build.

Nits: agreed on all three; leaving them as-is in this PR (the `close()` asymmetry and the log wording are pre-existing, and the loose `object | None` typing deliberately matches the surrounding code). Happy to fold the one-line `close_reranker_safely` unification into a follow-up if you feel strongly.

pandas-studio (commented):
## Delta re-review — PR #1779 (846d8d6, since 3df0dd0)

One new commit: `docs(search): pin the fresh-instance contract on swap_reranker` — a 9-line docstring addition adopting option (a) from the previous review for the same-instance-swap hazard.

Verified against the code:

- The contract statement matches the actual early-return (`old.reranker is None or old.reranker is reranker`) and correctly describes the failure mode it forbids.
- "Both production call sites always pass a fresh `create_reranker` product or `None`" — re-confirmed; still true for `_sync_reranker` and the PATCH handler.
- The closing rationale ("supporting identity re-install would take a per-instance, not per-generation, refcount") is **more correct than my own suggested alternative**: the "transfer `old.leases` onto the new entry" 2-line variant I floated is actually broken — `_lease_reranker` decrements the *captured* entry on release, so transferred counts on the new generation would never decrement (leak, never closes) while the old generation underflows its guard. Per-instance refcounting really is the minimum viable mechanism, and declining to build it for a caller that doesn't exist is the right YAGNI call.

Docs-only, no behavior change, previous non-blocking nits remain optional. **LGTM — the one open finding from the previous review is resolved.**

tsdata (approved):

PR #1780, opened 2026-07-16
fix(rerank): a closed reranker refuses to resurrect its resources (#1778)

pandas-studio:
## Summary

All three reranker providers treated `close()` as resettable state, so using
a closed instance silently re-created what close had released — cohere built
a brand-new `httpx.AsyncClient` that nothing on an orphaned instance ever
closes (connection leak), fastembed/local silently re-downloaded /
re-initialized the model. Closes #1778.

Since #1777/#1779 the swap sites lease-defer closes, so no production path
invokes a closed reranker — this is defense-in-depth that makes the
invariant loud instead of trusting it:

- A `_closed` flag set in `close()` makes the lazy loaders (`_get_client` /
  `_get_model`) raise `RuntimeError("<Provider> is closed")`. The guard sits
  at the loader — the single choke point shared by `rerank()`, warmup, the
  PATCH readiness validation (`_validate_reranker_ready`), and the
  hot-reload eager load. (The `GET /system/model-readiness` endpoint only
  inspects loader state; it never drives a load.)
- For the threaded providers (fastembed/local) the guard holds across the
  whole load, not just at entry: `_closed` is re-checked under `_load_lock`,
  and a model whose construction straddled a `close()` is
  published-then-verified — the loader unpublishes and raises rather than
  parking the finished model on the closed instance. `close()` never takes
  the load lock (it must not block the event loop behind an in-flight
  download).
- Through the search pipeline a stray call degrades to the existing
  un-reranked fallback with a warning (cohere/local propagate to Stage 3b's
  catch; fastembed's own `except Exception` fallback catches it — its
  `ImportError`/`ValueError` setup-error carve-out is untouched). Direct
  loader callers get the loud error, which is correct: reaching a closed
  instance there is a lifecycle bug worth surfacing.
- `close()` stays idempotent for all three providers.

Also adds the CHANGELOG entry for #1777 that #1779 landed without (separate
commit).

## Behavior change

- `rerank()` / lazy-load on a closed reranker previously "worked" (silent
  resurrect); it now raises at the loader. Pipeline callers see a logged
  warning + un-reranked results for the stray call; nothing changes for any
  live instance.
- `test_close_resets_model` pinned the opposite contract ("so the reranker
  can be reused") — wording from the provider's introduction (#222); `git
  log -S` shows no caller ever reused a closed instance. The test now pins
  release-and-refuse (`test_close_releases_model_and_refuses_reuse`).

## Testing

- Per-provider post-close tests with positive controls: cohere pins that no
  new client materializes after close (`_client` stays `None` — the leak);
  local/fastembed pin no model reload; fastembed additionally pins the
  original-order fallback through its own except path; double close is a
  no-op for all three.
- Red-first: with the guard neutered all three new tests fail — the
  neutered fastembed run visibly re-downloads the model from HuggingFace,
  demonstrating the exact cost being fenced.
- Full suite: 8,781 passed, 11 skipped; ruff clean; mypy unchanged from baseline (18
  pre-existing errors).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   an in-flight search still holds it; the close is deferred until the last
   such search releases its lease. Swaps with no search in flight close
   immediately, as before.
+- **Closed rerankers no longer resurrect their resources** (#1778) — using a
+  reranker after `close()` previously re-created what close had released
+  (cohere: a new `httpx` client nothing ever closes; fastembed/local: a
+  silent full model reload). The lazy loaders now raise `RuntimeError`
+  instead; a stray call through the search pipeline degrades to the existing
+  un-reranked fallback with a warning.
 
 ## [0.3.11] — 2026-07-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   Omitted/`true` follows `rerank.enabled`; `true` cannot force-enable
   reranking on a server that has it disabled.
 
+### Fixed
+
+- **Reranker hot-swap use-after-close** (#1777) — hot reload (disk config
+  edit or `PATCH /api/config`) no longer closes the outgoing reranker while
+  an in-flight search still holds it; the close is deferred until the last
+  such search releases its lease. Swaps with no search in flight close
+  immediately, as before.
+
 ## [0.3.11] — 2026-07-14
 
 ### Added
--- a/packages/memtomem/src/memtomem/search/reranker/cohere.py
+++ b/packages/memtomem/src/memtomem/search/reranker/cohere.py
@@ -20,8 +20,14 @@ class CohereReranker:
     def __init__(self, config: RerankConfig):
         self._config = config
         self._client: AsyncClient | None = None
+        self._closed = False
 
     def _get_client(self) -> AsyncClient:
+        # A closed instance must not resurrect: re-creating the client here
+        # would leak it — nothing ever closes a client born after close()
+        # on a swapped-out instance (#1778).
+        if self._closed:
+            raise RuntimeError("CohereReranker is closed")
         if self._client is None:
             import httpx
 
@@ -75,6 +81,7 @@ class CohereReranker:
         return reranked[:top_k]
 
     async def close(self) -> None:
+        self._closed = True
         if self._client:
             await self._client.aclose()
             self._client = None
--- a/packages/memtomem/src/memtomem/search/reranker/fastembed.py
+++ b/packages/memtomem/src/memtomem/search/reranker/fastembed.py
@@ -48,11 +48,14 @@ class FastEmbedReranker:
         """
         # A closed instance must not resurrect: re-downloading/re-initializing
         # the released ONNX model here is silent expensive work on an
-        # instance nobody owns (#1778).
+        # instance nobody owns (#1778). Cached reads go through a local
+        # snapshot so a concurrent close() nulling ``_model`` between the
+        # check and the return cannot hand the caller ``None``.
         if self._closed:
             raise RuntimeError("FastEmbedReranker is closed")
-        if self._model is not None:
-            return self._model
+        model = self._model
+        if model is not None:
+            return model
         with self._load_lock:
             # Re-check under the lock: a warmup/readiness thread that passed
             # the guard above can lose the race to a concurrent close() on
@@ -60,8 +63,9 @@ class FastEmbedReranker:
             # the closed instance (#1778).
             if self._closed:
                 raise RuntimeError("FastEmbedReranker is closed")
-            if self._model is not None:
-                return self._model
+            model = self._model
+            if model is not None:
+                return model
             try:
                 from fastembed.rerank.cross_encoder import (  # type: ignore[import-untyped]
                     TextCrossEncoder,
@@ -81,9 +85,7 @@ class FastEmbedReranker:
             self._loading = True
             self._load_error = None
             try:
-                self._model = TextCrossEncoder(
-                    model_name=self._config.model, cache_dir=str(cache_dir)
-                )
+                model = TextCrossEncoder(model_name=self._config.model, cache_dir=str(cache_dir))
             except ValueError as exc:
                 supported = [m.get("model", "") for m in TextCrossEncoder.list_supported_models()]
                 self._load_error = str(exc)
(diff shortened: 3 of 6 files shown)


tsdata (approved):

pandas-studio (commented):
## Review

### Overview

Adds a `_closed` flag to all three reranker providers so post-`close()` use raises `RuntimeError` at the lazy loader instead of silently re-creating released resources (cohere: leaked `httpx.AsyncClient`; fastembed/local: silent model re-download/re-init). Also backfills the missing #1777 CHANGELOG entry. Defense-in-depth on top of the #1777/#1779 lease-deferred closes.

### Verified claims

I checked each claim in the PR description against the code:

- **Loader is the single choke point** — confirmed. `CohereReranker.rerank()` calls `_get_client()` outside its try block (`cohere.py:46`), `LocalReranker.rerank()` calls `_get_model()` outside its try (`local.py:50`), and fastembed reaches `_get_model()` on every call via `_rerank_sync`. Warmup (`server/warmup.py:55`), the model-readiness eager load (`web/routes/system.py:138`), and the hot-reload eager load (`web/hot_reload.py:368`) all go through the same loaders, and all operate on live/pre-install instances only.
- **Pipeline degradation** — confirmed. Stage 3b wraps `reranker.rerank(...)` in `except Exception` → warning + `fused[:top_k]` (`pipeline.py:1186-1192`), so a stray cohere/local `RuntimeError` degrades as described. Fastembed's own `except Exception` fallback catches it (`RuntimeError` is not in the `(ImportError, ValueError)` re-raise carve-out), pinned by the updated test.
- **Idempotent close** — confirmed for all three; pinned in each test.
- **Tests** — ran both test files locally on this branch: 17 passed. Positive controls (live instance works pre-close) are a nice touch — they keep the tests from passing vacuously.

### Minor (optional, non-blocking)

- **TOCTOU window in the double-checked lock** (`fastembed.py:52`, `local.py:31`): the `_closed` check sits outside `_load_lock`, and `close()` doesn't take the lock. A thread that passes the guard, then loses the race to a concurrent `close()`, can still load a model onto the closed instance — the exact resurrect being fenced. This is production-unreachable for the same reason the whole guard is (#1777 lease-defer means no in-flight caller when close runs), but since this PR's entire purpose is making the invariant hold *without* trusting the call sites, re-checking `self._closed` inside the lock (before the load) would make it airtight for two extra lines, matching the double-checked pattern already in place. Cohere's `_get_client` has no lock at all, but that's pre-existing and its resurrect cost is a cheap client construction guarded the same way.
- `FastEmbedReranker.close()` leaves `_load_error`/`_loading` untouched — fine today since model-readiness only introspects live instances, just noting it's the same "trust the call sites" surface the guard otherwise removes.

### Verdict

LGTM. Correct, well-scoped, tests are red-first with positive controls, and the CHANGELOG backfill is a separate commit as it should be. The TOCTOU note above is the only thing I'd consider folding in, and it's optional.

pandas-studio:
**TOCTOU finding — confirmed and folded in as 62de86e2** (upgraded from optional on adjudication: the #1777 lease covers *searches* only, so the warmup / model-readiness eager-load surfaces can genuinely race a swap-triggered close — the window is reachable, not just theoretical). `_get_model` now re-checks `_closed` under `_load_lock` in fastembed/local, and the interleaving is pinned deterministically with a lock wrapper that flips `_closed` at the acquisition point instead of racing real threads.

Cohere left as-is per your note: its loader is lock-free on the event loop and the check-to-construct span contains no await, so close() cannot interleave inside it.

`_load_error`/`_loading` on close: agreed it's the same trust-the-call-sites surface, but clearing them would also erase the last observability signal for a failed instance — leaving them is a deliberate keep for now.

pandas-studio (commented):
## Review (round 2 — after 62de86e2)

### Delta since last review

`62de86e2` addresses the TOCTOU note from the previous round: `_get_model` now re-checks `_closed` **under** `_load_lock` in both fastembed and local, so a loader thread that passed the outer guard and then lost the race to a concurrent `close()` at the lock-acquisition point now refuses instead of loading. Each provider gains a deterministic pin (`test_close_landing_at_lock_acquisition_refuses_load`) that injects the interleaving via a lock wrapper flipping `_closed` at the exact acquisition point — no flaky real-thread racing. Nice.

### Verified on this branch

- Both test files pass locally: 19 passed (was 17 — the two new race pins). `ruff check` + `format --check` clean on all changed files.
- The wrapper-lock technique is sound: `_CloseOnAcquire.__enter__` flips the flag before delegating to a real lock, so the test exercises exactly the code path the inner guard protects, and the `_model is None` assertion pins that no resurrect happened.
- Cohere intentionally keeps only the outer guard (it has no load lock; its resurrect cost is a cheap client construction) — consistent with the previous round's scoping.

### Residual window (informational — no further round needed)

One narrower interleaving remains: `close()` never takes `_load_lock`, so if it lands **while a load is already in flight** (between the inner re-check and the `self._model = ...` publish — a window that is wall-clock wide, since model construction takes seconds), the loader still publishes a model onto the closed instance. Reaching it requires a loader actively mid-download at the moment of close (e.g. startup warmup racing an immediate config swap that retires the instance), and the resurrected model is freed with the orphaned instance at GC, so the cost is transient memory, not a durable leak like the cohere client.

Closing it fully would require `close()` to serialize on the load lock (`await asyncio.to_thread(self._load_lock.acquire)` — it can't block the event loop for a multi-second in-flight load), which adds real complexity to shutdown for a defense-in-depth guard. I don't think that trade is worth it; if you agree, this is fine to land as-is (optionally note the window in the loader comment). Flagging only so the boundary of the invariant is a decision, not an accident.

### Verdict

LGTM — the review feedback was folded in with deterministic (not probabilistic) race coverage, and the remaining gap is a conscious, low-impact boundary.

pandas-studio:
**Round-2 residual window + codex round 1 — resolved by 9d22acb3, adjudicating between the two.**

Both reviews found the same close-during-construction window; they disagreed on disposition (round 2 here: "conscious low-impact boundary, land as-is"; codex: Blocker). Codex wins on the fix, not the severity argument: the round-2 assessment only weighed `close()` serializing on `_load_lock` (correctly rejected — it would block the event loop behind a seconds-long download), but **publish-then-verify** needs no lock in `close()`: the loader constructs into a local, publishes, re-checks `_closed`, and unpublishes + raises if closure landed. Every interleaving of close()'s `(flag, _model=None)` writes now ends with the closed instance holding no model. Cached fast-path reads also return a local snapshot so a close() between check and return can't hand the caller `None`.

Per codex's test finding: the new tests pin the *real* interleaving — a pausing constructor gates on events while the actual async `close()` runs mid-construction (`sentence_transformers` stubbed via `sys.modules` for local). Red-first: with the verify neutered, both fail.

Also folded in codex's PR-body correction: the loader caller is the PATCH readiness validation (`_validate_reranker_ready`); `GET /system/model-readiness` only inspects state. Body updated.

Codex round 2 running for confirmation.

Issue #750, opened 2026-05-03
Allow tag-only search (relax empty-q guard in /api/search and doSearch)


pandas-studio:
## What

Follow-up to #672 / PR #749. After PR #749, `_searchByTag` no longer pollutes `search-input`, so a tag pill click on a fresh session leaves `search-input` empty. `doSearch()` (`packages/memtomem/src/memtomem/web/static/app.js:1675`) currently early-returns on empty `q`:

```js
async function doSearch() {
  const q = qs('search-input').value.trim();
  if (!q) return;  // ← blocks tag-only path
  const tf = qs('tag-filter').value.trim();
  ...
}
```

…and the backend (`packages/memtomem/src/memtomem/web/routes/search.py`) declares `q` as required:

```python
async def search(
    q: str = Query(...),          # ← min length enforced; empty q → 422
    tag_filter: str | None = Query(None),
    ...
):
```

So the click currently acts as a "filter prep" — Search tab opens with `tag-filter` populated, but no results render until the user types a query.

## Why

The intuitive UX (per #672 discussion) is that clicking a tag pill = "show me all memos with this tag". Today that requires an extra keystroke. Lifting the empty-`q` guard on both layers makes the click work as expected without any further `_searchByTag` change — PR #749 already calls `doSearch()` unconditionally, so this fix is purely additive.

## Suggested change

1. **Frontend** (`app.js:1673-1722`): replace `if (!q) return;` with `if (!q && !tf) return;` so the search runs whenever either axis has a value.
2. **Backend** (`packages/memtomem/src/memtomem/web/routes/search.py`): make `q` optional (`q: str | None = Query(None)`) and validate that at least one of `q` / `tag_filter` is present (return 400 otherwise — keeps the "no-op search" guard at the API layer).
3. **`pipeline.search`** (`packages/memtomem/src/memtomem/search/pipeline.py`): verify it tolerates `query=None` / `query=""`. The pipeline order is fixed (per `CLAUDE.md`); the BM25/dense stages should short-circuit cleanly when there's no query and the tag/source filter stage takes over as the primary selector. Add a regression test for `tag_filter`-only retrieval.
4. **CHANGELOG**: `[Unreleased] / Changed` — "Tag pill click now runs a tag-only search on a fresh session (was a no-op filter prep after #672)."

## Out of scope

- Multi-tag selection on `tag-filter` (separate UX question — issue #672 noted the field is currently single-value text).
[...]

PR #1781, opened 2026-07-16
feat(search): name the score scale in structured search output (#1767)

pandas-studio:
## Summary

`mem_search`'s structured output emits a bare `score` whose scale silently follows server config — RRF fusion (~0.033 ceiling) vs cross-encoder rerank (model-dependent: raw logits locally, [0,1] on Cohere). Programmatic consumers can't threshold portably and can't detect which scale they're on (#1767).

This PR names the scale in the payload. Additive fields only; no ranking or default changes.

- **`RetrievalStats.score_scale`** is recorded at the pipeline branch that produced the scores: `"rrf"` (fused legs), `"bm25"` / `"dense"` (single-retriever passthrough — raw scores that never touch RRF; a third and fourth scale beyond the issue's two-value framing), `"none"` (filter-only enumeration, #750 — no relevance scale), `"rerank"` (Stage 3b actually re-scored the list).
- **The `"rerank"` label is derived from the returned results** (every item re-stamped `source="reranked"`), not from the decision flag — both the pipeline's `except` fallback and the providers' internal error fallbacks silently keep the fused order, so a decision-based label would lie exactly when it matters. Verified against all three reranker implementations.
- **Structured payloads** (`mem_search`, `mem_agent_search`) gain top-level `score_scale` + `reranker` (rerank model ID, since ranges are model-dependent); both omitted when there are no results. Per-result shape unchanged.
- **`mm search --format json`** carries `score_scale` per item (its payload is a bare list).
- Stats ride the TTL cache, so cache hits report the scale of the cached scores; the #1766 per-call bypass keys the same path.
- Labels name the **base** scale: decay/access/importance modifiers (all default-off) multiply on top when enabled — documented in the tool docstring and `core-memory-tools.md`.

## Review

`codex` review round 1: NEEDS-FIX — flagged that "constant 1.0" / "raw" wording was false under enabled modifier stages, plus two coverage gaps (filter-only under modifiers, `mem_agent_search` wiring). All three addressed: wording reframed as base scale, `TestScoreScale.test_filter_only_stays_none_scale_under_modifiers` pins the "none ≠ constant" contract, and the agent-search integration test now asserts the new keys.

## Testing

- `uv run pytest -m "not ollama"` — 8747 passed. New coverage: `TestScoreScale` (10 pipeline-branch cases incl. both silent rerank fallbacks, per-call bypass, cache hit), formatter emission/omission, tool-level (`mem_search` + `mem_agent_search`), CLI e2e JSON.
- `uv run ruff check` + `ruff format --check` clean; no new mypy findings in touched files.
- Real-store smoke (4k+ chunks, rerank enabled): `mm search --format json` → `"score_scale": "rerank"` with logit scores; `--no-rerank` → `"score_scale": "rrf"` with scores ≤ 0.0323, matching the distribution table in #1767.

Closes #1767

### DIFF
--- a/docs/guides/reference/core-memory-tools.md
+++ b/docs/guides/reference/core-memory-tools.md
@@ -213,6 +213,20 @@ mem_search(query="deploy pipeline", as_of="2025-Q3")    # historical query
 
 > **Trust-UX hints**: when you don't pin a namespace, results are followed by a parenthesized hint if chunks were hidden in system namespaces (e.g. `archive:*`) or if the configured embedding dimension disagrees with what's in the database. A third hint — independent of namespace selection — appears when you pass `rerank=true` but the server has reranking disabled (`rerank.enabled=false`), since the parameter cannot force-enable it. In `output_format="structured"` those hints are emitted as a `hints` array instead.
 
+> **Score scale**: `score` values are only comparable within one scale, and the
+> scale follows server config. Structured output names the base scale in a
+> top-level `score_scale` key: `"rerank"` (cross-encoder output — range depends
+> on the model, reported alongside in a `reranker` key), `"rrf"`
+> (reciprocal-rank fusion), `"bm25"` / `"dense"` (unfused single-retriever
+> scores when only one retriever is enabled), or `"none"` (filter-only
+> enumeration — no relevance scale; the filter is the selector). Optional
+> modifier stages (time decay, access/importance boosts; all off by default)
+> multiply on top of the base scale when enabled. Pick score thresholds per
+> scale — or skip score gating for a scale you don't recognize — instead of
+> inferring the scale from the value range. Both keys are omitted when there
+> are no results. `mm search --format json` carries the same value as a
+> per-item `score_scale` key.
+
 ### Tuning search weights
 
 Use `bm25_weight` and `dense_weight` to shift between keyword and semantic matching:
--- a/docs/guides/reference/core-memory-tools.md
+++ b/docs/guides/reference/core-memory-tools.md
@@ -224,8 +224,8 @@ mem_search(query="deploy pipeline", as_of="2025-Q3")    # historical query
 > multiply on top of the base scale when enabled. Pick score thresholds per
 > scale — or skip score gating for a scale you don't recognize — instead of
 > inferring the scale from the value range. Both keys are omitted when there
-> are no results. `mm search --format json` carries the same value as a
-> per-item `score_scale` key.
+> are no results. `mm search --format json` carries the same values as
+> per-item `score_scale` / `reranker` keys.
 
 ### Tuning search weights
 
--- a/packages/memtomem/src/memtomem/cli/search.py
+++ b/packages/memtomem/src/memtomem/cli/search.py
@@ -199,12 +199,15 @@ async def _search(
 
     if fmt == "json":
         # The payload is a bare list, so score provenance rides per item
-        # (#1767): key omitted when the pipeline produced no ranked scale.
+        # (#1767): keys omitted when the pipeline produced no ranked scale.
+        # ``reranker`` accompanies the "rerank" scale because rerank score
+        # ranges are model-dependent — mirroring the MCP structured payload.
         out = [
             {
                 "rank": r.rank,
                 "score": round(r.score, 4),
                 **({"score_scale": stats.score_scale} if stats.score_scale is not None else {}),
+                **({"reranker": stats.reranker_model} if stats.reranker_model is not None else {}),
                 "source": str(r.chunk.metadata.source_file),
                 "content": r.chunk.content[:200],
             }
--- a/packages/memtomem/src/memtomem/cli/search.py
+++ b/packages/memtomem/src/memtomem/cli/search.py
@@ -198,10 +198,13 @@ async def _search(
         return
 
     if fmt == "json":
+        # The payload is a bare list, so score provenance rides per item
+        # (#1767): key omitted when the pipeline produced no ranked scale.
         out = [
             {
                 "rank": r.rank,
                 "score": round(r.score, 4),
+                **({"score_scale": stats.score_scale} if stats.score_scale is not None else {}),
                 "source": str(r.chunk.metadata.source_file),
                 "content": r.chunk.content[:200],
             }
--- a/packages/memtomem/src/memtomem/search/pipeline.py
+++ b/packages/memtomem/src/memtomem/search/pipeline.py
@@ -288,6 +288,25 @@ class RetrievalStats:
     # the per-call effective decision (#1766), not the live server config,
     # so hint builders stay accurate across concurrent hot reloads.
     rerank_applied: bool = False
+    # Which BASE scale the returned ``score`` values are on (#1767):
+    # "rerank" (cross-encoder output — range is model-dependent, see
+    # ``reranker_model``), "rrf" (reciprocal-rank fusion, bounded near
+    # ``legs/rrf_k``), "bm25"/"dense" (single-retriever passthrough:
+    # unfused BM25 / cosine scores), "none" (filter-only enumeration —
+    # no relevance scale, the filter is the selector), or None (no ranked
+    # results produced). Base means pre-modifier: decay / access /
+    # importance boosts (Stages 4/6/7, all default-off) multiply on top
+    # when enabled, so absolute thresholds are only portable across
+    # servers with the same modifier config. Unlike ``rerank_applied``
+    # (the pre-Stage-3b decision), this is derived from the results
+    # actually returned, so a reranker that fails and silently falls back
+    # to the fused order keeps the label truthful.
+    score_scale: str | None = None
+    # Rerank model identifier, set only when ``score_scale == "rerank"`` —
+    # rerank ranges differ per provider (local cross-encoders emit raw
+    # logits, Cohere emits [0, 1] relevance), so clients calibrating a
+    # threshold need the model, not just the scale family.
+    reranker_model: str | None = None
 
 
 if TYPE_CHECKING:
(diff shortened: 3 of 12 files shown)


tsdata (approved):

pandas-studio:
## Review — PR #1781 `feat(search): name the score scale in structured search output`

**Verdict: SHIP** (nits only, none blocking)

### Overview

Adds `RetrievalStats.score_scale` / `reranker_model`, recorded at the pipeline branch that produced the scores, and surfaces them as top-level `score_scale` / `reranker` keys in structured `mem_search` / `mem_agent_search` payloads and as a per-item `score_scale` in `mm search --format json`. Additive only; per-result shape, ranking, and defaults unchanged.

### Correctness — verified against source

- **Result-derived `"rerank"` label is the right call.** All three shipped rerankers (`local.py:82`, `fastembed.py:144`, `cohere.py:77`) re-stamp every returned item `source="reranked"` on success and return the input *unchanged* on their internal error fallbacks (`local.py:77`, `fastembed.py:140`, `cohere.py:66`) — so a decision-flag label would lie exactly on those silent fallbacks, and the `all(r.source == "reranked")` check cannot false-positive (RRF stamps leg labels via `_source_label`, passthrough keeps retriever labels).
- **No interference from later stages.** MMR re-stamps `source="mmr"` but runs at Stage 5, after the detection point; scores are preserved. Context-window expansion attaches neighbors via `ContextInfo` rather than as new scored items, so the per-item CLI claim stays truthful.
- **Cache claim holds.** Stats ride the `_search_cache` tuple, and the cache key already includes `apply_rerank` + `rerank_cfg`, so a `rerank=False` (#1766) call gets its own slot and a cache hit reports the scale of the cached scores. Pinned by `test_cache_hit_preserves_scale`.
- **Coverage is complete.** Both `RetrievalStats` construction sites are handled (filter-only → `"none"`; main path → branch-assigned). Empty-result structured paths (`server/tools/search.py:243`, `multi_agent.py:218`) omit the keys, matching the tests. The only other structured formatter (`_format_recall_structured`) carries no scores — no gap.

### Tests

Strong: 10 pipeline-branch cases including both silent-fallback shapes, the per-call bypass, cache hit, and the modifier-enabled filter-only case that pins `"none"` ≠ "constant 1.0". Formatter emission/omission, tool-level parity for both tools, and CLI e2e are all covered.

### Nits (non-blocking)

1. `RetrievalStats.score_scale` docstring says `None` means "no ranked results produced", but the passthrough branches set `"bm25"`/`"dense"` even when the retriever returned zero hits (scale non-None, results empty). Not user-visible — formatters gate on empty results — but direct pipeline consumers reading the field contract could be misled. A one-word docstring tweak ("no ranked *path* taken") would close it.
2. `stats.reranker_model = rerank_cfg.model if rerank_cfg is not None else None` — dead-defensive: the enclosing block is gated on `apply_rerank`, which already requires `rerank_cfg is not None`. `rerank_cfg.model` alone would do.
3. The new CLI JSON assertion lives in the in-process `CliRunner` class only; the sibling subprocess class wasn't extended. Fine for a shape-level assertion given the subprocess class already covers the same journey.

### Conventions / security / performance

- `docs/guides/reference/core-memory-tools.md` is on the public doc surface (existing-doc edit — correct venue), and the doc text matches the implementation.
- Additive keys only; the rerank model ID is server config exposed to MCP clients in the same trust domain — no leak concern. The `all(...)` scan is O(top_k) — negligible.

pandas-studio:
# Review — PR #1781: name the score scale in structured search output

## Overview

Adds `RetrievalStats.score_scale` / `reranker_model`, recorded at the pipeline branch that actually produced the scores, and surfaces them as top-level `score_scale` / `reranker` keys in `mem_search` / `mem_agent_search` structured output plus a per-item `score_scale` in `mm search --format json`. Purely additive — no ranking, ordering, or default changes. Closes #1767.

**Verdict: LGTM with minor suggestions.** Correctness verified against the pipeline source; the design decisions are sound and well-tested.

## What I verified against source

- **Result-derived `"rerank"` label is the right call.** Both the pipeline's `except` fallback and all three provider implementations' internal error fallbacks silently keep the fused order; only a genuinely re-scored list is re-stamped `source="reranked"` (`reranker/local.py:82`, `reranker/fastembed.py:144`, `reranker/cohere.py:77`). A decision-flag-based label would lie exactly in those cases — and both silent-fallback modes are pinned by tests.
- **Cache behavior is real, not assumed.** `_search_cache` stores `(ts, ver, results, stats)` and a hit returns the cached stats object (`pipeline.py:937-939`), so cache hits report the scale of the cached scores. `test_cache_hit_preserves_scale` pins it.
- **Label placement is safe w.r.t. later stages.** The label is assigned at Stage 3/3b, before filters/decay/MMR/boosts — consistent with the documented "base scale, modifiers multiply on top" contract. The rescue re-stamp at the boundary uses `dataclass_replace` and doesn't disturb `source`.
- **Empty-result omission is consistent across all four formatter call sites** (both tools' empty branches pass no scale; non-empty branches thread `stats` through).
- **The added `rerank_cfg is not None` narrowing** mirrors the existing pool-widening site (`pipeline.py:1115`) — stylistically consistent, no behavior change.

## Suggestions (minor)

1. **CLI JSON omits the `reranker` model ID.** The PR's own rationale for the MCP payload is that rerank ranges are model-dependent, so the scale alone isn't enough to calibrate a threshold — but `mm search --format json` carries only `score_scale`. A CLI consumer on the `"rerank"` scale has the same problem. Consider adding a per-item (or documented-elsewhere) `reranker` key for parity, or note in the guide why the CLI surface intentionally omits it.
2. **Scale vocabulary lives in free strings across five files** (dataclass comment, two docstrings, formatter, guide, tests). A `Literal["rerank", "rrf", "bm25", "dense", "none"]` alias on `RetrievalStats.score_scale` would give mypy a single source for the vocabulary and catch a future misspelled label at the assignment site rather than in a consumer.
3. **Key/field naming drift:** the stats field is `reranker_model` but the payload key is `reranker`. Fine if intentional (payload brevity), but worth a one-line note in the formatter docstring so a future field addition doesn't "fix" the mismatch and break consumers.

## Non-issues considered

- `all(r.source == "reranked")` is an O(pool) scan on ≤ `max_pool` items — negligible.
- A future partially-re-stamped reranker output would fall back to the base-scale label — the conservative direction.
- Cache hits hand the same mutable `RetrievalStats` to multiple callers — pre-existing pattern, and nothing in the tool layer mutates the new fields.
- Stage-order invariant (CLAUDE.md) untouched; Stage comments updated where behavior-adjacent.

## Test coverage

Strong. `TestScoreScale` covers all five labels, both silent rerank-fallback modes, the per-call bypass (#1766), the cache hit, and the modifiers-on filter-only case that pins "none ≠ constant". Formatter emission/omission, tool-level integration for both tools (including empty-payload omission), and a CLI e2e that ties the `"rrf"` label to the observable 1/(k+1) score value. No gaps I'd block on.

## Security / performance

No new input surfaces, no secrets in output, additive JSON keys only. Performance impact is a few string assignments and one small list scan per search.

pandas-studio:
## Review — score scale naming in structured output (#1767)

### Overview

Adds `RetrievalStats.score_scale` (+ `reranker_model`), recorded at the pipeline branch that produced the scores, and surfaces both as top-level keys in the `mem_search` / `mem_agent_search` structured payloads and per-item in `mm search --format json`. Additive only; no ranking or default changes. Docs (tool docstring + `core-memory-tools.md`) updated in the same PR.

### Correctness — verified against source

- **Results-derived `"rerank"` label is the right call.** All three providers (`local.py`, `fastembed.py`, `cohere.py`) rebuild every returned item with `source="reranked"` on success, and all three silently return their input (no re-stamp) on internal errors — so `all(r.source == "reranked")` distinguishes a real re-score from both the pipeline `except` fallback and the providers' internal fallbacks. A decision-flag label (`rerank_applied`) would lie in exactly those cases, and the test suite pins both fallback flavors.
- **MMR interaction is safe.** MMR re-stamps `source="mmr"` (`mmr.py:59,114`), but the label is captured immediately after Stage 3b, before MMR runs — and MMR preserves scores, so the label stays truthful for the returned values.
- **Modifier stages are correctly framed as "on top of base scale".** `decay`/`access`/`importance` multiply scores but preserve `source`; the docs say thresholds are only portable across servers with the same modifier config, which matches the code.
- **`rerank_cfg is not None` narrowing is behavior-neutral.** `apply_rerank` (pipeline.py:908) is defined as `reranker is not None and rerank_cfg is not None and rerank is not False`, so the added conjunct can't change which calls rerank.
- **Cache path holds.** `_search_cache` stores `(ts, ver, results, stats)`, so hits report the scale of the cached scores; pinned by `test_cache_hit_preserves_scale`.
- **Empty-result omission is consistent** across all three surfaces (MCP structured empty branch never passes the kwargs; formatter omits `None`; CLI emits a bare list).
- All `RetrievalStats` construction sites are covered: main path (branch-labeled), filter-only (`"none"`), early-empty return (`None` → keys omitted).

### Test coverage

Strong. The 10-case `TestScoreScale` covers every label branch including both silent-fallback shapes, per-call bypass (#1766), cache hit, and filter-only under enabled decay (the "none ≠ constant" contract from review round 1). Formatter emission/omission, tool-level parity (`mem_search` + `mem_agent_search`), and CLI e2e (with the nice touch of asserting the RRF label via the observable `1/(rrf_k+1)` score) are all pinned.

### Minor / non-blocking

- `_format_structured_results` types the new params as `str | None` rather than the pipeline's `ScoreScale` Literal. Keeping the server layer decoupled from a pipeline-internal type alias is defensible; just noting the Literal's assignment-site protection doesn't extend to the formatter boundary.
- CLI JSON repeats identical `score_scale`/`reranker` values on every item. The bare-list payload shape forces this and the code comment says so — fine, at the cost of a few bytes.
- `reranker` can technically be passed to the formatter without `score_scale`; both callers source from `stats` where `reranker_model` is only set alongside `"rerank"`, so no guard needed today.

### Conventions

- Pipeline stage order untouched; no new registry tools; public-doc edit (`core-memory-tools.md`) is one of the 17 public docs, edited in place as policy requires; line length fine.

**Verdict: LGTM.** Correctness claims in the PR description all check out against the source; no blocking findings.

PR #1789, opened 2026-07-16
fix: empty indexing.memory_dirs is a state, not a crash (#1768)

pandas-studio:
## Summary

`indexing.memory_dirs: []` passes config validation but crashed every path that
reads `memory_dirs[0]` unconditionally — over MCP the client saw only
`Error: internal error (IndexError: list index out of range)` (#1768).

Empty `memory_dirs` stays **valid** — it's a deliberate "index nothing" state
(the issue's own use case: keeping a throwaway server's watcher off live
memory dirs while benchmarking). Instead of rejecting it at validation:

- **Read surfaces degrade gracefully.** `PinnedContextStore` now stores
  `user_base: Path | None`; `list`/`get`/`_read_scope`/`search_exclusion_roots`
  skip the user tier when it's unconfigured, so `mem_context_compose` (the
  issue's verified crash) returns a normal bundle with no pinned blocks.
  Project-tier pinned blocks keep working.
- **Writes that need the user-tier base refuse with a named field.** New
  `memory_scope.require_user_base()` raises
  `ConfigError("indexing.memory_dirs is empty — …")`, used by every user-tier
  write-target resolution: `PinnedContextStore._base` (chokepoint for
  `set`/`delete`), the LangGraph store's default root, `mm review approve`'s
  dated file, and the default scratch-promotion target. Project-tier writes no
  longer evaluate `memory_dirs[0]` at all (it was passed as an unused
  `user_base` argument).
- **Every silent fallback is gone.** `_mem_add_core` / `mem_batch_add` /
  `_validate_path` fell back to `Path(".")` (writes under the server's cwd);
  web `/api/add`, `mm mem add`, and `mm context memory-migrate` fell back to
  the historical `~/.memtomem/memories` literal — the last one can **move**
  files with `--apply`. All now surface the same field-naming error before
  touching the filesystem.
- **Bespoke empty-list messages centralized.** `mem_fetch`, the Notion /
  Obsidian importers, `mm agent share-note`, the LangGraph tool wrapper, and
  the interactive shell now emit the shared `EMPTY_MEMORY_DIRS_ERROR` (each in
  its surface-native shape) instead of five divergent wordings that described
  the state as uninitialized.
- **Every surface renders it cleanly.** MCP already knows `ConfigError`
  (`_KNOWN_EXCEPTIONS`); the CLI routes it through `raise_cli_error` on the
  pinned/review/mem-add write commands (hint reworded — the old one claimed
  the config "failed to load", wrong for a loaded-but-unusable value); the web
  app gains a `ConfigError` handler → **409** with a path-redacted,
  field-naming detail (matching `deps.require_configured`'s convention for
  unusable-config states, vs the opaque generic 500).

## Review

Four `codex` gates. Design gate (report-only, pre-implementation): NEEDS-FIX —
caught the `_mem_add_core`/`mem_batch_add` cwd fallbacks, project-tier writes
needlessly indexing `memory_dirs[0]`, `delete` belonging on the error side (a
mutating op reporting "not found" would be a lie), the 409 convention, and the
stale CLI hint. Diff review round 2: NEEDS-FIX — the three
historical-literal fallbacks (web `/api/add`, `mm mem add`, memory-migrate)
contradicted the new contract, plus message centralization. Round 3:
NEEDS-FIX — default scratch promotion's `bases[0]` (a derived-variable
crash the plain `memory_dirs[0]` grep missed) and the shell messages. Final
convergence re-gate: **SHIP** — repository-wide scan of `memory_dirs[0]` /
`mdirs[0]` / `bases[0]` / `user_bases[0]` shows every remaining access
guarded.

## Testing

- `uv run pytest -m "not ollama"` — full suite.
- New `tests/test_empty_memory_dirs.py`: MCP compose / pinned_set / mem_add /
  mem_batch_add (including "nothing written under cwd" pins), `mm mem add`
  and memory-migrate refusals (no legacy-directory writes/moves, source file
  untouched), LangGraph default-root vs project-tier, `mm review approve`
  clean error with claim rollback.
- Pinned-store unit cases in `test_pinned_context.py` (graceful reads,
  ConfigError writes, project tier, exclusion roots), `_validate_path`
  empty-dirs case in `test_memory_write_scope.py`, and web 409 pins in
  `test_web_routes.py` (handler contract, `/api/add`, scratch promote —
  asserting no write and no promote mark).
- Real-flow smoke against an isolated `HOME` with the issue's exact
  `config.json`: the reported constructor repro passes, `mm pinned compose`
  exits 0 with a normal bundle, `mm pinned set` prints the one-line error +
  hint (no traceback).

Closes #1768

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -353,6 +353,16 @@ running).
 > behavior (config-write without the seed scan) — useful for staging a
 > large folder before a controlled `mm index` run.
 
+An **empty list** (`"memory_dirs": []`, set by editing `config.json` or
+via the env override) is a valid "index nothing" state — useful for
+keeping a throwaway server's file watcher off your live memory
+directories. Search and project-tier operations keep working, and read
+surfaces like `mem_context_compose` simply return no user-tier pinned
+blocks. Writes that need the user-tier base directory (`mem_add`
+without a project scope, `mem_pinned_set`, approving a review
+candidate) refuse with a configuration error naming
+`indexing.memory_dirs` instead of writing anywhere else.
+
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `MEMTOMEM_INDEXING__MEMORY_DIRS` | `["~/.memtomem/memories"]` (+ provider folders selected in `mm init`) | Directories watched for reactive re-index (see above) |
--- a/packages/memtomem/src/memtomem/cli/_errors.py
+++ b/packages/memtomem/src/memtomem/cli/_errors.py
@@ -84,8 +84,8 @@ def _hint_for(e: Exception) -> str | None:
         )
     if isinstance(e, ConfigError):
         return (
-            "configuration failed to load — inspect it with `mm config show`, check "
-            "~/.memtomem/config.json permissions, or re-run `mm init`."
+            "check the active configuration with `mm config show`, then edit "
+            "~/.memtomem/config.json or re-run `mm init`."
         )
     if isinstance(e, StorageError):
         return "storage backend error — run `mm status` to check the DB path and health."
--- a/packages/memtomem/src/memtomem/cli/agent_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/agent_cmd.py
@@ -297,12 +297,16 @@ async def _run_share(chunk_id: str, target: str, force_unsafe: bool = False) -> 
             else "Shared: memory"
         )
 
-        if not comp.config.indexing.memory_dirs:
-            raise click.ClickException("No memory directories configured. Run `mm init` first.")
         from datetime import datetime, timezone
-        from pathlib import Path
 
-        base = Path(comp.config.indexing.memory_dirs[0]).expanduser().resolve()
+        from memtomem.cli._errors import raise_cli_error
+        from memtomem.errors import ConfigError
+        from memtomem.memory_scope import require_user_base
+
+        try:
+            base = require_user_base(comp.config.indexing.memory_dirs)
+        except ConfigError as exc:
+            raise_cli_error(exc)
         date_str = datetime.now(timezone.utc).strftime("%Y-%m-%d")
         path = base / f"{date_str}.md"
         append_entry(path, chunk.content, title=title, tags=tags)
--- a/packages/memtomem/src/memtomem/cli/context_cmd.py
+++ b/packages/memtomem/src/memtomem/cli/context_cmd.py
@@ -4943,10 +4943,12 @@ async def _memory_migrate_run(
         _lock_path_for,
         async_file_lock,
     )
+    from memtomem.errors import ConfigError
     from memtomem.memory_scope import (
         MemoryScopeError,
         is_project_tier_registered,
         project_tier_registration_error,
+        require_user_base,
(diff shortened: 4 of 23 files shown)


tsdata (approved):

pandas-studio:
## Review — SHIP

Empty `indexing.memory_dirs` is now treated as a deliberate "index nothing" state rather than an `IndexError` landmine. The design split is exactly right: **reads degrade gracefully, writes refuse by name.** I traced every guarded access and the contract holds.

### What's strong
- **`require_user_base()` as the single chokepoint** for user-tier write targets. Centralizing `memory_dirs[0]` behind one function that raises a named `ConfigError` is the correct fix — the alternative (validation-time rejection) would have broken the issue's actual use case.
- **`PinnedContextStore.user_base: Path | None`** cleanly separates the two tiers. `_read_scope`/`get`/`list` short-circuit on `None`; `search_exclusion_roots` filters the user scope out of the tuple so it never hits the raising `_base("user")`. Project tier is fully independent. Verified against the tests in `test_pinned_context.py`.
- **Killing the silent fallbacks is the highest-value part.** `Path(".")` (cwd writes) and the `~/.memtomem/memories` literal (which `memory-migrate --apply` could *move files into*) were genuine data-placement hazards, not just crashes. Good instinct to gate those before any filesystem touch.
- **Claim rollback survives the refusal.** In `review_cmd._decide`, both `require_user_base` (L178) and the pinned `.set()` (L161) sit inside the `try` whose `except Exception → release_memory_candidate; raise` runs — the ConfigError path rolls back the claim, and the test pins `release_memory_candidate.assert_awaited_once()`. No dangling claims.
- **409 (not 500) for the web surface** matches `deps.require_configured`'s convention for loadable-but-unusable config, and the path-redaction regex is reused from the existing 400 handler. Consistent.
- Test coverage is thorough — MCP, CLI, LangGraph, web, and unit tiers each pin both the graceful-read and the refuse-without-writing behaviors (incl. "nothing under cwd" / "source untouched" assertions, which are the ones that actually catch a regressed fallback).

### Minor nits (non-blocking)
- **`shell.py` hand-rolls the guard** (`if not memory_dirs: … EMPTY_MEMORY_DIRS_ERROR` then `memory_dirs[0]`) instead of routing through `require_user_base`. Fine for `_cmd_index` (it indexes a dir, not a dated write, so `require_user_base` doesn't fit), but `_cmd_add` could have used the same chokepoint the rest of the PR standardizes on. Cosmetic divergence, not a bug.
- **`require_user_base` now `.resolve()`s** where `mm mem add`'s old `user_base = Path(mdirs[0]).expanduser()` did not. This is *more* correct and consistent with every other surface, but it's a behavior change worth being aware of (symlinked `memory_dirs` now resolve).
- **`import re` inside `config_error_handler`** duplicates the module's use of `re` in the sibling handler. Harmless; could hoist.

### Risk
Low. No correctness issues found. The `memory_dirs[0]` / `bases[0]` / `user_bases[0]` accesses are all now guarded (the final SHIP gate's repo-wide scan matches what I see in the diff), and non-empty behavior is preserved everywhere I checked (`scratch.py`'s `bases[0]` was identical to `memory_dirs[0]` resolved).

PR #1792, opened 2026-07-17
fix(indexing): cap embedding concurrency on the parallel index_path

pandas-studio:
## What

The bulk `index_path` run gated each file's **entire** `_index_file`
pipeline behind a single `Semaphore(8)`, so up to 8 `embed_texts` calls
ran concurrently. For the local CPU ONNX provider that multiplies peak
inference memory: the intra-op thread pool (`embedding.threads`) is the
real compute bottleneck, so the extra concurrency buys no throughput —
only RSS.

This adds a **dedicated embedding semaphore** sized from an optional
per-provider `preferred_concurrency` hint (`onnx=1`; `ollama/openai/noop=8`,
i.e. unchanged). Files still pipeline discovery / chunking / hash-diff /
DB work under the file-level semaphore; only the `embed_texts` call
itself queues.

The hint is a **documented optional duck-typed attribute** — the
`EmbeddingProvider` Protocol is left unchanged so existing fakes and
out-of-tree providers stay structurally conformant. `_resolve_embed_limit`
accepts only a real `int`, clamps to `[1, 8]`, and falls back to the
file-level cap for absent / malformed / Mock-valued hints (preserving
pre-change behavior).

Because cancelling a coroutine awaiting an inference **abandons** (does
not stop) the running `session.run` and frees the async slot immediately,
`OnnxEmbedder` enforces the cap in the worker layer with a **dedicated
single-worker `ThreadPoolExecutor`**: `max_workers=1` *is* the
serialization, and unlike the shared default executor it is
cancellation-aware — a queued inference whose awaiting task is cancelled
has its future cancelled before it starts and never runs, so cancelled
work cannot pile up as blocked threads in the process-wide pool. Accepted
trade-off: `embed_query` now waits behind an in-flight file's inference
rather than running as an extra concurrent `session.run`.

`close()` was hardened along the way (review rounds 4–5): a `_closed`
latch set under `_load_lock` makes **every** load path (inference and the
warmup load on the default executor) refuse to resurrect the model after
close; teardown runs in `_close_sync` off the loop, drains the inference
executor before clearing `_model` (preserving the #206 mmap-release
guarantee), and the await is `asyncio.shield`ed with settle-on-cancel
(same pattern as `server/warmup.py`) so a cancelled `close()` — even one
cancelled while `_close_sync` is still queued — cannot lose the teardown.

## Measured effect

Real bge-m3 (dim=1024) ONNX embedder, `index_path` concurrent path, peak
RSS via `ru_maxrss`:

| Corpus | Baseline (8× concurrent) | This PR (serialized) |
|---|---|---|
| 16 files × 48 chunks | **5.91 GB** | **2.32 GB** (2.5× lower) |
| 40 files × 4 chunks | 2.00 GB | 1.68 GB |

The multiplier scales with per-file batch size — exactly the
`8 × per-run activation peak` mechanism described in #1783. All files
index, no dimension/short-array errors, wall-clock unchanged (CPU-bound).

## Scope — why #1783 remains open

This covers the **concurrent** `index_path` (MCP `mem_index`, web
reindex, debounce drain). The `mm index <dir> --force` **CLI**
reproduction in the issue takes the **sequential** `index_path_stream`
path (`for fp in files:` — one embed at a time), which never ran
concurrent embeds. Its ~31 GB peak is dominated by a **single**
`session.run`'s onnxruntime BFC arena / sequence-length high-water mark
— the follow-up lever the issue tracks separately. Verified: on the real
auto-memory corpus the CLI peaked at ~30 GB on **both** `main` and this
branch. Filed as `Refs #1783`, issue left open with a root-cause comment.

## Tests

All concurrency pins are deterministic (event-driven, no timing windows)
and mutation-verified (each pin fails under the targeted regression):

- `TestEmbedConcurrency` (engine): hint honored (`embed peak == 1` while
  `files_inflight peak >= 2` — pins the *decouple* half), absent-hint
  overlap preserved, `0`-hint clamps without deadlock, bare-`AsyncMock`
  regression, direct `_resolve_embed_limit` unit pins.
- `test_embedding_providers` (`TestOnnxEmbedder`): per-provider hint
  values, side-effect-free hint read (no model load), abandoned-inference
  cap survives task cancellation, cancelled-while-queued inference never
  runs, close() drains a mid-construction loader (no resurrection),
  `_get_model` after close refuses, close() cancelled mid-drain and
  cancelled-while-queued both still complete teardown.
- Full suite `-m "not ollama"` green (8,795); ruff check + format clean;
  CI green on all 17 checks (3-OS tests + golden-path ONNX included).

### DIFF
--- a/packages/memtomem/src/memtomem/embedding/base.py
+++ b/packages/memtomem/src/memtomem/embedding/base.py
@@ -7,6 +7,19 @@ from typing import Protocol, Sequence
 
 
 class EmbeddingProvider(Protocol):
+    # Optional capability (deliberately NOT a required Protocol member, so
+    # existing duck-typed fakes and out-of-tree providers stay structurally
+    # conformant): providers MAY expose ``preferred_concurrency: int`` — the
+    # maximum number of concurrent ``embed_texts`` calls the provider wants
+    # across files during a bulk index run. The index engine clamps the value
+    # to its file-level concurrency cap and falls back to that cap when the
+    # attribute is absent or not a real ``int`` (see
+    # ``indexing.engine._resolve_embed_limit``). Reading the attribute must be
+    # side-effect-free — in particular it must not trigger a model load.
+    # Local CPU providers should advertise a low value (concurrent inference
+    # multiplies peak activation memory without adding throughput, #1783);
+    # remote latency-bound providers can match the file cap.
+
     @property
     def dimension(self) -> int: ...
 
--- a/packages/memtomem/src/memtomem/embedding/noop.py
+++ b/packages/memtomem/src/memtomem/embedding/noop.py
@@ -14,6 +14,11 @@ class NoopEmbedder:
     index engine skips vector storage entirely.
     """
 
+    # Inert — the engine skips embedding when ``dimension == 0`` — but kept
+    # equal to the engine's file-level cap so the hint contract (base.py)
+    # holds for every built-in provider.
+    preferred_concurrency = 8
+
     @property
     def dimension(self) -> int:
         return 0
--- a/packages/memtomem/src/memtomem/embedding/ollama.py
+++ b/packages/memtomem/src/memtomem/embedding/ollama.py
@@ -16,6 +16,12 @@ logger = logging.getLogger(__name__)
 
 
 class OllamaEmbedder:
+    # Remote HTTP embedding is latency-bound, so concurrent per-file
+    # ``embed_texts`` calls genuinely overlap; match the index engine's
+    # file-level cap to preserve pre-#1783 throughput. Intra-file batch
+    # fan-out is still self-limited by ``max_concurrent_batches`` below.
+    preferred_concurrency = 8
+
     def __init__(self, config: EmbeddingConfig) -> None:
         self._config = config
         self._client: httpx.AsyncClient | None = None
--- a/packages/memtomem/src/memtomem/embedding/onnx.py
+++ b/packages/memtomem/src/memtomem/embedding/onnx.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import logging
 import threading
 from collections.abc import Callable
@@ -272,14 +273,23 @@ class OnnxEmbedder:
         return embeddings[0]
 
     async def close(self) -> None:
-        # Teardown runs entirely in ``_close_sync`` on a worker thread, so the
-        # cleanup completes even if the awaiting coroutine is cancelled
-        # (cancelling ``run_in_executor`` abandons the await, not the thread) —
-        # a cancelled lifespan teardown still frees the model. Off the event
-        # loop also means the blocking ``_load_lock`` acquire / executor drain
-        # never stall it.
+        # Teardown runs entirely in ``_close_sync`` on a worker thread, off
+        # the event loop, so the blocking ``_load_lock`` acquire / executor
+        # drain never stall the loop. ``shield`` + settle-on-cancel (same
+        # pattern as ``server/warmup.py``) makes the teardown survive
+        # cancellation of the awaiting task in BOTH phases: a bare await
+        # would cancel a still-queued ``_close_sync`` before it starts
+        # (leaving the model and executor alive), and an already-running
+        # worker can't be interrupted anyway — either way, keep the future
+        # alive, wait for it to settle, then propagate the cancellation.
         loop = asyncio.get_running_loop()
-        await loop.run_in_executor(None, self._close_sync)
+        future = loop.run_in_executor(None, self._close_sync)
+        try:
+            await asyncio.shield(future)
+        except asyncio.CancelledError:
+            with contextlib.suppress(BaseException):
+                await future
+            raise
 
     def _close_sync(self) -> None:
         # Latch closed and drop the model under ``_load_lock`` — the same lock
--- a/packages/memtomem/src/memtomem/embedding/onnx.py
+++ b/packages/memtomem/src/memtomem/embedding/onnx.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
 import logging
 import threading
 from collections.abc import Callable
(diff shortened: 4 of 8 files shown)


tsdata (approved):

pandas-studio:
## Code review

### Overview

Adds a dedicated embedding semaphore to the bulk `index_path` run, sized from an optional duck-typed `preferred_concurrency` hint (`onnx=1`, others `=8`), plus a `threading.BoundedSemaphore` hard gate inside `OnnxEmbedder` so a cancelled-and-abandoned `asyncio.to_thread` inference cannot overlap a follow-up run. Files still pipeline chunking/hash/DB work; only `embed_texts` queues.

### Verdict: SHIP — design and tests are solid; notes below are minor.

**Verified against source (not just the diff):**
- `contextlib` already imported in `engine.py` (`nullcontext` OK, async CM on 3.12).
- Comment claims hold: `max_concurrent_batches` semaphores exist in ollama/openai; engine skips embedding at `dimension == 0` (`engine.py:1225`).
- `ruff check` + `format --check` green across src/tests/tools; `EmbeddingProvider` Protocol unchanged so existing fakes and out-of-tree providers stay structurally conformant.

**Strengths:**
- Two-layer defense is the right call: the asyncio semaphore is scheduling only; the in-worker `threading.BoundedSemaphore` is what actually survives task cancellation. The cancellation test proves this deterministically (instrumented gate counts attempts vs grants — no timing window, can't pass vacuously, `finally` prevents hangs).
- `_resolve_embed_limit` hardening (bool excluded, Mock auto-attribute rejected, clamp `[1, 8]`) with direct unit pins; the AsyncMock regression pin protects the existing test idiom in the repo.
- The engine test pins both halves of the contract: `peak == 1` for embeds AND `files_inflight peak >= 2`, so it can't pass by accidentally serializing whole files.
- Nested acquisition order is always file-sem → embed-sem with embed ≥ 1 — no deadlock; the `0`-hint clamp test covers the `Semaphore(0)` hazard.

**Minor notes (non-blocking):**
1. Single-file `index_file` calls (serialized under `_index_lock`) can still overlap a concurrent bulk run's embeds — bulk `_bounded` doesn't take `_index_lock` — so a remote provider can briefly see 8+1 concurrent `embed_texts`. For ONNX the `_infer_gate` hard cap makes this moot (the case that matters for memory). Not worth restructuring; maybe one line in the `embed_semaphore` comment if this area is touched again.
2. `embed_query` now queues behind an in-flight file's inference during a reindex — documented trade-off, bounded by one file's embed; fine. Worth remembering if search-latency-during-reindex complaints appear.
3. An abandoned worker's `on_progress` → `call_soon_threadsafe` on a torn-down loop is pre-existing behavior, unchanged here; the gate's context manager still releases on any raise.

**Scope:** correctly does NOT claim to close #1783 — the CLI sequential-path ~31 GB single-`session.run` peak is a separate lever; `Refs #1783` with the issue left open matches the measured both-branches-peak-~30 GB evidence.

pandas-studio:
## Review

### Overview

Splits embedding concurrency out of the file-level `Semaphore(8)` into a dedicated semaphore sized from an optional per-provider `preferred_concurrency` hint (ONNX=1; remote providers=8, unchanged), plus a single-worker `ThreadPoolExecutor` in `OnnxEmbedder` as the hard cap that survives cancel-abandon, with a `_closed`-latch + drain + shield rewrite of `close()`. 8 files, +611/−20. All 17 CI checks green (lint, 3-OS tests, golden-path ONNX included).

### Verified against the branch

- `contextlib` was already imported in `engine.py` (the `nullcontext()` call is safe). ✓
- `_loading` referenced by `test_preferred_concurrency_read_is_side_effect_free` exists (`onnx.py:73`). ✓
- `_closed` check sits inside `_load_lock`, consistent with the existing double-checked-lock pattern; the lock-free fast path only returns an already-built model, so no post-close resurrection. ✓
- `_close_sync` lock ordering: latch + null under the lock, release, then `shutdown(wait=True)` — no deadlock with a worker blocked in `_get_model` (it acquires the released lock, sees `_closed`, refuses). ✓
- `OnnxEmbedder` is constructed once via `factory.py`, so the per-instance executor cap is the process-wide cap. ✓

### Strengths

- Correct layering: the asyncio semaphore is the normal-path scheduler; `max_workers=1` is the hard guarantee that holds when a cancelled task abandons a running `session.run`. The comments explain why a threading semaphore was rejected (queued-work cancellation awareness).
- Protocol left unchanged — existing fakes and out-of-tree providers stay structurally conformant; `_resolve_embed_limit` handles `bool` exclusion, `[1, 8]` clamp, and Mock auto-attributes, with an `AsyncMock` regression pin.
- Deterministic, event-driven tests with no timing windows: abandoned-inference cap, queued-future cancellation, close-drains-mid-construction-loader, close-cancelled-while-queued. Pinning the *decouple* half (`files_inflight peak >= 2` while `embed peak == 1`) is a nice touch.
- Measured effect (5.91 GB → 2.32 GB) and the explicit scoping of why this does not close #1783 (sequential CLI path is a different lever) are both clear.

### Findings

1. **PR description is stale (fix before merge).** The body says the worker-side cap is a `threading.BoundedSemaphore(1)`, but the implementation is a dedicated single-worker executor — the code comments explicitly reject the semaphore approach. Looks like a leftover from an earlier iteration; please update the body.
2. **`embed_query` queues behind bulk indexing (accepted trade-off, noting for the record).** A search query issued mid-reindex waits for the in-flight file's inference — potentially seconds for a large file on CPU. If search-latency reports show up, this is the spot; a follow-up could give queries a priority lane.
3. **No cross-run cap (minor).** `embed_sem` is created per `_index_path_inner` call, so concurrent `index_path` runs don't share it. ONNX is still hard-capped by the executor; remote providers can reach runs × 8 — same as pre-change, so not a blocker, but worth a one-line comment.
4. **Nits.** The literal `8` in noop/ollama/openai can drift from `_FILE_CONCURRENCY`, but the clamp makes drift harmless and avoiding an embedding→indexing import justifies the literal. `embed_texts` after close wraps the executor's `RuntimeError` into `EmbeddingError` via the generic except — behavior is right but untested (nice-to-have pin). The queued-close test swaps the loop's default executor without restoring it — fine since anyio gives each test a fresh loop.

### Security / performance

No security surface touched (privacy gate / redaction paths untouched; the semaphore size never comes from external input). Remote providers keep 8 → no throughput change; ONNX is CPU-bound so wall-clock is unchanged while peak RSS drops 2.5× (measured).

### Verdict

**SHIP** — no correctness issues found. Update the stale `BoundedSemaphore` wording in the PR body before merge; findings 2–3 are follow-up considerations, not blockers.

PR #1793, opened 2026-07-17
fix(chunking): split oversized markdown parts

Sanjays2402:
Fixes #1785.

Refines oversized non-fence Markdown parts before packing them into chunks. Multiline parts split at line boundaries, prose falls back to sentence or word boundaries, and fenced code plus tables retain their existing atomic behavior.

Regression proof: the new bullet-list test fails on `main` and passes with this change; focused chunking and fence-protection tests pass.


pandas-studio (changes requested):
Thanks for the PR — and welcome! This is a well-aimed first contribution: the diagnosis matches #1785 exactly, the fix sits at the right place in the pipeline (refine before the size-based merge), and the regression proof holds up — I verified that the new test fails on `main` and passes on this branch, and the full chunking suite passes.

Two findings surfaced while exercising the change, one of which is a merge blocker. Both come with repro snippets below.

## 1. Blocker: sub-chunk `start_line`/`end_line` drift, including inverted out-of-bounds ranges

The merge loop in `_split_section` advances its line counter with `line_offset += part.count("\n") + 2` — the `+2` assumes consecutive parts were separated by a blank line in the source. That holds for paragraph parts, but the pieces `_split_oversized_part` produces are separated by a single `\n` (line mode) or nothing at all (sentence/word mode). Each refinement boundary therefore accumulates +1 (or +2) of drift in the recorded line ranges.

Repro:

```python
from pathlib import Path
from memtomem.chunking.markdown import MarkdownChunker

class Cfg:
    max_chunk_tokens = 50          # max_chars = 200
    min_chunk_tokens = 10
    chunk_overlap_tokens = 0
    paragraph_split_threshold = 30

content = "## Notes\n\n" + "\n".join(f"- item {i:02d} " + "w" * 30 for i in range(20))
chunks = MarkdownChunker(indexing_config=Cfg()).chunk_file(Path("/t.md"), content)
for c in chunks:
    print(c.metadata.start_line, c.metadata.end_line, repr(c.content.splitlines()[0][:12]))
```

Observed: the second chunk records L8–11 but its content (`item 04`…`item 07`) actually lives at L7–10; drift grows by +1 per boundary; the final chunk records **`start_line=23 > end_line=22`** — an inverted range pointing past the end of the 22-line file.

This matters because these fields are not display-only: `mem_edit` / `mem_remove` mutate the source file by exactly this range (`replace_lines(meta.source_file, meta.start_line, meta.end_line, …)` in `server/tools/memory_crud.py`). And blank-line-free bullet lists — the headline case this PR fixes — are precisely the auto-memory files those tools edit. On `main` these files produce one oversized chunk with *correct* line ranges, so as written the PR trades a size regression for a mutation-targeting regression.

Suggested direction: make the merge accounting separator-aware — e.g. have `_split_oversized_part` return pieces together with the number of source lines each piece + its separator consumes (blank-line-separated paragraph parts keep the current `+2`), or track offsets by locating parts in the original text. Related: pieces that get re-merged into one chunk are joined with `"\n\n"`, inserting a phantom blank line that isn't in the file — worth fixing in the same pass since it has the same root cause.

## 2. Fence guard only checks the first line — a mid-part fence gets sliced

`_split_oversized_part` bails when the part *starts* with a fence opener, but a part can contain a fence after a prose line. That shape reaches the new line-splitting whenever paragraph splitting is skipped (`est_tokens < paragraph_split_threshold` while the part still exceeds `max_chars` — with default config, sections between ~2048 and ~3200 chars):

```python
class Cfg:
    max_chunk_tokens = 50            # max_chars = 200
    min_chunk_tokens = 10
    chunk_overlap_tokens = 0
    paragraph_split_threshold = 200  # 800 chars: paragraph split skipped below this

code = "\n".join(f"    step_{i} = do_thing_number_{i}()" for i in range(12))
content = "## Code\n\nintro line without punctuation\n" + "```python\n" + code + "\n```"
```

Observed: three chunks, two of them with an odd number of ``` ``` ``` markers — the code block is cut mid-fence. On `main` the same section is emitted oversized but intact. A conservative one-line fix matches the existing accept-as-oversize convention: bail with `[text]` when *any* line matches `_FENCE_OPEN_RE`, not just the first.

## Minor

- `first_line = text.lstrip("\n").splitlines()[0] if text else ""` recomputes what `lines` already holds, and raises `IndexError` on a part made only of newlines (unreachable in practice, and `main` has the same latent pattern at the sentence-split guard — fine to just reuse `lines` here).
- The docstring says "non-fence part" but the function also keeps pipe-tables atomic — worth a mention.

## Tests

Please add assertions that each emitted chunk's recorded `start_line`/`end_line` actually match the file lines its content came from — that assertion would have caught finding 1. Coverage for the fence bail (finding 2) and the sentence/word fallback path would also be welcome; the current test only exercises line-mode splitting.

Out of scope for this PR (no action needed): the pre-existing line-drift on `main`'s sentence-split path — I'll file that separately — and the CJK token-ratio consideration already split out in #1785.

Happy to take another look once the accounting fix is in — the overall shape of the change is right.

Sanjays2402:
Addressed the review in `e04c854`:

- replaced inferred `+2` line accounting with exact source spans and original separators, so refined/re-merged chunks no longer gain phantom blank lines and their mutation ranges stay aligned with source lines;
- made fence/table-containing parts atomic regardless of which line opens the fence, removing the newline-only `splitlines()[0]` edge case and updating the docstring;
- added regressions for exact bullet-list source ranges/content, a mid-part fence, and sentence/word fallback spacing and line metadata.

Verification: all 25 markdown chunking tests pass; Ruff check and format pass for both changed files; the staged scope/security gate passes with exactly those two files. The full fork CI run is currently awaiting maintainer approval (GitHub created zero jobs), rather than reporting a test failure.

Sanjays2402:

pandas-studio (approved):
Round 2 verified — approving. Thanks for the thorough follow-up in `e04c854`.

## What I checked (locally, isolated worktree)

- **Both round-1 findings are resolved.** Re-running the line-accounting repro from the previous review now yields exact ranges — no drift, no inverted ranges — and every chunk's recorded `start_line`/`end_line` matches the source lines its content came from, with overlap both off and on. The mid-part fence repro now stays atomic (one chunk, balanced fence markers).
- **Regression proof still holds**: the new bullet-list test fails against `main`'s `markdown.py` and passes on this branch.
- **`ruff check` + `ruff format --check` clean**; the chunking/edit/index-related suite passes (798 tests).
- **All four requested test additions are in** (per-chunk line assertions, fence bail, sentence and word fallback paths).

The exact-source-span approach (`_TextSpan` + verbatim separators from the original text) is a nicer resolution than the separator-counting direction I sketched — it also fixes the phantom `\n\n` join noted in round 1.

## Follow-up (non-blocking)

`should_refine` gates refinement on `"\n" in part.text`, so an oversized **single-line** paragraph among multiple parts still escapes the cap:

```python
# max_chunk_tokens = 50 → max_chars = 200
content = "## X\n\npara one small\n\n" + "Sentence one is here. " * 30 + "\n\npara three small"
# → middle chunk is 659 chars, over the cap
```

That's `main`'s behavior too — no regression — but since `_split_oversized_part` already no-ops on in-cap parts, dropping the gate would close the residual gap of #1785. Happy to have that as a tiny follow-up PR (yours if you'd like it, otherwise we'll file it) — no need to grow this one.

## Housekeeping

CI required maintainer approval to run on a fork (approved now), and the CLA check last ran before your signature landed (re-triggered). Once both are green I'll update the branch and merge.

tsdata (approved):
LGTM

PR #1795, opened 2026-07-17
fix(indexing): avoid sibling re-embedding on memory CRUD

pandas-studio:
## Summary

- refresh start_line and end_line for hash-matched chunks with a metadata-only SQLite update
- switch MCP and Web chunk edit/delete rollback paths to incremental re-indexing
- preserve UUIDs, vectors, FTS rows, timestamps, and personalization for unchanged siblings
- clarify that force indexing is reserved for embedding changes or index recovery

## Verification

- full Python suite: 8830 passed, 11 skipped
- focused indexing, storage, CRUD concurrency, cross-process, and memory doctor suites passed
- Ruff check and format check passed across src, tests, and tools
- MyPy reports the same 18 gradual baseline errors as origin/main; no new errors

Closes #1788


tsdata (approved):

pandas-studio:
## Review

Verified against source (differ, sqlite schema, engine transaction scope, doctor/index paths) — not just the diff.

**What checks out:**

- `differ.py` reuses the stored UUID for hash-matched chunks while carrying the fresh parse's line metadata, so the `WHERE id=?` UPDATE in `update_chunk_line_ranges` hits the right rows.
- The two-phase negative-temp update is correct against `idx_chunks_unique_content` `(namespace, source_file, content_hash, start_line)`: `-rowid-1` is unique per row, can't collide with real (non-negative) line values, and handles the identical-content swap case the docstring describes.
- `engine.py` runs delete → line-range update → upsert inside one `storage.transaction()`, and the write connection is not in autocommit, so the transient negative values are never visible to readers.
- The doctor/docs change from `mm index <dir> --force` to plain `mm index <dir>` is accurate: `index_path_stream` has no file-level hash short-circuit — the diff is computed against DB chunk hashes, so a zero-chunk `db_coverage` file gets fully upserted by a plain index.
- Incremental rollback is safe: the diff is state-based, so it converges on the restored file regardless of where the forward pass failed; embedding failures return early without touching the DB at all.
- Test coverage pins the contract well: FTS/vec/`updated_at`/`access_count` invariance, duplicate-hash swap, no-sibling-embedding assertions, and `force=False` on both forward and rollback paths.

**Findings:**

1. **Heading renames no longer reach the index (regression on CRUD surfaces).**
   `content_hash` hashes the body only (`models.py`), and the markdown chunker keeps the heading line out of `content` — it lives only in `heading_hierarchy`. `mem_edit` documents a heading override (`## ` prefix, `memory_crud.py`). Rename a heading with an unchanged body and the chunk hash-matches → classified `unchanged` → only line ranges refresh. The `heading_hierarchy` column, the FTS row (which indexes heading terms), the embedding (`retrieval_content` prepends the hierarchy), and descendants' `parent_context` all stay stale: search by the new heading misses, search by the old heading still hits, DB-backed surfaces show the old title.
   This gap pre-exists on the watcher/plain-index path, but the CRUD paths were immune via `force=True` until this PR. Suggested fix: extend `get_chunk_hashes` to also return the stored hierarchy and have the differ promote hash-matched chunks with a hierarchy mismatch into `to_upsert` (targeted re-embed; the common body-edit savings are preserved). Folding the hierarchy into the hash itself would invalidate every stored hash and force a full re-embed on upgrade — avoid that. If out of scope here: add a failing/xfail regression test, note the limitation in the ADR-0005 amendment, and file a follow-up.

2. **Broken comment grammar in `mem_delete`** (`memory_crud.py` ~890): the edit left a double possessive — "without both, a locked CRUD span's the edit's incremental `index_file`". The preceding "a locked CRUD span's" needs rewording too.

3. **`--force` use-case list is incomplete.** The new docs/CLI help scope force to "embedding changes or index recovery", but the ADR-0011 PR-D note in `engine.py` (and issue #1788 itself) also lists scope/namespace re-resolution after project deregistration as a legitimate force case — and if finding 1 stands, heading-rename recovery becomes one as well. Worth reflecting in `organization-maintenance.md` and the `cli/indexing.py` help text.

**Verdict:** implementation, tests, and doc fanout are solid; finding 1 needs a decision before merge (fix in the differ here, or document + follow-up issue). Findings 2–3 are one-liners.

pandas-studio:
Addressed all three review findings in commit `708be97d`:

1. Heading-only changes now compare stored `heading_hierarchy` alongside the body hash. Hash matches with a hierarchy mismatch keep their chunk ID but are promoted to a targeted upsert, refreshing hierarchy/parent context, FTS terms, and embeddings. Duplicate-body matching reserves exact hierarchy matches before assigning renamed fallbacks, so IDs remain stable.
2. Reworded the broken `mem_delete` lock comment.
3. Expanded `--force` help/docs to include explicit scope/namespace re-resolution; documented that plain indexing now handles heading-only changes.

Regression coverage includes heading-only FTS + embedding refresh, stable IDs, and duplicate-body hierarchy matching.

Validation:
- `8834 passed, 11 skipped` (full Python suite)
- `320 passed` (post-adjustment indexing/storage/CRUD focused suite)
- Ruff + `git diff --check` passed
- mypy: no new errors; existing 18-error gradual baseline only

PR #1796, opened 2026-07-17
feat(compose): name the score scale in the composed bundle (#1791)

pandas-studio:
## Summary

#1781 added top-level `score_scale` + `reranker` to `mem_search` /
`mem_agent_search` structured payloads (closing #1767), but `context_compose`
— the surface memtomem-stm prefers whenever the negotiated schema is >= 2 —
still emitted bare `score`s. The consumer most likely to be scale-sensitive
was blind on its primary path while the fallback path reported the scale.

Mirror the fields additively on the composed bundle:

- **No new derivation.** Compose already calls the same `SearchPipeline.search()`
  and received the `RetrievalStats` carrying `score_scale` / `reranker_model`
  (all five labels and the results-derived `"rerank"` rule established by
  #1781) — it just discarded the tuple element. Thread it into `ContextBundle`.
- **Top-level placement**, matching the issue: `score_scale` plus `reranker`
  (rerank model id) on the bundle envelope; `retrieved` items and `pinned`
  blocks are untouched — pinned blocks carry no relevance scale.
- **Omission rule**: both keys are omitted when the final post-budget
  `retrieved` list is empty (budget truncation can empty it even when the
  pipeline returned hits; a query-less compose has no retrieval leg at all).
  Enforced structurally in `as_dict()`, not only on the assembler path.
- **Pair invariant at construction**: `ContextBundle.__post_init__` rejects
  inconsistent direct construction loudly in both directions (`reranker`
  without the `"rerank"` scale, `"rerank"` scale without a model id) —
  mirroring the invariant `RetrievalStats` already guarantees on the
  pipeline path.
- **`context_compose` capability bumps to `schema_version: 4`** (schemas 2/3
  stay immutable released contracts). Because the keys are omitted for empty
  results, clients cannot distinguish "server too old" from "no results" by
  key presence — the advertised version is the discovery mechanism, same
  rationale as the 2→3 bump (#1747). Both compose surfaces (MCP
  `mem_context_compose`, `mm pinned compose`) share `ContextBundle.as_dict()`,
  so one emission site covers both.
- `ScoreScale` is imported at runtime (not `TYPE_CHECKING`) so
  `typing.get_type_hints(ContextBundle)` resolves for reflection-based
  consumers; measured cold-import parity with `pinned.py`'s existing deps and
  no cycle, so no new types module.

No ranking or default changes; additive fields only.

## Review

Four `codex` gates. Design gate (report-only, pre-implementation): **SHIP** —
with the field-ordering nit (scale fields appended after every pre-existing
field so legacy positional construction keeps its meaning; regression-tested)
and the docs nit (preserve the schema 2→3 ladder, add 4 as a new rung) folded
in. Diff review round 1: NEEDS-FIX — `TYPE_CHECKING`-only `ScoreScale` broke
`get_type_hints(ContextBundle)`, and `as_dict()` didn't enforce the
empty-retrieved omission on direct construction. Round 2 (full re-gate):
NEEDS-FIX — a stamped `reranker` escaped with non-rerank scales. Round 3
(full re-gate): NEEDS-FIX — the inverse pair (`"rerank"` scale without a
model id) → superseded serialization-side normalization with the
`__post_init__` invariant; a second finding (tolerate `stats=None` via
`getattr`) was **declined**: `SearchPipeline.search()` types the stats as
non-Optional, `(results, None)` was never a valid pipeline contract (only
lazy test doubles, upgraded to real `RetrievalStats` per the design gate).
Round 4 (full re-gate, decline rationale re-examined): **SHIP**, zero
findings — "schema-4 metadata is correctly propagated, validated, documented,
and negotiated; the `stats=None` decline is sound."

## Testing

- `uv run pytest -m "not ollama"` — full suite, 8790 passed (run again after
  the `__post_init__` behavior change).
- New cases in `test_pinned_context.py`: top-level emission (not per-item,
  not on pinned), rerank scale + model id, omission without a query, omission
  when budget truncation empties `retrieved`, structural omission and
  inconsistent-pair rejection on direct construction, runtime-resolvable
  annotations, and legacy positional-construction compatibility. Existing
  compose fakes upgraded from `(results, None)` to real `RetrievalStats`.
- `test_meta_tool.py` capability pin updated to `schema_version: 4`.
- Real-flow smoke on an isolated worktree + isolated `HOME` (fresh index,
  `mm pinned set`, `mm pinned compose`): query compose emits top-level
  `"score_scale": "rrf"` with no per-item/pinned leakage and no stray
  `reranker`; query-less compose omits both keys. Re-run at the final HEAD.

## Follow-up (separate repo, not this PR)

memtomem-stm consumption for its threshold-mismatch diagnostic
(memtomem-stm#727): the design gate enumerated STM's hard-coded schema-3
ceilings to lift when it adopts schema 4 — daemon adapter defaults
(`surfacing/daemon_adapter.py:67`, `:84`), the daemon server clamp
(`daemon/server.py:635`), and `ContextComposeResult` lacking scale fields
(`surfacing/mcp_client.py:89`).

Closes #1791

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### Added
 
+- **Context Compose schema 4** (#1791) — the composed bundle now names the
+  retrieval leg's score scale at the top level (`score_scale`: `rrf` / `bm25`
+  / `dense` / `none` / `rerank`, plus `reranker` with the model id when
+  reranked), mirroring the fields `mem_search` structured output gained in
+  #1781. Both keys are omitted when `retrieved` is empty, so clients should
+  gate on the advertised `context_compose` schema version, not key presence.
+  Schemas 2 and 3 remain immutable released contracts.
 - **Per-call rerank bypass** (#1766) — `mem_search` and `mem_context_compose`
   accept `rerank=false` (CLI: `mm search --no-rerank`, `mm pinned compose
   --no-rerank`) to skip the cross-encoder rerank stage and its candidate-pool
--- a/docs/guides/pinned-context.md
+++ b/docs/guides/pinned-context.md
@@ -34,6 +34,11 @@ never cut in the middle and omitted ids are returned explicitly.
 `mem_context_compose` schema 2 accepts the same optional `namespace` and
 `context_window` retrieval controls used by search. Schema 3 additionally
 returns adjacent chunks under each retrieved item's optional `context` object.
+Schema 4 additionally names the retrieval leg's score scale at the top level
+of the bundle: `score_scale` (`rrf`, `bm25`, `dense`, `none`, or `rerank` —
+the same labels `mem_search` reports) plus `reranker` with the rerank model id
+when the scale is `rerank`. Both keys are omitted when `retrieved` is empty;
+pinned blocks carry no relevance scale.
 Matched hits retain schema 2 budget priority; de-duplicated adjacent chunks use
 only the remaining composed bundle character budget, selected globally by
 distance with hit rank and before/after as deterministic tie-breakers. Pinned
@@ -46,6 +51,10 @@ Schema 3 is a new capability instead of an in-place schema 2 change because
 schema 2 shipped in memtomem 0.3.9 without adjacent chunks in its response.
 Keeping that released contract immutable lets clients distinguish scoped
 composition from composition that also preserves visible context windows.
+Schema 4 follows the same rule: because `score_scale`/`reranker` are omitted
+for an empty `retrieved` list, a client cannot tell "server too old" from
+"no results" by key presence — the advertised schema version is the
+discovery mechanism.
 
 ## Review-first candidates
 
--- a/docs/guides/reference/operations.md
+++ b/docs/guides/reference/operations.md
@@ -173,7 +173,9 @@ a separate package; it communicates with memtomem core entirely through the MCP
 protocol — no direct code coupling. Enhanced composition is negotiated from
 `mem_do(action="version").capabilities`, not inferred from an installed package
 version: `context_compose` schema 2 enables scoped Pinned Context composition,
-schema 3 additionally preserves adjacent context-window chunks, and
+schema 3 additionally preserves adjacent context-window chunks, schema 4
+additionally names the retrieval leg's score scale (top-level
+`score_scale`/`reranker`, omitted when `retrieved` is empty), and
 `candidate_propose` schema 1 independently enables review-first proposals.
 
 ### How it works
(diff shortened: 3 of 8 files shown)


tsdata (approved):

pandas-studio:
## Review

### Overview

Mirrors the #1781 `score_scale` / `reranker` fields onto the composed bundle (`mem_context_compose` / `mm pinned compose`), additively: compose threads the `RetrievalStats` it already received (previously discarded) into `ContextBundle`, `as_dict()` enforces the empty-`retrieved` omission structurally, `__post_init__` rejects inconsistent scale/model pairs at construction, and the `context_compose` capability bumps to `schema_version: 4`. 8 files, +185/−11.

### Verified

- Single `schema_version` emission site for `context_compose` (`status_config.py`) — grepped the tree; the capability pin in `test_meta_tool.py` is updated to match. ✓
- CHANGELOG wording matches the source vocabulary (`rrf`/`bm25`/`dense`/`none`/`rerank` all stamped in `search/pipeline.py`; `reranker_model` documented as rerank-only). ✓
- Field ordering: the two new fields sit after every pre-existing field, and the legacy positional-construction test pins it. ✓
- CI green across the board at review time except `windows-test-shard (0)` still pending — wait for it before merge (required check).

### Strengths

- **No new derivation** — compose reuses the exact stats the pipeline already computed, so the label vocabulary and the results-derived `"rerank"` rule can't drift from `mem_search`'s.
- **Defense in depth is layered correctly**: the assembler stamps only when the post-budget `retrieved` is non-empty, `as_dict()` enforces the same rule structurally for direct construction, and `__post_init__` rejects inconsistent pairs loudly in both directions — mirroring the invariant `RetrievalStats` already guarantees on the pipeline path.
- The `stats=None` tolerance decline is sound: `SearchPipeline.search()` types stats as non-Optional, and upgrading the lazy test doubles to real `RetrievalStats` (rather than papering over with `getattr`) keeps the test doubles strict.
- Runtime (non-`TYPE_CHECKING`) `ScoreScale` import with a `get_type_hints` regression pin; the schema 2→3→4 immutable-ladder rationale is carried consistently through `pinned-context.md`, `operations.md`, and the CHANGELOG.
- Test coverage is thorough: top-level-not-per-item placement, rerank + model id, no-query omission, budget-truncation omission, structural omission on direct construction, pair rejection, positional compatibility.

### Findings

1. **CHANGELOG conflict with #1797 (process, not code).** Both PRs insert into the same `[Unreleased] → Added` spot; whichever merges second needs a rebase. Merge serially and re-check the strict up-to-date state — an auto-merge queued behind the other will silently sit BEHIND. Entry order after both land reads fine either way (the #1791 entry references #1781 by number, not by position).
2. **Nit — `as_dict()` reranker guard is redundant but harmless.** Given the `__post_init__` invariant, `self.reranker is not None` inside the `score_scale is not None` branch can only be true when the scale is `"rerank"`; the nested check reads as belt-and-braces rather than dead logic. Fine to keep.

### Security / performance

No new derivation, no ranking or default changes, no external input reaching the new fields; additive payload keys only. No perf impact (the stats tuple element was already computed and returned).

### Verdict

**SHIP** once `windows-test-shard (0)` completes green. Coordinate merge order with #1797 (finding 1); no code changes requested.

PR #1800, opened 2026-07-17
feat(search): persist durable Quality Lab observations

pandas-studio:
## Summary

- persist a durable UUID `query_run_id` for each ranked search invocation after the local SQLite commit succeeds
- store a retrieval profile/fingerprint, latency, cache/degradation state, and a content-minimized result snapshot
- expose committed run IDs through MCP structured output and the Web search API while preserving the existing CLI JSON result shape
- migrate pre-observation `query_history` tables additively without a schema-generation bump
- document the local data shape, privacy boundary, and existing 90-day history retention

## Behavioral boundary

- normal and zero-result ranked searches are observed by the SQLite capability
- SQLite cache hits receive distinct invocation IDs
- filter-only browsing is not observed
- observation persistence failure never fails search and returns no run ID
- snapshots omit result content and absolute source paths; heading hierarchy remains content-derived metadata and existing query history still contains query text
- alternate storage backends keep the legacy fire-and-forget history path, do not advertise a run ID, and preserve the prior behavior of not writing history on cache hits

## Commit-before-advertise performance tradeoff

The SQLite observation INSERT+COMMIT is awaited so a returned run ID always names a durable row. On a local file DB with WAL / `synchronous=NORMAL`, 60 cache-hit invocations measured:

- commit path: p50 `0.057ms`, p95 `0.071ms`
- same path with observation commit replaced by an async no-op: p50 `0.013ms`, p95 `0.014ms`
- observed delta: p50 `0.044ms`, p95 `0.057ms`

This intentionally spends a small amount of cache-hit latency for the durable-ID contract. The synchronous SQLite call still runs on the event-loop thread; if real workloads show materially higher storage latency, a dedicated serialized writer/ack path should be designed rather than moving the thread-affine connection through `to_thread`.

## Validation

- `8838 passed, 11 skipped` — full Python suite before rebasing onto current main
- `481 passed` — search/Web/MCP/schema/Compose boundary suite after rebase
- `606 passed` — post-review search/legacy-backend/schema regression suite
- Ruff and `git diff --check` pass
- mypy adds no errors; existing 18-error gradual baseline remains

Closes #1799
Refs #1798

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,17 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   #1781. Both keys are omitted when `retrieved` is empty, so clients should
   gate on the advertised `context_compose` schema version, not key presence.
   Schemas 2 and 3 remain immutable released contracts.
+
+- **Durable local search-run observations (Quality Lab foundation)** — ranked
+  searches now persist a unique `query_run_id`, retrieval profile, latency,
+  cache/degradation state, and a content-free result snapshot in SQLite. MCP
+  structured output and the Web search API expose the ID only after the local
+  commit succeeds; cache hits get distinct IDs, zero-result ranked searches are
+  recorded, filter-only browsing is excluded, and persistence failure never
+  interrupts search. Snapshots omit result content and absolute source paths;
+  existing query-history retention and CLI JSON output remain unchanged.
+  (#1799)
+
 - **Per-call rerank bypass** (#1766) — `mem_search` and `mem_context_compose`
   accept `rerank=false` (CLI: `mm search --no-rerank`, `mm pinned compose
   --no-rerank`) to skip the cross-encoder rerank stage and its candidate-pool
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 - **Durable local search-run observations (Quality Lab foundation)** — ranked
   searches now persist a unique `query_run_id`, retrieval profile, latency,
-  cache/degradation state, and a content-free result snapshot in SQLite. MCP
+  cache/degradation state, and a content-minimized result snapshot in SQLite. MCP
   structured output and the Web search API expose the ID only after the local
   commit succeeds; cache hits get distinct IDs, zero-result ranked searches are
   recorded, filter-only browsing is excluded, and persistence failure never
--- a/docs/guides/reference/core-memory-tools.md
+++ b/docs/guides/reference/core-memory-tools.md
@@ -227,6 +227,19 @@ mem_search(query="deploy pipeline", as_of="2025-Q3")    # historical query
 > are no results. `mm search --format json` carries the same values as
 > per-item `score_scale` / `reranker` keys.
 
+> **Quality Lab run ID**: every ranked search persisted by the local SQLite
+> backend receives a durable `query_run_id`. MCP structured output and the Web
+> search API expose it only after the observation commit succeeds, so it can be
+> used to attach later feedback without guessing which invocation produced a
+> result set. Cache hits receive distinct IDs. Filter-only browsing does not
+> create an observation, and an observation write failure never fails search.
+> The local snapshot stores ranks, scores, chunk IDs, content hashes, heading
+> hierarchy, namespaces, languages, and source **basenames**—not result content
+> or absolute paths. Existing query history still records the query text and is
+> pruned by the existing 90-day history policy. `mm search --format json` keeps
+> its current per-result schema; use MCP structured output or the Web API when
+> the run ID is needed.
+
 ### Tuning search weights
 
 Use `bm25_weight` and `dense_weight` to shift between keyword and semantic matching:
--- a/packages/memtomem/src/memtomem/cli/search.py
+++ b/packages/memtomem/src/memtomem/cli/search.py
@@ -134,6 +134,7 @@ async def _search(
             scope=scope,
             project_context_root=project_context_root,
             rerank=rerank,
+            origin="cli",
         )
 
     if not results and fmt in ("table", "plain"):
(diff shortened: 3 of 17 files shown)


tsdata (approved):

pandas-studio:
## Review — feat(search): persist durable Quality Lab observations

### Overview

Ranked searches now persist a UUID `query_run_id` plus a retrieval profile/fingerprint, latency, cache/degradation state, and a content-minimized result snapshot into `query_history` (additive columns, no schema-generation bump). The ID is advertised through MCP structured output and the Web API only after the local SQLite commit succeeds; CLI JSON shape is unchanged. Cache hits get distinct IDs, zero-result ranked searches are recorded, filter-only browsing is excluded, and persistence failure degrades to "no run ID" without failing search.

### What's good

- **Commit-before-advertise contract** is cleanly enforced: the run ID only appears on `RetrievalStats` after `save_search_observation` returns, and the failure-isolation test (`test_observation_failure_never_fails_search`) pins it.
- **Cache hygiene**: `dataclass_replace(stats, query_run_id=None, cache_hit=False, latency_ms=None)` before caching prevents a stale run ID from leaking to later cache hits — easy to get wrong, done right here.
- **Additive migration** with a real pre-observation-table test (legacy row survives, defaults land, schema version stamped), plus a partial unique index on `run_id` that tolerates legacy NULL rows.
- **Test coverage is strong**: zero-result, cache-hit distinctness (order-independent set comparison — avoids the same-second `created_at` tie trap), failure isolation, filter-only exclusion, MCP structured exposure, Web route threading, storage round-trip.
- `original_query` captured before any downstream query mutation; docs/CHANGELOG land in the same PR and match the code.

### Findings

**1. Observation write is now awaited on the hot search path — including cache hits** (perf, medium)

`save_search_observation` runs synchronous `db.execute` + `db.commit` on the event loop, and the pipeline now awaits it inline (previously history was fire-and-forget off the response path). WAL + `synchronous=NORMAL` keeps a single INSERT cheap, but:

- Cache hits went from **zero I/O** to one INSERT+COMMIT per hit — the cache's latency benefit is partially spent on telemetry. For agents calling `mem_search` in tight loops this is the common path.
- The write blocks the event loop, so concurrent Web/MCP searches serialize behind it.

The awaiting is forced by the commit-before-advertise design, so this may be an accepted tradeoff — but it deserves a measurement (cache-hit p50 before/after) and a note in the PR body. `asyncio.to_thread` isn't a drop-in fix since `self._db` is a single write connection (thread affinity), so if the numbers are fine, documenting the tradeoff is enough.

**2. Legacy-backend behavior quietly changed on cache hits** (correctness of stated contract, low-medium)

For backends without `save_search_observation`, `_record_ranked_search` fires the legacy `save_query_history` fallback **on cache hits too** (with an empty embedding). Previously cache hits wrote no history at all. The PR body says alternate backends "keep the legacy fire-and-forget history path", which is subtly inaccurate — the path is kept but now fires in a new situation. Either skip the legacy fallback when `stats.cache_hit` is true, or update the PR/CHANGELOG wording to state that cache hits now also produce legacy history rows. There's also no test for the legacy-backend branch (saver is `None` → no run ID, history still saved) — worth one since it's a distinct code path with its own closure.

**3. `save_search_observation` is absent from the `MemoryStorage` protocol** (design, low)

The pipeline duck-types via `getattr(type(self._storage), ...)` for detection but calls through `getattr(self._storage, "save_search_observation")(...)` — mypy can't see the call signature, and the class-vs-instance asymmetry means a backend that attaches the method per-instance is silently treated as legacy. Consider declaring it on the protocol (optional capability) or at least making detection and dispatch use the same lookup. `getattr` with a constant string on the call site can just be a direct attribute access once detection has passed.

**4. Startup runs three failing ALTER TABLEs forever** (nit)

Every `create_tables` on an already-migrated DB executes three ALTERs that raise and get string-matched on `"duplicate column"`. Harmless and the message is stable in practice, but a single `PRAGMA table_info(query_history)` gate would avoid relying on error-message text.

**5. Privacy: `heading_hierarchy` is content-derived** (note, no action needed)

The snapshot omits content and absolute paths, but heading hierarchies often carry topic-sensitive titles. This is local-only SQLite and the docs disclose exactly what's stored, so it's fine — just noting it's not strictly "content-free" in the way "basenames only" is.

**6. History-consumer side effects** (note)

Cache hits inserting rows means `mem_search_history` and `suggest_queries` now surface duplicate entries for repeated queries, and history volume grows faster (prune interval/90-day retention unchanged). This is the documented intent ("cache hits get distinct IDs"), just flagging the downstream UX. Relatedly, `get_query_history` orders by second-precision `created_at` only — a secondary `ORDER BY id DESC` would make same-second ordering deterministic now that rows are more likely to collide.

### Verdict

No blocking correctness bugs found; the observation envelope, migration, and exposure contract are all pinned by tests. Main pre-merge asks: decide/document the cache-hit write tradeoff (#1), and fix or re-word the legacy-backend cache-hit behavior plus add its missing test (#2).

pandas-studio:
Addressed the review asks in `1bd76d83` (rebased onto current main):

1. **Measured and documented the awaited commit tradeoff.** Local WAL / `synchronous=NORMAL` file-DB benchmark over 60 cache hits: commit path p50/p95 `0.057/0.071ms`; async no-op in place of the commit `0.013/0.014ms`; delta `0.044/0.057ms`. The PR body now records both the result and the event-loop/thread-affinity tradeoff.
2. **Preserved legacy cache-hit behavior.** Alternate backends still fire-and-forget history for non-cached searches, but cache hits again skip legacy history writes. Added a test covering the legacy branch, no public run ID, and exactly one history write across miss + hit.
3. **Made capability lookup/dispatch symmetric.** A real class method or explicitly attached instance method resolves to one bound callable used for dispatch; dynamic mock/proxy attributes do not fabricate support. Added an instance-capability regression test.
4. **Removed exception-driven migrations.** Startup now gates each additive ALTER through one `PRAGMA table_info(query_history)` snapshot.
5. Changed the changelog wording from `content-free` to `content-minimized`; the PR body explicitly notes heading hierarchy is content-derived metadata.
6. Added deterministic `ORDER BY created_at DESC, id DESC` and pinned same-second history ordering.

Validation after the changes:
- `606 passed` broad search/legacy/schema regression set
- `181 passed` after rebasing onto current main
- Ruff and `git diff --check` pass

PR #1805, opened 2026-07-17
fix(warmup): survive repeated cancellation while settling

pandas-studio:
## Summary

- shield every executor-future settle await so repeated task cancellation cannot cancel a queued model load
- preserve cancellation as the caller-visible outcome after the load settles
- add a deterministic single-worker regression for the double-cancel interleaving
- align the ONNX close comment with the now-hardened warmup path

Closes #1803

## Tests

- `uv run pytest packages/memtomem/tests/test_warmup.py -q`
- `uv run pytest packages/memtomem/tests/test_embedding_providers.py -q -k "close_double_cancel_still_tears_down"`
- `uv run ruff check packages/memtomem/src/memtomem/server/warmup.py packages/memtomem/src/memtomem/embedding/onnx.py packages/memtomem/tests/test_warmup.py`
- `uv run ruff format --check packages/memtomem/src/memtomem/server/warmup.py packages/memtomem/src/memtomem/embedding/onnx.py packages/memtomem/tests/test_warmup.py`
- `uv run mypy packages/memtomem/src/memtomem/server/warmup.py packages/memtomem/src/memtomem/embedding/onnx.py`

### DIFF
--- a/packages/memtomem/src/memtomem/embedding/onnx.py
+++ b/packages/memtomem/src/memtomem/embedding/onnx.py
@@ -280,9 +280,9 @@ class OnnxEmbedder:
         # still-queued ``_close_sync`` before it starts (which would leave
         # the model and executor alive; an already-running worker can't be
         # interrupted anyway). On cancellation we keep settling until the
-        # future is done, then propagate. NOTE: ``server/warmup.py`` uses a
-        # weaker settle (bare ``await future`` after the first cancel) that
-        # a second cancellation can still pierce — don't copy that shape.
+        # future is done, then propagate. ``server/warmup.py`` uses the same
+        # repeated-shield settlement for model loads; keep both paths aligned
+        # so neither loses queued executor work (#1803).
         loop = asyncio.get_running_loop()
         future = loop.run_in_executor(None, self._close_sync)
         cancelled = False
--- a/packages/memtomem/src/memtomem/server/warmup.py
+++ b/packages/memtomem/src/memtomem/server/warmup.py
@@ -60,17 +60,30 @@ async def _warm_one(
         return WarmupOutcome(component, provider, model, "already-loaded")
     loop = asyncio.get_running_loop()
     future = loop.run_in_executor(None, load_model)
-    try:
-        await asyncio.shield(future)
-    except asyncio.CancelledError:
-        # A worker thread can't be interrupted — cancelling only the
-        # awaiting task would leave the load running while shutdown
-        # closes components under it. Wait for the in-flight load to
-        # settle (``close()`` awaits this task after cancelling), then
-        # propagate the cancellation.
+    # A worker thread can't be interrupted — cancelling only the awaiting
+    # task would leave the load running while shutdown closes components
+    # under it. Every settle-await stays shielded so repeated cancellation
+    # cannot cancel a load that is still queued in the executor. Once the
+    # load settles, preserve cancellation as the caller-visible outcome.
+    cancelled = False
+    while True:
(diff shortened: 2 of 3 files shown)


pandas-studio:
## Review

### Overview

Fixes #1803. Previously `_warm_one` shielded only the *first* await; after catching the first `CancelledError` it settled with a bare `await future`, which a second cancellation could pierce — cancelling a load still *queued* in the executor (a not-yet-started inner future is cancellable), so shutdown could proceed believing the load had settled while the model/executor state was lost. The fix loops over shielded awaits until the executor future actually settles, latches `cancelled`, and re-raises `CancelledError` afterward — the same repeated-shield shape `OnnxEmbedder.close()` already uses. The onnx comment is updated from "don't copy that shape" to "both paths aligned", and a deterministic single-worker regression pins the double-cancel interleaving.

### Correctness — verified

- Every path into the `if cancelled:` block has `future.done()` true (shield success, explicit `future.done()` check, or the load's exception propagating through the shield), so the non-blocking `future.result()` can never raise `InvalidStateError`.
- Cancellation-wins contract preserved: a load failure racing a cancel is consumed (`suppress(BaseException)`) and `CancelledError` is raised, matching the old behavior; a load failure *without* cancellation still propagates via the `BaseException` branch.
- The regression test is genuinely deterministic: `task.cancel()` synchronously cancels the shield's outer future and queues the task wakeup *before* the subsequent `sleep(0)`'s own wakeup, so each single `sleep(0)` reliably advances the task to its next suspension. `assert load_calls == []` after the second cancel pins exactly the #1803 failure mode.
- Ran locally on the PR head: `test_warmup.py` 17 passed; `test_embedding_providers.py -k close_double_cancel_still_tears_down` 1 passed.

### Minor points (non-blocking)

1. **Test leaves the loop's default executor replaced and shut down.** `loop.set_default_executor(small)` is never restored, and `small.shutdown(wait=False)` runs in `finally`. Safe today because pytest-asyncio's function-scoped loop is discarded per test, but if this file ever moves to a module/session loop scope, any later `run_in_executor(None, …)` on that loop dies on a shut-down executor. Cheap hardening: swap in a fresh/original executor in the `finally` block.
2. **Fresh `CancelledError` drops the cancel message.** `raise asyncio.CancelledError` discards any `task.cancel(msg)` message and the original traceback context that the old bare `raise` preserved. `close()` in onnx.py does the same, so the paths are consistent — but capturing the first caught `CancelledError` and re-raising it would keep fidelity in both.
3. **Comment slightly oversells symmetry.** The onnx comment now says warmup uses "the same repeated-shield settlement", but the onnx loop has no `except BaseException` arm — if `_close_sync` raised after a cancel there, the exception (not `CancelledError`) would propagate, unlike warmup's cancellation-wins handling. Pre-existing and out of scope here; just noting the two shapes differ in that one respect in case a future sweep literalizes "aligned".
4. `except BaseException` with `cancelled=True` also converts a hypothetical `KeyboardInterrupt`/`SystemExit` from the load into `CancelledError`. Vanishingly unlikely inside a task and consistent with the pre-existing `suppress(BaseException)` shape — fine to keep.

### Conventions / coverage

- Comment density and style match the file; 100-col respected; test slots into the existing `TestShutdownThreadSettlement` class.
- PR is one focused change with a regression test and an accurate cross-reference update — matches the one-change-per-PR and comment-alignment conventions.

**Verdict: SHIP** — the fix is correct and well-pinned; all findings above are optional follow-ups.

tsdata (approved):

Issue #1567, opened 2026-07-03
config: config.json read-modify-write has no cross-process lock — concurrent writers silently lose updates


memtomem:
## Summary

Every writer of `~/.memtomem/config.json` follows read → merge in memory → `_atomic_write_json`. The atomic helper prevents torn/corrupt JSON, but **nothing locks the read→merge→write window across processes**, so two concurrent writers silently lose one side's update. The file stays valid JSON, so the loss produces no error — just a setting that quietly reverts.

## Where

Unguarded read-modify-write sites (all end in `_atomic_write_json` / an atomic write, none hold a lock):

- `packages/memtomem/src/memtomem/config.py:2114-2152` — `save_config_overrides` (also the Web UI `PATCH /api/config` path).
- `packages/memtomem/src/memtomem/cli/config_cmd.py:163-166` — `mm config unset`.
- `packages/memtomem/src/memtomem/cli/init_cmd.py:2154-2156, 2328-2331` — `_write_config_and_summary`.
- `packages/memtomem/src/memtomem/config.py:1865-1890` — `_persist_auto_discover_migration`.

Locks exist for the DB and pid files (`portalocker` in `cli/_liveness.py`, `cli/web.py`) but not for config writes.

## Failure scenario

The Web UI saves a settings change via `PATCH /api/config` (→ `save_config_overrides`) at the same moment the user runs `mm config set search.default_top_k 20` in a terminal. Both processes read the pre-change file, each merges only its own delta, and whichever `os.replace`s second discards the other's change. No corruption, no error — one of the two changes just never happened.

## Suggested fix

A `portalocker` sidecar lock (e.g. `config.json.lock`) held across read→merge→write in a single shared helper, mirroring the existing pid-file lock pattern. All four sites should route through it so the lock discipline can't drift.

Also: the `_atomic_write_json` docstring (`config.py:2004-2014`) still says "Used by `mm config unset`. Follow-up work will migrate `save_config_overrides` ... onto this helper" — the migration already happened; worth fixing the stale comment in the same PR.

Found during the 2026-07-03 operational-stability audit.

PR #1809, opened 2026-07-17
fix(embedding): bound ONNX batch and sequence memory

pandas-studio:
## Summary

- cap local FastEmbed inference at onnx_batch_size=8 and max_sequence_tokens=1024 by default
- configure and verify the real FastEmbed tokenizer limit while keeping stored content and BM25 text complete
- apply runtime batch updates atomically across Web, MCP, and config hot reload
- fingerprint the vector-generation policy and enter safe degraded mode on mismatch: block index writes and fall back to BM25
- expose field-level settings controls, mismatch status, reset guidance, docs, and regression coverage

## Migration and safety

Pre-fingerprint ONNX indexes are conservatively treated as max_sequence_tokens=0. Resolve a mismatch with mm embedding-reset --mode apply-current, then run mm index --force for each memory directory (or mem_index with force=true). No automatic destructive re-index is performed.

## Verification

- ruff check: passed
- focused regression after latest main rebase: 946 passed
- golden/docs/schema checks: 82 passed
- real FastEmbed tokenizer cap contract: passed
- measured oversized-input peak RSS: about 5.02 GB

Closes #1786
Refs #1783

#1783 intentionally remains open: the roughly 30 GB sequential CLI path and broader ORT arena/sequence-length tuning are outside this PR and remain the next slice.


tsdata (approved):

pandas-studio:
## Review — fix(embedding): bound ONNX batch and sequence memory

### Overview

Bounds local FastEmbed/ONNX inference memory on two axes: an explicit `onnx_batch_size=8` passed to `model.embed(batch_size=...)` (replacing FastEmbed's implicit 256), and a real tokenizer-level `max_sequence_tokens=1024` cap applied via `enable_truncation`. A vector-generation policy fingerprint is stored in DB meta; on mismatch the system enters safe degraded mode (index writes blocked, dense retrieval falls back to BM25). Pre-fingerprint ONNX DBs are conservatively stamped as `max_sequence_tokens=0`, forcing an explicit `embedding-reset` + forced re-index. Runtime batch updates fan out atomically across Web PATCH, MCP `mem_config`, and disk hot reload; the sequence cap is restart-only. UI, i18n (en/ko), docs, CHANGELOG, and regression tests are included.

### Correctness — verified paths

- **Degraded mode is consistent across all write/read surfaces**: `IndexEngine._index_file` early-returns with an actionable error; `SearchPipeline.search` disables dense per-call via the storage property; `mem_add`/CRUD paths are covered because `_check_embedding_mismatch` reads the same `embedding_mismatch` property, which now includes `policy_mismatch`; `component_factory` reuses the existing `embedding_broken` signal so watchers/startup soft-sync can't write mixed vectors.
- **`reset_embedding_meta` clears the new flag**: it ends in `clear_embedding_mismatch()`, which now also resets `_policy_mismatch` — the storage migration test confirms the full legacy→degraded→reset→clean cycle.
- **`_revert_to_stored` stays coherent**: `max_sequence_tokens` is copied into config *before* `create_embedder`, so the rebuilt embedder configures its tokenizer at the stored cap; mismatch flags are cleared afterward. Runtime-only (non-persisted) revert matches the pre-existing provider/model behavior.
- **Tokenizer layout is pinned defensively**: `_configure_tokenizer_limit` refuses to run uncapped if FastEmbed's internal layout changes (`refusing unsafe ONNX fallback`), and never widens a smaller model limit. The golden-path test exercises the *real* FastEmbed tokenizer (native 128 narrowed to 64), which is the right contract pin for a private-attr dependency.
- **`chunk_positions` keyed by `id(chunk)` is safe**: `compute_diff` appends the same `Chunk` objects from `new_chunks` into `to_upsert` (no copies), and `new_chunks` outlives the lookup.
- **Persist-timeout atomicity**: the Web PATCH fanout for `onnx_batch_size` is deferred until after a successful persist (same pattern as #1567), and the timeout test proves the live embedder is untouched on 503.
- **Validator bypass covered**: `embedding.onnx_batch_size` is in the PATCH validation table (1–256) *and* re-validated in `set_onnx_batch_size`, so the known setattr-bypasses-validators hole doesn't apply.

### Suggestions (minor, non-blocking)

1. **Triplicated setter guard** — the `callable(setter) and (callable(class_setter) or "set_onnx_batch_size" in vars(embedder))` Mock-detection dance appears verbatim in `status_config.py`, `hot_reload.py`, and `web/routes/system.py`. Extract a small helper (e.g. `publish_onnx_batch_size(embedder, value) -> bool`) so the three call sites can't drift; the subtle Mock rationale then lives in one docstring.
2. **Preflight double-tokenization** — `_truncated_input_indexes` encodes every input once for warning detection, then FastEmbed tokenizes again for inference. That's ~2× tokenizer CPU per embed call even when nothing truncates. Fine at current chunk sizes, but worth a note: an alternative is checking only inputs whose `len(text)` exceeds ~`4 × applied_limit` chars, which skips the preflight entirely for the common short-chunk case.
3. **Bulk-index error duplication** — with a policy mismatch, `index_path` over N files will collect N identical "indexing is disabled" error strings. Consider short-circuiting at the directory level (or deduplicating in the aggregate) so `mm index` output stays readable on large trees.
4. **`import inspect` placement** in `engine.py` lands between `os` and `stat` (non-alphabetical). Ruff's isort rules aren't enabled so lint passes, but it reads as an accidental insertion.
5. **Banner edge**: `banner.emb_mismatch_policy` renders `stored.max_sequence_tokens` directly; for exotic meta states this can print `null`. Cheap guard: `?? t('common.model_limit')`-style fallback. Low priority since the adopt path always writes the max alongside the fingerprint.

### Tests

Coverage is thorough and hits the right seams: real-tokenizer contract pin (golden path), fake-tokenizer unit matrix (cap applied / zero preserved / never-widens / fail-loud), legacy-DB migration to degraded mode and recovery, BM25-only pipeline fallback with `dense_search`/`embed_query` assert-not-awaited, duck-typed capability negative tests (Mock auto-attribute, legacy signature subclass), persist-timeout atomicity, restart-only rejection of `max_sequence_tokens`, and cache-bust pins. The truncation-warning test asserting content *absence* in logs is a nice touch.

### Security

- Truncation warnings log only path + chunk indices, never content — consistent with the no-secrets-in-logs policy, and pinned by a test.
- The end-user impact of truncation (dense vector represents prefix only, BM25/stored content complete) is honestly documented in three places.

### Verdict

**LGTM.** The invariants (degraded-mode gating on every surface, tokenizer contract pinned against a private FastEmbed layout, conservative legacy migration with no automatic destructive re-index) are all sound and test-backed. The suggestions above are cleanups, none blocking.

pandas-studio:
Addressed all five review follow-ups in c211ea4d:

1. Extracted publish_onnx_batch_size into a shared helper with centralized Mock/AsyncMock capability detection.
2. Added a conservative tokenizer-preflight fast path for provably short ASCII inputs; non-ASCII and long inputs retain exact overflow detection (avoids unsafe 4x char heuristics for CJK/code).
3. Deduplicated aggregate bulk and streaming index errors while preserving order.
4. Restored alphabetical import placement in indexing/engine.py.
5. Added localized model-limit fallbacks for missing sequence-cap metadata and bumped the app cache key.

Verification: Ruff passed, 626 focused tests passed, and the real FastEmbed golden tokenizer-cap contract passed.

pandas-studio:
## Follow-up review — c211ea4 ("refactor(embedding): address PR review follow-ups")

All five suggestions from the previous review are addressed, each with test coverage. No new blocking issues found.

### Verified

1. **Setter guard extracted to `embedding/runtime.py::publish_onnx_batch_size`** — all three call sites (`status_config.py`, `hot_reload.py`, `web/routes/system.py`) now share the helper. The helper is also an improvement over the originals: `getattr(embedder, "__dict__", {})` makes it safe for `embedder=None` (the old hot-reload variant relied on short-circuit order alone), and the instance-dict lookup is evaluated *before* the `getattr` probe that would trigger Mock attribute fabrication. Pinned by both the AsyncMock-rejection and concrete-OnnxEmbedder-acceptance tests.
2. **ASCII fast path for the truncation preflight** — `text.isascii() and len(text) + 8 <= max_sequence_tokens` skips the extra encode. The bound is sound: practical tokenizers emit at most one token per character, and +8 is a comfortable margin for special tokens. Non-ASCII inputs keep the exact preflight (pinned with a Korean input). Nice catch on the interaction with the existing log-content test: the short ASCII sentinel would have been silently skipped by the fast path, and lengthening it ×50 keeps that test exercising the real truncation branch. `encode_calls == []` pins the fast path itself.
3. **Bulk error dedup** — order-preserving `dict.fromkeys(all_errors)` at both aggregate sites, with a 2-file test asserting a single error.
4. **`import inspect` ordering** fixed.
5. **Banner null guard** — `?? t('common.model_limit')` with new en/ko locale keys, `app.js?v=154` cache bust, and the refresh-contract pin updated in the same commit. Leaving `settings-config.js` at v=18 is correct since it's untouched here.

### Minor observations (no action needed)

- The dedup applies to *all* identical error strings, not just the mismatch message. Most error strings embed the file path so the practical risk is nil, but it is a small contract change ("N identical messages → 1").
- The `publish_onnx_batch_size` docstring slightly overstates the hazard: a plain `getattr` on an `AsyncMock` fabricates a callable child mock, but the un-awaited coroutine only appears if it's *called*. Wording only.

### Verdict

**LGTM — ready to merge.** The core invariants from the first pass (degraded-mode gating on every surface, real-tokenizer contract pin, conservative non-destructive migration) are unchanged, and the follow-up lands all five cleanups with tests.

Issue #1692, opened 2026-07-10
context(web/cli): surface read failures instead of empty defaults — registry, probes, error-vs-drift


pandas-studio:
## Problem

Several Context Gateway read paths collapse `error` / `unknown` / `no-op` into
`empty` / `clean` / `drift`, which gives a first-time user false confidence
about what was actually inspected. All verified against source at `d087cf15`:

1. **Registry silent degradation** — a corrupt or unreadable
   `known_projects.json` reduces the registered roster to `[]` on the read
   path (`KnownProjectsStore.load(strict=False)`,
   `context/projects.py:176-274`; used by `discover_project_scopes:819`), so
   `GET /api/context/projects` returns 200 with an empty roster and only a
   server-side log line. (Mutating routes are already strict → 500 with
   original bytes preserved — no change needed there.)
2. **Unknown fields dropped on rewrite** — `_write`
   (`context/projects.py:407-423`) re-renders the document from four known
   fields, so unknown document- and row-level JSON keys are lost on any
   mutation.
3. **Probe failure is wire-identical to a real zero** — per-kind count-probe
   exceptions collapse to `0` (`_counts_for`,
   `web/routes/context_projects.py:276-347`) and runtime probes collapse to
   `[]` (`web/routes/context_gateway.py:425-429,470-475`). The Portal renders
   a failed probe as the "empty" pill. Bonus: `compute_runtime_coverage` has
   no exception guard in `_scope_to_dict` (`context_projects.py:374-376`), so
   one bad scope 500s the whole Projects response under
   `?include=runtime_coverage`.
4. **Diff-collection errors count as drift** — `bool(diff_errors)` sets the
   project drift flag (`context/status.py:696-700`), so a probe failure is
   reported as Sync-remediable drift in the fleet summary and the CLI
   (`status --all-projects` exits 0 for it). The Portal renders
   `status:"error"` entries with no badge at all.

## Scope and constraints

- Additive wire changes only; existing field names/types stay stable through
  the next minor release. New stable reason codes: `registry_corrupt`,
  `status_unavailable`.
- No cross-surface state enum — per-probe availability (`ok`/`unavailable`)
  plus UI presentation states only; existing vocabularies
  (`ok|drift|error|skipped`, `ok|partial|failed`) unchanged; the fleet
  summary stays derivable (`drifted + errors == 0`).
- `known_projects.json` stays schema v1; refusal of unsupported future
  versions must never rewrite the original bytes.
[...]

PR #1813, opened 2026-07-17
fix(web): document SnapshotEntryOut as a deliberate allowlist (#1812)

pandas-studio:
Closes #1812. Follow-up from the #1810 review.

## Problem

`SnapshotEntryOut` is built via `SnapshotEntryOut(**entry, judgment=..., feedback_updated_at=...)` in `GET /api/search/runs/{run_id}`. Its docstring promised to "mirror the observation snapshot fields" and stay "permissive on optional metadata" — but Pydantic's default `extra="ignore"` made that permissiveness mean *accepted and discarded*, not *passed through*. A field added to the snapshot writer later (#1799/#1800 surface) would land in storage and be silently dropped from the response, with nothing to flag the drift.

## Resolution — closed allowlist, documented as intent

The issue offered two directions; this resolves toward the **closed allowlist**, not verbatim passthrough:

- This response is the **privacy boundary** for the surface. The search pipeline holds raw result content and only *projects* the safe fields (basename + hash, never content or absolute paths — `search/pipeline.py`), and storage round-trips whatever dict it is handed without validating it. So `extra="allow"` would let a future writer regression (a stray `content` or absolute-path key) auto-surface here.
- Keep the default `extra="ignore"` and **document it as load-bearing intent**: new snapshot fields are dropped until added to this allowlist on purpose — which is the moment to make the privacy call.
- Matches the repo convention that reserves `extra="allow"` for bodies carried *verbatim from another surface* (#1692 PR7), which a privacy-scrubbed projection is not.

A test injects a benign future field **plus a `content`/absolute-path leak** and asserts none reach the response while the declared safe fields still render. Mutation-checked: re-introducing `extra="allow"` fails it.

> Note: an earlier revision of this PR took the opposite option (`extra="allow"` passthrough). Codex review flagged that it weakens the privacy boundary on an unenforced writer assumption; reversed here.

## Notes

- **Stacked on `feat/1801-search-feedback` (PR #1810)** — the `SnapshotEntryOut` code only exists there, not yet on `main`. Retarget to `main` if #1810 merges first.

Verified: `pytest test_web_routes_search_runs.py` (16 passed), `ruff check`/`format`, `mypy` clean.

### DIFF
--- a/packages/memtomem/src/memtomem/storage/mixins/history.py
+++ b/packages/memtomem/src/memtomem/storage/mixins/history.py
@@ -245,21 +245,32 @@ class HistoryMixin:
             # BEGIN IMMEDIATE makes these unreachable in practice; classify
             # anyway so a constraint hit never leaks as an internal error.
             db.rollback()
-            return self._classify_feedback_integrity_error(db, exc, run_id, chunk_id, judgment)
+            return self._classify_feedback_integrity_error(
+                db, exc, run_id, chunk_id, judgment, replace=replace
+            )
         except Exception:
             db.rollback()
             raise
 
     def _classify_feedback_integrity_error(
-        self, db, exc: sqlite3.IntegrityError, run_id: str, chunk_id: str, judgment: str
+        self,
+        db,
+        exc: sqlite3.IntegrityError,
+        run_id: str,
+        chunk_id: str,
+        judgment: str,
+        *,
+        replace: bool = False,
     ) -> dict:
         """Map a constraint hit on the feedback write to its domain meaning.
 
         FK violation → the run was pruned between validation and insert
         (KeyError, same as an unknown run). Unique violation → a concurrent
         writer landed this (run_id, chunk_id) first: re-read and classify
-        exactly like the existing-row branch (idempotent success or
-        conflict). Anything else stays a StorageError.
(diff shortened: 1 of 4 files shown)


tsdata (approved):

pandas-studio:
Second Codex review round: the prior `extra="allow"` Major is resolved and the unknown-key drop is confirmed as a genuine pin. Codex then flagged that `source_name` (`str | None`) isn't content-validated at this model, so the docstring's "never absolute paths" read as a boundary guarantee it doesn't make. Resolved by precision, not scope: the docstring now separates the two guarantees at their two layers — *which* fields appear is bounded here (the allowlist); the *values* of allowed fields (`source_name` = basename) are guaranteed upstream by the writer's projection (`search/pipeline.py` snapshots `source_file.name`), which this model relies on and does not re-sanitize. Re-validating field contents at read would duplicate the writer's guarantee and mask a writer regression rather than surface it; if that hardening is wanted it belongs at the writer (or the centralized persistence-boundary sanitizer discussed earlier), tracked separately from this unknown-key fix.

pandas-studio:
Superseded by #1816 — this PR auto-closed when its base branch `feat/1801-search-feedback` was deleted on the #1810 merge. Same commit, rebased onto `main`.

PR #1814, opened 2026-07-17
fix(storage): keep feedback reclassification read-only, document replace collapse (#1811)

pandas-studio:
## Problem

Follow-up from the #1810 review (#1801). `HistoryMixin._classify_feedback_integrity_error` re-reads the row a concurrent writer landed after a `UNIQUE` hit and reclassifies it. The issue notes it never received the caller's `replace` flag, so a losing `replace=True` call surfaces a conflict instead of replacing.

## Why the fix is "document the collapse", not "perform the replace"

The helper runs **after `db.rollback()`** — outside the `BEGIN IMMEDIATE` that serializes the main write. Adding an `UPDATE`+`commit` there (the other suggested fix) would introduce an **unlocked read-modify-write**: a concurrent judgment change or a prune's cascade-delete could race the re-read, and a zero-row `UPDATE` would still commit and report `replaced=True` for a row that no longer exists. `BEGIN IMMEDIATE` keeps this branch unreachable in practice, so the right move is to keep it **read-only** and document the contract rather than harden an unreachable write.

## Fix

- Keep the reclassification path replace-agnostic and read-only: same judgment → idempotent success; different judgment → `FeedbackConflictError` (even when the caller passed `replace=True`); FK → `KeyError`; anything else → `StorageError`.
- Document in the helper's docstring that replacement intent **deliberately collapses to a conflict** on this defense-in-depth branch, and that recovery is a plain retry through the normal locked path — which performs the replace correctly.

## Test

`TestIntegrityClassification` gains `test_replace_intent_collapses_to_conflict_and_leaves_row_unchanged`, pinning: the collapse to a conflict, the winner's row left byte-for-byte unchanged (no unlocked write), and a subsequent retry through the locked path replacing correctly (`created_at` pinned, `updated_at` advanced).

Reviewed by Codex (dev-trio), which flagged the unlocked-write race in the initial "perform the replace" approach; this PR takes the read-only/document route it recommended.

Closes #1811.

### DIFF
--- a/packages/memtomem/src/memtomem/storage/mixins/history.py
+++ b/packages/memtomem/src/memtomem/storage/mixins/history.py
@@ -257,9 +257,20 @@ class HistoryMixin:
 
         FK violation → the run was pruned between validation and insert
         (KeyError, same as an unknown run). Unique violation → a concurrent
-        writer landed this (run_id, chunk_id) first: re-read and classify
-        exactly like the existing-row branch (idempotent success or
-        conflict). Anything else stays a StorageError.
+        writer landed this (run_id, chunk_id) first: re-read and classify.
+        Same judgment is an idempotent success; a different judgment is a
+        conflict; anything else stays a StorageError.
+
+        This path is **read-only by design** (#1811). It runs *after*
+        ``db.rollback()``, i.e. outside the ``BEGIN IMMEDIATE`` that
+        serializes the main write, so it must not perform a write here: an
+        unlocked read-modify-write would race a concurrent judgment change
+        or a prune's cascade-delete (a zero-row UPDATE would still report
+        success). A different judgment therefore raises a conflict even when
+        the losing caller passed ``replace=True`` — the caller's replacement
+        intent deliberately collapses to a conflict on this defense-in-depth
+        branch. Recovery is a plain retry through the normal locked path,
+        which then performs the replace correctly.
         """
         message = str(exc).upper()
         if "FOREIGN KEY" in message:
@@ -273,6 +284,8 @@ class HistoryMixin:
             if landed is not None:
                 if landed[0] == judgment:
                     return self._feedback_row(run_id, chunk_id, judgment, landed[1], landed[2])
+                # Different judgment: see docstring — replace intent
+                # collapses to a conflict here rather than an unlocked write.
                 raise FeedbackConflictError(
                     f"feedback for run {run_id!r} chunk {chunk_id!r} is already "
                     f"{landed[0]!r}; pass replace=true to overwrite"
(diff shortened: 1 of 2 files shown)


tsdata (approved):

pandas-studio (commented):
## Review — fix(storage): keep feedback reclassification read-only (#1811)

### Overview

Follow-up to #1810/#1801. Instead of threading the caller's `replace` flag into `_classify_feedback_integrity_error` and performing the replace there, this PR keeps the reclassification branch **read-only** and documents that a losing `replace=True` call deliberately collapses to `FeedbackConflictError`, with recovery via a plain retry through the locked path. Docstring + inline comment + one pinning test; no behavior change.

### Correctness — the design call is right

- Verified against `history.py`: the helper runs after `db.rollback()` (`history.py:247-248`), i.e. outside the `BEGIN IMMEDIATE` that serializes `save_search_feedback`. An `UPDATE`+`commit` here would be an unlocked read-modify-write — it could race a concurrent judgment change, and a prune's cascade-delete would make a zero-row `UPDATE` commit and falsely report `replaced=True`. Read-only + documented collapse is the safer contract for a branch that `BEGIN IMMEDIATE` makes unreachable in practice.
- The new test pins all three contract points: the collapse itself, the winner's row byte-for-byte untouched (catches any future unlocked write), and the retry-through-locked-path recovery (`created_at` pinned, `updated_at` strictly advanced).

### Verified locally

- `pytest packages/memtomem/tests/test_search_feedback.py` — 27 passed.
- `ruff check` + `ruff format --check` (full src/tests/tools scope) — clean.

### Nits (non-blocking)

1. **Docstring summary sentence dropped the conflict outcome.** The first paragraph now reads "Same judgment is an idempotent success; anything else stays a StorageError" — but a *different judgment* raises `FeedbackConflictError`, not `StorageError`; only unrecognized constraint messages stay `StorageError`. The pre-PR wording ("idempotent success or conflict") was accurate here. The long paragraph below clarifies it, but the summary is now internally inconsistent with the code three lines down. Suggest: "Same judgment is an idempotent success; a different judgment is a conflict; anything else stays a StorageError."
2. **Error message vs. losing `replace=True` caller.** The conflict message says "pass replace=true to overwrite" — which the losing caller may have already done. Harmless on an unreachable branch, and changing the message would diverge it from the main-path conflict, so fine to leave; just noting the docstring ("recovery is a plain retry") is the real guidance.
3. `test_replace_intent_collapses_to_conflict_and_leaves_row_unchanged` overlaps with `test_unique_violation_different_judgment_conflicts` on the raise itself; the extra assertions justify the separate test, no change requested.

### Verdict

**SHIP** modulo nit 1 (one-line docstring fix worth doing before merge — it's the sentence future readers will skim).

PR #1817, opened 2026-07-17
fix(search): deterministic id tie-break in BM25/dense legs (#516)

pandas-studio:
## What

Append the unique `c.id` as the final `ORDER BY` key on both the BM25 and dense search legs so tied rows come back in a fixed, id-sorted order.

## Why

Equal `fts.rank` (BM25) or equal `distance` (dense) rows within the same scope priority took arbitrary SQLite order (#516). Repeated queries or reopened connections could reorder them — harmless for interactive search, but it breaks the byte-deterministic replay diffs the Memory Quality Lab (#1798) needs. This is the first, independently-landable slice of the #1802 replay/eval work.

## Scope note

The dense fix stabilizes only rows the inner KNN actually returned. sqlite-vec 0.1.9 prunes to the inner `LIMIT` inside the vec0 virtual table with an unstable distance-only sort, so equal-distance rows straddling that inner cutoff are a separate concern — handled by the exhaustive-K evaluation path that lands with replay mode. The comment on the dense leg records this.

## Tests

`tests/test_search_tiebreak.py`:
- BM25 tied-rank rows return in id-sorted order.
- Dense tied-distance rows (identical embeddings) return in id-sorted order.
- Order stable across repeated queries.
- Order stable across a fresh backend / connection-pool reopen against the same DB.

`ruff check` + `ruff format --check` (src/tests/tools) and the storage + search suites pass; `mypy` clean on the edited file.

---

### Scope note (recall path)

This fixes the tie non-determinism on the **BM25 and dense ranked legs** only. The empty-query filter-only path (`recall_chunks`, `ORDER BY created_at DESC, {scope_sort_priority_case()}`) still lacks a unique final tiebreak, so the original same-second `created_at` flip from #516 remains there. That path is out of scope here — the Quality Lab replay set is keyword-ranked, so this PR unblocks #1802 — and closing it fully can be a separate follow-up. Read this as "deterministic ranked search legs", not "#516 fully closed on every path".

### DIFF
--- a/packages/memtomem/src/memtomem/storage/sqlite_backend.py
+++ b/packages/memtomem/src/memtomem/storage/sqlite_backend.py
@@ -1428,11 +1428,15 @@ class SqliteBackend(
             # importance, validity) activate. Score sits at the
             # trailing position after the chunk columns — see
             # ``_chunks_table_column_count`` consumer below.
+            # Trailing ``c.id`` is the unique final tiebreak (#516): equal
+            # ``fts.rank`` rows within the same scope priority would otherwise
+            # take arbitrary SQLite order, so repeated queries / reopened
+            # connections could reorder them and destabilize replay diffs.
             sql = f"""SELECT c.*, fts.rank
                    FROM chunks_fts fts
                    JOIN chunks c ON c.rowid = fts.rowid
                    WHERE chunks_fts MATCH ? {ns_clause} {scope_clause} {metadata_clause}
-                   ORDER BY fts.rank, {tie_break}
+                   ORDER BY fts.rank, {tie_break}, c.id
                    LIMIT ?"""
 
             # Try AND first (default FTS5 behaviour)
@@ -1522,8 +1526,14 @@ class SqliteBackend(
                    LIMIT ?
                ) sub
                JOIN chunks c ON c.rowid = sub.rowid {ns_clause} {scope_clause} {metadata_clause}
-               ORDER BY sub.distance, {tie_break}
+               ORDER BY sub.distance, {tie_break}, c.id
                LIMIT ?"""
+        # Trailing ``c.id`` gives the outer ordering a unique final tiebreak
+        # (#516). Note this only stabilizes rows the inner KNN actually
+        # returned; sqlite-vec 0.1.9 prunes to the inner ``LIMIT`` with an
+        # unstable distance-only sort, so equal-distance rows straddling that
+        # inner cutoff are a separate concern handled by the exhaustive
+        # evaluation path (see ``dense_search`` replay mode).
 
         # Total embedding rows — the upper bound for a meaningful
         # KNN K. Cheap; sqlite stores chunks_vec row counts in its
--- a/packages/memtomem/tests/test_search_tiebreak.py
+++ b/packages/memtomem/tests/test_search_tiebreak.py
@@ -1,0 +1,75 @@
+"""Deterministic tie-break in the BM25 and dense search legs (#516).
+
+Equal ``fts.rank`` (BM25) or equal ``distance`` (dense) rows within the same
+scope priority previously took arbitrary SQLite order, so repeated queries or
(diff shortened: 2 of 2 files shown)


tsdata (approved):

pandas-studio:
## Review

Minimal, well-scoped change: append the unique `c.id` (`TEXT PRIMARY KEY`, stored as `str(uuid4())`) as the final `ORDER BY` key on both search legs. Verified locally: the new test file passes (4/4), `ruff check` + `ruff format --check` clean.

### Verified correct

- `chunks.id` is a real unique key (`sqlite_schema.py` — `id TEXT PRIMARY KEY`); lowercase hex+dash ASCII means SQLite BINARY collation and the test's `sorted(key=str)` agree exactly.
- The BM25 OR-fallback reuses the same `sql` string, so the tie-break applies on that path too.
- The dense-leg scope note is accurate: the inner sqlite-vec KNN `LIMIT` cutoff can't be fixed by the outer `ORDER BY`, and deferring equal-distance rows straddling that cutoff to the exhaustive-K path is the right call.
- Perf impact negligible (one extra sort key on an already-sorted candidate set); no new injection surface (`c.id` is a literal).

### Suggestions (non-blocking)

1. **`recall_chunks` still lacks an id tie-break** — `ORDER BY created_at DESC, {scope_sort_priority_case()}` leaves the originally-observed same-second `created_at` flip (#516) in place on the recall path. Worth a note in the PR body (or a follow-up issue) so "#516 fix" isn't read as closing the recall path as well.
2. **Dense reopen coverage** — the connection-reopen stability test only exercises BM25. Parametrizing it over both legs would make the coverage symmetric; low risk since the mechanism is identical.
3. **Forward reference** — the dense-leg comment points at the `exhaustive` evaluation path, which lands with the stacked #1802 PR; if this merges first the reference dangles briefly. Adding "lands with #1802" to the comment would make that explicit.
4. Style nit: the BM25 comment sits before its `sql =`, the dense one after — worth unifying.

**Verdict: SHIP** — suggestion 1 is the only one I'd act on before merge (a one-line scope note).

pandas-studio:
Thanks for the review + approval. Addressed **#1** with a scope note in the PR body: this fixes the ranked BM25/dense legs only; the `recall_chunks` filter-only path still carries the #516 same-second flip and is explicitly out of scope (the replay set is keyword-ranked). Read as "deterministic ranked legs", not "#516 fully closed".

The non-blocking items (#2 dense reopen parametrization, #3 forward-ref wording, #4 comment placement) I've left as-is to avoid force-pushing onto an approved branch that #1819 is stacked on — happy to fold them in if you'd prefer before merge.

pandas-studio:
## Re-review

The diff is byte-identical to the previously reviewed revision (`5be6d86e..dc87ec0f`, test file `0d27ab43`); the change since last review is the new **"Scope note (recall path)"** section in the PR body. All prior verification stands: `test_search_tiebreak.py` 4/4 passing, `ruff check` + `ruff format --check` clean, `chunks.id` confirmed as a unique `TEXT PRIMARY KEY` whose `str(uuid4())` storage makes SQLite BINARY collation and the test's `sorted(key=str)` agree.

### Previous suggestions — status

1. **`recall_chunks` scope disclosure** → **resolved.** The new scope note records exactly what was asked: the same-second `created_at` flip remains on the filter-only recall path, the Quality Lab replay set is keyword-ranked so this PR unblocks #1802, and the PR should read as "deterministic ranked search legs", not "#516 fully closed".
2. Dense reopen coverage (reopen test exercises BM25 only) — still open, non-blocking.
3. Forward reference to the `exhaustive` path landing with #1802 — still open, non-blocking; the PR body discloses it.
4. Comment placement asymmetry (BM25 before `sql =`, dense after) — still open, style nit.

**Verdict: SHIP** — the only pre-merge ask (recall-path scope note) is addressed; items 2–4 are non-blocking and fine to leave for the stacked #1802 work or a follow-up.

PR #1818, opened 2026-07-17
refactor(quality): promote IR metrics into the package (#1802)

pandas-studio:
## What

Move the four pure IR-metric functions from `tests/ir_metrics.py` into a new runtime-importable module `memtomem.quality.metrics`, and leave `tests/ir_metrics.py` as a re-export shim. Add the metrics the Quality Lab reports need.

## Why

The replay/evaluation work in #1802 (parent #1798) computes hit rate / MRR / precision at runtime, so the metrics can no longer live under `tests/`. The shim keeps every existing consumer working with no change: test imports (`from ir_metrics import ...`) and the `tools/retrieval-eval` file-path loaders both still resolve.

## New metrics

- `hit_rate_at_k` — 1.0 if any relevant item is in the top-k.
- `precision_at_k` — returns `None` unless **every** top-k item carries a judgment, so partial qrels surface as `incomplete_labels` rather than a downward-biased number that an aggregate would silently average in.
- `recall_labeled_at_k` — an explicitly named alias signaling the denominator is the labeled-relevant set, not true corpus recall.

## Duplicate-safety

The Quality Lab keys results by `content_hash`, and several chunks can share one hash, so a ranking may contain duplicate IDs. `recall_at_k` / `recall_labeled_at_k` / `precision_at_k` now count **distinct** hits, so a duplicated relevant item cannot push a rate above 1.0. For the existing unique-id callers this is identical to the prior occurrence-counting behavior.

## Tests

`tests/test_quality_metrics.py` covers precision `None`-on-incomplete-labels, boundary `k`, duplicate-hash bounding at 1.0, an aggregate that excludes `None` precision, and shim parity. Existing `tests/test_ir_metrics.py` continues to pass through the shim; the `tools/retrieval-eval` loader was verified to resolve the shim transparently. `ruff` (src/tests/tools) and `mypy` (quality package) are clean.

### DIFF
--- a/packages/memtomem/src/memtomem/quality/__init__.py
+++ b/packages/memtomem/src/memtomem/quality/__init__.py
@@ -1,0 +1,6 @@
+"""Memory Quality Lab (#1798): local search-evaluation building blocks.
+
+This package holds the runtime-importable pieces of the Quality Lab —
+starting with pure-function IR metrics (:mod:`memtomem.quality.metrics`).
+Replay, fingerprints, and comparison land in later slices of #1802.
+"""
--- a/packages/memtomem/src/memtomem/quality/metrics.py
+++ b/packages/memtomem/src/memtomem/quality/metrics.py
@@ -10,10 +10,11 @@ The ``retrieved`` argument is an ordered list of IDs from rank 1 downward.
 Identity note (#1802): the Quality Lab keys results by ``content_hash``, and
 several chunks can share one hash, so a ranking may contain duplicate IDs. The
 replay layer deduplicates by identity before calling these functions; as defense
-in depth, ``recall_at_k`` / ``recall_labeled_at_k`` / ``precision_at_k`` count
-*distinct* hits so a duplicated relevant item cannot push a rate above 1.0. For
-rankings of unique IDs (the existing retrieval-regression callers) this is
-identical to the previous occurrence-counting behavior.
+in depth, every metric here — ``recall_at_k``, ``recall_labeled_at_k``,
+``precision_at_k``, ``ndcg_at_k``, ``hit_rate_at_k`` — credits each distinct ID
+at most once within the top-k window, so a duplicated item cannot push a score
+above 1.0. For rankings of unique IDs (the existing retrieval-regression
+callers) this is identical to the previous occurrence-counting behavior.
 """
 
 from __future__ import annotations
@@ -41,7 +42,7 @@ def recall_at_k(retrieved: list[str], relevant: set[str] | frozenset[str], k: in
     """
     if k <= 0 or not relevant:
         return 0.0
-    hits = len(set(retrieved[:k]) & set(relevant))
+    hits = len(set(retrieved[:k]) & relevant)
     return hits / len(relevant)
 
 
@@ -76,11 +77,19 @@ def ndcg_at_k(retrieved: list[str], relevance: Mapping[str, float], k: int) -> f
 
     Missing IDs in ``relevance`` count as zero gain. Returns 0.0 when the ideal
     DCG is zero (no positive-relevance items known at all).
+
+    Each distinct ID is credited once, at its first-occurrence rank within the
+    top-k window — a repeated ID (hash conflation, see module note) neither
+    double-counts nor promotes a rank-(k+1) item, so NDCG stays bounded by 1.0.
     """
     if k <= 0 or not relevance:
         return 0.0
     dcg = 0.0
+    seen: set[str] = set()
     for rank, item in enumerate(retrieved[:k], start=1):
+        if item in seen:
+            continue
+        seen.add(item)
         gain = relevance.get(item, 0.0)
         if gain > 0:
             dcg += gain / math.log2(rank + 1)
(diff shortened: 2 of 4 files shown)


pandas-studio:
## Review

Structure, tests, and docs are all solid. One item (#1) is worth a decision before merge; the rest are minor/nits.

**Verified**

- All three shim consumers keep working unchanged: `tests/test_ir_metrics.py` (`from ir_metrics import ...`), and the file-path loaders in `tools/retrieval-eval/benchmark_v2.py:131` and `calibrate_portfolio.py:167`. The shim now requires `memtomem` to be importable at `exec_module` time, but both tools already import memtomem (`create_components`), so this holds in every environment they run in.
- Existing `test_ir_metrics.py` recall tests use unique IDs only — no pin conflicts with the occurrence→distinct change.
- Packaging: src layout + `__init__.py` means the `quality` subpackage ships automatically.
- The `None`-on-incomplete-labels design for `precision_at_k` is the right call, and the aggregate-excludes-`None` test locks in the contract.

**Findings**

1. **`ndcg_at_k` is the one metric left out of duplicate-safety** (medium). The module docstring promises recall/recall_labeled/precision can't exceed 1.0 under duplicated IDs, but NDCG still counts occurrences: `ndcg_at_k(["a", "a"], {"a": 1.0}, 2)` → DCG = 1/log₂2 + 1/log₂3 ≈ 1.63 over IDCG = 1.0 → **≈1.63**. The "replay layer dedups first" premise covers it today, but the asymmetry is easy to miss when the Quality Lab later adds NDCG to reports — and `benchmark_v2.py:234` / `calibrate_portfolio.py:214` already call this function. Suggest either deduping `retrieved[:k]` by first occurrence in this PR, or at minimum stating in the docstring that NDCG is the exception and requires pre-deduped input.

2. **`precision_at_k` denominator semantics deserve one more docstring line** (minor). The denominator is the *distinct* window size — not `k`, not `len(window)`. A duplicated non-relevant item therefore *raises* precision: `(["a", "b", "b"], {"a"}, {"b"}, 3)` → 1/2 = 0.5 vs 1/3 occurrence-based. That's consistent with the dedup story, but the deviation from standard P@k (including the shrinking denominator when fewer than k items are retrieved) is worth spelling out explicitly.

3. **Potential retroactive effect on benchmark floors** (minor, verify only). The occurrence→distinct change in `recall_at_k` flows through the shim into both retrieval-eval benchmarks. If any ranking ever contained duplicate portable keys, past calibrated floors were computed on inflated recall. Likely a no-op, but one post-merge floor check would confirm.

4. **Shim parity test covers only the original 4 functions** (nit). Adding `precision_at_k`, `hit_rate_at_k`, and `recall_labeled_at_k` to `TestShimParity` completes it — three lines.

5. **Style** (nit, optional): `set(relevant)` in `recall_at_k` / `hit_rate_at_k` copies an argument that's already a `set | frozenset`; `set(retrieved[:k]) & relevant` suffices.

With #1 addressed (dedup or an explicit docstring exception), this is ready to ship.

pandas-studio:
Addressed in 0c21fa11:
- **#1 (NDCG duplicate-safety)** — `ndcg_at_k` now credits each distinct ID once at its first-occurrence rank within top-k, so it's bounded by 1.0 like the others; no-op for unique-id rankings (incl. the retrieval-eval benchmarks). Added a duplicate-bounding test.
- **#2** — spelled out the precision distinct-window denominator (shrinks below k; a duplicated non-relevant item raises the score).
- **#4** — extended `TestShimParity` to the three added functions.
- **#5** — dropped the redundant `set(...)` copies.
- **#3 (benchmark floors)** — verify-only; left for a post-merge floor check as you suggested. Existing rankings use unique chunk ids, so recall is unchanged there.

pandas-studio:
## Re-review (after 0c21fa11)

**SHIP.** Every finding from the previous round is addressed, and the fixes check out. No new issues introduced.

**Previous findings**

1. **NDCG duplicate-safety — resolved.** `ndcg_at_k` now credits each distinct ID once, at its first-occurrence rank, via a `seen` set. Verified the bound: distinct items' actual ranks are always ≥ their ranks in the ideal placement (ranks strictly increase), so by the rearrangement inequality DCG ≤ IDCG and the 1.0 bound holds. Regression test added (`["a", "a"]` → 1.0), and the docstring states the semantic choice that duplicates occupy their rank slots rather than promoting a rank-(k+1) item.
2. **Precision denominator semantics — resolved.** The docstring now spells out both deviations from textbook P@k: the denominator shrinks below `k` when fewer items are retrieved, and a duplicated non-relevant item raises the score (`["a", "b", "b"]` → 1/2, not 1/3).
3. **Benchmark-floor retroactive effect — still a post-merge action.** Nothing to fix in the diff; one retrieval-eval floor check after merge would confirm the occurrence→distinct change is a no-op for the calibrated floors.
4. **Shim parity — resolved.** `TestShimParity` now covers all 7 functions with `is` checks.
5. **Redundant `set()` copies — resolved.** Now `set(retrieved[:k]) & relevant` / `considered & relevant`. (`labeled = set(relevant) | set(not_relevant)` could be `relevant | not_relevant`, but that's taste-level.)

**Additional checks on the new commit**

- The module docstring now claims distinct-ID crediting for *every* metric — accurate; `reciprocal_rank_at_k` only looks at the first hit, so it was safe all along. No remaining asymmetry.
- NDCG dedup edges verified by hand: `["a", "b", "a"]` (k=3, both relevant) → 1.0; `["a", "a", "b"]` (k=2) → 0.613, i.e. the duplicate wastes a slot — matches the documented intent.
- Diff delta (279 → 300 additions) is entirely the fix commit; no other files touched.

The only blocker from round 1 is cleared — good to merge.

tsdata (approved):

PR #1819, opened 2026-07-17
feat(search): no-side-effects replay mode on SearchPipeline (#1802)

pandas-studio:
> Stacked on #1817 (base branch `fix/516-search-tiebreak`). The exhaustive-KNN determinism below relies on that PR's outer `c.id` tiebreak. Retarget to `main` once #1817 merges.

## What

Add a keyword-only `record: bool = True` to `SearchPipeline.search`. The default is today's behavior; `record=False` is the no-side-effects replay/evaluation mode the Memory Quality Lab (#1798 → #1802) needs.

## `record=False` contract

A read that touches no persistent or cross-call state:

- neither reads nor writes the TTL **result cache** — a concurrent interactive search's cache is never served to, nor evicted by, a replay;
- neither reads nor writes the LLM **expansion cache**, so replay can't depend on nor mutate hidden prior pipeline state;
- does not **increment access counters** (they feed the access/importance boosts, so mutating them would drift a later baseline-vs-candidate comparison);
- persists **no query-run observation** (`stats.query_run_id` stays `None`).

Threaded through the empty-query filter-only path too, so the flag never lies.

## Two determinism fixes ride along

- **Decay pinning.** `apply_score_decay` defaulted to wall clock at both call sites, so a replay with an explicit `as_of_unix` still drifted with real time (decay ages from `updated_at`). It now receives the effective `as_of` as `now`, so a delayed replay pinned to one instant scores identically.
- **Exhaustive dense KNN.** `record=False` switches **every** dense leg reachable from `search` — primary retrieval, heading expansion, and session-summary rescue — to `dense_search(exhaustive=True)`. sqlite-vec 0.1.9 prunes to the adaptive inner `LIMIT` with an unstable distance-only sort, so equal-distance rows straddling that cutoff are dropped nondeterministically. Scanning every embedding removes the inner cutoff, so the outer stable `ORDER BY sub.distance, …, c.id` (from #1817) fully determines selection. Production keeps the adaptive schedule; the exhaustive pass costs one full-table KNN scan and runs only off the interactive hot path.

## Tests

- `test_pipeline_replay_mode.py` — after `record=False`: access counts unchanged, no observation row, both caches empty; `record=True` still increments + persists + caches; a warm cache entry is neither served to nor mutated by a replay (`cache_hit` stays False); decay pinned to `as_of` is stable across a delayed replay and moves with a later `as_of`.
- `test_dense_exhaustive_replay.py` — with 120 tied-distance chunks (a real boundary past the first adaptive cutoff), exhaustive mode returns the id-sorted top-k deterministically across runs; a spy verifies the `exhaustive` flag threads through the primary leg (via `search`) and is off for normal search, and through heading expansion and the rescue leg (`_rescue_retrieval`) by driving each directly.

`ruff` (src/tests/tools), `mypy` (edited sources), and the search/storage/temporal suites pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,17 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   Omitted/`true` follows `rerank.enabled`; `true` cannot force-enable
   reranking on a server that has it disabled.
 
+### Changed
+
+- **Historical searches decay to the requested instant** (#1802) — a search
+  with an explicit `as_of_unix` now applies time decay relative to that
+  instant instead of the wall clock, matching how validity filtering already
+  treats `as_of`. Default-path searches (no `as_of`) are unchanged. This makes
+  point-in-time queries reproducible and underpins the Quality Lab's
+  no-side-effects replay mode (new `record=False` on the internal search
+  pipeline), which additionally skips access-counter increments, run
+  observations, and both search caches.
+
 ### Fixed
 
 - **Incremental CRUD re-indexing** (#1788) — `mem_edit`, `mem_delete`, and Web
--- a/packages/memtomem/src/memtomem/search/expansion.py
+++ b/packages/memtomem/src/memtomem/search/expansion.py
@@ -52,6 +52,7 @@ async def expand_query_headings(
     max_terms: int = 3,
     *,
     project_context_root: Path | None = None,
+    exhaustive: bool = False,
 ) -> str:
     """Expand query by appending related heading terms from dense search.
 
@@ -61,6 +62,11 @@ async def expand_query_headings(
     ``SearchPipeline.search`` call. Without it, expansion would
     silently sample only user-tier headings even when the outer
     search is pinned to a project context.
+
+    ``exhaustive`` propagates the caller's deterministic-evaluation mode
+    (#1802) into this leg's own dense probe: heading expansion changes the
+    query text, so a nondeterministic dense cutoff here would perturb the
+    whole downstream ranking during replay.
     """
     try:
         embedding = await embedder.embed_query(query)
@@ -68,6 +74,7 @@ async def expand_query_headings(
             embedding,
             top_k=3,
             project_context_root=project_context_root,
+            exhaustive=exhaustive,
         )
     except Exception:
         logger.warning("Heading expansion failed; returning original query", exc_info=True)
--- a/packages/memtomem/src/memtomem/search/pipeline.py
+++ b/packages/memtomem/src/memtomem/search/pipeline.py
@@ -760,6 +760,7 @@ class SearchPipeline:
         scope_filter: ScopeFilter | None = None,
         project_context_root: Path | None = None,
         metadata_filter: SearchMetadataFilter | None = None,
+        exhaustive: bool = False,
     ) -> list[SearchResult]:
         """Parallel BM25+dense retrieval restricted to boost_sources.
 
@@ -811,6 +812,7 @@ class SearchPipeline:
                     namespace_filter=None,
                     scope_filter=scope_filter,
                     project_context_root=project_context_root,
+                    exhaustive=exhaustive,
                     **metadata_kwargs,
                 )
             except Exception:
@@ -891,6 +893,7 @@ class SearchPipeline:
         project_context_root: Path | None = None,
         metadata_filter: SearchMetadataFilter | None = None,
         exclude_source_roots: tuple[Path, ...] = (),
+        record: bool = True,
     ) -> tuple[list[SearchResult], RetrievalStats]:
         """Empty-query path (#750): enumerate by filter, skip retrievers.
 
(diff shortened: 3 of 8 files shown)


pandas-studio:
## Review — no-side-effects replay mode

Verified against source on the PR head (`76dde429`): new tests 7/7 pass locally (0.25s), `ruff check` + `format --check` clean on the changed files.

### Confirmed correct

- **All four side-effect surfaces are gated symmetrically.** TTL result cache is blocked in both directions — read (`pipeline.py:1133`) and write (`:1606`) — and the expansion cache likewise (`.get() if record` + gated write). Access increments are gated on both the keyword path (`:1566`) and the filter-only path (`:1006`). The filter-only path never touched the caches or observation store to begin with, so the increment gate alone completes the contract there — "the flag never lies" holds.
- **Decay pinning is consistent.** `apply_score_decay(now=)` exists, both call sites pass the same `effective_as_of` the validity filter uses, imports are in place, and no pipeline stage is reordered (CLAUDE.md invariant intact).
- **`exhaustive` threading is complete.** Primary (`:1267`), heading expansion (`expansion.py:73`), and rescue (`:1332` → `:812`) all receive `exhaustive=not record`. The `base.py` protocol change is keyword-only with a default, so the four other `dense_search` call sites (dedup, conflict, web chunks route, indexing engine) are unaffected.
- **Empty-table edge in the exhaustive path is safe**: `attempts=[1]` → empty result → `inner_k >= 0` breaks immediately.

### Findings (all minor, none blocking)

1. **Test name overstates coverage** — `test_primary_and_rescue_legs_thread_exhaustive` never actually exercises the rescue leg. Rescue only fires with a non-empty `boost_sources` from session summaries (`pipeline.py:1313–1325`), which the `components` fixture doesn't seed, so the spy observes only the primary leg and `all(seen)` is vacuously true for legs that didn't run. The wiring is a literal one-liner so risk is low, but either drop the rescue claim from the name/docstring or seed a session summary so it genuinely fires.
2. **Replay is deterministic but not production-faithful** — switching heading expansion's dense probe to exhaustive can sample different headings than the adaptive cutoff did in production, changing the expanded query itself. The tradeoff is stated in the PR body and is fine; the Quality Lab consumer (#1798) should just keep this gap in mind when diffing replays against logged production runs.
3. **A `record=True`-visible behavior change rides along** — historical searches with an explicit `as_of_unix` now decay to the pinned instant instead of wall clock. It's the correct fix and consistent with validity filtering, and historical queries aren't cached so blast radius is small, but it deserves a CHANGELOG line since it changes ranking outside replay mode.
4. **Fixture config mutation** — `test_decay_pinned_to_as_of` sets `pipeline._decay_config.enabled = True` without restoring it. Harmless while the fixture is function-scoped; would leak across tests if the scope ever widens.

### Assessment

Production keeps the adaptive dense schedule, so no interactive hot-path regression; the exhaustive full-table KNN runs only off the hot path. Conventions (registry, stage order, comment density, line length) are all respected. No security surface.

**Verdict: SHIP** — the four notes above can land as follow-ups; only #1 (vacuous rescue assertion) is worth fixing in this PR. The stacked-base retarget to `main` after #1817 merges is already noted in the body.

pandas-studio:
Addressed in d1d897b4:
- **#1 (vacuous rescue assertion)** — split the combined test: a primary-leg pipeline test, plus a direct `_rescue_retrieval` spy test that genuinely drives the rescue dense leg (the pipeline path never fires rescue without seeded session-summary boost sources, exactly as you noted). Now all three legs — primary, heading expansion, rescue — have honest coverage.
- **#3 (CHANGELOG)** — added a "Changed" entry noting that an explicit `as_of_unix` now decays to that instant on the normal path too.
- **#4 (fixture mutation)** — switched the decay-enabled toggle to `monkeypatch.setattr`.
- **#2 (production-faithfulness)** — agreed; it's a documented replay-vs-production gap for the #1798 consumer, left as-is.

Note the full CI matrix won't run here until this is retargeted to `main` after #1817 merges (CI only triggers on main-targeted PRs). Locally: updated tests 8/8, ruff + mypy clean.

pandas-studio:
## Re-review — delta since `76dde429` (`d1d897b4`)

Verified against source on the new head: replay test files now 8/8 locally (0.60s), `ruff check` + `format --check` clean.

The follow-up commit resolves three of the four notes from the [previous review](https://github.com/memtomem/memtomem/pull/1819#issuecomment-5002401772):

- **#1 (vacuous rescue assertion) — resolved.** The overstated test is renamed to `test_primary_leg_threads_exhaustive` with an honest assertion message, and a new `test_rescue_leg_threads_exhaustive` exercises `_rescue_retrieval` directly with a non-empty `boost_sources`, proving the dense leg receives `exhaustive=True` (`seen == [True]`). Calling the private method is a fair tradeoff here and the comment explains why the fixtures can't fire the leg organically; the call matches the real signature (positional `query`/`query_embedding`, defaults cover the scope/metadata params).
- **#3 (CHANGELOG) — resolved.** The `Changed` entry correctly leads with the `record=True`-visible part (historical `as_of_unix` searches now decay to the pinned instant) and notes the default path is unchanged.
- **#4 (fixture mutation) — resolved.** `monkeypatch.setattr` now restores `_decay_config.enabled` automatically.
- **#2 (replay ≠ production-faithful heading expansion)** was informational; no action needed.

One trivial residual: the PR body still says the spy verifies rescue "under replay" through `search()` — it's now verified via a direct `_rescue_retrieval` call instead. Worth a one-word tweak when retargeting the base, not worth a push on its own.

**Verdict: SHIP** — no remaining blockers; pending only the stacked-base retarget to `main` after #1817 merges.

tsdata (approved):

PR #1823, opened 2026-07-17
fix(embedding): release ONNX CPU arena memory

pandas-studio:
## Summary

- disable the ONNX Runtime CPU memory arena by default for local FastEmbed sessions
- verify the loaded ORT session applied the requested allocator option, with an explicit compatibility escape hatch
- expose the restart-only setting in config APIs and Web UI, and document that it does not require re-indexing

## Evidence

- ORT 1.27 synthetic A/B, three pairs: arena-on median retained excess 1.57 GiB; arena-off median retained excess below baseline
- vectors were bit-identical in all three A/B pairs
- 189-source validation: 560 chunks/vectors, 0 errors; post-implementation peak 4.68 GiB and final RSS below the pre-workload baseline
- pinned ORT 1.24 golden path confirms the loaded session has the arena disabled

## Validation

- 8,645 passed, 11 skipped, 314 deselected (non-live suite)
- 216 focused tests passed after rebasing onto current main
- ruff check, changed-file format check, focused mypy, JS syntax, locale JSON, and git diff checks passed

Closes #1784


tsdata (approved):

pandas-studio:
## Review

### Overview

Disables ORT's CPU memory arena by default for the local FastEmbed provider (`embedding.onnx_cpu_mem_arena=false`), verifies the loaded session actually applied the setting, and surfaces the restart-only field across config API, Web UI, and docs. Fixes multi-GB retained RSS after sequential indexing (#1784).

### Correctness — verified against sources

- **The kwarg is an officially supported fastembed surface, not a lucky side channel.** fastembed 0.8.0 declares `EXPOSED_SESSION_OPTIONS = ('enable_cpu_mem_arena',)`; `OnnxTextEmbedding.__init__` filters top-level kwargs through `_select_exposed_session_options` and `_load_onnx_model` applies it to `ort.SessionOptions` via `add_extra_session_options`. The `fastembed>=0.8,<0.9` floor in `pyproject.toml` guarantees this path exists, so `_verify_cpu_mem_arena` is a drift guard rather than load-bearing plumbing — the right shape.
- **Fail-closed asymmetry is right.** Unknown session layout with the memory-safety default (`false`) raises `EmbeddingError`; layout drift under an explicit `true` opt-in degrades to a warning. The `isinstance(actual, bool)` check makes the subsequent `is not` identity comparison safe (a non-bool from a future pybind change fails loud instead of comparing wrong).
- **Fingerprint exclusion is correct by construction.** `embedding_policy_fingerprint` enumerates only `max_sequence_tokens` for onnx, so the allocator toggle can't leak into vector identity — and `test_cpu_mem_arena_does_not_change_vector_policy` pins that, matching the bit-identical A/B evidence.
- **No server-side read-only change was needed and none was made:** the field is simply absent from `MUTABLE_FIELDS`, so the PATCH handler rejects it as `read-only field`; `test_patch_cpu_mem_arena_is_restart_only` pins that behavior.
- **Load-failure integration is clean:** `_verify_cpu_mem_arena` raising inside `_get_model`'s try block sets `_load_error` and resets `_loading` via the existing `except`/`finally`.

### Test coverage

Strong. Forwarding + verification parametrized over both values, fail-closed / escape-hatch / mismatch unit-tested, fingerprint invariance pinned, and — most importantly — the golden path asserts `enable_cpu_mem_arena is False` on a **real** loaded session, so the mock-based tests can't false-green the whole feature. Web schema, PATCH rejection, and the cache-bust contract (`settings-config.js?v=19`) are all updated; en/ko locale keys added in pairs.

### Minor notes (no changes requested)

- The `model.model.model` attribute walk is coupled to fastembed internals, but that coupling is the entire purpose of the verify function, both unknown-layout branches are handled, and the golden path pins the real layout under the pinned dependency. Acceptable.
- If a future 0.8.x patch release changes the session layout, the default config refuses to load ONNX embeddings entirely rather than run unsafe. That's a deliberate availability-for-safety trade-off with a documented escape hatch (`onnx_cpu_mem_arena=true`) — worth remembering when bumping fastembed.
- `test_cpu_mem_arena_true_is_unknown_layout_escape_hatch` relies on caplog's default WARNING propagation without `caplog.set_level`; fine as-is, just slightly implicit.

### Verdict

**SHIP.** Mechanism verified against fastembed 0.8.0 source, defaults are memory-safe with a fail-loud guard, policy fingerprint untouched, and the real-session golden-path assertion closes the mock false-green gap.

PR #1825, opened 2026-07-18
feat(quality): eval case storage, promotion, export/import (#1802)

pandas-studio:
## What

PR-3 of the #1802 Quality Lab Q3 series. Adds the durable **evaluation-case** layer: promote a labeled search run into a self-contained case that outlives the 90-day `query_history` prune, plus the profile/corpus/index **fingerprints** that will gate replay-report comparison in PR-4. Storage + fingerprints + tests only — no replay/CLI yet.

## Changes

**Storage** (`storage/mixins/eval_cases.py`, `sqlite_schema.py`, `sqlite_backend.py`, `errors.py`)
- New `eval_cases` + `eval_case_labels` tables (additive; `SCHEMA_VERSION` stays 1 — see follow-up below), added to `reset_all()`'s deletion list (children before parents) with a test.
- `promote_search_run` copies query/filters/snapshot + per-`content_hash` labels under one `BEGIN IMMEDIATE`, so a concurrent history prune or feedback replace can't tear it. `source_run_id` is provenance only (no FK) so cases survive pruning. Refuses runs with no feedback, unreplayable-filter runs (overridable), and runs that reach a **project tier** — project-tier detection reuses `scope_context_sql`'s own clause builders evaluated by SQLite, so `LIKE` case-insensitivity and `%` semantics match production exactly. Conflicting per-hash judgments raise `EvalCaseError`.
- `list` / `get` (case_id-then-name precedence) / `set_status` / `export` / `import`. Import runs the whole batch under one `BEGIN IMMEDIATE`, validates the full envelope up front (raising `EvalCaseError`, never a raw `TypeError`/SQLite error), and **preserves the payload `case_id` + `version`** (minting only when absent) so an export→import round-trip keeps case-set identity across machines.

**Fingerprints** (`quality/fingerprints.py`)
- `profile_fingerprint` — explicit non-secret allowlist (`api_key` never read); each ranking stage contributes only its enable flag when inactive, so two ranking-identical profiles that differ only in a disabled stage's dead parameters don't false-drift (MMR additionally gated on `enable_dense`).
- `corpus_fingerprint` — multiplicity-preserving; includes `project_root` (retrievability depends on it).
- `index_fingerprint` — vectors/FTS/links/access keyed by **durable** chunk identity (`content_hash` + namespace + hashed source + start_line), never the storage-local rowid/uuid a re-index rotates, so a detached vector is drift while an equivalent rebuild is not.

## Tests

`tests/test_eval_cases.py` + `tests/test_quality_fingerprints.py` (101 tests): promotion atomicity/refusals/dup-hash, prune survival, project-scope detection (incl. uppercase/`%` globs), export/import round-trip + identity preservation + failed-batch rollback + malformed-input rejection, `reset_all` coverage, per-dependency fingerprint drift, rekeyed-rebuild stability, disabled-stage invariance, and a sentinel-secret scan.

Full suite green (`pytest -m "not ollama"`: 9063 passed), ruff clean, mypy clean on changed files. Design and implementation both reviewed via the repo's Codex gate.

## Follow-up

- #1826 tracks the additive-table `reset_all`/downgrade gap **repo-wide** (it affects `search_feedback` #1801 identically): an older binary can open a same-version DB and its `reset_all` won't clear tables it predates. Kept out of this PR to stay consistent with the established additive-table convention; tracked for a consistent policy decision.
- PR-4 (replay engine + `compare` + `mm quality` CLI) consumes these fingerprints.

Part of #1802 (series continues through PR-5).


pandas-studio:
## Review — PR #1825 (eval case storage, promotion, export/import)

### Overview

PR-3 of the #1802 series: two additive tables (`eval_cases`, `eval_case_labels`), an `EvalCaseMixin` (promotion / retrieval / export-import / fingerprint read helpers), and pure fingerprint functions in `quality/fingerprints.py`. Storage-only, no CLI/replay surface yet. Overall this is a well-designed slice — the review found **one confirmed contract bug** and a few minor issues.

### Findings

**1. [Bug — confirmed] Import: an explicit `null` fingerprint value leaks a raw `sqlite3.IntegrityError`** — `eval_cases.py`

`_validate_import_case` accepts `None` fingerprint values (`if value is not None and not isinstance(value, str)`), but `_import_one_case` reads them with `fingerprints.get("profile", "")` — the `.get` default applies only when the key is **absent**; a present-but-null key returns `None`, which hits the `NOT NULL` column. Reproduced against this branch:

```py
"promoted_fingerprints": {"profile": None}
# → sqlite3.IntegrityError: NOT NULL constraint failed: eval_cases.promoted_profile_fingerprint
```

This violates the documented contract ("raising `EvalCaseError`, never a raw `TypeError`/SQLite error") and the validate-everything-up-front intent. The batch does roll back cleanly, so there's no torn state — it's the exception type that's wrong. Fix either way: reject `None` in `_validate_import_case`, or normalize with `fingerprints.get(key) or ""` at insert; and add this payload to the `test_import_rejects_malformed_domain_fields` parametrize list.

**2. [Minor — silent drop] `export_eval_cases(case_ids=[...])` silently skips unknown ids**

A misspelled or stale `case_id` just never matches the `wanted` set, so the envelope comes back smaller with no error — the classic round-trip silent-drop shape. Suggest raising `EvalCaseError` for unmatched ids after the loop (cheap: `wanted - {c["case_id"] for c in summaries}`). Also a small API asymmetry: `get`/`set_status` accept name-or-id, export accepts only case_id.

**3. [Minor] Duplicate `case_id` within one payload + `replace=True` is silent last-wins**

Two cases in the same envelope sharing a `case_id`: the second deletes and rewrites the first with no error. Duplicate *names* in one payload do fail (insert-time owner check + batch rollback), so ids are the inconsistent one. A duplicate-id check in the up-front validation loop would close it.

**4. [Nit] `top_k = int(observation.get("top_k", ...))` in `promote_search_run` can raise a raw `ValueError`/`TypeError`** if a stored observation carries a non-numeric `top_k`. Observations are internally written, so this is trusted input in practice — but a guard raising `EvalCaseError` would match the module's error discipline.

**5. [Nit] Label-less cases pass import validation** (`labels` defaults to `[]`; then `top_k` just has to be explicit and positive). A case with zero labels is unusable for PR-4 metrics — either require ≥1 label at import or document that empty is intentional.

**6. [Nit — for the PR-4 consumer] `case_set_fingerprint` can `TypeError` on mixed missing/present `case_id`** across cases (`None` vs `str` is unorderable in the outer sort). Unreachable via stored cases (import always mints an id), only via hand-built dicts — worth remembering when PR-4 wires it up.

### What's solid

- **Atomicity**: promotion and import both run under one `BEGIN IMMEDIATE`, mirror `save_search_feedback`'s `_in_transaction` guard, and the rollback paths are tested (name collision, failed batch replacing a case).
- **Project-scope detection** delegating to `scope_context_sql`'s own clause builders evaluated by SQLite is the right call — the `PROJECT_*` / `project%*` parametrized tests pin exactly the semantics a hand-rolled matcher would miss.
- **Durable chunk identity** for index fingerprints (content_hash + namespace + hashed source + start_line, never rowid/uuid) with both directions tested: rekeyed rebuild stable, vector detach drifts.
- **Privacy**: source paths and project roots hashed before entering fingerprints; profile fingerprint is an explicit allowlist with a sentinel-secret + `is_secret_key` walk test.
- **Schema discipline**: `reset_all` children-before-parents with a count-asserting test; the partial unique index on `name` explicitly justified against the #1801 FK-parent constraint in a comment; `list_eval_cases` ordering has a deterministic tie-break.
- Validation-before-mutation for the import batch (so `replace` never strands a deleted case) is exactly right.

### Test coverage

95 tests covering refusals, atomicity, prune survival, round-trip identity, malformed-input rejection, and per-dependency fingerprint drift — strong. Gaps: the finding-1 payload (`{"profile": null}`), and unknown-id export behavior.

### Performance / security

- O(corpus) fingerprint reads are documented and kept off the search path; export's per-case query loop is fine at eval-case scale.
- SQL is parameterized throughout; the one f-string interpolates only internal clause fragments with user values bound via params (`nosec B608` justified in a comment).

### Verdict

**NEEDS-FIX (small)** — finding 1 is a confirmed contract violation with a two-line fix plus one test row; findings 2–3 are worth a decision before the export/import surface gets a CLI in PR-4. Everything else is ship-shape.

pandas-studio:
Thanks — addressed all six findings in `1204d186` (force-pushed). Full suite green (9063 passed), ruff/bandit/mypy clean, 101 tests in the two eval-case files (+6 new).

**1. [Bug] null fingerprint → raw `IntegrityError`** — fixed. `_validate_import_case` now rejects a *present* non-string fingerprint value via `if key in fingerprints and not isinstance(fingerprints[key], str)` (distinguishing present-null from absent, since `.get(key, "")` only defaults on absence). Reproduced the reported `{"profile": null}` payload → now raises `EvalCaseError: imported fingerprint 'profile' must be a string`; added to the `test_import_rejects_malformed_domain_fields` parametrize.

**2. [Minor] export silently skips unknown ids** — fixed. `export_eval_cases` now raises `EvalCaseError("unknown eval case ids: [...]")` for any requested id not present, before building the envelope (`test_export_unknown_id_raises`). Left the export-by-id vs get/set-by-name-or-id asymmetry as-is — bulk export by id reads fine, and I didn't want to widen the surface without a reason; happy to revisit when the CLI lands in PR-4.

**3. [Minor] duplicate `case_id` in one payload + replace** — fixed. The up-front validation loop now tracks explicit case_ids and raises `EvalCaseError("duplicate case_id ...")`, matching the existing duplicate-*name* rejection (`test_import_duplicate_case_id_in_payload_rejected`, asserts nothing persists).

**4. [Nit] `promote_search_run` raw `ValueError` on non-numeric `top_k`** — fixed. Wrapped the `int(...)` in a guard raising `EvalCaseError("... has a non-numeric top_k ...")` (`test_non_numeric_observation_top_k_raises`, seeds a corrupt observation).

**5. [Nit] label-less cases** — chose to require ≥1 label at import (`EvalCaseError("imported case must have at least one label")`), matching promotion's feedback floor since a zero-label case is unusable for PR-4 metrics. Test row added.

**6. [Nit/PR-4] `case_set_fingerprint` `TypeError` on mixed missing/present `case_id`** — hardened now rather than deferring: every ordering field is coerced through a None-safe `_ns()` (None→""), so a hand-built mix can't `TypeError` the outer sort (`test_mixed_missing_and_present_case_id_does_not_crash`). Stored cases are unaffected (case_id/version always present).

Findings 2/5 involved a small judgment call (export API scope; require-vs-document empty labels) — flagging those two explicitly in case you'd prefer a different resolution.

pandas-studio:
## Re-review — delta since [previous review](https://github.com/memtomem/memtomem/pull/1825#issuecomment-5009155315) (head `9ca6ab04` → `1204d186`)

**Verdict: SHIP.** The amended head addresses all six findings from the previous review, each with a regression test; CI is fully green on the new head (all required checks pass) and the PR is mergeable. No new defects introduced by the fixes.

### Finding-by-finding verification (from the inter-head diff)

| Previous finding | Fix | Regression test |
|---|---|---|
| 1. Present-but-null fingerprint leaked raw `IntegrityError` (confirmed bug) | Validation now rejects a present non-string: `key in fingerprints and not isinstance(fingerprints[key], str)` | `{"promoted_fingerprints": {"profile": None}}` parametrize row |
| 2. `export_eval_cases` silently skipped unknown ids | Up-front set difference raises `EvalCaseError` with the sorted missing ids | `test_export_unknown_id_raises` |
| 3. Duplicate `case_id` in one payload + `replace=True` was silent last-wins | Rejected in the up-front validation loop | `test_import_duplicate_case_id_in_payload_rejected` |
| 4. `int(observation top_k)` could raise raw `ValueError`/`TypeError` | try/except → `EvalCaseError` (`from None`) | Corrupt-observation test via direct `UPDATE` |
| 5. Label-less cases passed import validation | Import now requires ≥1 label, matching promotion's feedback floor | `{"labels": []}` parametrize row |
| 6. `case_set_fingerprint` could `TypeError` on mixed None/str fields | `_ns` None-safe string coercion on every ordering field | `test_mixed_missing_and_present_case_id_does_not_crash` |

### Review of the fixes themselves

- **Fix 1 took the right branch**: rejecting at validation rather than papering over with `or ""` at insert keeps the "validate the entire envelope up front" design intact; an *absent* key still defaults to `""` for hand-authored sets.
- **Fix 3's boundary is correct**: cases that both omit `case_id` each mint a fresh uuid at import, so excluding `None` from the duplicate check is right.
- **Fix 5 interaction is benign**: `top_k`'s `len(labels)` default is now guaranteed ≥1, which also removes the previously misleading "top_k must be positive" path for label-less cases.
- **Fix 6 changes the canonical form** (ints coerced to strings), but `case_set_fingerprint` has no callers outside tests yet and nothing persists its output before PR-4, so there's no compatibility concern. The theoretical `_ns(None) == _ns("")` collision is unreachable — import validation forbids empty-string `case_id`.
- One harmless leftover: the `if isinstance(case, dict)` guard in the duplicate-id loop is dead code (`_validate_import_case` already rejected non-dicts one line earlier). Not worth a respin.

Two non-blocking notes carried forward: export still accepts only `case_id` (not name) — fine to settle with the PR-4 CLI design; and the PR body's "95 tests" count is now slightly behind the amended suite (+6).

tsdata (approved):

